PR #33, opened 2012-12-11
Ref #3001 : add event log filters

VinceMacBuche:
The main goal of that feature is to provide a way to filter events logs easily

see http://www.rudder-project.org/redmine/issues/3001 for more détails!


fanf:
I'm closing that one because I don't thing there is much which would be done today like it was done at the time, and we aren't even sure we want to to it anymore like that. That is without counting the merge conflicts to solve.

PR #760, opened 2015-01-02
Fixes #6063: Performance of retrieving last node run reports time impact...

fanf:
The main idea is just to encapsulate the read/write of node runs into a proxy service which holds a cache. 

A little synchronisation is needed to avoid loosing some runs, but that's 20 lines of code that massively speed up things.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
@@ -37,7 +37,7 @@ package com.normation.rudder.reports.execution
 import com.normation.inventory.domain.NodeId
 import net.liftweb.common._
 import org.joda.time.DateTime
-
+import com.normation.rudder.repository.CachedRepository
 
 /**
  * Service for reading or storing execution of Nodes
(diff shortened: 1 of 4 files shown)


fanf:
I, forgot to bind the clear-cache to the user clear-cache action.

fanf:
It's OK, PR updated

VinceMacBuche:
Looks very cool to me, just making a last review but it should be ok  !!

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
```diff
@@ -71,4 +71,61 @@ trait WoReportsExecutionRepository {
...
+    }
+  }
+
+
+  override def updateExecutions(executions : Seq[AgentRun]) : Box[Seq[AgentRun]] = this.synchronized {
+    for {
+      runs <- writeBackend.updateExecutions(executions)
+    } yield {
+      cache match {
+        case None => //ok, does nothing
```
VinceMacBuche: Why not set cache to last run here ?

fanf: Because we would miss all the node run that are not updated here and already in backend. So we must always initialise cache (from None to Some...) with a full read. 

That make me thing that it is actually NOT what I did, since I initialized the cache with a partial read only. So dumb. 

Let me correct that.

fanf:
PR updated. Thanks for the remark.

VinceMacBuche:
Great ! Merging!

PR #761, opened 2015-01-02
Fixes #5905: When adding a line of search criteria, the category and field from the line above should be the same on the new line

fanf:


VinceMacBuche:
Awesome :)

PR #764, opened 2015-01-05
Fixes #6073: Compute compliance asynchronously when dispalying Rule page

VinceMacBuche:


fanf:
Something does not work as expected here, but I don't know what for now. I don't see real async computation for compliance, and switching from page to page on the rule grid takes ages (10s ?) with lots of CPU used (one core 100%). 
So, I'm going to try to find more information on that.

VinceMacBuche:
Activate the debug logger, you will see what is taking you so much time

VinceMacBuche:
@fanf updated !

fanf:
Now, that's nice !

PR #768, opened 2015-01-07
Fixes #5452: Performance issue for node list

fanf:
Major performance improvment on "nodeInfo" with a "cached" implementation of the nodeInfoService in place of the dumb one. 
The propose change lead to two major perf improvment:
- for dynGroup, reduce the number of request from "Number of node in group" to one (1000x improvment with > 1000 nodes in groups)
- when getting the list of all nodeInfo, time from ~4000ms to ~20ms

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
@@ -71,7 +71,7 @@ trait NodeInfoService {
   /**
    * Retrieve minimal information needed for the node info
    */
-  def getLDAPNodeInfo(nodeId: NodeId) : Box[LDAPNodeInfo]
+  def getLDAPNodeInfo(nodeIds: Set[NodeId]) : Box[Set[LDAPNodeInfo]]
 
   /**
    * Return a NodeInfo from a NodeId. First check the ou=Node, then fetch the other data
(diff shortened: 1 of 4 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -111,84 +111,97 @@ object NodeInfoServiceImpl {
...
+      TimingDebugLogger.debug(s"Cache for nodes info expire ?: ${n1-n0}ms")
+      res
+    }
 
-    ldap.map { con =>
+    /**
+     * Get all relevant info from cache along with the date of the
+     * last modification
+     */
+    def getInfo(): Box[(Map[NodeId, (LDAPNodeInfo, NodeInfo, Node)], DateTime)] = {
```
ncharles: could you give a more meaningfull name to this method, like getAllNodesInfo ?

ncharles:
This is a much welcomed change !

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -111,84 +111,97 @@ object NodeInfoServiceImpl {
...
-                       , ldapEntries.machineObjectClasses
-                     )
-    } yield {
-      nodeInfo
-    }
-  }
+  /*
+   * Our cache
+   */
+  private[this] var nodeCache = Option.empty[Map[NodeId, (LDAPNodeInfo, NodeInfo, Node)]]
```
VinceMacBuche: Why not None ? and typing the node cache ? (if needed)

fanf: It's exactly the same as:
private[this] var nodeCache: Option[Map[NodeId, (LDAPNodeInfo, NodeInfo, Node)]] = None

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -196,54 +209,113 @@ class NodeInfoServiceImpl(
       val machineInventories = MutMap[String, LDAPEntry]() // machine_dn -> entry
 
       val n1 = System.currentTimeMillis
```
fanf: n1 is actually a very bad name (I'm used to "tN"). I don't like too much long variable name for short leaving variable, they tend to clutter the code and make it less readable (it's the same for function like map/filter on collection)

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -196,54 +209,113 @@ class NodeInfoServiceImpl(
...
+    val t0 = System.currentTimeMillis
+
+    val boxInfo = (if(nodeCache.isEmpty || !isUpToDate(lastModificationTime)) {
+      getDataFromBackend match {
+        case Full((info, lastModif)) =>
+          nodeCache = Some(info)
+          lastModificationTime = lastModif
+          Full(info)
+        case eb: EmptyBox =>
+          nodeCache = None
```
fanf: no particular reason, I'm going to add it

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -111,84 +111,97 @@ object NodeInfoServiceImpl {
...
+          , "1.1"
+        )
+      } match {
+        case Full(seq) =>
+          //filter DN only in ou=Nodes,ou=Accepted Inventories,ou=Inventories,cn=rudder-configuration
+          // or in ou=Nodes,cn=rudder-configuration
+          val res = !seq.exists { e => e.dn.getParent == nodeDit.NODES.dn || e.dn.getParent == inventoryDit.NODES.dn }
+          logger.trace(s"Cache check for node info gave '${res}' (${seq.size} entry returned)")
+          res
+        case _        => false
```
fanf: Not a warning, because if it's because the server is down, there will be already other messages (from connection pool and other) and it will be just noise. But a debug message to understand why the cache is not use, why not.

fanf:
Given the remarks, I'm self-merging it.

PR #772, opened 2015-01-09
Fixes #6090: Node breakdown pie chart on homepage may be of different size

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6090

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/homePage.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/homePage.js
@@ -85,57 +85,93 @@ function homePage (
   $('#nodeCompliance').append(chart.element);
 }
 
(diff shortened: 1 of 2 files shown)


fanf:
I'm not sure about that correction because now, even if the are of the same size, they are not aligned as soon as the number of line of legend is not the same on all nodes, and the result is very strange.

I looked and didn't find any option to give the donut OR the legend a fixed heigh. Sad. So, it would be nice to look if a better solution is needed (else, we will habe to choose between the two bad ones).

VinceMacBuche:
@fanf updated with new idea (two graphs, one withou legend, one with only legend)

fanf:
The result is perfect!

PR #773, opened 2015-01-09
Fixes #6114: Accept new node screen has some minor display issue

fanf:
The PR seems big, because I cleaned-up a little the template (indentation and remove comments that were either outdated or not bringing any interesting information given the template).

The actual changes are:
- the place where the button() js is called; 
- the addition of <div> for lift lazy-load snippet so that they don't interact with the div used for tabs.


ncharles:
great !

PR #774, opened 2015-01-12
Fixes #6116: IE bug: groups cannot be saved nor deleted

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6116


fanf:
For the posterity, the problem was that depending of the call site of displayQuery(), we had a Form in a Form, that rises amok on IE.

PR #778, opened 2015-01-14
Fixes #4359: It is overly slow to show the list of Pending Review or Pending Deployment even if the number is 0

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/4359


fanf:
Could you try to remove methods from services/repositories that were used to make thing slow ? 
I'm not sure there are stull used, and we should force the future us to use the quick way.

VinceMacBuche:
They are still used (only once when looking for one ochange request detailed info

fanf:
OK

fanf:
oups, bad button

PR #781, opened 2015-01-15
Fixes #6133: Performance of RuleCategory repository request are bad

fanf:
Massivelly reduce the number of LDAP request for anything related to rules/ruleCategories - even with a very low number of categories.


ncharles:
This is awesome !
But doesn't that happen on groups as well ?

PR #784, opened 2015-01-16
Fixes #6146: Rudder stores too much or too big cookies, causing white page with HTTP status: 413: FULL head

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6146

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
@@ -796,10 +796,10 @@ class RuleEditForm(
                 }
               }
             }
-          ) (reportsGridXml) ++ Script( JsRaw("""
-                  %s
+          ) (reportsGridXml) ++ Script( JsRaw(s"""
+                  ${FormatDetailsJSFunction}
                   var anOpen = [];
-                  var oTable = $('#reportsGrid').dataTable( {
+                  var oTable = $$('#reportsGrid').dataTable( {
                     "asStripeClasses": [ 'color1', 'color2' ],
                     "bAutoWidth": false,
                     "bFilter" : true,
@@ -807,8 +807,6 @@ class RuleEditForm(
                     "bLengthChange": true,
                     "sPaginationType": "full_numbers",
                     "bJQueryUI": true,
-                    "bStateSave": true,
-                    "sCookiePrefix": "Rudder_DataTables_",
                     "oLanguage": {
                       "sSearch": ""
                     },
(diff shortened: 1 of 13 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -1194,7 +1204,12 @@ class RuleEditForm(
                "sPaginationType": "full_numbers",
                "bJQueryUI": true,
                "bStateSave": true,
-               "sCookiePrefix": "Rudder_DataTables_",
+                    "fnStateSave": function (oSettings, oData) {
+                      localStorage.setItem( 'DataTables_reportsGrid', JSON.stringify(oData) );
+                    },
+                    "fnStateLoad": function (oSettings) {
+                      return JSON.parse( localStorage.getItem('DataTables_changeHistory') );
```
fanf: DataTables_reportsGrid no ? And I'm not sur it's really the one we want since it's already used in line 812 of RuleEditForm above (I can't find my way on all the table :/)

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -1440,7 +1455,12 @@ class RuleEditForm(
                "bLengthChange": true,
                "sPaginationType": "full_numbers",
                "bStateSave": true,
-               "sCookiePrefix": "Rudder_DataTables_",
+                    "fnStateSave": function (oSettings, oData) {
+                      localStorage.setItem( 'DataTables_${gridId}', JSON.stringify(oData) );
+                    },
```
fanf: I don't think we want one option by grid. The used will expect to have AT MOST one settings for each grid he sees - by for him, all "details" table are the same. 

In the same idea, all node grid details are the same, etc.

fanf: Hum, in fact gridId is a static value, isn't it ? Sorry for the noise

VinceMacBuche: Yes it's a static value

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
VinceMacBuche: Maybe we don't wan't to keep reports grid at all, and maybe should remove from here

VinceMacBuche:
@fanf Updated!

fanf:
Great!

PR #789, opened 2015-01-20
Fixes #6156: Cleanup expected Reports (branch 3.0)

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6156

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/system/DatabaseManager.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/system/DatabaseManager.scala
@@ -41,6 +41,7 @@ import org.springframework.jdbc.core.JdbcTemplate
 import net.liftweb.common.Loggable
 import com.normation.rudder.repository.ReportsRepository
 import com.normation.rudder.repository.UpdateExpectedReportsRepository
+import com.normation.utils.Control
 
 trait DatabaseManager {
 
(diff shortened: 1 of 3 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/system/DatabaseManager.scala
```diff
@@ -124,24 +124,44 @@ class DatabaseManagerImpl(
...
-         }
-         val nodeConfigError = b match {
-           case f:Failure => f
-           case Empty => Failure("An error occured while deleting old nodeConfig", Empty, Empty)
-         }
-         Failure(nodeConfigError.msg, nodeConfigError.exception, Full(reportError))
+     val expectedReports = expectedReportsRepo.deleteExpectedReports(date)
+
+     // Maybe we could use a reduce here
+     (reports, nodeConfigs, expectedReports) match {
```
VinceMacBuche: Here I can do that instead:

     Seq(reports, nodeConfigs, expectedReports).reduce[Box[Int]] {
       case (Full(n1),Full(n2)) => Full(n1+n2)
       case (Full(_),f:Failure) => f
       case (f:Failure,Full(_)) => f
       case (f1:Failure,f2:Failure) =>
         Failure(f1.messageChain,f1.exception,Full(f2))

       // dummy cases
       case (a,Empty) => a
       case (Empty,a) => a

     }

VinceMacBuche: I allow myself to do that because I prevent Empty case:


     val reports = reportsRepository.deleteEntries(date) ?~! "An error occured while deleting reports"
     val nodeConfigs = expectedReportsRepo.deleteNodeConfigIdInfo(date) ?~! "An error occured while deleting reports"
     val expectedReports = expectedReportsRepo.deleteExpectedReports(date) ?~! "An error occured while old expected reports"

VinceMacBuche: @fanf What do you think of it ?

fanf: what about:
Control.bestEffort(Seq(reports ?~! "...", nodeConfigs ?~! "...", expectedReports ?~! "...."))(identity) ?

VinceMacBuche: And like that ? 


     (Control.bestEffort(Seq(reports, nodeConfigs, expectedReports)) (a => a)).map(_.sum)

PR #794, opened 2015-01-21
Fixes #6165: In rule details->compliance, reload should reload all graphes at once

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6165


fanf:
With that correction, you still have the two reload buttons on grid. 
I'm not sur it's what we want. More likelly, we just need one reload button. But I don't know where to put it, and moreover, with several button we are always able to click on them. 
So, perhaps add one other for recent change (between the title orange line and the graphe) ?

VinceMacBuche:
@fanf updated! There is room for improvement, but what do you think of it ?

fanf:
That seems far better, but could the "refresh" on compliance also refresh last changes, and "refresh" on last change also refresh compliance ?

VinceMacBuche:
All refresh button refresh al data ? That can be done without problem and would even be easier

VinceMacBuche:
@fanf updated once more :)

fanf:
Perfect!

PR #796, opened 2015-01-22
Bug 6021/pending state is not managed as user could expect in changes only mode

fanf:
Both compliance mode works in a similar way regarding pending states,
most notably now, in changes only, you get a pending state that is
expected to be updated in the agent run-lenght+grace period interval
for a given node.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
@@ -1,0 +1,121 @@
+/*
+*************************************************************************************
+* Copyright 2015 Normation SAS
+*************************************************************************************
+*
+* This program is free software: you can redistribute it and/or modify
+* it under the terms of the GNU Affero General Public License as
+* published by the Free Software Foundation, either version 3 of the
+* License, or (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU Affero GPL v3, the copyright holders add the following
+* Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU Affero GPL v3
+* licence, when you create a Related Module, this Related Module is
+* not considered as a part of the work and may be distributed under the
+* license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/agpl.html>.
+*
+*************************************************************************************
+*/
+
+package com.normation.rudder.domain.logger
+
+import org.slf4j.LoggerFactory
+import net.liftweb.common.Logger
+import com.normation.rudder.services.reports._
+import com.normation.rudder.repository.NodeConfigIdInfo
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.reports.ComplianceMode
+import com.normation.rudder.reports.ResolvedAgentRunInterval
+import com.normation.rudder.reports.ComplianceMode
+import com.normation.rudder.reports.FullCompliance
+import com.normation.rudder.reports.ResolvedAgentRunInterval
+import com.normation.rudder.reports.ChangesOnly
+
+
+/**
+ * Log information about compliance.
+ * This log is intended to be used in debug & trace level.
+ * It allows to get trace explaining why the compliance is what it
+ * is.
+ */
+object ComplianceDebugLogger extends Logger {
+  override protected def _logger = LoggerFactory.getLogger("explain_compliance")
+
+
+  implicit class NodeConfigIdInfoToLog(n: NodeConfigIdInfo) {
+    val toLog: String = s"${n.configId.value}/[${n.creation}-${n.endOfLife.fold("now")(_.toString)}]"
+  }
+
+  implicit class RunAndConfigInfoToLog(c: RunAndConfigInfo) {
+    val toLog: String = c match {
+      case NoRunNoInit =>
+        "[NoRunNoInit: expected NodeConfigId: not found | last run: not found]"
+
+      case VersionNotFound(lastRunDateTime, lastRunConfigId) =>
+         s"[VersionNotFound: expected NodeConfigId: not found |"+
+         s" last run: nodeConfigId: ${lastRunConfigId.fold("none")(_.value)} received at ${lastRunDateTime}]"
+
+      case NoReportInInterval(expectedConfigId) =>
+         s"[NoReportInInterval: expected NodeConfigId: ${expectedConfigId.toLog}|"+
+         s" last run: none available (or too old)]"
+
+      case Pending(expectedConfigId, optLastRun, expirationDateTime, missingStatus) =>
+        s"[Pending: until ${expirationDateTime} expected NodeConfigId: ${expectedConfigId.toLog} |"+
+        s" last run: ${optLastRun.fold("none (or too old)")(x => s"nodeConfigId: ${x._2.toLog} received at ${x._1}")}]"
+
+      case UnexpectedVersion(lastRunDateTime, lastRunConfigId, lastRunExpiration, expectedConfigId, expectedExpiration) =>
+        s"[UnexpectedVersion: expected NodeConfigId: ${expectedConfigId.toLog} |"+
+        s" last run: nodeConfigId: ${lastRunConfigId.toLog} received at ${lastRunDateTime} |"+
+        s" expire at ${lastRunExpiration}]"
+
+      case ComputeCompliance(lastRunDateTime, lastRunConfigId, expectedConfigId, expirationDateTime, missingStatus) =>
+        s"[CheckChanges: expected NodeConfigId: ${expectedConfigId.toLog} |"+
+        s" last run: nodeConfigId: ${lastRunConfigId.toLog} received at ${lastRunDateTime} |"+
+        s" expire at ${expirationDateTime}]"
+
+    }
+
+    val logName =  c match {
+      case    NoRunNoInit       => "NoRunNoInit"
+      case _: VersionNotFound   => "VersionNotFound"
+      case _: NoReportInInterval=> "NoReportInInterval"
+      case _: Pending           => "Pending"
+      case _: UnexpectedVersion => "UnexpectedVersion"
+      case _: ComputeCompliance => "ComputeCompliance"
+    }
+  }
+
+  implicit class AgentRunConfigurationToLop(map: Map[NodeId, (ComplianceMode, ResolvedAgentRunInterval)]) {
+
+    private[this] def log(c: ComplianceMode, r: ResolvedAgentRunInterval): String = {
+      val h = c match {
+        case FullCompliance => ""
+        case ChangesOnly(_) => s", hearbeat every ${r.heartbeatPeriod} run(s)"
+      }
+      s"run interval: ${r.interval.toStandardMinutes.getMinutes} min${h}"
+    }
+
+    val toLog: String = map.map { case (id, (c, r)) =>
+      s"[${id.value}:${c.name}, ${log(c, r)}]"
+    }.mkString("")
+  }
+
+
+}
+
+
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
@@ -44,8 +44,25 @@ import com.normation.inventory.domain.NodeId
 
 
 /**
+ * The main class helping maps expected
+ * reports for a node
+ */
+case class RuleNodeExpectedReports(
+    ruleId    : RuleId
+  , serial    : Int // the serial of the rule
+  , directives: Seq[DirectiveExpectedReports]
+  // the period where the configuration is applied to the servers
+  , beginDate : DateTime = DateTime.now()
+  , endDate   : Option[DateTime] = None
+) extends HashcodeCaching {
+  val interval = new Interval(beginDate, endDate.getOrElse(DateTime.now))
+}
+
+/**
  * The representation of the database object for the expected reports for
- * a rule and a list of nodes
+ * a rule and a list of nodes.
+ *
+ * Its use is discouraged in favor of RuleNodeExpectedReports
  *
  */
 case class RuleExpectedReports(
@@ -69,6 +86,7 @@ case class DirectivesOnNodes(
   , nodeConfigurationIds    : Map[NodeId, Option[NodeConfigId]]
   , directiveExpectedReports: Seq[DirectiveExpectedReports]
 ) extends HashcodeCaching
+
 
 
 /**
--- a/rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
@@ -37,75 +37,106 @@ package com.normation.rudder.reports
 import com.normation.utils.HashcodeCaching
 import net.liftweb.json.JBool
 import net.liftweb.common._
+import org.joda.time.Duration
+import com.normation.inventory.domain.NodeId
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.services.nodes.NodeInfoService
+import com.normation.rudder.domain.Constants
 
 /**
  * Class that contains all relevant information about the reporting configuration:
  * * agent run interval
  * * compliance mode (soon)
  * * heartbeat configuration
+ *
+ * Everything is option because that information is at the node level and
+ * may not be defined (for example if the node only use default global values)
+ *
+ * For the node "resolved reportingconfiguration", see:
  */
-final case class ReportingConfiguration (
-    agentRunInterval      : Option[AgentRunInterval]
-  , heartbeatConfiguration: Option[HeartbeatConfiguration]
-) extends HashcodeCaching
+final case class ReportingConfiguration(
+  agentRunInterval: Option[AgentRunInterval], heartbeatConfiguration: Option[HeartbeatConfiguration]) extends HashcodeCaching
 
-final case class HeartbeatConfiguration (
-   overrides : Boolean
- , heartbeatPeriod : Int
-)
+final case class HeartbeatConfiguration(
+  overrides: Boolean, heartbeatPeriod: Int)
 
-final case class AgentRunInterval (
-    overrides  : Option[Boolean]
-  , interval   : Int
-  , startMinute: Int
-  , startHour  : Int
-  , splaytime  : Int
-) extends HashcodeCaching {
+final case class AgentRunInterval(
+  overrides: Option[Boolean], interval: Int //in minute
+  , startMinute: Int, startHour: Int, splaytime: Int) extends HashcodeCaching {
 
- def json() = {
+  def json() = {
 
-   val overrideValue = overrides.map(_.toString).getOrElse("null")
-   val splayHour = splaytime / 60
-   val splayMinute = splaytime % 60
+    val overrideValue = overrides.map(_.toString).getOrElse("null")
+    val splayHour = splaytime / 60
+    val splayMinute = splaytime % 60
 
-   s"""{ 'overrides'   : ${overrideValue}
+    s"""{ 'overrides'   : ${overrideValue}
        , 'interval'    : ${interval}
        , 'startHour'   : ${startHour}
        , 'startMinute' : ${startMinute}
        , 'splayHour'   : ${splayHour}
        , 'splayMinute' : ${splayMinute}
        }"""
- }
+  }
 }
+
+final case class ResolvedAgentRunInterval(
+  interval: Duration, heartbeatPeriod: Int)
 
 trait AgentRunIntervalService {
-  def getAgentRun : Box[AgentRunInterval]
+  def getGlobalAgentRun(): Box[AgentRunInterval]
+
+  /**
+   * For each node Id passed as argument, find the corresponding
+   * run interval and heartbeat value (using global default if needed)
+   */
+  def getNodeReportingConfigurations(nodeIds: Set[NodeId]): Box[Map[NodeId, ResolvedAgentRunInterval]]
+
 }
 
-class AgentRunIntervalServiceImpl (
-    readOverrides : () => Box[Option[Boolean]]
-  , readInterval  : () => Box[Int]
-  , readStartHour  : () => Box[Int]
-  , readStartMinute  : () => Box[Int]
-  , readSplaytime  : () => Box[Int]
-) extends AgentRunIntervalService {
+class AgentRunIntervalServiceImpl(
+  nodeInfoService: NodeInfoService, readGlobalInterval: () => Box[Int], readGlobalStartHour: () => Box[Int], readGlobalStartMinute: () => Box[Int], readGlobalSplaytime: () => Box[Int], readGlobalHeartbeat: () => Box[Int]) extends AgentRunIntervalService {
 
-  override def getAgentRun : Box[AgentRunInterval] = {
+  override def getGlobalAgentRun(): Box[AgentRunInterval] = {
     for {
-      overrides   <- readOverrides()
-      interval    <- readInterval()
-      startHour   <- readStartHour()
-      startMinute <- readStartMinute()
-      splaytime   <- readSplaytime()
+      interval <- readGlobalInterval()
+      startHour <- readGlobalStartHour()
+      startMinute <- readGlobalStartMinute()
+      splaytime <- readGlobalSplaytime()
     } yield {
       AgentRunInterval(
-          overrides
-        , interval
-        , startHour
-        , startMinute
-        , splaytime
+        None, interval, startHour, startMinute, splaytime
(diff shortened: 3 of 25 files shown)


rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
```diff
@@ -0,0 +1,112 @@
+/*
+*************************************************************************************
+* Copyright 2012 Normation SAS
```
ncharles: Wrong year

rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
```diff
@@ -0,0 +1,112 @@
...
+import com.normation.rudder.reports.ResolvedAgentRunInterval
+import com.normation.rudder.reports.ChangesOnly
+
+
+/**
+ * Log information about compliance.
+ * This log is intended to be used in debug & trace level.
+ * It allows to get trace explaining why the compliance is what it
+ * is.
+ */
```
ncharles: This is awesome \o/

rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
```diff
@@ -37,12 +37,22 @@ package com.normation.rudder.reports
...
+import com.normation.rudder.domain.Constants
 
 /**
  * Class that contains all relevant information about the reporting configuration:
  * * agent run interval
  * * compliance mode (soon)
  * * heartbeat configuration
+ *
+ * Everything is option because that information is at the node level and
+ * can not be defined (for example if the node only use default global values)
```
ncharles: Could you change "can" for "may"; at first i thought it was illegal to define it

rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
```diff
@@ -107,5 +130,32 @@ class AgentRunIntervalServiceImpl (
...
+
+  override def getNodeReportingConfigurations(nodeIds: Set[NodeId]): Box[Map[NodeId, ResolvedAgentRunInterval]] = {
+    for {
+      gInterval  <- readGlobalInterval()
+      gHeartbeat <- readGlobalHeartbeat()
+      nodeInfos  <- nodeInfoService.getAll()
+    } yield {
+      nodeIds.map { nodeId =>
+
+        if(nodeId == Constants.ROOT_POLICY_SERVER_ID) {
```
ncharles: it should be also every node with a server role

rudder-core/src/main/scala/com/normation/rudder/repository/ExpectedReportsRepository.scala
```diff
@@ -118,14 +117,27 @@ trait FindExpectedReportRepository {
...
+   *
+   * Box[  // the method can fails
+   *   Map[NodeId, Option[ // for each node on the query, either return None if the node expected reports were not found or the couple:
+   *      (NodeConfigId,   // the node config id for the expected reports
+   *        Map[(RuleId, Int), DirectiveExpectedReports]) // and the expected reports by rule/serial
+   *   ]]
+   * ]
    */
-  def getLastExpectedReports(nodeIds: Set[NodeId], filterByRules: Set[RuleId]): Box[Set[RuleExpectedReports]]
+//   def getLastExpectedReports(nodeIds: Set[NodeId], filterByRules: Set[RuleId]): Box[Map[NodeId, Option[(NodeConfigId, Map[SerialedRuleId, RuleNodeExpectedReports])]]]
```
ncharles: could you remove this commented function ? or is there a reason for it ?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -225,16 +232,148 @@ class FindExpectedReportsJdbcRepository(
...
+        }
+      }
 
-      for {
-        entries <- tryo ( jdbcTemplate.query(query, ReportAndNodeMapper).asScala )
+
+      //to get the actual query as passed to pg
+      //println(query_subscript_configids)
+      //println(query_subscript_configids2)
+
```
ncharles: could you remove these commented lines?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -193,31 +212,19 @@ class FindExpectedReportsJdbcRepository(
...
-       val query = s"""select
-            E.pkid, E.nodejoinkey, E.ruleid, E.directiveid, E.serial, E.component, E.componentsvalues
-          , E.unexpandedComponentsValues, E.cardinality, E.begindate, E.enddate
-          , NNN.nodeid, NNN.nodeconfigids
+      /**
+       * Trying to reduce the number of returned rows by the query.
+       * But it does make so little difference on the quantity of data transfered
+       * that it does not matter.
+       */
+      val query_subscript_configids_3 = s"""select X.pkid, array_agg(array_to_string(X.nodeid || X.nodeconfigids, ',')) as nodeconfigs from (
```
ncharles: here you kept unuased sql query; is that voluntary ? 
It make the code harder to read, since the name of the strings are very similar

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -225,16 +232,148 @@ class FindExpectedReportsJdbcRepository(
...
+      val t0 = System.currentTimeMillis()
+
+      val res = for {
+        reportsAndConfigIds <- tryo ( jdbcTemplate.query(query_subscript_configids, ReportAndConfigIdMapper).asScala )
+//        reportsAndConfigIds <- tryo ( jdbcTemplate.query(query_subscript_configids_3, ReportAndConfigIdMapper2).asScala.flatten )
+        t1                  =  System.currentTimeMillis
+        _                   =  TimingDebugLogger.debug(s"GetExpectedReports: configIds: ${t1-t0}ms")
+        reportIds           =  Set[Int]() ++ reportsAndConfigIds.map( _._1)
+        t2                  =  System.currentTimeMillis
+        _                   =  TimingDebugLogger.debug(s"GetExpectedReports: map reportsIds: ${t2-t1}ms")
```
ncharles: is it interesting to log the time to transform the result into a set of Int ?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -225,16 +232,148 @@ class FindExpectedReportsJdbcRepository(
...
+      val query_subscript_configids2 = s"""select
+            E.pkid, NNN.nodeid, NNN.nodeconfigids
+        from expectedreports E
+        inner join expectedreportsnodes NNN
+        on E.nodejoinkey = NNN.nodejoinkey
+        where  NNN.nodeconfigids && Array[${nodeConfigIds.map(_.version.value).mkString("'", "','", "'")}]
+       ${if(filterByRules.isEmpty) "" else " and " + in("ruleid", filterByRules.map(_.value))}"""
+
+
+      def getExpectedReports(reportIds: Set[Int]): Box[Seq[ReportMapping]] = {
```
ncharles: Inception of getExpectedReports !
could you rename one of the two getExpectedReports method please ? It makes it hard to find which one is called below

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -225,16 +232,148 @@ class FindExpectedReportsJdbcRepository(
...
+
+      res
+    }
+  }
+
+  object ReportAndConfigIdMapper extends RowMapper[(Int, NodeId, List[NodeConfigId])] {
+    def mapRow(rs : ResultSet, rowNum: Int) : (Int, NodeId, List[NodeConfigId]) = {
+      def notNullS[T](name: String): String = {
+        rs.getString(name) match {
+          case null => throw new IllegalArgumentException(s"Column '${name}' is null (illegal value)")
```
ncharles: "Column '${name}' has a null value (illegal value)" maybe ?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -225,16 +232,148 @@ class FindExpectedReportsJdbcRepository(
...
+      )
+    }
+  }
+
+  /**
+   * A specialized version for the query where nodeid is aggredated to the mkString of nodeconfigid.
+   */
+  object ReportAndConfigIdMapper2 extends RowMapper[Seq[(Int, NodeId, List[NodeConfigId])]] {
+    private[this] def unserialize(versions: java.sql.Array): Seq[(NodeId, List[NodeConfigId])] = {
+      if(null == versions) throw new IllegalArgumentException(s"ERROR")
```
ncharles: if there is no previous value (in case of migration), it the array null or is it empty ?

ncharles: could tou return a more meaningful error message?
IllegalArgumentException(s"ERROR: Null nodeconfigsids")

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -295,10 +434,67 @@ class FindExpectedReportsJdbcRepository(
...
+  /**
+   * Group a list of reports by node and rule
+   * On the resulting map, we only have nodeId for nodeId with actual node config.
+   * The returned nodeConfigId is the most recent available.
+   */
+  private[this] def toNodeExpectedReports(nodeConfigIds: Set[NodeAndConfigId], entries:Seq[ReportAndNodeMapping]) : Map[NodeId, Map[NodeConfigId, Map[SerialedRuleId, RuleNodeExpectedReports]]] = {
+
+    val t0 = System.currentTimeMillis
+
+    val nodeIds = nodeConfigIds.groupBy { _.nodeId }.mapValues { _.toSeq.map(_.version) }
```
ncharles: the name nodeIds is misleading, as it contains nodeIds -> Seq[NodeConfigIds]

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -295,10 +434,67 @@ class FindExpectedReportsJdbcRepository(
...
+    TimingDebugLogger.debug(s"GetExpectedReports: toNodeExpectedReports: groupBy: ${t1-t0}ms")
+
+    var t_byConfigId = 0l
+    var n = 0
+
+    val res = nodeIds.map { case (nodeId, configIds) =>
+      val tx_0 = System.currentTimeMillis
+
+      val byConfigId = configIds.map { configId =>
+        val reportsForId = reportsForNode.getOrElse(nodeId, Seq()).filter { _.nodeConfigVersions.exists { _ == configId } }
```
ncharles: I know that i'm a bit slow today, but a comment on what this algorithm do would help the future maintainer

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -758,6 +968,44 @@ object ReportAndNodeMapper extends RowMapper[ReportAndNodeMapping] {
...
+object ReportMapper extends RowMapper[ReportMapping] {
+
+  //handler to raise exception on null: we DON'T have better means here,
+  //so be sure to handle the exception above
+
+
+  def mapRow(rs : ResultSet, rowNum: Int) : ReportMapping = {
+    def notNullS[T](name: String): String = {
+      rs.getString(name) match {
+        case null => throw new IllegalArgumentException(s"Column '${name}' is null (illegal value)")
```
ncharles: same coment as above Column '${name}' has a null value (illegal value)

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -44,6 +44,22 @@ import com.normation.inventory.domain.NodeId
 
 
 /**
+ * Remark on above: RuleExpectedReport does not
+ * map to anything usefull today. We would like to
+ * have something like:
+ */
+case class RuleNodeExpectedReports(
```
VinceMacBuche: I am not very fan of the name as it has no links with Node, and I think it may be hard to maintain with time. (very confusing having two class with almost the same name)

Is it possible to remove "RuleExpectedReports" and rename it or is it still used somehow ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -260,126 +294,151 @@ object ExecutionBatch extends Loggable {
...
 
-      nodeIds.map { case (nodeId, graceDuration) => (nodeId, {
+    /**
+     * Utility method to factor out common logging task and be assured that
+     * the log message is actually sync with the info type.
+     */
+    def runType(traceMessage: String, info: RunAndConfigInfo)(implicit nodeId: NodeId): RunAndConfigInfo = {
+      val msg = if(traceMessage.trim.size == 0) "" else ": " + traceMessage
+      val x = info.getClass.getCanonicalName
+      val name = x.split('.').lastOption.getOrElse(x)
```
ncharles: Wouldn't it be better to define a human name in each class, and use it ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -521,30 +565,48 @@ object ExecutionBatch extends Loggable {
    * only can be missing, and if we have reports for other directives,
    * they are unexpected.
    *
+   * We have a to level comparison to do:
```
ncharles: a two level

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -583,15 +645,43 @@ object ExecutionBatch extends Loggable {
...
+    //now, for all current expected reports, choose between the computed value and the default one
+
+    // note: isn't there something specific to do for unexpected reports ? Keep them all ?
+
+    val nil = Seq[RuleNodeStatusReport]()
+    val (computed, newStatus) = ((nil, nil)/: currentExpectedDirectives) { case ( (c,n), (k, expectedReport)) =>
+      complianceForRun.get(k) match {
+        case None => //the whole rule is new!
+          //here, the reports are ACTUALLY pending, not missing.
+          val x = buildRuleNodeStatusReport(mergeInfo, Map(k -> expectedReport), PendingReportType)
```
ncharles: i'm not sure i understand here the k, and why the reports are pendings
But does that mean that the pending state will never expire is we don't receive reports ?

ncharles:
This is a welcome change, thank you very much !
I asked some question, certain very minor, other more complex, but they won't change the overall logic

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: The expiration status is at the report level, not individual component, because expiration is linked to a not sending report or not. 

That last part of the algorithm is managing things (rule for now, directive in the future) whose serial changed since the last run. If serial changed, it means config was updated, and since we dealt with all the macro status (unexpect version, etc) before, the status is pending and not missing or success. 

k is just the name of the map _k_eys (don't care much of what they are, ruleAndserialId certainly)

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
fanf: I DOES have link with node, it's the expected reports for a given node - or at least the intent is that, contrary to expected reports for a rule, and that's the reason why there is no nodeId/configId on it. 
The name was chosen to match the corresponding compliance report (RuleNodeStatusReport). 

I believe we can't remoce RuleExpectedReport for now because of some method that still use it (for plugin?). I could be renamed (DeprecatedExpectedReportForRule ?)

rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
fanf: I'm not sure it's what is done on the promise side: https://github.com/Normation/rudder/blob/branches/rudder/3.0/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala#L285

if(nodeInfo.isPolicyServer) {....

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
fanf: This mapper won't be user as the benefits are null, it will just be removed with the surnumerary sql queries.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: why not

rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
ncharles: iit is at least done for all policy server

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
ncharles: Or maybe simply a comment saying it is deprecated ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
ncharles: ok, this makes more sense

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
ncharles: ok :)

rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
fanf: What is the correct thing to do ?

ncharles: Le 24 janvier 2015 15:04:22 GMT+01:00, Francois Armand <notifications@github.com> a écrit :
>> @@ -107,5 +130,32 @@ class AgentRunIntervalServiceImpl (
>>        )
>>      }
>>    }
>> +
>> +  override def getNodeReportingConfigurations(nodeIds: Set[NodeId]):
>Box[Map[NodeId, ResolvedAgentRunInterval]] = {
>> +    for {
>> +      gInterval  <- readGlobalInterval()
>> +      gHeartbeat <- readGlobalHeartbeat()
>> +      nodeInfos  <- nodeInfoService.getAll()
>> +    } yield {
>> +      nodeIds.map { nodeId =>
>> +
>> +        if(nodeId == Constants.ROOT_POLICY_SERVER_ID) {
>
>What is the correct thing to do ?
>
>---
>https://github.com/Normation/rudder/pull/796/files#r23496507

To be consistent, it should set the value to 5 is isPolicyServer is true
We'll see later for the server roles
-- 
Envoyé de mon téléphone Android avec K-9 Mail. Excusez la brièveté.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -260,126 +294,149 @@ object ExecutionBatch extends Loggable {
...
     }
 
-      nodeIds.map { case (nodeId, graceDuration) => (nodeId, {
+    /**
+     * Utility method to factor out common logging task and be assured that
+     * the log message is actually sync with the info type.
+     */
+    def runType(traceMessage: String, info: RunAndConfigInfo)(implicit nodeId: NodeId): RunAndConfigInfo = {
+      val msg = if(traceMessage.trim.size == 0) "" else ": " + traceMessage
+      ComplianceDebugLogger.trace(s"Run cfonfig for node ${nodeId.value}: ${info.logName} ${msg}")
```
ncharles: samll typo here - config

ncharles:
This look great ! From my point of view, you should rebase and selfmerge; and we'll need to start testing in depth

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -44,8 +44,25 @@ import com.normation.inventory.domain.NodeId
 
 
 /**
+ * The main class helping maps expected
+ * reports for a node
+ */
+case class RuleNodeExpectedReports(
```
VinceMacBuche: In fact what's bothering me is that there is no reference to a Node in that structure, and we say that it is related to a Node, meaning it is something external that defines it and we have no way to check it (and maybe that's good, but yet that makes me tickle)

ncharles: actually, when you drill down to the DirecteiveExpectedReports, you have nodes

VinceMacBuche: Then I find even more strange that we don't have the information at Rule level

VinceMacBuche: From what i have experienced when dealing with compliance (mostly as display) I always need to have as much informations in top level

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -89,6 +90,14 @@ class PostgresqlInClause(
     //use IN ( VALUES (), (), ... )
     else s"${attribute} IN(VALUES ${values.mkString("('","'),('","')")})"
   }
+
+  def inNumber(attribute: String, values: Iterable[AnyVal]): String = {
+    //with values, we need more ()
+    if(values.isEmpty) ""
+    else if(values.size < inClauseMaxNbElt) s"${attribute} IN (${values.mkString("", ",", "")})"
```
VinceMacBuche: The problem here is that VALUES have a max number of elems ?

VinceMacBuche: Misread, It's "IN" clause that have a limit value ?

fanf: As explain in the comment above, it's an optimization with postgres interval behaviour.

VinceMacBuche: Ha ok, make sense when opening it :)

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
fanf: It's a datastructure above node. In the same idea that components don't have reference to directive, that datastracture does not have reference to node. 
The actual full structure at top level looks like: Map[NodeId, Map[NodeConfigId, Set[RuleNodeExpectedReports]]

So, we could add VersionnedNodeExpectedReports with a config Id, but I didn't feel the need in the code. Perhaps it's a fault.

VinceMacBuche: Ok, I understand the structure, that they are above the Node, and that expected reports are now completely "by node". 

I thinks it's the naming that confuses me ( and i would rename it RuleExpectedReports, but it still exists and it's goot that it creates a link with RuleNodeStatusReports) So maybe I'll be happier with a refactoring, but I think it's not the place to do it right now

The full datastrucuture is complicated with its 4 level map (there is one more level RuleSerialedId) And i fear it might be hard to maintain ( like in 3 months ;) ), but i have really NO idea of how to do it better.

rudder-web/src/main/webapp/templates-hidden/common-layout.html
```diff
@@ -208,7 +208,7 @@ <h4 class="modal-title">Participate in usage survey?</h4>
     </div>
   <hr class="spacer" />
 <div id="footer">
-  <p>Rudder v2.3 - &copy; <a href="http://www.normation.com/">Normation</a></p>
+  <p>Rudder v3.0 - &copy; <a href="http://www.normation.com/">Normation</a></p>
```
VinceMacBuche: Yeah \o/

VinceMacBuche:
This is great to me! Can be merged to be tested intensively!

rudder-web/src/main/webapp/templates-hidden/common-layout.html
fanf: well, in fact that does not bring anything for now safe potential pb with integration. In all case, it should not be done siliently here (was a mistake: forgotten test).

fanf:
Self merging as requesting by other.

PR #799, opened 2015-01-23
Fixes #6176: cannot use \"enabled\" json parameter in Directive API

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6176

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
@@ -573,7 +573,7 @@ case class RestExtractorService (
       name             <- extractOneValueJson(json, "displayName")(convertToMinimalSizeString(3))
       shortDescription <- extractOneValueJson(json, "shortDescription")()
       longDescription  <- extractOneValueJson(json, "longDescription")()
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -573,7 +573,7 @@ case class RestExtractorService (
       name             <- extractOneValueJson(json, "displayName")(convertToMinimalSizeString(3))
       shortDescription <- extractOneValueJson(json, "shortDescription")()
       longDescription  <- extractOneValueJson(json, "longDescription")()
-      enabled          <- extractOneValueJson(json, "enabled")( convertToBoolean)
+      enabled          <- extractJsonBoolean(json, "enabled")
```
ncharles: doesn't it prevent it from using it outside the json ?

VinceMacBuche: json and 'params' are treated separetely (look  line 547)

ncharles: oh, ok !

PR #800, opened 2015-01-27
Fixes #6187: Some tables are broken in compliance popup and inventory 

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6187

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
@@ -1412,6 +1412,7 @@ class RuleEditForm(
                   } );
                 }
           } );} );""".format(tabid, gridId + "Grid", S.contextPath, innerJsFun)
+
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -1419,14 +1420,14 @@ class RuleEditForm(
              * Everything is based what has been done for the previous dataTable
              */
         Script(JsRaw(s"""
-            function fnFormatDetail${tabid} ( oTable, nTr ) {
+            function fnFormatDetails${tabid} ( oTable, nTr ) {
```
ncharles: this method is already defined on line 636; doesn't it cause any issue ?

VinceMacBuche: They are on a different scope so there is no problem.

Plus they are not named the same ;) on 636its only fnFormatDetails, here is if suffixed by ${tabid} (0 1 2 3 ...)

PR #802, opened 2015-01-27
Fixes #6190: Refreshing rules grid make them all disappear

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6190


fanf:
Good catch!

PR #804, opened 2015-01-27
Fixes #6184: Add a cache for recent changes

fanf:
The code seems to work, but one graphe does not go to the edge of its div in the rule grid... 
Something seems broken in the displaying :(

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
@@ -33,33 +33,114 @@
 */
 
 package com.normation.rudder.services.reports
+
+import scala.concurrent._
+import scala.concurrent.ExecutionContext.Implicits.global
+
+
+
 import org.joda.time.DateTime
-import com.normation.rudder.domain.reports.ResultRepairedReport
-import net.liftweb.common.Box
-import com.normation.rudder.repository.ReportsRepository
 import org.joda.time.Interval
-import com.normation.rudder.domain.reports.ResultRepairedReport
-import net.liftweb.common.Loggable
-import com.normation.rudder.domain.reports.ResultRepairedReport
+
 import com.normation.rudder.domain.policies.RuleId
-import org.joda.time.format.PeriodFormatterBuilder
+import com.normation.rudder.domain.reports.ResultRepairedReport
+import com.normation.rudder.repository.CachedRepository
+import com.normation.rudder.repository.ReportsRepository
+
+import net.liftweb.common._
+import net.liftweb.common.Box
+import net.liftweb.common.Loggable
 import net.liftweb.http.js.JE._
+
 
 /**
  * This service is responsible to make available node changes for nodes.
- * The hard part is on implementation, to make it efficient.
+ *
+ * The changes are grouped by rules and by the intervals of time.
+ * The intervals of time are always the 4 following, for each days:
+ * - [0h-6h[ , [6h-12h[ , [12h-18h[, [18h-24[.
  */
 trait NodeChangesService {
 
-  /**
-   * Get all changes until a date specified as parameter
-   */
-  def getChanges(notBefore: Option[DateTime] = None): Box[Seq[ResultRepairedReport]]
+  type ChangesByRule  = Map[RuleId, Map[Interval,Seq[ResultRepairedReport]]]
 
   /**
-   * Get all changes until a date specified as parameter and regroup them by interval of hours
+   * Get all changes since a date specified as parameter and regroup them by interval of hours.
+   *
+   * If no starting date is specified, a default duration will be used to compute it.
+   *
    */
-  def getChangesByInterval(intervalLength : Int = 6, notBefore: Option[DateTime]  = None) : Box[Map[Interval,Seq[ResultRepairedReport]]]
+  def getChangesByInterval(since: Option[DateTime]) : Box[ChangesByRule]
+
+
+  /**
+   * For the given service, what are the intervals currently valid?
+   * Optionnally give a starting date, or else the service is in charge
+   * to provide a default value.
+   */
+  def getCurrentValidIntervals(since: Option[DateTime]): Seq[Interval]
+
+  /*
+   * The two following methods are pure algorithm methods and we want that all
+   * implementation of getChangesByInterval use them.
+   */
+
+  /**
+   * Group the given set of reports by rules and based on the allowed intervals
+   * given in parameter. All report outside any of the intervals is discared.
+   */
+  def filterGroupByInterval(intervals: Seq[Interval], changes: Seq[ResultRepairedReport]): ChangesByRule = {
+    changes.groupBy(_.ruleId).mapValues { ch =>
+      ( for {
+         interval <- intervals
+      } yield {
+        (interval,ch.filter(interval contains _.executionTimestamp))
+      } ).toMap
+    }
+  }
+
+
+  /**
+   * Stable computation of intervals of 6h starting at 0,6,12 or 18
+   * between two dates.
+   *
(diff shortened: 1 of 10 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -33,102 +33,268 @@
...
     // Compute intervals of changes recursively from the last interval lower
-    def computeIntervals(previousLowerBound : DateTime)  : List[Interval] = {
-      if ( previousLowerBound isBefore limit) {
+    def computeIntervals(previousLowerBound: DateTime): List[Interval] = {
+      if ( previousLowerBound isBefore since) {
         // Our last bound is before the limit, do not create a interval
         Nil
       } else {
-        val nextLowerBound = previousLowerBound.minusHours(intervalLength)
+        val nextLowerBound = previousLowerBound.minusHours(6)
```
ncharles: the value is now hardcoded ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -33,102 +33,268 @@
...
+   */
+  private[this] def initCache() : Box[ChangesByRule] = this.synchronized {
+    cache match {
+      case None =>
+        for {
+          changes <- changeService.getChangesByInterval(None)
+        } yield {
+          cache = Some(changes)
+          logger.debug("NodeChanges cache initialized")
+          logger.trace("NodeChanges cache content: " + cacheToLog(changes))
```
ncharles: cacheToLog looks expensive. Maybe put this in a if (logger.isTrace) ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -33,102 +33,268 @@
...
+
+  /**
+   * For intervals "on", merge change1 and change2 (respective to rules). Intervals not in "on"
+   * are removed from the result.
+   * For a given rule, on a given interval, changes fomr change1 and change2 are added.
+   */
+  private[this] def merge(on: Seq[Interval], changes1: ChangesByRule, changes2: ChangesByRule): ChangesByRule = {
+    //shortcut that get map(k1)(k2) and return an empty seq if key not found
+    def get(changes: ChangesByRule, ruleId: RuleId, i: Interval): Seq[ResultRepairedReport] = {
+      changes.getOrElse(ruleId, Map()).getOrElse(i, Seq())
```
ncharles: isn't the interval of the on always changing ?
if so, how can the getOrElse(i, Seq()) return the proper value ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -33,102 +33,268 @@
...
+   */
+  def update(changes: Seq[ResultRepairedReport]): Box[Unit] = this.synchronized {
+    for {
+      existing <- initCache()
+    } yield {
+      val intervals = changeService.getCurrentValidIntervals(None)
+      val newChanges = this.filterGroupByInterval(intervals, changes)
+      logger.debug("NodeChanges cache updated")
+      cache = Some(merge(intervals, existing, newChanges))
+      logger.trace("NodeChanges cache content: " + cacheToLog(cache.get))
```
ncharles: again, here could you if (logger.isTrace) ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
fanf: It was already, but saying it wasn't. All computation were assuming 6h, or even more preciselly the 4 intervals [0h-6h[, [6h-12h[, [12h-18h[, [18h-24h[

So at least it's explicit, and if one day we want to change that, we DO know what are the hypothesis.

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
fanf: logger. trace is call by name, that won't change anything

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
fanf: Intervals are stable, I will make it more explicit on the trait "getInterval" methods. They always are [0h-6h[, [6h-12h[, [12h-18h[, [18h-24h[ for a given day. So they can actually be used as key for a map.

ncharles: oh, so we don't see the last 6 hours, but the last interval of 6 hours ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
ncharles: but isn't the string always computed ?

fanf: That would be against the definition of call by name, but I'm checking it

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
fanf: I think I don't understand. Trying to infer, please correct is badly done: 
If "now" is 13:17:53.947, then the most recent interval for changes is [12:00:00.000-18:00:00.000[, not [7:17:53.947-13:17:53.947[

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
fanf: No it's ok, the parameter is not evaluated if the log level is not in use.

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
ncharles: oh, ok. So that's relevant then

So how are "old" interval cleaned? (when time goes from 11:59 to 12:00, how the three day old interval get removed ?)

fanf: on the next update for these rules, the merge get the list of current valid interval and filter out other

fanf:
PR updated, also fixe the width of the graphe to 168px so that the glitch on size disappear. 168px seems to be a good value, it is ok on 1024px screen and not to tiny on HD one

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
ncharles: but what if the rule is never updated ? won't we have an evergrowing cache that will eat all the memory ?

fanf: no, you have a max value depending of the number of rules. But you can have old values not in use, it's not that important.

fanf: of course, the number of rules * number * intervals * (mean number of changes), not only rules. So you could have a degenerative case where on interval on a rule never updated after that consume a lot of memory. Seems sufficiently unusual to let it be managed by either "clean cache" or webapp restart. Rules are not so valite things in Rudder.

ncharles:
Great !

PR #805, opened 2015-01-28
Fixes #6194: Add a cach for compliance by node

fanf:


rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
```diff
@@ -178,6 +181,27 @@ class ReportsExecutionService (
       case scala.util.Failure(ex) =>
         logger.error("An error occured when trying to update the cache of last changes", ex)
     }
+
+    future {
+      for {
+        updated <- cachedCompliance.checkAndUpdateCache(updatedNodeIds)
```
ncharles: i'm not sure of why you are doing this:
- need to check if cache is filed for these nodes
- need to invalidate the cache for expired nodes
- but why with a list of nodeid as parameters, and shouldn't it be the job of https://github.com/Normation/rudder/pull/805/files#diff-14ce2ba69947867c1979f20a63aa2d2bL163 ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -50,18 +50,40 @@ import com.normation.rudder.reports.ResolvedAgentRunInterval
...
+ */
 class ReportingServiceImpl(
-    confExpectedRepo   : FindExpectedReportRepository
-  , reportsRepository  : ReportsRepository
-  , agentRunRepository : RoReportsExecutionRepository
-  , nodeConfigInfoRepo : RoNodeConfigIdInfoRepository
-  , runIntervalService : AgentRunIntervalService
-  , getComplianceMode  : () => Box[ComplianceMode]
-) extends ReportingService with Loggable {
+    val confExpectedRepo: FindExpectedReportRepository
```
ncharles: wrong indentation

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -50,18 +50,40 @@ import com.normation.rudder.reports.ResolvedAgentRunInterval
...
 
+class CachedReportingServiceImpl(
+    val defaultFindRuleNodeStatusReports: ReportingServiceImpl
+  , val nodeInfoService                 : NodeInfoService
+) extends ReportingService with RuleOrNodeReportingServiceImpl with CachedFindRuleNodeStatusReports {
+  val confExpectedRepo = defaultFindRuleNodeStatusReports.confExpectedRepo
+}
+
+/**
+ * Two of the reporting srevices methods are just utilities above
```
ncharles: typo on services

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -85,6 +107,132 @@ class ReportingServiceImpl(
...
+       * For simplicity (and keeping computation logic elsewhere than in a cache that already has its cache logic
+       * to manage), we will group 2 and 3, but it is well noted that it seems that a RuleNodeStatusReports that
+       * expired could be computed without any more logic than "every thing is missing".
+       *
+       */
+
+      val (expired, upToDate) = nodeIds.partition { id =>  //two group: expired id, and up to date info
+        cache.get(id) match {
+          case None =>
+            if(id == NodeId("36DB125C-8F69-11E4-B8D0-072EA8D53D78")) { println("not in cache ????????")}
```
ncharles: tests part that should not remain

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -85,6 +107,132 @@ class ReportingServiceImpl(
...
+
+      /*
+       * Three cases: 1/ cache does not exist, 2/ cache exists but expiration date expired, 3/ cache does note exists
+       * For simplicity (and keeping computation logic elsewhere than in a cache that already has its cache logic
+       * to manage), we will group 2 and 3, but it is well noted that it seems that a RuleNodeStatusReports that
+       * expired could be computed without any more logic than "every thing is missing".
+       *
+       */
+
+      val (expired, upToDate) = nodeIds.partition { id =>  //two group: expired id, and up to date info
```
ncharles: expired may not be the best name, has the comment explain nicely that it is not only expired
missingOrExpired ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -85,6 +107,132 @@ class ReportingServiceImpl(
...
+      for {
+        newStatus <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(expired, Set())
+      } yield {
+        /*
+         * Here, it's missing all the UnexpectedVersion/NoInitNoVersion results, because, well, they
+         * actually don't have rulenodestatutreport. So.
+         * We add them in NodeId -> Set()
+         */
+        val updated = newStatus.groupBy { _.nodeId }
+        val invalidVersion = expired -- updated.keySet
```
ncharles: i don't understand how a node could not be in the result of findRuleNodeStatusReports; unless it has been deleted

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
fanf: For point 1/ & 2/, see update coming. 

For 3, no because we need to know anything about new runs, not only changes - so we really want all new (finished) runs

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: invalide version to. All in all, it's a case that must be managed

ncharles: but if i understand correctly, newStatus should contains all the RuleNodeStatus  from expired; so updated would contain all the nodeid 
So you are removing everything from something, resulting in invalidVersion being empty ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -85,6 +106,143 @@ class ReportingServiceImpl(
...
+  }
+
+
+  /**
+   * Invalidate some keys in the cache. That won't charge them again
+   * immediately
+   */
+  def invalidate(nodeIds: Set[NodeId]): Unit = this.synchronized {
+    logger.debug(s"Compliance cache: invalidate cache for nodes: [${nodeIds.map { _.value }.mkString(",")}]")
+    cache = cache -- nodeIds
```
ncharles: shouldn't it be --= ?
http://www.scala-lang.org/api/current/#scala.collection.mutable.Map says
 --=(xs: TraversableOnce[A]): Map.this.type
Removes all elements produced by an iterator from this shrinkable collection.


 --(xs: GenTraversableOnce[A]): Map[A, B]
Creates a new map with all the key/value mappings of this map except mappings with keys equal to any of those provided by the specified traversable object.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: In certains cases (like, a new node added right now without promise generated yet), the set of nodeId in status reports (newStatus) is NOT the same as nodeId.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: No, I really want a new immutable collection - cache is immutable

ncharles: ok ! merging

PR #808, opened 2015-01-30
Fixes #6207: Deleting a directive is VERY slow

fanf:


rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -112,8 +112,8 @@ class RoLDAPDirectiveRepository(
 
   override def getDirectiveWithContext(directiveId:DirectiveId) : Box[(Technique, ActiveTechnique, Directive)] = {
     for {
-      directive         <- this.getDirective(directiveId) ?~! "No user Directive with ID=%s.".format(directiveId)
-      activeTechnique   <- this.getActiveTechnique(directiveId) ?~! "Can not find the Active Technique for Directive %s".format(directiveId)
+      (activeTechnique
```
VinceMacBuche: Why not putting all on the same line ?

It makes me sad :( ( and the space on the directive val too)

fanf: because, too wide

PR #819, opened 2015-02-05
Fixes #6162: You can now sort compliances ! and recent changes

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6162


VinceMacBuche:
Don't merge !

VinceMacBuche:
Resort function was called for every Rule, on compliance Bar, we want it to be called once

PR #828, opened 2015-02-12
Fixes #6260: Cannot apply a Directive to all visible Rules

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6260

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
@@ -294,13 +294,13 @@ class RuleGrid(
     directiveApplication match {
       case Some(directiveApp) =>
         def moveCategory(arg: String) : JsCmd = {
-        //parse arg, which have to  be json object with sourceGroupId, destCatId
+          //parse arg, which have to  be json object with sourceGroupId, destCatId
           try {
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -314,12 +314,18 @@ class RuleGrid(
             }
             }
           } catch {
-            case e:Exception => Alert("Error while trying to move group :"+e)
+            case e:Exception =>
+              val msg = s"Error while trying to apply directive ${directiveApp.directive.id.value} on visible Rules"
+              logger.error(s"$msg, cause is: ${e.getMessage()}")
+              logger.debug(s"Error details:")
+              e.getStackTrace().foreach(st => logger.debug(st))
```
fanf: You should juste use: logger.debug(s"Error details:", e) - it will duplicates the exception first line, but it's ok since the log will be in debug and so, when debugging it is easier to understand the whole exception.

VinceMacBuche:
PR updated

PR #830, opened 2015-02-18
Fixes #6282: Cannot remove node properties when setting value to \"\"

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6282

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
@@ -99,7 +99,14 @@ class WoLDAPNodeRepository(
    * what is given in paramater
    */
   def updateNodeProperties(nodeId: NodeId, properties: Seq[NodeProperty], modId: ModificationId, actor:EventActor, reason:Option[String]) : Box[Node] = {
-    val updateNode = (node: Node) => node.copy(properties = properties)
+    val updateNode = (node: Node) => {
+      val newProperties = properties.groupBy(_.name)
+      val oldProperties = node.properties.groupBy(_.name)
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
PR updated

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
```diff
@@ -99,7 +99,14 @@ class WoLDAPNodeRepository(
    * what is given in paramater
    */
   def updateNodeProperties(nodeId: NodeId, properties: Seq[NodeProperty], modId: ModificationId, actor:EventActor, reason:Option[String]) : Box[Node] = {
-    val updateNode = (node: Node) => node.copy(properties = properties)
+    val updateNode = (node: Node) => {
+      val newProperties = properties.groupBy(_.name)
+      val oldProperties = node.properties.groupBy(_.name)
+      // Merge all properties, since this a Map old ones will be overwritten by new since they have the same key
+      // Filter to remove properties with empty values
+      val allProperties = (oldProperties ++ newProperties).values.flatten.toSeq.filterNot(_.value == "")
```
ncharles: If you do this, you need to correct the comment which is wrong

VinceMacBuche: If we keep up with the same logic we should rename the function 'setNodeProperties', I found it more logic with the behavior (with update i did not feel it would overwrite the current params)

fanf:
In fact, that bug was introduced in commit:ea16f601ceb098ed7a72bb688e93944a42406aec
The code already exists in method updateProperties in file rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService5.scala, it is just no more called. 

Closing that PR, and proposing as alternative the old code (which was working).

PR #831, opened 2015-02-27
Fixes #6282: Cannot remove node properties when setting value to ""

fanf:
Revert bad code introduce in commit ea16f601ceb098ed7a72bb688e93944a42406aec


VinceMacBuche:
Great!

PR #833, opened 2015-03-02
Fixes #6323: Missing rule and group category ID

fanf:
I'm not sure it should go in 2.10, but perhaps only in 3.0 ?


VinceMacBuche:
Or it's a bug and it should be merged in 2.10 or it's a feature and should be merged on master

This is so lame that we do not display that information, and it will force people to make ldap requests on their server which we do not want ...

fanf:
OK, could you please double check that only displaying is impacted, and that "things run as expected" with that ?

VinceMacBuche:
I'm on it!

VinceMacBuche:
This works fine! I'm merging it I think it will really help our users!

PR #834, opened 2015-03-02
Fixes #6297: Display UUIDs in rudder web interface in lowercase so that API can be used.

fanf:
Are we really targetting 2.10 for that ?


VinceMacBuche:
Yeah \o/ FINALLY Thanks @fanf

VinceMacBuche:
Yes we want it in 2.10, having uppercasse breaks all our script (ie node_to_relay.sh) ...

PR #836, opened 2015-03-12
Fixes #6381: File /var/log/rudder/compliance/non-compliant-reports.log is not correctly rotated

fanf:
This ticket only add the log rotation from logback. We still need to remove the log rotation done by logrotate, and the two pull request must be merged together.

### DIFF
--- a/rudder-web/src/main/resources/logback.xml
+++ b/rudder-web/src/main/resources/logback.xml
@@ -79,9 +79,15 @@ along with this program. If not, see <http://www.gnu.org/licenses/agpl.html>.
     
     'report' is expected to be the ops logger name.
    -->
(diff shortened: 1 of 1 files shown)


ncharles:
Looks good to me !

fanf:
I added /var/log/rudder/core/rudder-webapp.log, because it is in exactly the same case.

ncharles:
@fanf : these are 2 completely different cases. the rudder-web logs are not causing any issue for the moment, as they are not logrotated. However, they are gziped by rudder-agent

fanf:
I'm not talking about /var/log/rudder/webapp/xxxx.sterrout.log. But closing that PR because it should target 2.10

PR #839, opened 2015-04-01
Fixes #6454: networks.conf empty on relay

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6454

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
@@ -227,17 +227,30 @@ object RuleTarget extends Loggable {
     , groups  : Map[NodeGroupId, Set[NodeId]]
   ) : Set[NodeId] = {
 
+    /* A Node with a server role is either:
+     *   * A node with a role (easy)
+     *   * A node that is a policy Server
+     *   * A node with the root id (in fact: Constants.ROOT_POLICY_SERVER_ID), the root policy server
+     */
+    def isANodeWithRole (nodeId : NodeId, isPolicyServer : Boolean ,serverRoles : Set[ServerRole]) : Boolean = {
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
Something need to be done for the "policy_server" class too, but it should be done in another issue I think

rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
```diff
@@ -232,12 +232,11 @@ object RuleTarget extends Loggable {
       case AllTargetExceptPolicyServers => nodes ++ allNodes.collect { case(k,n) if(!n._1) => k }
       case PolicyServerTarget(nodeId) => nodes + nodeId
       case AllServersWithRole =>
-        //we have a special case for the root node that always belongs to that group, even if some weird
-        //scenario lead to the removal (or non addition) of them
-        nodes ++ allNodes.collect { case(k,n) if(n._2.size>0 || n._1 == Constants.ROOT_POLICY_SERVER_ID) => k }
+        // all nodes with server roles or policyServer (ie relay)
+        nodes ++ allNodes.collect { case(k,n) if(n._2.size>0 || n._1) => k }
```
ncharles: If I understand correctly, you are always adding k in the node; for there always is a n._1
this looks plain wrong

VinceMacBuche: N._1 is a boolean meaning that the node is a policyServer ( either the root server or a relay)

Maybe I should explicit the tuple here by having ( nodeId, ( isPolicyServer, server roles)) inster of (k,n)

The code before was completely wrong ( comparing a boolean and a string)

ncharles: ho, so it was before in another file ( see PR https://github.com/Normation/rudder/pull/824/files ) and that broke it.

VinceMacBuche:
PR updated

PR #841, opened 2015-04-07
Fixes #6451: Missing error message when trying to accept a node with an already existing hostname

fanf:
The problem is http://webmail.dev411.com/t/gg/liftweb/12cmapx88f/lift-posible-bug-in-lazyload-snippet, so I just change the lazy-load part to not enclose errors.


fanf:
Bad branche :(

PR #844, opened 2015-04-09
Fixes #6482: Node API cannot get pending and refused node details

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6482

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService4.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService4.scala
@@ -40,6 +40,7 @@ import com.normation.rudder.web.rest.RestExtractorService
 import com.normation.rudder.web.rest.RestDataSerializer
 import com.normation.inventory.domain.NodeId
 import com.normation.inventory.domain._
+import com.normation.rudder.domain.nodes._
 import net.liftweb.http.Req
 import net.liftweb.common._
 import com.normation.rudder.web.rest.RestUtils._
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService4.scala
```diff
@@ -63,13 +64,13 @@ class NodeApiService4 (
 
   def getNodeDetails(nodeId: NodeId, detailLevel: NodeDetailLevel, state: InventoryStatus) = {
     for {
-      node          <- nodeInfoService.getNode(nodeId)
-      nodeInventory <- inventoryRepository.get(nodeId,state)
+      inventory <- inventoryRepository.get(nodeId,state)
+      node      = nodeInfoService.getNode(nodeId).getOrElse(Node(inventory))
```
fanf: I'm not sure that it is what we want, because node in error also get the defaults information, not only the missing one. 
That's clearly a design problem with getNode that sould return a Box[Option[Node]]. 
In the meantime, could you code it like that:
node <- nodeInfoService.getNode(nodeId) match { case Full(node) => Full(node) ; case eb: EmptyBox => if(inventory /* check status different than accepted */) Full(Node(inventory) else eb }

 ?

VinceMacBuche: In fact it should depend on the state: 

If accepted status: getNode may fail
other status, make no request

fanf: yeah, could coded like that to

VinceMacBuche:
PR updated

PR #845, opened 2015-04-09
Fixes #6481: Create a rudder variable containing all IP of agents

fanf:


rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -353,6 +353,8 @@ class SystemVariableServiceImpl(
       //IT IS VERY IMPORTANT TO SORT SYSTEM VARIABLE HERE: see ticket #4859
       val varManagedNodesAdmin = systemVariableSpecService.get("MANAGED_NODES_ADMIN").toVariable(children.map(_.localAdministratorAccountName).distinct.sorted)
 
+      //IT IS VERY IMPORTANT TO SORT SYSTEM VARIABLE HERE: see ticket #4859
+      val varManagedNodesIp = systemVariableSpecService.get("").toVariable(children.flatMap(_.ips).distinct.sorted)
```
ncharles: the ips contains all the ip of the node ?
aren't we supposed also to make it empty if we have activated NAT ?

fanf: no, we discussed that there should be a config somewhere to say "not secure" and use authorized network in place. 

But you're right, I'm not sure where the logic for that should go. It does not seems right to put it in the variable, else it's name will be misleading a lot. What do you think ?

ncharles: indeed, it make more sense to config it somewhere else, and keep the variable with the proper values

ncharles: and what about the _.ips ? does it really contain all ips, or only the primary one ?

fanf: Hum, I have not the least idea :) How can I test that ?

ncharles:
ok for me

rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
ncharles: systemVariableSpecService.get("").

the name is not correct, it should be MANAGED_NODES_IP

VinceMacBuche:
Merging!

PR #848, opened 2015-04-16
Fixes #6511: Add an API to manage Rule categories

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6511


VinceMacBuche:
Need to target 3.1, cloising

PR #849, opened 2015-04-19
Fixes #6518: Correct the detection on change on nodes to prevent constan...

ncharles:
...t redeploy

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
@@ -65,7 +65,17 @@ case class NodeConfigurationCache(
   , nodeContextCache: Int
   , policyCache: Set[PolicyCache]
 
-)
+) {
+  // We need a method to correctly compare a NodeConfigurationCache that was serialized
+  // from a NodeconfigurationCache created from a NodeConfiguration (so without writtenDate)
+  def equalWithoutWrittenDate(other: NodeConfigurationCache) : Boolean = {
+    id               == other.id &&
+    nodeInfoCache    == other.nodeInfoCache &&
+    parameterCache   == other.parameterCache &&
+    nodeContextCache == other.nodeContextCache &&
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
```diff
@@ -65,7 +65,17 @@ case class NodeConfigurationCache(
   , nodeContextCache: Int
   , policyCache: Set[PolicyCache]
 
-)
+) {
+  // We need a method to correctly compare a NodeConfigurationCache that was serialized
+  // from a NodeconfigurationCache created from a NodeConfiguration (so without writtenDate)
+  def equalWithoutWrittenDate(other: NodeConfigurationCache) : Boolean = {
```
VinceMacBuche: NITPICKING time: 

What do you think of the solutions from this thread ? (need @fanf eye on this too!) 

http://stackoverflow.com/questions/10373715/scala-ignore-case-class-field-for-equals-hascode

I like the first solution (having case class NodeConfigurationCache(id : NodeId, nodeInfo : ...)(val writtenDate : ...) but it would prevent comparing on writtenDate when it is defined, so we may want to exclude this 

Maybe rewriting equals and checking None cases would be better ? 


Maybe we should have transform all our NodeConfigurationCache into NodeConfiguration and compare them instead of comparing NodeConfigurationCache ?

VinceMacBuche:
This is a great imporvement! Thank you for finding this @ncharles !! Just a little question on its implementation

rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
VinceMacBuche: I think the best solution would be to transform NodeConfigCache into Nodeconfig and compare them to actual config with equals

fanf:
I double checked that updates in techniques are correctly reported, and it seems to be ok!
So, good for me :)

PR #850, opened 2015-04-19
Bug 6519/don t increase serial when adding nodes

ncharles:
The rational here is to prevent undue serial upgrade when adding a node in a rule.
The serial change should be only for "change" of config

The huge benefit of that is that we don't lose compliance info of the rule when accepting node
The second great benefit is that promise generation is faster: less change

This picture show the rule screen after having accepted a node
rules

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationServiceImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationServiceImpl.scala
@@ -90,9 +90,10 @@ class DetectChangeInNodeConfiguration extends Loggable {
     logger.trace(s"Checking changes in node '${targetConfig.nodeInfo.id.value}'")
     val changedRuleIds = currentOpt match {
       case None =>
-        //what do we do if we don't have a cache for the node ? All the target rules are "changes" ?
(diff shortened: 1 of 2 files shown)


ncharles:
to be merge after https://github.com/Normation/rudder/pull/849

fanf:
It's ok with me. good explanation and motivational justification

rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationServiceImpl.scala
```diff
@@ -90,9 +90,10 @@ class DetectChangeInNodeConfiguration extends Loggable {
     logger.trace(s"Checking changes in node '${targetConfig.nodeInfo.id.value}'")
     val changedRuleIds = currentOpt match {
       case None =>
-        //what do we do if we don't have a cache for the node ? All the target rules are "changes" ?
+        //what do we do if we don't have a cache for the node ? All the target rules are "changes" ? No, we skip,
+        // as all changes will be caught later. Otherwise, it will increase the serial of all rules applied to this node
         logger.trace("`-> No node configuration cache availabe for that node")
-        targetConfig.policyDrafts.map( _.ruleId ).toSet
+        Set[RuleId]()
```
ncharles: ok, in case of clear cache, this doesn't do as expected

PR #852, opened 2015-04-27
Fixes #6470: Add an API to query Nodes

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6470

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
@@ -42,6 +42,7 @@ import com.normation.rudder.domain.policies._
 import com.normation.rudder.domain.queries.Query
 import com.normation.rudder.repository._
 import com.normation.rudder.services.queries.CmdbQueryParser
+import com.normation.rudder.services.queries.CmdbQueryParser._
 import com.normation.rudder.web.rest.directive.RestDirective
 import com.normation.rudder.web.rest.group.RestGroup
 import com.normation.rudder.web.rest.node._
@@ -53,6 +54,7 @@ import com.normation.rudder.web.services.UserPropertyService
 import com.normation.utils.Control._
 import net.liftweb.common._
 import net.liftweb.json._
+import net.liftweb.json.JsonDSL._
 import com.normation.rudder.api.ApiAccountId
 import com.normation.rudder.web.rest.parameter.RestParameter
 import com.normation.rudder.domain.parameters.ParameterName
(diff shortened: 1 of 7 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPI6.scala
```diff
@@ -0,0 +1,89 @@
...
+
+    case Get(Nil, req) =>
+      implicit val prettify = restExtractor.extractPrettify(req.params)
+      restExtractor.extractNodeDetailLevel(req.params) match {
+        case Full(level) =>
+          restExtractor.extractQuery(req.params) match {
+            case Full(None) =>
+              serviceV6.listNodes(AcceptedInventory, level, None)
+            case Full(Some(query)) =>
+              serviceV6.queryNodes(query,AcceptedInventory, level)
```
fanf: You forget the failure case here, that blows up to the client with a "scala.MatchError" if reach

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -679,5 +713,27 @@ case class RestExtractorService (
...
   }
 
+  def extractQuery (params : Map[String,List[String]]) : Box[Option[Query]] = {
+    extractOneValue(params,"query")(convertToQuery) match {
+      case Full(None) =>
+        extractOneValue(params,CRITERIA)(convertToQueryCriterion) match {
+          case Full(None) => Full(None)
+          case Full(Some(criterion)) =>
+            for {
+              optType <- extractOneValue(params,TARGET)(convertToQueryReturnType)
```
fanf: I'm pretty sure that it is not what we want, because with that, in case of a bad extract for target, we get a Full(none) that buble up to ignoring the query and returning the full list of nodes. 
This is extremelly hard to understand that Rudder ignored your request. 
In fact, each time the user gives (valid) param names, we should assure that we are not ignoring them - if it bored itself to specify a rather complexe query object, he DOES want to know that the string he gaves what not OK. 

This place is only the tip of the iceberg, there is other places that behaves the same. Our motto should be "fail early, fail explicitly".

VinceMacBuche: We get only a Full(None) only if target is not set (and so we set the default value below)

If target is not correctly parsed we get a Failure, so we wont have a problem

Semantic of the extract function are:
Full(Some(x)) => The param was present and correct
Full(None) => The param was missing and should be treated (keep as None, use a default value, error for mandatory params)
Failure/Empty => the param was present but could not correctly been parsed

VinceMacBuche: ie: a request /rudder/api/latest/nodes?where=myCriterion&select=notagoodvalue will end in::

    {
  "action": "listAcceptedNodes",
  "result": "error",
  "errorDetails": "Query return type 'notagoodvalue' is not valid"
* Connection #0 to host localhost left intact

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

PR #855, opened 2015-04-28
Fixes #6540: Webdav allow from any network not only allowed network

peckpeck:
http://www.rudder-project.org/redmine/issues/6540


Kegeruneku:
OK good.

PR #856, opened 2015-05-05
Fixes #6558: Update test in Rudder so it is ok with new inventory data model

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6558


VinceMacBuche:
Needs https://github.com/Normation/ldap-inventory/pull/62 to be merged before

PR #857, opened 2015-05-06
Fixes #6560: Display key used to sign inventory and if the Node is \"Certified\"

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6560


VinceMacBuche:
PR updated

PR #858, opened 2015-05-11
Fixes #6577: In group API, attribute and comparator names are case sensitive

fanf:


rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -92,8 +92,9 @@ case object HasKey extends KeyValueComparator { override val id = "hasKey" }
 trait ComparatorList {
   def comparators : Seq[CriterionComparator]
   def comparatorForString(s:String) : Option[CriterionComparator] = {
+    val lower = s.toLowerCase
     for(comp <- comparators) {
-      if(comp.id == s) return Some(comp)
+      if(lower == comp.id) return Some(comp)
```
VinceMacBuche: I think comp.id should ne lowercased

VinceMacBuche:
Apart the small comment this can be merged

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
fanf: ah yes, it should.

PR #859, opened 2015-05-17
Fixes #6601: Remove invalid default public key for root server

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6601


fanf:
Depending of the way #6000 if fiexed, that one may not be necessary at all

PR #862, opened 2015-05-19
Fixes #6619: When only adding directives to Rules, reporting is not displayed for added directives

VinceMacBuche:
http://www.rudder-project.org/redmine/issues/6619


fanf:
This PR does not solve the previous problem of updating serial on rule that are not actually changed. 

I'm closing it to propose an alternative solution.

VinceMacBuche:
replaced by #864, closing

PR #863, opened 2015-05-22
Fixes #6575: When we accept a new node, we have two promises generati…

ncharles:
…on instead of one


rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
```diff
@@ -203,10 +206,15 @@ class UpdateDynamicGroups(
...
               if(diff.added.nonEmpty || diff.removed.nonEmpty) {
                 logger.info("Dynamic group %s: added node with id: %s, removed: %s".format(id.value, addedNodes, removedNodes))
-                asyncDeploymentAgent ! AutomaticStartDeployment(modId, RudderEventActor)
+                // we need to trigger a deployment in this case
+                needDeployment = true
               }
           }
         }
+        if (needDeployment)
+        {
```
fanf: Could you use our standard formatting here ? 
if(needDeployment) {
...
} ?

fanf:
Appart for the formatting details, it's OK with me, you can self-merge after the udpate

PR #864, opened 2015-05-22
Fixes #6619: When only adding directives to Rules, reporting is not displayed for added directives

fanf:


rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationServiceImpl.scala
```diff
@@ -100,54 +98,75 @@ class DetectChangeInNodeConfiguration extends Loggable {
...
+             * and no modification in directives). On any change, update the serial.
+             */
+            case (ruleId, Some(cDrafts), Some(tDrafts)) =>
+
+              //here, we can just compare the sets for equality
+
+              if(cDrafts == tDrafts) {
+                Set.empty[RuleId]
+              } else {
+                val diff = ((cDrafts &~ tDrafts) ++ (tDrafts &~ cDrafts)).map { case PolicyCache(RuleId(rid), Cf3PolicyDraftId(did), _) =>
```
ncharles: i don't guet the meaning of this line ... could you comment a bit on it ?

fanf: It's the diff on two set to get what directives changed - ha, and actually I could use --, it could be clearer than &~

ncharles:
\o/

ncharles:
oups, wrong click

PR #865, opened 2015-05-22
Fixes #6663: Compliance in dashboard and rule page are not consistant

fanf:


VinceMacBuche:
All right ! merging !

PR #868, opened 2015-05-28
Issue #6137: add an API for compliance by node or rule

fanf:
See http://www.rudder-project.org/redmine/issues/6137 for details about what is added

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
@@ -1,0 +1,255 @@
+/*
+*************************************************************************************
+* Copyright 2015 Normation SAS
+*************************************************************************************
+*
+* This program is free software: you can redistribute it and/or modify
+* it under the terms of the GNU Affero General Public License as
+* published by the Free Software Foundation, either version 3 of the
+* License, or (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU Affero GPL v3, the copyright holders add the following
+* Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU Affero GPL v3
+* licence, when you create a Related Module, this Related Module is
+* not considered as a part of the work and may be distributed under the
+* license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* This program is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+* GNU Affero General Public License for more details.
+*
+* You should have received a copy of the GNU Affero General Public License
+* along with this program. If not, see <http://www.gnu.org/licenses/agpl.html>.
+*
+*************************************************************************************
+*/
+
+package com.normation.rudder.web.rest.compliance
+
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.RuleId
+import com.normation.rudder.domain.reports._
+import net.liftweb.json.JsonDSL._
+import com.normation.rudder.domain.policies.DirectiveId
+
+
+
+/**
+ * Here, we want to present two view of compliance:
+ * - by node
+ * - by rule
+ *
+ * For "by node", we want to have a list of node, then rules,
+ * then directives, then component and then values.
+ *
+ * For "by rule", we want to have a list of rules, then
+ * directives, then components, THEN NODEs, then values.
+ * Because we want to be able to analyse at the component level
+ * if we have problems on several nodes, but values are only
+ * meaningful in the context of a node.
+ *
+ * So we define two hierarchy of classes, with the name convention
+ * starting by "ByRule" or "ByNode".
+ */
+
+
+/**
+ * Compliance for a rules.
+ * It lists:
+ * - id: the rule id
+ * - compliance: the compliance of the rule by node
+ *   (total number of node, repartition of node by status)
+ * - nodeCompliance: the list of compliance for each node, for that that rule
+ */
+case class ByRuleRuleCompliance(
+    id             : RuleId
+    //compliance by nodes
+  , compliance     : ComplianceLevel
+  , directives     : Seq[ByRuleDirectiveCompliance]
+)
+
+case class ByRuleDirectiveCompliance(
+    id        : DirectiveId
+  , compliance: ComplianceLevel
+  , components: Seq[ByRuleComponentCompliance]
+)
+
+case class ByRuleComponentCompliance(
+    name      : String
+  , compliance: ComplianceLevel
+  , nodes     : Seq[ByRuleNodeCompliance]
+)
+
+case class ByRuleNodeCompliance(
+    id        : NodeId
+  , values    : Seq[ComponentValueStatusReport]
+)
+
+/**
+ * Compliance for a node.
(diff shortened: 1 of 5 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
```diff
@@ -0,0 +1,248 @@
...
+
+  def apply(d: DirectiveStatusReport): ByNodeDirectiveCompliance = {
+    new ByNodeDirectiveCompliance(d.directiveId, d.compliance, d.components)
+  }
+}
+
+object JsonCompliance {
+
+
+  implicit class JsonbyRuleCompliance(rule: ByRuleRuleCompliance) {
```
VinceMacBuche: Interesting, I compared with how we display compliance for a Rule in the webui

In the web ui, We do:

* Directives
  * Components
    * Values

And here, in the API, we do

* Directives
  * Components
    * Nodes
      * Values


Do you think we should add the "Node" level also in the webui, in a near future ? (3.2 ?)

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
```diff
@@ -0,0 +1,248 @@
...
+  private[this] def statusDisplayName(r: ReportType): String = {
+    r match {
+      case NotApplicableReportType => "successNotApplicable"
+      case SuccessReportType => "successAlreadyOK"
+      case RepairedReportType => "successRepaired"
+      case ErrorReportType => "error"
+      case UnexpectedReportType => "unexpectedUnknownComponent"
+      case MissingReportType => "unexpectedMissingComponent"
+      case NoAnswerReportType => "noReport"
+      case PendingReportType => "applaying"
```
VinceMacBuche: Typo here "applaying"

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
```diff
@@ -0,0 +1,248 @@
...
+                      )
+                    })
+                )
+             })
+         )
+        })
+    )
+  }
+
+  private[this] def statusDisplayName(r: ReportType): String = {
```
VinceMacBuche: I was unconfortable in the beginning with some names (success and unknown ones) but after is understook what could be achieved:

* Match on success/unexpected, and ignore the subtype (repair, N/A, ...)
* Match on a specific subtype

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
```diff
@@ -0,0 +1,248 @@
...
+  }
+}
+
+object JsonCompliance {
+
+
+  implicit class JsonbyRuleCompliance(rule: ByRuleRuleCompliance) {
+    def toJson = (
+        ("id" -> rule.id.value)
+      ~ ("compliance" -> percents(rule.compliance))
```
VinceMacBuche: I think we should provide the global percent of compliance without details: (success  + na + repaired) / global

some ideas: 

have them in a compliance entry:

    "compliance" -> { 
        "global" => "global percent"
      , "details" => percents object
     }

Have them in the root object in two separeted entries:

    "compliance" -> "global percent"
    "detailedCompliance" => pecents object

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIService.scala
```diff
@@ -0,0 +1,265 @@
...
+              rule.id
+            , ComplianceLevel(noAnswer = nodeIds.size)
+            , Seq()
+          ))
+        }).toMap
+      }
+
+      //reports by rules
+      //that's a *little* gruikkk because, well,
+      //we just plainly ignore errors
```
VinceMacBuche: Could we log them here ? do we want to log the error? 

using partition function to get failures and success into two separated collections

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIService.scala
```diff
@@ -0,0 +1,265 @@
...
+                  , rule.directiveIds.map { id => ByNodeDirectiveCompliance(id, ComplianceLevel(noAnswer = 1), Map())}.toSeq
+                )
+              }).toSeq
+          ))
+        }.toMap
+      }
+
+      //reports by rules
+      //that's a *little* gruikkk because, well,
+      //we just plainly ignore errors
```
VinceMacBuche: Same remark here about the error logging

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
fanf: I'm not sur, because in the UI, you have the two view availables, so having nodes and value merge may not be a limitation. 
On the other hand, we could have a way to see which nodes are in error for a given value (and only errors), what could be nice

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
fanf: You're right, and it kind of matches what guided me to them: I wanted to give the API user the same kind of grouping with the name that one has with the color in the UI. And moreover, people are really easelly missleaded by unexpected / missing and take them for error and no data, and repaired is not obviously a success. 

So in the end, do you agree with them ?

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
fanf: Really good idea, and in fact we should have it at each level with the pair compliance / complianceDetails so that user can either just deal with a number or process the details. 

Moreover, it allows a "short" format in the future with just the compliance number, and up to a given level of object.

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIService.scala
fanf: We certainly want to do something less extrem than that, yes. I'm not sur of the log level we want to use, thought.

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
VinceMacBuche: Understanding the need it fulfills, I agree with them!

And i do not find a better way to name them and provide the same usefullness

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIData.scala
VinceMacBuche: To me it's clearly missing on the webapp, you can't know which nodes is wrong on a component, until you go to the last level of each nodes

The other way that could be done is:

* Directives
  * Components
    * Values
      * Nodes

fanf: This other way feels less logical, because a value may be node dependant (ex: file path depending of the hostname), and the error are by value and by node (node foo reported error bla on value val).

rudder-web/src/main/scala/com/normation/rudder/web/rest/compliance/ComplianceAPIService.scala
VinceMacBuche: hmmm What do you think of...

    If there was an error (so partition not empty)
      logger.error "there was error when   computing compliance for rules x, y z, please enable debug output for more details"
      for each error 
        logger.debug(" blah blah error")

VinceMacBuche:
Great feature for the web api ! Don't forget to add the API documentation !!

PR #869, opened 2015-05-28
Fixes #6680: Bad RegEx in Groups causes Server to not respond

fanf:


VinceMacBuche:
Another day explaining why your errors should be managed by the type system! merging!

PR #872, opened 2015-06-11
Fixes #6715: Add info about API

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6715

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestAPI.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestAPI.scala
@@ -41,6 +41,7 @@ import net.liftweb.http.rest.RestHelper
 import net.liftweb.common.Full
 import net.liftweb.common.EmptyBox
 import net.liftweb.json.JString
+import net.liftweb.json.JValue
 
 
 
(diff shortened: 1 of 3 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/RestAPI.scala
```diff
@@ -119,4 +121,19 @@ case class APIDispatcher (
...
 
+    serve {
+      case Get("api" :: "info" :: Nil,req) =>
+        implicit val action = "listChangeRequests"
+        implicit val prettify = restExtractor.extractPrettify(req.params)
+
+        val availableVersions = apisByVersion.keySet.map(_.value).toList.sorted
+        val max = availableVersions.max
+        import net.liftweb.json.JsonDSL._
+        val versions: JValue =
```
fanf: I think we need to have versions { latest = ... , available } else it will be hard to extend the info in the future. 
Perhaps it's apiVersions ? version is shorter, thought

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

fanf:
Great !

PR #876, opened 2015-06-12
Fixes #6277: Why are search results (search nodes) grayed out in 3.0?

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6277


fanf:
Great !

PR #878, opened 2015-06-12
Fixes #6633: The Global compliance is buggy when there is no data

fanf:


ncharles:
you are removing  rudder-web/src/main/webapp/javascript/c3/c3-0.4.8.js

this is probably an error, as it is included in main/webapp/templates-hidden/common-layout.html :

<script type="text/javascript" src="/javascript/c3/c3-0.4.8.js"></script>

VinceMacBuche:
I think it's the minified version of the file that is included in common-layout not this one which can be used for debugging

ncharles:
github ate my line:
 src="/javascript/c3/c3-0.4.8.js"
so no minified is used....

fanf:
PR updated to restore the js. 
For information, I tried to updated c3.js to version 0.4.10 to see if it corrects the problem, and it wasn't, but I didn't correctly restored the js files in the commit.

VinceMacBuche:
The gague does not come with c3.js but another library http://bernii.github.io/gauge.js/

PR #883, opened 2015-06-25
Fixes #6729: After a partial promise generation, some nodes being a r…

ncharles:
…elay may have their promises stored in a non reachable location

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputer.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputer.scala
@@ -41,26 +41,12 @@ import com.normation.rudder.services.policies.nodeconfig.NodeConfiguration
 
 /**
  * Utilitary tool to compute the path of a server promises (and others information) on the rootMachine
- *
- * @author nicolas
+
  *
  */
 trait PathComputer {
 
-  /**
-   * Compute the base path for a server, i.e. the full path on the root server to the data
-   * the searched server will fetch, and the backup folder
-   * Finish by the server name, with no trailing /
-   * Ex : /opt/hive/cfserved/serverA/served/serverB, /opt/hive/backup/serverA/served/serverB
-   * @param searchedNode : the server we search
-   * @return
-   */
-  def computeBaseNodePath(searchedNodeId : NodeId, rootNodeId: NodeId, allNodeConfigs:Map[NodeId, NodeConfiguration]): Box[((String, String))]
+  def computeBaseNodePath(searchedNodeId : NodeId, rootNodeId: NodeId, allNodeConfigs:Map[NodeId, NodeConfiguration]): Box[((String, String, String))]
 
-  /**
-   * Return the path of the promises for the root (we directly write its promises in its path)
-   * @param agent
-   * @return
-   */
   def getRootPath(agentType : AgentType) : String
 }
--- a/rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
@@ -57,25 +57,32 @@ class PathComputerImpl(
 ) extends PathComputer with Loggable {
 
 
-  private[this] val promisesPrefix = "rules/"
+  private[this] val promisesPrefix = "/rules"
+  private[this] val newPostfix = ".new"
 
   private[this] val baseFolder = Constants.NODE_PROMISES_PARENT_DIR_BASE
   private[this] val relativeShareFolder = Constants.NODE_PROMISES_PARENT_DIR
 
   /**
-   * Compute the base path for a machine, i.e. the full path on the root server to the data
-   * the searched machine will fetch, and the backup folder
-   * Finish by the machine uuid, with no trailing /
-   * It does not contain the root machine, except if we search the root machine
-   * Ex : /var/rudder/share/uuid-a/share/uuid-b, /var/rudder/backup/uuid-a/share/uuid-b
+   * Compute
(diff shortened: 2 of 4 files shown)


ncharles:
This modification doesn't delete promises of deleted nodes (like before), and they leave a dandlling empty rules.new folder for each node (which should not be an issue, as it is not accessed)

rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
```diff
@@ -57,25 +57,32 @@ class PathComputerImpl(
...
+   *  the new folder, i.e. the full path where promises are written and checked before being updated
+   *  the backup folder, the path were promises are backuped
+   * Finish with no trailing /
+   * It does not contain the root machine, and does not work for root
+   * Ex : /var/rudder/share/uuid-a/share/uuid-b/rules,/var/rudder/share/uuid-a/share/uuid-b/rules.new, /var/rudder/backup/uuid-a/share/uuid-b
    * @param searchedNodeConfiguration : the machine we search
    * @return
    */
-  def computeBaseNodePath(searchedNodeId : NodeId, rootNodeId: NodeId, allNodeConfigs: Map[NodeId, NodeConfiguration]): Box[((String, String))] = {
+  def computeBaseNodePath(searchedNodeId : NodeId, rootNodeId: NodeId, allNodeConfigs: Map[NodeId, NodeConfiguration]): Box[((String, String, String))] = {
```
fanf: I tried to see if we can create a datastructure for that horrible (string, string, string) but the changes are non trivial, and so lead to a little to much refactoring for a bug fixe. So let be it like that.

rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
```diff
@@ -57,25 +57,32 @@ class PathComputerImpl(
...
     for {
```
fanf: Before the for, we should test for rootID and fails, no ? Because if we don't support it, that can lead to surprising results. Something like:

if(searchedNodeId == rootNodeId) {
      Failure("Path for promises of root node can't be computed with the same method than other ndoes")
} else for {

ncharles: i'm not sure this would behave as expected, as 'im pretty sure we call this path compute also for root
However, for root we override the value afterward (magic inside)

rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
```diff
@@ -57,25 +57,32 @@ class PathComputerImpl(
...
       path <- recurseComputePath(rootNodeId, searchedNodeId, "/"  + searchedNodeId.value, allNodeConfigs)
```
fanf: In fact, it's more the comment that we need to change:
"It does not contain the root machine, and does not work for root"
"The path calculated for the root server is not the one expected by cfengine and so must not be used. A special case must be done for root server where the method is called".

rudder-core/src/main/scala/com/normation/rudder/services/policies/RudderPromiseWriterServiceImpl.scala
```diff
@@ -156,13 +156,16 @@ class RudderCf3PromisesFileWriterServiceImpl(
    * From a base path and a base backup path, plus a node configuration, write the rules
```
fanf: On that file, could you explain exactly what is done today, and if possible which file are doing that, so that the next time we know how to corrrect it:
- root server is special cases in class#method, new promises are written in /var/rudder/cfengine-community/inputs.new, then move to ... and old promises are moved to ....bkp
- normal nodes new promises etc etc

Wilth the quirks is any reminds. 

Thanks!

rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
ncharles: indeed, that is a took big change for 2.10 :(

ncharles:
PR updated

rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputerImpl.scala
fanf: Ah, in fact I misused github, and following comment should have been in response to that one: so yes, the code is not to change, only the comment about root machine is.

ncharles: i already changed it

rudder-core/src/main/scala/com/normation/rudder/services/path/PathComputer.scala
```diff
@@ -41,21 +41,21 @@ import com.normation.rudder.services.policies.nodeconfig.NodeConfiguration
...
-   * Ex : /opt/hive/cfserved/serverA/served/serverB, /opt/hive/backup/serverA/served/serverB
-   * @param searchedNode : the server we search
-   * @return
+   * Compute
+   *  the path of promises for a node, i.e. the full path on the root server to the data
+   *  the new folder, i.e. the full path where promises are written and checked before being updated
+   *  the backup folder, the path were promises are backuped
+   * Finish with no trailing /
+   * It does not contain the root machine, except if we search the root machine
+   * Ex : /var/rudder/share/uuid-a/share/uuid-b/rules,/var/rudder/share/uuid-a/share/uuid-b/rules.new, /var/rudder/backup/uuid-a/share/uuid-b
```
fanf: Could you synchronize the comment witht the one on implementation ? Or only put the comment in the interface, if you prefer ?

PR #886, opened 2015-06-30
Fixes #6912: Update Rudder authentication to allows plugins

fanf:
https://www.rudder-project.org/redmine/issues/6912

### DIFF
--- a/rudder-web/pom.xml
+++ b/rudder-web/pom.xml
@@ -84,11 +84,16 @@ along with this program. If not, see <http://www.gnu.org/licenses/agpl.html>.
       <version>1.2.1</version>
     </dependency>
 
-     <dependency>
+    <dependency>
       <groupId>com.normation.rudder</groupId>
       <artifactId>rudder-core</artifactId>
       <version>${rudder-version}</version>
     </dependency>
+<!--     <dependency> -->
+<!--       <groupId>com.normation</groupId> -->
+<!--       <artifactId>authorization-api</artifactId> -->
+<!--       <version>${rudder-version}</version> -->
+<!--     </dependency> -->
     
     <dependency>
       <groupId>net.liftmodules</groupId>
--- a/rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
+++ b/rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
@@ -142,6 +142,7 @@ import com.normation.rudder.web.rest.compliance.ComplianceAPIService
 import com.normation.rudder.services.policies.write.Cf3PromisesFileWriterServiceImpl
 import com.normation.rudder.services.policies.write.PathComputerImpl
 import com.normation.rudder.services.policies.write.PrepareTemplateVariablesImpl
+import com.typesafe.config.ConfigException
 
 /**
  * Define a resource for configuration.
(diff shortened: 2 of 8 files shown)


rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -134,6 +186,27 @@ class AppConfigAuth extends Loggable {
     }
   }
 
+  @Bean def fileAuthenticationProvider : AuthenticationProvider = {
```
VinceMacBuche: There is no ldapAuthenticationProvider ?

rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
```diff
@@ -185,6 +192,23 @@ object RudderProperties {
       case FileSystemResource(file) => ConfigFactory.load(ConfigFactory.parseFile(file))
     }
   }
+
+  //some logic for the authentication providers
+  val authenticationMethods: Seq[AuthenticationMethods] = {
+    val names = try {
+      //config.getString can't be null by contract
+      config.getString("rudder.auth.type").split(",").toSeq.map( _.trim).collect { case s if(s.size > 0) => s}
```
VinceMacBuche: Is the property fetched from the configuration.property file ? if so it misses the updated configuration file  (and we should have a migration script)

rudder-web/pom.xml
```diff
@@ -78,7 +80,7 @@ along with this program. If not, see <http://www.gnu.org/licenses/agpl.html>.
       <groupId>com.normation.rudder</groupId>
       <artifactId>rudder-core</artifactId>
       <version>${rudder-version}</version>
-    </dependency>
+     </dependency>
```
VinceMacBuche: I Think this space should not be added (not aligned with the 'dependency' below)

rudder-web/src/main/scala/bootstrap/liftweb/LiftInitContextListener.scala
```diff
@@ -105,32 +103,4 @@ class LiftInitContextListener extends ContextLoaderListener {
     super.contextDestroyed(sce)
   }
 
-  private[this] def initSpringAuthentication(): Unit = {
-    import RudderProperties.{config => c}
-
-    val springConfigFile = if( c.getBoolean("rudder.auth.ldap.enable")) {
```
VinceMacBuche: That property should be removed from configuration file but myabe it is already removed ;)

VinceMacBuche:
@fanf I made some feedback can you look at them ?.

rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
fanf: Yes, you're right

rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
fanf: No, it is still defined in rudder-web/src/main/resources/applicationContext-security-auth-ldap.xml

fanf: So you can see what you need to either defined a code-based provider, or a file based one. And you get a tast of the horrors of spring.

rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
fanf: Added in last PR update

fanf:
PR updated

rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
VinceMacBuche: That seems great ( I mean great like the great depression of 1929)

VinceMacBuche:
Should we wait for the migration script to be done before merging this ?

fanf:
Yes, we should, else most likely it will lead to inconsistencies. But not sure when I will be able to do it.

fanf:
Merging this one since the migration script was merged.

PR #893, opened 2015-07-08
Fixes #6939: event logs pages stay empty and cannot load

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6939

### DIFF
--- a/rudder-web/src/main/webapp/secure/utilities/eventLogs.html
+++ b/rudder-web/src/main/webapp/secure/utilities/eventLogs.html
@@ -10,13 +10,11 @@
   <div class="portlet-content"> 
     <hr class="spacer"/>
 
-    <lift:lazy-load>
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/webapp/secure/utilities/eventLogs.html
```diff
@@ -10,13 +10,11 @@
   <div class="portlet-content"> 
     <hr class="spacer"/>
 
-    <lift:lazy-load>
     <div class="lift:administration.EventLogsViewer.display">
     </div>
     <span id="eventLogsError"></span>
     <table id="eventLogsGrid" class="display" cellspacing="0">
-      </table>
```
fanf: why removing the </table> ?

fanf: ah no, sorry, git diff is playing with me

PR #895, opened 2015-07-09
Fixes #6948: Merge cf-clerk into Rudder

fanf:
And everything's still working !

Ticket: http://www.rudder-project.org/redmine/issues/6948


ncharles:
\o/

PR #896, opened 2015-07-21
Fixes #7016: Adapt Rudder to Scala 2.11 and Lift 2.6

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7016


VinceMacBuche:
Still WIP

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

rudder-web/pom.xml
```diff
@@ -87,8 +87,8 @@ along with this program. If not, see <http://www.gnu.org/licenses/agpl.html>.
     
     <dependency>
       <groupId>net.liftmodules</groupId>
-      <artifactId>widgets_2.5_${scala-binary-version}</artifactId>
-      <version>1.3</version>
+      <artifactId>widgets_2.6_${scala-binary-version}</artifactId>
+      <version>1.4-SNAPSHOT</version>
```
fanf: we need to do something for that, we can't rely on snapshot

rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
```diff
@@ -170,14 +167,16 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
...
 
-        <li class="dropdown">
-          <a href="#" class="dropdown-toggle"  data-toggle="dropdown">
-            <span>Status</span> <span class="badge" id="generation-status"> {statusIcon}</span><span class="caret" style="margin-left:5px"></span></a>
-          <ul class="dropdown-menu" role="menu">
-            {lastStatus}
-            <li>{currentStatus}</li>
-          </ul>
-        </li>
+    <lift:authz role="deployment_read">
```
fanf: That seems to be a real bug correction, no ? Deserve its own ticket, I think

VinceMacBuche: In fact its has changed because I removed the lift:authz in common-layout to fix the comet actor correctly and ada it back here, so I don't think it needs it specific bug (lift:authz elem are not accepted in ul html5...

PR #899, opened 2015-07-29
Fixes #7043: Make Rudder code use BundleOrder where it's needed

fanf:
https://www.rudder-project.org/redmine/issues/7043


rudder-core/src/main/scala/com/normation/rudder/services/policies/RudderPromiseWriterServiceImpl.scala
```diff
@@ -252,6 +252,8 @@ class RudderCf3PromisesFileWriterServiceImpl(
 
       val container = node.toContainer(newNodeRulePath)
 
+      println(s"======> current node: [${node.nodeInfo.id.value}]; agent: ${node.nodeInfo.agentsName}")
```
ncharles: you don't want this println

PR #902, opened 2015-07-30
Fixes #7055: Fix reporting issue (multiple reports for one component) on branch 2.11

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7055

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/reports/bean/ExecutionBatch.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/reports/bean/ExecutionBatch.scala
@@ -51,6 +51,7 @@ import com.normation.rudder.domain.logger.ReportLogger
 import com.normation.rudder.domain.policies.Directive
 import com.normation.rudder.domain.reports.ReportComponent
 import com.normation.inventory.domain.NodeId
+import net.liftweb.common.Loggable
 
 /**
  * An execution batch contains the node reports for a given Rule / Directive at a given date
@@ -89,12 +90,6 @@ trait ExecutionBatch {
   def getRepairedReports() : Seq[Reports] = {
     executionReports.collect { case x : ResultRepairedReport => x }
   }
-
-  /* Warn is temporarly unused*/
-  /*
-  def getWarnReports() : Seq[Reports] = {
-    executionReports.filter(x => x.isInstanceOf[WarnReport])
-  }*/
 
   def getErrorReports() : Seq[Reports] = {
     executionReports.collect { case x : ResultErrorReport => x }
@@ -136,7 +131,7 @@ case class ConfigurationExecutionBatch(
   , val beginDate               : DateTime
   , val endDate                 : Option[DateTime]
   , val agentExecutionInterval : Int
-) extends ExecutionBatch {
+) extends ExecutionBatch with Loggable {
 
   // A cache of the already computed values
   val cache = scala.collection.mutable.Map[String, Seq[NodeId]]()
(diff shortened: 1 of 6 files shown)


VinceMacBuche:
PR updated

rudder-core/src/main/scala/com/normation/rudder/domain/reports/bean/ExecutionBatch.scala
```diff
@@ -489,36 +485,57 @@ case class ConfigurationExecutionBatch(
...
-         val nodes = componentValues.map(value => NodeReport(value.nodeId,value.reportType,value.message))
+ def getComponentValuesRuleStatus(directiveid:DirectiveId, component:String, values:Map[Option[String], Set[String]], components:Seq[ComponentStatusReport]) : Set[ComponentValueRuleStatusReport]={
+
+   (values.flatMap{
+     case (unexpanded,values) =>
+       // Get all reports that match a value from our unexpandedValue, and group them by value
+       val componentValues = components.flatMap(_.componentValues.filter(v => values.contains(v.componentValue))).groupBy(_.componentValue)
+       for {
+         // for each value, process a reporting date
+         (value,values) <- componentValues
```
ncharles: what about those that don't match ?
plus, you are using twice values in 3 lines, so when rereading, it's hard to know which one we use

rudder-core/src/main/scala/com/normation/rudder/domain/reports/bean/ExecutionBatch.scala
```diff
@@ -489,36 +485,57 @@ case class ConfigurationExecutionBatch(
...
- def getComponentValuesRuleStatus(directiveid:DirectiveId, component:String, values:Seq[(String, Option[String])], components:Seq[ComponentStatusReport]) : Seq[ComponentValueRuleStatusReport]={
-     values.map{
-       case (value, unexpanded) =>
-         val componentValues = components.flatMap(_.componentValues.filter(_.componentValue==value))
-         val nodes = componentValues.map(value => NodeReport(value.nodeId,value.reportType,value.message))
+ def getComponentValuesRuleStatus(directiveid:DirectiveId, component:String, values:Map[Option[String], Set[String]], components:Seq[ComponentStatusReport]) : Set[ComponentValueRuleStatusReport]={
+
+   (values.flatMap{
+     case (unexpanded,values) =>
+       // Get all reports that match a value from our unexpandedValue, and group them by value
```
ncharles: is this comment true ???

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

fanf:
OK, so there is a lots of things strange or that I don't understand on the pr. For one, the tests are not passing. Second: it seems that we don't have values anymore, at all, in ComponentStatusReport, but only the unexpanded value. I'm pretty sure we don't want to throw it away (and only rely on the content of reports to hope for it). 
I spent several hours trying to understand what we really want, and I'm not sure of the answer now. I think that one of the reason is that our data model is not really clear, and we really miss a "report key" - what kind of unexpandedValue is used for, here, but not quite because we are still unable to correctly merge back reports on lots of edge cases. 
Redisigning the data model to assert that a we have "unexpandedvalue that can have several values that can have several messages" is a rather big change, not sure we want it to go in 3.0. 

Or we assume the fact that we are only managing unexpanded values and that we loose the value along the way, and we address the problem in 3.2. 

Ideas ?

fanf:
Ignore the preceding comment, it was for branch 3.0 / pr #904

fanf:
Closing this one because need a rebase, replaced by #911

PR #906, opened 2015-08-04
Fixes #7059: In UI, directives in a rule are not sorted as their bundle order

fanf:
https://www.rudder-project.org/redmine/issues/7059

Some information about that pull request: AngularJS needs to have stable JS object or it goes in infinite loop. So I need to have both the data I need, with the model I want to handle AND a copy of them on an angularjs-no-infinity-loop compatible way. That su**s. Not found anything else.

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/angular/ruleDirectives.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/angular/ruleDirectives.js
@@ -1,7 +1,46 @@
 var ruleDirectives = angular.module('ruleDirectives', []);
 
 ruleDirectives.controller('DirectiveCtrl', ['$scope', function($scope) {
-    $scope.directives = [];
+    $scope.directives = {};
+    
+    //really, angular... 
+    //a call to sortDirectives is needed. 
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/webapp/javascript/rudder/angular/ruleDirectives.js
```diff
@@ -1,7 +1,46 @@
...
+    $scope.sortedDirectives = [];
+    
+    //needed to correctly display directive by alphaNum
+    $scope.sortDirectives = function() { 
+      var arr = [];
+      for (var o in $scope.directives) {   
+        arr.push({"id":o, "name": $scope.directives[o] });
+      } 
+      var sorted = arr.sort(function(a, b) {
+        return a.name.localeCompare(b.name);
```
ncharles: this is unfortunately not compatible IE10 :(

ncharles: oh, my bad, this existed already in IE6
http://stackoverflow.com/questions/51165/how-do-you-do-string-comparison-in-javascript

PR #911, opened 2015-08-06
Fixes #7055: Reporting cannot be used when there is several component with the same value and several messages

fanf:
This PR is just a version of #902 but rebased to current state of branche 2.11.


fanf:
Self-merging, because in fact it means merging #902

PR #914, opened 2015-08-21
Fixes #7123: Correct, clarify and simplify descriptive texts

jooooooon:
https://www.rudder-project.org/redmine/issues/7123

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
@@ -215,7 +215,7 @@ $$("#${detailsId}").bind( "show", function(event, ui) {
   }
 
   /**
-   * Show details about the server in a tabed fashion if
+   * Show details about the server in a tabbed fashion if
    * the server exists, display an error message if the
    * server is not found or if a problem occurred when fetching it
    *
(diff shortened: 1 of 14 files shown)


jooooooon:
PR updated

jooooooon:
PR updated

rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -269,10 +269,10 @@ $$("#${detailsId}").bind( "show", function(event, ui) {
     <li><a href={htmlId_#(jsId,"sd_fs_")}>File systems</a></li>
     <li><a href={htmlId_#(jsId,"sd_net_")}>Network interfaces</a></li>
     <li><a href={htmlId_#(jsId,"sd_soft_")}>Software</a></li>
-    <li><a href={htmlId_#(jsId,"sd_props_")}>Properties</a></li>
-    <li><a href={htmlId_#(jsId,"sd_var_")}>Environment variables</a></li>
+    <li><a href={htmlId_#(jsId,"sd_var_")}>Environment</a></li>
     <li><a href={htmlId_#(jsId,"sd_process_")}>Processes</a></li>
     <li><a href={htmlId_#(jsId,"sd_vm_")}>Virtual machines</a></li>
+    <li><a href={htmlId_#(jsId,"sd_props_")}>Properties</a></li>
```
jooooooon: Note that this change doesn't fix the English, it moves the "Properties" tab to the end of the list, to avoid it being "lost" in the middle of inventory data (which it is not).

rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -284,10 +284,10 @@ $$("#${detailsId}").bind( "show", function(event, ui) {
     val jsId = JsNodeId(sm.node.main.id,salt)
     displayTabFilesystems(jsId, sm) ++
     displayTabNetworks(jsId, sm) ++
-    node.map(displayTabProperties(jsId, _)).getOrElse(Nil) ++
     displayTabVariable(jsId, sm) ++
     displayTabProcess(jsId, sm) ++
     displayTabVM(jsId, sm) ++
+    node.map(displayTabProperties(jsId, _)).getOrElse(Nil) ++
```
jooooooon: I moved this line because I moved the order of tabs above, but I suspect it's not necessary. If it could be dangerous, and is not necessary, just let me know and I'll put it back.

rudder-web/src/main/webapp/templates-hidden/server/server_details.html
```diff
@@ -37,18 +37,18 @@
...
-    <div class="nodeheadercontent ui-corner-top"> Node Details - <span id="node_name"></span>
+    <div class="nodeheadercontent ui-corner-top"> Node: <span id="node_name"></span>
     </div>
   </div>
 
 
     <ul id="NodeDetailsTabMenu">
-      <li><a href="#node_summary">Node summary</a></li>
+      <li><a href="#node_summary">Summary</a></li>
+      <li><a href="#node_reports">Compliance</a></li>
```
jooooooon: This line is both a rename (Reports --> Compliance) and a re-order, so that the Compliance tab is displayed right at the beginning of Node details. This seems more logical than hiding at the end, since Rudder is all about compliance.

PR #918, opened 2015-08-25
Fixes #7132: Setting page broken: missing lift changes

fanf:
https://www.rudder-project.org/redmine/issues/7132


fanf:
It seems that something is broken with error message (don't appear anymore), but I can't find what.

VinceMacBuche:
good to me! We will open another issue about the missing messages!

PR #919, opened 2015-08-26
Fixes #7108: Rules are not serialized correctly anymore

ncharles:


VinceMacBuche:
Good to me !! Merging!

PR #920, opened 2015-08-28
Fixes #7146: Move event log and authz into Rudder (adding into Rudder)

fanf:
https://www.rudder-project.org/redmine/issues/7146


ncharles:
\o/

PR #922, opened 2015-09-04
Fixes #7154: Agent schedule is not historised, so we can't know what …

ncharles:
…was the agent run interval in the past


Kegeruneku:
OK !

PR #924, opened 2015-09-17
Fixes #7199: Read only role is not correctly interpreted in technique editor.

fanf:
https://www.rudder-project.org/redmine/issues/7199


ncharles:
looks good to me

PR #932, opened 2015-10-21
Fixes #7244: Change request list page, link to Rules are invalids (extra rudder)

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7244

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
@@ -63,7 +63,7 @@ object JsInitContextLinkUtil extends Loggable {
   private[this] val roGroupRepository     = RudderConfig.roNodeGroupRepository
   private[this] val roDirectiveRepository = RudderConfig.roDirectiveRepository
 
-  private[this] def baseGroupLink(id:NodeGroupId) =
+  def baseGroupLink(id:NodeGroupId) =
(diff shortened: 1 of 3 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
```diff
@@ -130,7 +130,7 @@ object JsInitContextLinkUtil extends Loggable {
 
   def createGroupLink(id:NodeGroupId) = {
     roGroupRepository.getNodeGroup(id) match {
-      case Full((group,_)) => <span> {SHtml.a(() => S.redirectTo(groupLink(id)), Text(group.name))} (Rudder ID: {id.value})</span>
+      case Full((group,_)) => <span> <a href={baseGroupLink(id)}>{group.name}</a> (Rudder ID: {id.value})</span>
```
ncharles: did you check all the calls made from this function ? i'm pretty sure only change requests are broken, but others part call createRuleLink (like, the diff displayer)

VinceMacBuche:
PR updated

rudder-web/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
VinceMacBuche: I verified the other calls are ok!

I still don't know why it adding /rudder ...)

fanf:
I'm really not at ease to to know why the context is added in that case. Still, the correction work, and I don't know where to look anywhere else. 
So let's merge it.

PR #934, opened 2015-10-23
Fixes #7303: RuleTargetTest.scala takes >40s where it should take 2s

fanf:
https://www.rudder-project.org/redmine/issues/7303


rudder-core/src/test/scala/com/normation/rudder/domain/policies/RuleTargetTest.scala
```diff
@@ -124,31 +129,28 @@ class RuleTargetTest extends Specification with Loggable {
...
+          case None        => gt.target == gt
         }
-      }
+      } === true
     }
 
     "Have their group target removed in composite targets" in {
-      groupTargets.forall{
-        case (gt,_) =>
-         (allComposite ++ allTargetExclusions).forall {
```
ncharles: ouch

PR #936, opened 2015-10-27
Fixes #7316: duplicate key value error on Rudder

fanf:
https://www.rudder-project.org/redmine/issues/7316

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
@@ -54,8 +54,6 @@ import com.normation.rudder.reports.execution.AgentRun
 import com.normation.rudder.domain.reports.Reports
 import com.normation.rudder.reports.execution.AgentRunId
 import com.normation.rudder.reports.execution.AgentRun
-import com.normation.rudder.reports.execution.AgentRun
-import com.normation.rudder.reports.execution.AgentRun
 
 class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsRepository with Loggable {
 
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -382,12 +380,54 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
...
 
+
+    /*
+     * here, we may have several runs with same nodeId/timestamp
+     * In that case, we need to keep the one with a configId, if such
+     * exists.
+     */
+    def distinctRuns(seq: Seq[AgentRun]): Seq[AgentRun] = {
+      //that one is for a list of agentRun with same id
+      def recDisctinct(runs: List[AgentRun]): AgentRun = {
```
ncharles: is that tail recursive ?
you may have a lot of agent run for one id

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -382,12 +380,54 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
...
+        runs match {
+          case Nil => throw new IllegalArgumentException("Error in code: distinctRuns methods should never call the recDistinct one with an empty list")
+          //easy, most common case
+          case h :: Nil => h
+          case a :: b :: t =>
+            if(a == b) {
+              recDisctinct(a :: t)
+            } else (a, b) match {
```
ncharles: there is no order in the list, so how can you ensure it would dedup :
run1, run56, run47, run1

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
fanf: How so ? It is an exceptionnal case to have more than one run for one node at a given second. I can't see how it could be more than a few ?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
fanf: I don't want to de-up run1, run56, run47, run1, I want to deup:
run( Node1, time1, None[configId], serial1) :: run( Node1, time1, Some(config), serial2) :: Nil

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
ncharles: my bad, it is grouped by node and time,not only node

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
ncharles: so no more than 2 runs at the same time

fanf: (oups, missing the remaining): and so, for a given node id and time, there is a total order about what I want - arbitrary decided in the case of two run with a different id, but it exists.

fanf: it could be more than 2, I implemented (at least that's what I wanted) a total order, there is always an unique "best" solution

ncharles: ok

PR #940, opened 2015-10-29
Fixes #6244: Feature idea: automatic directive naming

fanf:
https://www.rudder-project.org/redmine/issues/6244


fanf:
Not what we want

PR #944, opened 2015-11-02
Fixes #7349: Support different BSD flavor

fanf:
https://www.rudder-project.org/redmine/issues/7349


rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -315,7 +315,7 @@ case object MemoryComparator extends CriterionType {
 }
 
 case object OstypeComparator extends CriterionType {
-  val osTypes = List("Linux","Windows","Solaris", "AIX")
+  val osTypes = List("Linux", "BSD", "Solaris", "AIX", "Windows")
```
ncharles: Why changing the ordering ?

PR #945, opened 2015-11-02
Fixes #7353: Missing debuging information for explain_compliance

fanf:
https://www.rudder-project.org/redmine/issues/7353

That PR is best view with ?w=1 (lots of code displaced / spaces changed). 

There is two main things done here to help debug compliance:
- the first one is to add logging possibility to use : explain_compliance.nodeid logger to debug a specific node
- the second is to split the method that find the big case, function of compliance mode, last run, expected config, so that the comparison time and the last valid time are more understandable.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
@@ -45,6 +45,9 @@ import com.normation.rudder.reports.ComplianceMode
 import com.normation.rudder.reports.FullCompliance
 import com.normation.rudder.reports.ResolvedAgentRunInterval
 import com.normation.rudder.reports.ChangesOnly
+import com.google.common.cache.CacheBuilder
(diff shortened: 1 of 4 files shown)


fanf:
I'm adding a missing comment in logback

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/domain/logger/ComplianceDebugLogger.scala
```diff
@@ -57,6 +60,21 @@ object ComplianceDebugLogger extends Logger {
   override protected def _logger = LoggerFactory.getLogger("explain_compliance")
 
 
+  //we have one logger defined by node.
+  //they automatically expires after some time.
+
+  val nodeCache = CacheBuilder.newBuilder().maximumSize(1000).expireAfterWrite(10, TimeUnit.MINUTES).build(
```
ncharles: what happen if we have 2000 nodes, and want to report on only one ?
will we have to hope that this one will be in the first 1000 logger created, or i am not understanding how it works ?

PR #949, opened 2015-11-03
Fixes #7357: Inventory query test are not passing anymore in Rudder 3.0

fanf:
https://www.rudder-project.org/redmine/issues/7357

Some explanation: in fact, the tests don't pass anymore because we are using OpenLDAP specific filters for efficiency, that are not available in the test directory. 

So we need to split the "nodeinfoservice" service into an efficient version used in the real web app, and a "naive" one, that is used in test. 

Of course, we must share most of the logic, else we don't test anything. So the split is done only on the way we retrieve entries, and if we should update cache (always true).


ncharles:
We shall call that the "VW-Pull Request"

PR #954, opened 2015-11-06
Fixes #7383: Improve Rule compliance details loading

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7383

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
@@ -115,7 +115,6 @@ object RuleEditForm {
       chooseTemplate("component", "form", xml)
     }) openOr Nil
 
-
   val htmlId_groupTree = "groupTree"
   val htmlId_activeTechniquesTree = "userPiTree"
 }
@@ -198,7 +197,6 @@ class RuleEditForm(
                 s"#${htmlId_EditZone} *" #> { (n:NodeSeq) => SHtml.ajaxForm(n) } andThen
                 ClearClearable &
               "#ruleForm" #> formContent &
-              "#details"  #> new RuleCompliance(rule,directiveLib, nodeInfos, rootRuleCategory).display &
               actionButtons()
             ).apply(body)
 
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/snippet/HomePage.scala
```diff
@@ -332,6 +329,8 @@ class HomePage extends Loggable {
       n2               =  System.currentTimeMillis
       agentSoftEntries =  con.searchOne(acceptedNodesDit.SOFTWARE.dn, EQ(A_NAME, "rudder-agent"))
       agentSoftDn      =  agentSoftEntries.map(_.dn.toString).toSet
+      // Skip if there is no rudder-agent packages in software DN
+      if (! agentSoftDn.isEmpty)
```
fanf: With that, you are returning Empty, which was forbiden in the Tablet Of Law for "wtf is the semantic of Empy in an Error handling, Failable sentence ?".
Please just return Full(Map()) in that case, which should be the case without the if, I think (agentSoftEntries == Seq(), agentSoftDn == Set(), agentSoft == Seq(), perhaps just add a if(agentSoftDn.isEmpty) Seq() else { ... } for the nodeEntries part).

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -228,6 +226,7 @@ class RuleEditForm(
             | $$("#editRuleZone").bind( "show", function(event, ui) {
             | if(ui.panel.id== 'ruleComplianceTab') { ${ruleComplianceTabAjax}; }
             | });
+            |${Replace("details", new RuleCompliance(rule,directiveLib, nodeInfos, rootRuleCategory).display).toJsCmd};
```
fanf: OK, that's nice. Moreover, if it takes too long, it won't block the "update form" part, and let user do stuff in the meantime.

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

PR #955, opened 2015-11-10
Fixes #7375: Add an option in the web interface to completely disable reporting

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7375

### DIFF
--- a/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
@@ -42,7 +42,7 @@ import com.normation.cfclerk.domain.IntegerVType
 import com.normation.cfclerk.domain.BooleanVType
 import com.normation.cfclerk.domain.BasicStringVType
 import com.normation.cfclerk.domain.RegexConstraint
-
+import com.normation.rudder.reports.ComplianceModeName
 
 class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
 
@@ -164,8 +164,14 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                           , constraint = Constraint(
                                               typeName = BasicStringVType(
                                                   regex = Some(RegexConstraint(
-                                                      pattern = "(full-compliance|changes-only)"
-                                                    , errorMsg = s"Forbiden value, only 'full-compliane' and 'changes-only' are authorized"
+                                                      pattern = {
+                                                        val allModes = ComplianceModeName.allModes.map(_.name).mkString("(", "|", ")")
+                                                        allModes
+                                                      }
+                                                    , errorMsg = {
+                                                        val allModes = ComplianceModeName.allModes.map(_.name).mkString("'", "' or '", "'")
+                                                        s"Forbiden value, only ${allModes} are authorized"
+                                                      }
                                                   ))
                                               )
                                             , default=Some("full-compliance")
--- a/rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
@@ -44,25 +44,54 @@ import net.liftweb.json._
  * - error_only: only report for repaired and error reports.
  */
 
-sealed trait ComplianceMode {
-  def name: String
+sealed trait ComplianceModeName {
+  val name : String
 }
 
-final case object FullCompliance extends ComplianceMode {
+case object FullCompliance extends ComplianceModeName {
   val name = "full-compliance"
 }
-final case class ChangesOnly (
-  heartbeatPeriod : Int
-)extends ComplianceMode {
-  val name = ChangesOnly.name
-}
 
-object ChangesOnly {
+case object ChangesOnly extends ComplianceModeName {
   val name = "changes-only"
 }
 
+case object ReportDisabled extends ComplianceModeName {
+  val name = "reports-disabled"
+}
+
+object ComplianceModeName {
+  val allModes : List[ComplianceModeName] = FullCompliance :: ChangesOnly :: ReportDisabled :: Nil
+
+  def parse (value : String) : Box[ComplianceModeName] = {
+    allModes.find { _.name == value } match {
+      case None =>
+         Failure(s"Unable to parse the compliance mode name '${value}'. was expecting ${allModes.map(_.name).mkString("'", "' or '", "'")}.")
+      case Some(mode) =>
+        Full(mode)
+    }
+  }
+}
+
+sealed trait ComplianceMode {
+  def mode: ComplianceModeName
+  def heartbeatPeriod : Int
+  val name = mode.name
+}
+
+case class GlobalComplianceMode (
+    mode : ComplianceModeName
+  , heartbeatPeriod : Int
+) extends ComplianceMode
+
+case class NodeComplianceMode (
+    mode : ComplianceModeName
+  , heartbeatPeriod : Int
+  , overrideGlobal : Boolean
+) extends ComplianceMode
+
 trait ComplianceModeService {
-  def getComplianceMode : Box[ComplianceMode]
+  def getGlobalComplianceMode : Box[GlobalComplianceMode]
 }
 
 class ComplianceModeServiceImpl (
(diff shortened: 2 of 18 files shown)


rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
```diff
@@ -164,8 +164,11 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                           , constraint = Constraint(
                                               typeName = BasicStringVType(
                                                   regex = Some(RegexConstraint(
-                                                      pattern = "(full-compliance|changes-only)"
-                                                    , errorMsg = s"Forbiden value, only 'full-compliane' and 'changes-only' are authorized"
+                                                      pattern = {
+                                                        val allModes = ComplianceModeName.allModes.map(_.name).mkString("(", "|", ")")
+                                                        allModes
+                                                      }
+                                                    , errorMsg = s"Forbiden value, only 'full-compliance' and 'changes-only' are authorized"
```
ncharles: the error message seems incomplete - reports-disabled should be authorized

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
```diff
@@ -44,23 +44,66 @@ import net.liftweb.json._
...
-  val name = ChangesOnly.name
-}
 
-object ChangesOnly {
+final case object ChangesOnly extends ComplianceModeName {
   val name = "changes-only"
 }
 
+final case object ReportDisabled extends ComplianceModeName {
+  val name = "report-disabled"
```
ncharles: this should be reports-disabled

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
```diff
@@ -44,23 +44,66 @@ import net.liftweb.json._
...
+  override val overrideGlobal = true
+}
+
+case class NodeComplianceMode (
+    mode : ComplianceModeName
+  , heartbeatPeriod : Int
+) extends ComplianceMode {
+  override val overrideGlobal = true
+}
+*/
```
ncharles: why the comments ?

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
```diff
@@ -68,23 +111,22 @@ trait ComplianceModeService {
...
-       case eb : EmptyBox =>
-         val fail = eb ?~! "Could not get compliance mode name"
-         fail
-     }
+
+    for {
+      modeName <- readComplianceMode()
+      mode <- ComplianceModeName(modeName)
+      heartbeat <- readHeartbeatFreq()
+      overrideGlobal <- readOverride()
```
ncharles: could you align the arrows ?

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
```diff
@@ -110,19 +110,19 @@ trait ReadConfigService {
...
    */
-  def rudder_compliance_mode(): Box[(String,Int)] = {
+  def rudder_compliance_mode(): Box[ComplianceMode] = {
     for {
         name <- rudder_compliance_mode_name
+        modeName <- ComplianceModeName(name)
         period <- rudder_compliance_heartbeatPeriod
     } yield {
-      (name,period)
+      ComplianceMode(modeName,period, true)
```
ncharles: i'm a bit confused about the "true" at the end

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
```diff
@@ -110,19 +110,19 @@ trait ReadConfigService {
    */
   def rudder_store_all_centralized_logs_in_file(): Box[Boolean]
 
-
   /**
    * Compliance mode:
    * - "compliance": full compliance mode, where "success" execution reports are sent
    *   back from node to server, and taken into account for compliance reports,
    * - "error_only": only error and repaired are going to the server.
```
ncharles: this comment looks not up to date ...

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
VinceMacBuche: I was tricked because I looked into the first email about the issue which states 'report-diabled' ;)

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
VinceMacBuche: It was a test to have an different API Thank you

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
VinceMacBuche: In fact it has no meaning to have an override for global compliance mode, so we could set any value ...

I have to made a design choice here: 

either only have one 'ComplianceMode' class with 3 param ( name, hearbeat, override)
or 1 trait 'complianceMode' with 2 param (name, heartbeat) and 2 sublcass GlobalComplianceMode(nothing more than complianceMode), NodeComplianceMode with a third parameter (overrides) (In fact you can see an example of this in commented section on which you said "Why the comments ??" ) 


I don't know which one would be better ???

VinceMacBuche:
PR updated

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
```diff
@@ -44,23 +44,66 @@ import net.liftweb.json._
...
   val name = "full-compliance"
 }
-final case class ChangesOnly (
-  heartbeatPeriod : Int
-)extends ComplianceMode {
-  val name = ChangesOnly.name
-}
 
-object ChangesOnly {
+final case object ChangesOnly extends ComplianceModeName {
```
fanf: no need for final on object, they are already (you can look at the bytecode for that, or it's certainly specified somewhere in some scala spec). But you can let it here, I does no arm.

rudder-core/src/main/scala/com/normation/rudder/reports/ComplianceMode.scala
```diff
@@ -44,23 +44,66 @@ import net.liftweb.json._
...
 }
 
+final case object ReportDisabled extends ComplianceModeName {
+  val name = "report-disabled"
+}
+
+object ComplianceModeName {
+  val allModes : List[ComplianceModeName] = FullCompliance :: ChangesOnly :: ReportDisabled :: Nil
+
+  def apply (value : String) : Box[ComplianceModeName] = {
```
fanf: I'm not at ease with apply being able to fail - it looks to much like an object instanciation, and object instanciation failing are surprising (and a PITA to debug). 
All that to say: could you name you method "parse" ? (parse is totally expected to fail. In fact, most of the time, "parse" methods fail :)

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
fanf: "make impossible things unrepresentable": one trait, two implementations.

VinceMacBuche:
PR updated

VinceMacBuche:
Still some things to do (last @fanf remark) 

And also waiting new text about the new mode in the webapp should be ok verrryyyy soon

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

rudder-web/src/main/webapp/templates-hidden/components/ComponentComplianceMode.html
```diff
@@ -12,7 +12,11 @@
               network traffic, but provides more precise reporting and may be necessary to prove compliance in your organization.<br/>
               In <b>Changes only</b> mode, reports will only be sent when the agent makes a change or an error occurs on a node
               (these are 'repair' or 'error' reports). This mode saves a lot of log space and bandwidth, but leads to some assumptions
-              about actual configuration status in reporting.
+              about actual configuration status in reporting.<br/>
+              In <b>Disabled</b>b> mode, no reports will be sent, and rudder-agent will not re-configure the local syslog to send reports.
```
fanf: There is a b> that should not be here

rudder-web/src/main/webapp/templates-hidden/components/ComponentComplianceMode.html
```diff
@@ -32,10 +36,16 @@
           Changes only
         </label>
      </div>
+      <div class="radio">
+        <label>
+          <input type="radio" ng-model="complianceMode.name"  value="reports-disabled"/>
```
fanf: Should be report-disabled

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

PR #957, opened 2015-11-13
Fixes #7406: Dashboard incorrectly says \"0 node\" (should be nodes, in plural)

jooooooon:
https://www.rudder-project.org/redmine/issues/7406


VinceMacBuche:
0 is plural ok, right, get it.  

Thanks Jon !

PR #959, opened 2015-11-13
Fixes #7407: Display failed cf-promises to make it easier to re-run while debugging

jooooooon:
https://www.rudder-project.org/redmine/issues/7407

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
@@ -330,7 +330,7 @@ class Cf3PromisesFileWriterServiceImpl(
 
     sequence(toCheck) { case (agentType, paths) =>
       // Check the promises
-      val (errorCode, errors) =  {
+      val (errorCode, errors, checkCommand) =  {
         if(logger.isDebugEnabled) {
           //current time millis for debuging info
           val now = System.currentTimeMillis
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
```diff
@@ -356,6 +356,7 @@ class Cf3PromisesFileWriterServiceImpl(
          * and "error/technical details"
          */
         val completeErrorMsg = ( s"The generated promises are invalid!errormessage!cf-promise check fails for promises generated at '${paths.newFolder}'"
+                               + "<-Test command is: '" + checkCommand + "'"
                                + (if(errors.isEmpty) "" else errors.mkString("<-", "<-", ""))
```
jooooooon: This change should work because the "client side" cuts the message up using the "<-" separator, to replace it by a new line (```<br/>```).

jooooooon:
This whole PR is just a proposal for a quick fix to display the command run. This may not be the best approach - as you know my Scala experience is rather limited. Any feedback welcome.

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
```diff
@@ -356,6 +356,7 @@ class Cf3PromisesFileWriterServiceImpl(
          * and "error/technical details"
          */
         val completeErrorMsg = ( s"The generated promises are invalid!errormessage!cf-promise check fails for promises generated at '${paths.newFolder}'"
+                               + "<-Test command is: '" + checkCommand + "'"
```
VinceMacBuche: I'm wondering if the message could be better with "promise check command is:" 
What do you think @jooooooon ?

VinceMacBuche:
Apart the question on the text, it's ok to me and can be merged (once we are ok about the text ;) )

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
jooooooon: I propose "Command to check generated promises is:"

jooooooon:
PR updated

PR #960, opened 2015-11-13
Fixes #7408: Empty inputs list is generated with double quotes even if empty

jooooooon:
https://www.rudder-project.org/redmine/issues/7408

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
@@ -291,7 +291,14 @@ class PrepareTemplateVariablesImpl(
     }
 
     //utilitary method for formating an input list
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -291,7 +291,14 @@ class PrepareTemplateVariablesImpl(
     }
 
     //utilitary method for formating an input list
-    def formatInputs(x: Seq[(Technique, String)]) = x.map(_._2).distinct.mkString("\"", s"""",\n${" "*14}"""", s""""\n  """)
+    def formatInputs(x: Seq[(Technique, String)]) = {
+      val inputs = x.map(_._2).distinct
+      if (inputs.isEmpty) {
+        ""
+      } else {
+        x.map(_._2).distinct.mkString("\"", s"""",\n${" "*14}"""", s""""\n  """)
```
VinceMacBuche: You can replace 'x.map(_._2).distinct' by 'inputs' since it is already computed :)

PR #964, opened 2015-11-15
Fixes #7414: Allow user to select level of compliance details in API response

fanf:
https://www.rudder-project.org/redmine/issues/7414

### DIFF
--- a/rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
+++ b/rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
@@ -137,7 +137,7 @@ import com.normation.rudder.reports.ComplianceModeService
 import com.normation.rudder.reports.ComplianceModeServiceImpl
 import com.normation.rudder.reports.AgentRunIntervalService
 import com.normation.rudder.reports.AgentRunIntervalServiceImpl
-import com.normation.rudder.web.rest.compliance.ComplianceAPI6
+import com.normation.rudder.web.rest.compliance.ComplianceAPI7
 import com.normation.rudder.web.rest.compliance.ComplianceAPIService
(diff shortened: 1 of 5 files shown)


fanf:
The API version should be updated, updating PR

fanf:
PR updated

VinceMacBuche:
Here, all v6 api call will have 'mode' and could have 'compliance level' set which is not expected ... IT should be in two separated output (or you don't have to create a new version ...)

VinceMacBuche:
Plus I'm not at ease with 'integer' levels, they have no meaning to me
What about using the same models like 'categories' or 'nodes' with string levels ?

Levels should be 
"Rules"
"Directives"
"Components"
"Nodes"
"Full"



In that order for Rules and for Nodes:

"Nodes"
 "Rules"
"Directives"
"Components"
"Full"


Maybe No needs for "Nodes" in the first list, and maybe no need for 'components' in the second list

fanf:
Well, that's not a problem, I think: a tool built for v6 will still work as expected on Rudder 3.2. And on 3.2, people will have v7 at disposition. So I'm really not sure it's interesting.

fanf:
I didn't used named level because it's cumbersome, error prone, and not at all what is expected with "tree like" structure. Level is common word for that, and the level semantic is rather standard.

VinceMacBuche:
Yeah, I agreee with you that it won't be a problem for tools built over v6, but the documentation would be wrong for v6 (on 3.2)

VinceMacBuche:
Ok for 'integer' level, not 100% agree, but ok!

And if people are confused, we could still change in a next version, but maybe they will not be and i'll be the alone confused guy in the world :(

fanf:
ok for the version

fanf:
PR updated

VinceMacBuche:
Great \o/ ! Merging

PR #965, opened 2015-11-16
Fixes #7416: Compliance detail and compliance summary use different colours

jooooooon:
https://www.rudder-project.org/redmine/issues/7416


rudder-web/src/main/webapp/javascript/rudder/angular/apiAccount.js
```diff
@@ -113,7 +113,7 @@ $scope.columnDefs = [
       , 'bSortable': false
       , "sWidth": "10%"
     , "fnCreatedCell" : function (nTd, sData, oData, iRow, iCol) {
-      var color = sData ? "#CCFFCC" : "#FF6655";
+      var color = sData ? "#5cb85c" : "#d9534f";
```
jooooooon: I changed the colour here too, even though it's not related to compliance, because it was the same green, and it's better to be consistent across the whole UI.

PR #966, opened 2015-11-17
Fixes #7376: Authorize both  path relative to technique and to config-repos in technique metadata.xml descriptor

fanf:
https://www.rudder-project.org/redmine/issues/7376


rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueTemplate.scala
```diff
@@ -40,18 +40,21 @@ import com.normation.utils.HashcodeCaching
...
-  override def toString() = techniqueId.toString + "/" + name
+sealed trait TechniqueResourceId {
+  def name: String
 }
 
+//the one relative to the technique
+final case class TechniqueResourceIdByName(techniqueId: TechniqueId, name: String) extends TechniqueResourceId
+
+//the one relative to configuration-repository - actually, that resource is not really link to a technique
+final case class TechniqueResourceIdByPath(parentDirectories: List[String], name: String) extends TechniqueResourceId
```
ncharles: you should explain why it takes a list of parent directory, and that if the list is empty, the behaviour is different

rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -384,18 +384,29 @@ class GitTechniqueReader(
...
-    //build a treewalk with the path, given by Cf3PromisesFileTemplateId.toString
-    val path = cf3PromisesFileTemplateId.toString + Cf3PromisesFileTemplate.templateExtension
+  override def getTemplateContent[T](techniqueResourceId: TechniqueResourceId)(useIt : Option[InputStream] => T) : T = {
+    //build a treewalk with the path, given by TechniqueTemplateId
+    val filenameFilter = {
+      val name = techniqueResourceId.name + TechniqueTemplate.templateExtension
+      techniqueResourceId match {
+        case TechniqueResourceIdByName(tid, _) =>
+          new FileTreeFilter(canonizedRelativePath, s"${tid.name}/${tid.version.toString}/${name}")
+        case TechniqueResourceIdByPath(Nil, _) =>
```
ncharles: am I understanding correctly that the extension will always be .st ?
I thought we decided we could have any extension ?
Or is it part of the next PR ?

fanf: It will be on the PR for ticket http://www.rudder-project.org/redmine/issues/7402

fanf:
As discussed, I'm self-merging this one, as the comment will be addressed in an other PR.

PR #967, opened 2015-11-19
Fixes #7402: Add a <FILE> tag in metadata.xml to allow simple file copy into techniques

fanf:
https://www.rudder-project.org/redmine/issues/7402

This ticket add the `<FILE>` tag in Techniques for resources that we just want to copy for each node. 

It also bring some refactoring in the "write promise" pipeline, trying to make it a little clearer (create some datastructure to holds things, rename other things to denote what they actually are, and not what they were 5 years ago, etc...) and more correct (for instance, expected_reports.csv is now written only *one* time for each node/agent, not one time for each template of each technique...).


ncharles:
This looks correct, thank you !

PR #968, opened 2015-11-23
Fixes #7437: Embedded docs don't display the correct index.html file

jooooooon:
https://www.rudder-project.org/redmine/issues/7437


rudder-web/src/main/resources/rudder-apache-common.conf
```diff
@@ -151,7 +151,6 @@ SetEnvIf X-API-Version "[^1]" api_deny
 # Link to Rudder documentation
 Alias /rudder-doc /usr/share/doc/rudder/html
 <Directory /usr/share/doc/rudder/html>
-  DirectoryIndex rudder-doc.html
```
jooooooon: Removing this line causes it to revert to the default, which is index.html, which is what our docs now use.

PR #969, opened 2015-11-23
Fixes #7429: Files added with the <FILE> tag in rudderify are not included in the input list

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7429


ncharles:
awesome, thank you

PR #970, opened 2015-11-23
Fixes #7442: License file is not generated any more

ncharles:


fanf:
This really look like dead code, the next time a refactoring occurs, it will be deleted again. 
So either use the seq of licences as a parameter of the copyLicence method, or change the method name to something like "load licence from fs" or something clearly stating that you don't want to use the result but are only interested on the side effect. 

I'm going to propose a modification and merge it so that it goes in the next nightly

ncharles:
actually, before we simply called the loadLicenses, but the change to 3.2 rendered that part private

fanf:
We merged #971

PR #971, opened 2015-11-23
Fixes #7442: License file is not generated any more

fanf:
https://www.rudder-project.org/redmine/issues/7442

Alternative to #970 with some refactoring so that: 
- licences are read only ONE time by generation process, and actually one (with the current implementation, they are not, since the getalllicences methods wasn't called at the begining of the generation process)
- all dead code regarding licenses and licenses repository is removed

Summary: we are getting all the licenses info at the begining of the generation process, and give that map of (nodeId, licenses) to systemVariableService and promiseWritter so they can use it directly without re-reading XML and with consistent data all along generation process.


rudder-core/src/main/scala/com/normation/rudder/repository/xml/LicenseRepositoryXML.scala
```diff
@@ -45,49 +44,46 @@ import scala.xml._
...
-    licenseMap.clear
-
-    logger.debug("Loading document {}", licenseFile)
-    val doc = loadLicenseFile()
-
-    for (elt <- (doc \\"licenses" \ "license")) {
-      logger.debug("Loading License")
-      val license = NovaLicense.parseXml(elt)
-      licenseMap.put(license.uuid, license)
+  private[this] def loadLicenses(): Box[Unit] = {
```
VinceMacBuche: Why not returning the map here and get rid of the mutable var which saddens me :(

ncharles: i indeed fail to see the benefit of the map, a def returning the result would do the same, especially since we are now loading the licences at each generation

PR #974, opened 2015-11-25
Fixes #7454: Directives with double quotes in their name need to be escaped before going in rudder_directives.cf and rudder_expected_reports.csv

fanf:
https://www.rudder-project.org/redmine/issues/7454

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
@@ -86,6 +86,74 @@ trait PrepareTemplateVariables {
 
 }
 
+object PrepareTemplateVariables extends Loggable {
+
+  /*
+   * utilitary method for formating list of "promisee usebundle => bundlename;"
+   */
+  def formatMethodsUsebundle(x:Seq[(String, Bundle)]): String = {
+
+    val alignWidth = if(x.size <= 0) 0 else x.map(_._1.size).max
+    x.map { case (promiser, bundle) =>
+      val escaped = ParameterEntry.escapeString(promiser)
+      //the promiser value (may) comes from user input, so we need to escape "
+      s""""${escaped}"${" "*Math.max(0, alignWidth - escaped.size)} usebundle => ${bundle.name};"""
+    }.mkString( "\n")
+  }
+
+  /*
+   * utilitary method for formating an input list
+   */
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -86,6 +86,76 @@ trait PrepareTemplateVariables {
...
+  /*
+   * utilitary method for formating an input list
+   */
+  def formatBundleFileInputs(x: Seq[String]) = {
+    val inputs = x.distinct
+    if (inputs.isEmpty) {
+      ""
+    } else {
+      inputs.mkString("\"", s"""",\n${" "*4}"""", s"""",""")
+    }
```
VinceMacBuche: In fact you don't need the spaces

current output (with spaces)is: 

    inputs => {
      "rpmPackageInstallation/7.0/rpmPackageInstallation.cf",
          "test_3_2/1.0/test_3_2.cf",
    };

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -86,6 +86,76 @@ trait PrepareTemplateVariables {
...
+   * Sort the techniques according to the order of the associated BundleOrder of Cf3PolicyDraft.
+   * Sort at best: sort rule then directives, and take techniques on that order, only one time
+   * Sort system directive first.
+   *
+   * CAREFUL: this method only take care of sorting based on "BundleOrder", other sorting (like
+   * "system must go first") are not taken into account here !
+   */
+  def sortTechniques(nodeId: NodeId, techniques: Seq[Technique], container: Cf3PolicyDraftContainer): Seq[(Technique, List[BundleOrder])] = {
+
+    def sortByOrder(tech: Seq[Technique], container: Cf3PolicyDraftContainer): Seq[(Technique, List[BundleOrder])] = {
```
VinceMacBuche: Why using this function ? you call it directly afterwars without treatment before :(

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
VinceMacBuche: Mybe you'll need to adapt the test :)

fanf:
PR updated

PR #981, opened 2015-11-30
Fixes #7514: Read technique access does not give access to techniques

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7514


VinceMacBuche:
PR updated

PR #984, opened 2015-12-03
Fixes #7536: Make possible to delete ANY active technique

fanf:
https://www.rudder-project.org/redmine/issues/7536

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
@@ -110,6 +110,7 @@ object TechniqueEditForm {
   val htmlId_addToActiveTechniques = "addToActiveTechniques"
   val htmlId_userCategoryDetails = "userCategoryDetails"
   val htmlId_addUserCategoryForm = "addUserCategoryForm"
+
 }
 
 /**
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
```diff
@@ -210,7 +212,9 @@ class TechniqueEditForm(
         case e:EmptyBox => NodeSeq.Empty
```
VinceMacBuche: There should be a log here or at lower level if the request failed (if it had to be launched)

fanf: I'm not sure, the code seems to use the box as an Option. I still can log on failure.

VinceMacBuche: You are right, it use empty as "there is no active technique", so don't log on emptybox but only on failure

VinceMacBuche: It's just that it's impossible to know that it fails

fanf:
pr updated

rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
```diff
@@ -147,9 +149,18 @@ class TechniqueEditForm(
 //  }
 
 
-  private[this] var currentActiveTechnique = roActiveTechniqueRepository.getActiveTechnique(technique.id.name)
+  private[this] var currentActiveTechnique: Box[ActiveTechnique] =  Box(activeTechnique).or(Box(technique).flatMap(t => roActiveTechniqueRepository.getActiveTechnique(t.id.name)))
   private[this] var uptCurrentStatusIsActivated = currentActiveTechnique.map( _.isEnabled)
 
+  currentActiveTechnique match {
+    case f: Failure =>
+      logger.warn(s"An error was encoutered when trying to find a technique in user library: ${f.messageChain}")
```
VinceMacBuche: encountered :)

fanf:
PR updated

PR #994, opened 2015-12-09
Fixes #7566: Create a test on promise write to check template filling

fanf:
https://www.rudder-project.org/redmine/issues/7566


rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
```diff
@@ -210,6 +210,10 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                           , multivalued = false
                                           , isUniqueVariable = true
       )
+//    , SystemVariableSpec("RUDDER_NODE_GROUPS" , "The array of group_UUID => group_NAME for the node, in a formatted string"
+//                                          , multivalued = false
+//                                          , isUniqueVariable = true
+//      )
```
ncharles: why the comments ?

fanf: ah, you're right, they should not be here.

rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
```diff
@@ -210,6 +210,10 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                           , multivalued = false
                                           , isUniqueVariable = true
       )
+//    , SystemVariableSpec("RUDDER_NODE_GROUPS" , "The array of group_UUID => group_NAME for the node, in a formatted string"
```
VinceMacBuche: This needs to be added in anothe PR (but since it's commented it's not that important ...) 

Just dead code until next PR

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -255,32 +254,37 @@ class PrepareTemplateVariablesImpl(
...
     // fill the variable
-    (for {
-      technique <- techniques
-    } yield {
-      val ptValues = variablesValues(technique.id)
+    (bestEffort(techniques) { technique =>
 
-      val variables:Seq[Variable] = (for {
-        variableSpec <- technique.getAllVariableSpecs
+      val ptValues = variablesValues(technique.id)
```
VinceMacBuche: ptvalues :(

VinceMacBuche:
A little pull request, I expect at least one per week like that :)

VinceMacBuche:
mergin

PR #995, opened 2015-12-11
Fixes #7424: Add cfengine classes for each group of a node

fanf:
https://www.rudder-project.org/redmine/issues/7424

You can safely ignore all the file renaming at the end, it's just that I changed the name of the root dir for the default installation promises. 

So, the interesting part is to know if the files expected-share/test-rudder-groups/{no-group.cf, some-groups.cf} are the awaited output for the need. 

That PR must be merded at the same time as http://www.rudder-project.org/redmine/issues/7599


rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -356,12 +361,57 @@ class SystemVariableServiceImpl(
...
+      }
+      (id, info.name)
+    }
+    val nodeMaxString = if(nodeGroups.isEmpty) 0 else nodeGroups.flatMap { case (a,b) => a.size :: b.size :: Nil }.max
+    val stringNodeGroupsVars = if(nodeGroups.isEmpty) {
+      ""
+    } else {
+      nodeGroups.map { case (id, name) =>
+        s"""by_uuid["${id}"] ${" "*(nodeMaxString-id.size)} string => "${name}",\n""" +
+        s"""            ${" "*(nodeMaxString)        }   meta => { "inventory", "attribute_name=rudder_groups" };"""
```
ncharles: i thought we decided to have the specifics content in the template, so that it was easier to evolve?

fanf: It was much, much harder to code and forced to put specific system variable cases outside all (most) of them are proceeded. 
More precisely, it would have supposed to build objects a little like we do for global parameters, and these objects must be hand-feeded to string template by building dedicated STVariables. That implies to have access to node and group info at that point of the code (quite far in the writting process), which we don't have.

rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -356,12 +361,57 @@ class SystemVariableServiceImpl(
...
+     */
+    //build the list of nodeId -> names, taking care of special nodeIds for special target
+    val nodeGroups = allGroups.getTarget(nodeInfo.id, allNodeInfos).map { case(target, info) =>
+      val id = info.target.target match {
+        case GroupTarget(id) => id.value
+        case t => t.target
+      }
+      (id, info.name)
+    }
+    val nodeMaxString = if(nodeGroups.isEmpty) 0 else nodeGroups.flatMap { case (a,b) => a.size :: b.size :: Nil }.max
```
ncharles: why do you need the size of "name" ? isn't it only the left-hand-side of the => that needs to be aligned ?

ncharles: ha, it's for the classes, ok

rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
```diff
@@ -198,17 +216,12 @@ class WriteSystemTechniqueSpec extends Specification with Loggable with ContentM
   //////////// set-up auto test cleaning ////////////
   override def afterAll(): Unit = {
     logger.info("Deleting directory " + abstractRoot.getAbsoluteFile)
-    FileUtils.deleteDirectory(abstractRoot)
+    //FileUtils.deleteDirectory(abstractRoot)
```
ncharles: don't leave commented code here

rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
```diff
@@ -97,23 +107,31 @@ import net.liftweb.common.Loggable
 @RunWith(classOf[JUnitRunner])
 class WriteSystemTechniqueSpec extends Specification with Loggable with ContentMatchers with AfterAll {
 
+  //just a little sugar to stop hurting my eyes with new File(blablab, plop)
+  implicit class PathString(root: String) {
+    def /(child: String) = new File(root + "/" + child)
```
ncharles: you are changing the semantic, by replacing File(path, filename) to File(path+"/"+filename)

rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
fanf: oups. And finally, I thought that the code may not evolve that much, because the name of the var will be freezed as soons as we release the version (public API). And other changes would probably need a rudder update, to (so not grave if in scala code). 
Also, it seems that the pasing of template with variables/loop is much longer (we are talking about a few ms, and in a unique file, so that may not be relevant - especially given the fiability of the test). 
And vertical alignement was much harder to do, to (not a real blocker ;)

So, all in all, it seems that the hypothetic futur gain compared present loss was not in favor of the previous choice. I can still try to implement it if you are thinking it is the best solution.

rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
fanf: rha, that's three time I remove the comment and forget it in the next push...

rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
fanf: I don't think so - how ?

ncharles: File(path, filename) is cross platform, while File(path+"/"+filename) requieres having / as a separator

ncharles: http://docs.oracle.com/javase/7/docs/api/java/io/File.html

fanf: Ah, we don't care, tests won't work on windows due to the "/tmp" in all case. And there is a whole lot of hard codede hypothesys that tests, even others, are run on a unix like plateform.

ncharles: still, there is no reasons to worsen the situation by using string with hardcoded separator

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
ncharles: ok

rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
```diff
@@ -387,6 +387,25 @@ object RuleTarget extends Loggable {
...
+   * prefixed by "group_"
+   */
+  def toCFEngineClassName(target: String) = {
+    //normalisation process:
+    // 1) to asccii
+    // 2) cfengine normalisation, replacing all non [alphanum-]
+    ///   char by _
+
+    // from http://stackoverflow.com/a/2413228/436331
+    def toAscii(s: String) = java.text.Normalizer.normalize(s, java.text.Normalizer.Form.NFD).replaceAll("""[\p{Mn}\p{Me}]+""", "")
```
ncharles: the reference from stackoverflow doesn't exhibit the same example; is it on purpose ?

fanf: ah yes, missing the second reference: http://stackoverflow.com/a/5697575/436331

fanf:
pr updated

PR #1002, opened 2015-12-15
Fixes #7596: Add the classes names for the group in API

fanf:
https://www.rudder-project.org/redmine/issues/7596


fanf:
PR updated with the correct name for json attribute

PR #1003, opened 2015-12-15
Fixes #7612: Keep 4 days of reports by default on new install

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7612


jooooooon:
:+1:

PR #1007, opened 2015-12-16
Fixes #7613: Remove popup when creating Directives

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7613


fanf:
Could you rebase and then, ok for self-merging!

PR #1009, opened 2015-12-17
Fixes #7634: Make NodeInfo more dependant on Node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7634


fanf:
I'm OK with that modification, but I don't understand why you want to put it in master ? It lead to bugs in the past, it should be corrected everywhere. And it's not an architectural modification, just some factorisation avoiding copy error...

VinceMacBuche:
CLosing replacing by new PR on 2.11,

PR #1010, opened 2015-12-17
Fixes #7636: Put \"Not applicable\" report in green status in compliance bar

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7636


ncharles:
technically, this is ok, but removing the N/A is a bit strange, isn't it ?

ncharles:
this does need a functionnal review

jooooooon:
@VinceMacBuche and I decided about this together. A lot of feedback has shown that having N/A as a separate block is confusing, and prevents ever seeing "100%" in a progress bar, which causes frustration.

We are not however removing the concept or the information about N/A, since the "tooltip" you see when you hover your mouse over the green section of the bar provides this detail, and the detailed compliance still shows N/A as the result for applicable components.

PR #1014, opened 2015-12-21
Fixes #7647: CFEngine does not parse group classes with uuid

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7647


peckpeck:
Ok merging

PR #1016, opened 2015-12-24
Fixes #7689: Error on promise generation after accepting a node

ncharles:
https://www.rudder-project.org/redmine/issues/7689

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/servers/NewNodeManager.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/servers/NewNodeManager.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/servers/NewNodeManager.scala
```diff
@@ -458,6 +458,7 @@ trait ComposedNewNodeManager extends NewNodeManager with Loggable {
       case Full(seq) => //ok, cool
         logger.info(s"New node accepted and managed by Rudder: ${id.value}")
         cacheToClear.foreach { _.clearCache }
+        updateDynamicGroups.startManualUpdate
```
ncharles: this is pretty much adding back https://github.com/Normation/rudder/pull/986/files#diff-fb9876df49b189786ae8f6c12bb71679L454

VinceMacBuche: Maybe cache building should trigger an update of groups ! But great ! Thanks ncharles

PR #1017, opened 2015-12-28
Fixes #7692: Renaming AND moving the \"Reports\" tab to \"Compliance\" in node overview is too confusing

jooooooon:
https://www.rudder-project.org/redmine/issues/7692


peckpeck:
OK, merging

PR #1020, opened 2016-01-08
Fixes #7743: Compliance take into account expired run

fanf:
https://www.rudder-project.org/redmine/issues/7743


ncharles:
Yeah !

PR #1022, opened 2016-01-12
Fixes #7634: Make NodeInfo more dependant on Node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7634

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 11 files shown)


VinceMacBuche:
PR updated

rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
```diff
@@ -62,13 +60,21 @@ case class NodeInfo(
...
   //TODO: remove isPolicyServer, and pattern match on
   //      on role everywhere.
   , serverRoles   : Set[ServerRole]
-) extends HashcodeCaching
+) extends HashcodeCaching {
+
+  val id                         = node.id
+  val name                       = node.name
+  val description                = node.description
+  //val creationDate               = node.creationDate
```
fanf: Why do you remove creationDate, nodeReportingConfiguration and properties which are not in Node ?

fanf: OK, at least creationDate (not sure the other two are actually used)

VinceMacBuche:
PR updated

PR #1025, opened 2016-01-12
Fixes #7625: Not accepted node reports trigger compliance processing

fanf:
https://www.rudder-project.org/redmine/issues/7625

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
-* 
+*
 * In accordance with the terms of section 7 (7. Additional Terms.) of
 * the GNU General Public License version 3, the copyright holders add
 * the following Additional permissions:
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -167,29 +167,31 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
-        }
-      }
-
       for {
-        newStatus <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(expired, Set())
+        allNodeIds        <- nodeInfoService.getAll.map( _.keySet)
+        //only try to update nodes that are accepted in Rudder
+        nodeIds            =  nodeIdsToCheck.intersect(allNodeIds)
+        /*
+         * Three cases: 1/ cache does not exist, 2/ cache exists but expiration date expired, 3/ cache does note exists
```
VinceMacBuche: Hmmm seems 1/ and 3/ are the same case (3 with typo ) one of them should be "cache does exists ? "

VinceMacBuche: From what I Read, 3/ should be  "cache exists"

VinceMacBuche: I need to read more :)

VinceMacBuche: 1 should be : cache exists, 3 should be: cache does not exist ?

PR #1026, opened 2016-01-12
Fixes #7753: Improve description of compliance mode

amousset:
https://www.rudder-project.org/redmine/issues/7753


jooooooon:
@amousset I already improved this text in the 3.2 branch. I agree this could be improved back in 3.0, but why don't you take the same text, at least as a working base? Then we can improve it everywhere.

amousset:
I had not noticed it had been changed in 3.2, I will use the new version as working base.

amousset:
PR updated

jooooooon:
In fact maybe it would be worthwhile you going through other changes from that same PR and selecting any other improved texts you think would benefit 3.0/3.1 ?

amousset:
Yes, preparing another PR.

PR #1028, opened 2016-01-13
Fixes #7757: Blue colour in \"Nodes by overall compliance\" chart is confusingly the same as \"No report\" blue

jooooooon:
https://www.rudder-project.org/redmine/issues/7757


VinceMacBuche:
So badass in Scala @jooooooon ;)

PR #1029, opened 2016-01-13
Fixes #7766: Too long directive name breaks directive serialization

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7766

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
-* 
+*
 * In accordance with the terms of section 7 (7. Additional Terms.) of
 * the GNU General Public License version 3, the copyright holders add
 * the following Additional permissions:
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -152,7 +151,7 @@ class DirectiveUnserialisationImpl extends DirectiveUnserialisation {
       fileFormatOk          <- TestFileFormat(directive)
       id                    <- (directive \ "id").headOption.map( _.text ) ?~! ("Missing attribute 'id' in entry type directive : " + xml)
       ptName                <- (directive \ "techniqueName").headOption.map( _.text ) ?~! ("Missing attribute 'techniqueName' in entry type directive : " + xml)
-      name                  <- (directive \ "displayName").headOption.map( _.text ) ?~! ("Missing attribute 'displayName' in entry type directive : " + xml)
+      name                  <- (directive \ "displayName").headOption.map( _.text.trim ) ?~! ("Missing attribute 'displayName' in entry type directive : " + xml)
```
ncharles: this means that if we store space in names, they will get lost on archive restore
like, after an upgrade, space will disapear

ncharles: ok, everything is indeed trimed at creation and modification

VinceMacBuche: I think we don't want space at the beginning and at the end of names in Rudder ( and almost sure for anything that is not a configuration parameter, directive parameter )

PR #1033, opened 2016-01-14
Fixes #7758: When several directives value have a CFEngine var, we get \"unexpected\" due to bad regex matching

fanf:
https://www.rudder-project.org/redmine/issues/7758
The PR is quite big but there is in fact 4 things in it:
- 1. I factored out some logic in test so that it's much more easy to write test for value with CFEngine variables. Most of ExecutionBatchTest.scala changes are due to that, but there was ALSO uncoherencies and things that were not good, showed by 2. and 3. above
- 2. I found a bug with the logic for getting simple value from reports. The algo was too greedy and was not letting "unused" reports to possible match for cfengine variables
- 3. We weren't merging to worst type the values for the same component/directive/rule/node. In practice, it may have never been seen, because that case is only for value with cardinality > 1, but somes tests were testing a bad expectation
- 4. And finally, I rebuild most of the logic for matching CFEngine vars. The high level view is that we are testing all reports for all patterns, with a lot of pruning along the side (i.e: each time we are finding a pattern matching exactly one report, we remove that report for testing, pruning any new case where a pattern precendently matching two reports now match one)

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


ncharles:
looks good (but this is complex...)

PR #1036, opened 2016-01-20
Fixes #7814: Promise generation never finishes if jetty is restarted mid-generation

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7814

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
-* 
+*
 * In accordance with the terms of section 7 (7. Additional Terms.) of
 * the GNU General Public License version 3, the copyright holders add
 * the following Additional permissions:
@@ -22,12 +22,12 @@
 * documentation that, without modification of the Source Code, enables
 * supplementary functions or services in addition to those offered by
 * the Software.
-* 
+*
 * Rudder is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
-* 
+*
 * You should have received a copy of the GNU General Public License
 * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
 
@@ -36,7 +36,6 @@
 */
 
 package com.normation.rudder.batch
-
 
 import net.liftweb.common._
 import net.liftweb.actor._
@@ -57,6 +56,7 @@ import com.normation.eventlog.ModificationId
 import com.normation.eventlog.ModificationId
 import com.normation.eventlog.ModificationId
 import com.normation.eventlog.ModificationId
+import java.io.File
 
 //ask for a new deployment - automatic deployment !
 //actor: the actor who asked for the deployment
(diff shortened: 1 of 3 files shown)


rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
```diff
@@ -162,14 +157,50 @@ final class AsyncDeploymentAgent(
     )
   }
 
-  override protected def lowPriority = {
+  // That flag file is used to determine if a policy update was running when the webapp stopped so we can launch a new one when starting
+  val policyUpdateRunningFlagPath = "/opt/rudder/etc/policy-update-running"
```
fanf: That must be a parameter, perhaps even something coming from config (with a default value if not provided), and documented. 
More over, it should go to /var/rudder, no ?

rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
```diff
@@ -162,14 +157,50 @@ final class AsyncDeploymentAgent(
...
+    }
+  }
+  override protected def lowPriority = {
 
     //
     // Start a new deployment
     //
     case AutomaticStartDeployment(modId, actor) => {
       implicit val a = actor
+      createFlagFile
```
fanf: I believe that we should fail and stop the process if we don't succeed in file creation, no ?
Else, it defeat the purpose of the flag ?

rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
VinceMacBuche: other 'flags' go to /opt/rudder/etc so we decided with @jooooooon to keep the same pattern

fanf: erk. 
OK

rudder-web/src/main/scala/bootstrap/liftweb/checks/ResumePolicyUpdateRunning.scala
```diff
@@ -0,0 +1,81 @@
...
+    val filePath = asyncGeneration.policyUpdateRunningFlagPath
+    // Check if the force technique reload file is present, and reload technique library if needed
+    val file =  new File(filePath)
+    try {
+      if (file.exists) {
+        // File exists, reload
+        logger.info(s"Flag file '${filePath}' found, Start a new policy update now")
+        asyncGeneration ! AutomaticStartDeployment(ModificationId(uuidGen.newUuid), RudderEventActor)
+      } else {
+        logger.info(s"Flag file '${filePath}' does not exists, No need to start a new policy update")
```
fanf: Not sure we want that at info level. Is it the same level for the equivalent message for technique library reload ?

rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
fanf: Does the file path for technique reload is configurable ?

rudder-web/src/main/scala/bootstrap/liftweb/checks/ResumePolicyUpdateRunning.scala
VinceMacBuche: Yes same level

rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
VinceMacBuche: No, it's fixed like this

rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
VinceMacBuche: What's bothering me with stopping the process is that it prevents generation...

And so it would prevent new policies to be generated in a normal case (if the flag creation failed)

fanf: ok, in all case that promise generation must succeed because it won't be restarted.

VinceMacBuche:
PR updated

rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentAgent.scala
jooooooon: I agree it's a bit ugly to have a state file under etc, but you can look at it the other way round, and see it as a flag to signal to the webapp it has to do something. Linux uses a similar mechanism during shutdown, when in touches /etc/nologin to prevent sshd/telnetd/etc to allow new users to login - it is both a config file and a state file, from different perspectives.

PR #1043, opened 2016-02-05
Fixes #6722: Order of button on \"accept nodes in rudder\" is the exact opposite of the rest of the webapp

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/6722

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
-* 
+*
 * In accordance with the terms of section 7 (7. Additional Terms.) of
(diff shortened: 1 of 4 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
```diff
@@ -356,20 +352,23 @@ class AcceptNode {
...
-      "accept" -> {if (seq.size > 0 ) { SHtml.ajaxButton("Accept into Rudder", {
-        () =>  showConfirmPopup(acceptTemplate, "confirmPopup")
-      }) % ("style", "width:170px")} else NodeSeq.Empty},
-      "refuse" -> {if (seq.size > 0 ) { SHtml.ajaxButton("Refuse", {
-        () => showConfirmPopup(refuseTemplate, "refusePopup")
-      }) } else NodeSeq.Empty},
-      "errors" -> (errors match {
+      )
+    , "accept" ->
+          SHtml.ajaxButton(
```
fanf: Why removing the guard when no node selected here ? (the if( seq.size > 0) ...)

fanf: For testimonial: having these test breaks the UI when no node are present, but does not change the behaviour when there was no node selected (i.e: it was just an UI inconsistancy when no nodes where available). So it's an enhancement to remove them.

PR #1044, opened 2016-02-05

VinceMacBuche:

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
 * Rudder is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
-* 
+*
 * In accordance with the terms of section 7 (7. Additional Terms.) of
 * the GNU General Public License version 3, the copyright holders add
 * the following Additional permissions:
@@ -22,12 +22,12 @@
 * documentation that, without modification of the Source Code, enables
 * supplementary functions or services in addition to those offered by
 * the Software.
-* 
+*
 * Rudder is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
-* 
+*
 * You should have received a copy of the GNU General Public License
 * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
 
@@ -59,8 +59,6 @@ import com.normation.utils.HashcodeCaching
  *
  */
 class Section2FieldService(val fieldFactory: DirectiveFieldFactory, val translators: Translators) extends Loggable {
-
-
 
   /**
    * Fully initialize a DirectiveEditor from a list of variables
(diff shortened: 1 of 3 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
```diff
@@ -254,8 +250,8 @@ class Section2FieldService(val fieldFactory: DirectiveFieldFactory, val translat
         for {
           nameValues <- seqOfNameValues
         } yield {
-          val valueOpt = try Some(nameValues.values(i)) catch { case e: Exception => None }
-          (nameValues.name, valueOpt)
+          val valueOpt = try Some(nameValues.values(i)) catch { case e: Exception => nameValues.name.constraint.default }
```
VinceMacBuche: I want to try some refactoring here before merging

VinceMacBuche:
PR updated

VinceMacBuche:
PR updated

rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
```diff
@@ -223,41 +220,40 @@ class Section2FieldService(val fieldFactory: DirectiveFieldFactory, val translat
...
-  // to
-  // Seq( Map((A -> "A1"), (B -> "B1")),
-  //      Map((A -> "A2"), (B -> "B2")) )
-  //If there is no value, a None is returned
+  // Transforms map of all values by Variable Name
+  // Into the Sequence of all values by section
+  // A section will be defined as a Map of value by variable name
+  //   * ie: Map(A -> Seq("A1", "A2"), B -> Seq("B1", "B2"))  => Seq( Map((A -> "A1"), (B -> "B1"))
+  //                                                                , Map((A -> "A2"), (B -> "B2")) )
+  // If there is no value, a Section with default value is returned is returned
```
fanf: is returned overflow

rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
```diff
@@ -223,41 +220,40 @@ class Section2FieldService(val fieldFactory: DirectiveFieldFactory, val translat
...
-    // seq of variable values with same name correctly ordered
-    val seqOfNameValues : Seq[NameValuesVar] = {
-      for {
-        varSpec <- section.getAllVariables
-      } yield {
-        NameValuesVar(varSpec.name, valuesByName.getOrElse(varSpec.name, Seq[String]()))
-      }
-    }
+    // Variable with max value determine the actual number of section
+    // If a varaible has 6 values, We need to have 6 sections at the end
```
fanf: varaible -> variable

fanf: Are you sure that holds? What about variables in sub-sections ? (perhaps it holds only because we don't have multi-valued sub section of multi-valued section, is it ?)

rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
```diff
@@ -223,41 +220,40 @@ class Section2FieldService(val fieldFactory: DirectiveFieldFactory, val translat
...
+              case None => varSpec.constraint.default
+              // Get value
+              case value => value
+            }
+          }
+          (varSpec.name,value)
         }
-      }.toMap
+      ).toMap
     }
```
fanf: What about the case where you have a section, which contains a variable A and a multi-valued subsection: in that case, you don't have numberMaxofSections for the A if there is several subsection, and you even can have 0 values for subsection (all optionnal) with numberMaxofSections == 1 (because of A), and you don't want to put defaults values in that case (because the user want 0 subsection, not one with defaults values)?

fanf: (not sure we have that case today in our technique, but nothing seems to prevent it to be in used in some user technique repos)

VinceMacBuche:
Closing, replacing by a better idea

jooooooon:
See #1046

PR #1045, opened 2016-02-11

VinceMacBuche:


jooooooon:
Replaced by #1046

PR #1046, opened 2016-02-11

VinceMacBuche:

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
@@ -4,12 +4,12 @@
 *************************************************************************************
 *
 * This file is part of Rudder.
-* 
+*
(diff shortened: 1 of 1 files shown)


fanf:
I much prefer that version :)

VinceMacBuche:
You are not funny :)

rudder-web/src/main/scala/com/normation/rudder/web/services/Section2FieldService.scala
```diff
@@ -156,6 +156,11 @@ class Section2FieldService(val fieldFactory: DirectiveFieldFactory, val translat
 
               valueOpt match {
                 case None =>
+                  varSpec.constraint.default match  {
+                    case None =>
+                    case Some(default) =>
```
fanf: could you use the syntax:
varSpec.constraint.default.foreach { default => 
setValueForField(default, field, t.from)
}

So that we don't end up with a case branch doing nothing ?

VinceMacBuche:
PR updated

PR #1050, opened 2016-02-24
Fixes #7960: API: add an option to filter directives based on their technique, status, etc

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7960


VinceMacBuche:
PR updated

jooooooon:
We recently discussed this, and agreed that we should not change any existing APIs in existing Rudder API versions. Therefore, if and when we introduce these parameters, it will only be in Rudder 3.3 (or 4.0, as it will most likely be called). Please re-target this PR.

VinceMacBuche:
Closing that PR that should be rebased on master

PR #1053, opened 2016-02-27
Fixes #2543: Allow to make a group with only 32bits or 64bits nodes

fanf:
See http://www.rudder-project.org/redmine/issues/2543

There is litterary only 3 things added:
- the attribute is added as a searcheable one, 
- configure its name in the dropdown, 
- add a line in the "OS" part of node details to show it.


ncharles:
\o/
Yay

PR #1054, opened 2016-02-29
Fixes #8001: create an index for detection of eventlog version

ncharles:
https://www.rudder-project.org/redmine/issues/8001


ncharles:
PR updated

PR #1056, opened 2016-03-01
Fixes #7961: API: add an option to just run a \"pre-check\" on directive modifications to change technique version

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/7961

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/directive/DirectiveAPIService2.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/directive/DirectiveAPIService2.scala
@@ -302,30 +302,20 @@ case class DirectiveAPIService2 (
     }
   }
 
-  def updateDirective(id: String, req: Req, restValues : Box[RestDirective]) = {
-
     // A function to check if Variables passed as parameter are correct
-    def checkParameters (paramEditor : DirectiveEditor) (parameterValues : (String,Seq[String])) = {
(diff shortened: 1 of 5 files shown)


jooooooon:
This is ready for review and could be merged as soon as the current releases are finished.

rudder-web/src/main/scala/com/normation/rudder/web/rest/directive/DirectiveAPIService2.scala
```diff
@@ -302,7 +302,7 @@ case class DirectiveAPIService2 (
     }
   }
 
-  def updateDirective(id: String, req: Req, restValues : Box[RestDirective]) = {
+  def updateDirective(id: String, req: Req, restValues : Box[RestDirective], check : Boolean) = {
```
fanf: I would much, much more prefer to have two separated method here, because:

- boolean parameters are quite obscure when seen at call site, 
- the method has two really different behavior depending of the boolean value: one read only + check, the other modify stored value. 
- and even the implementation is completly different for the two sides of the else.

fanf:
Appart from the comment, the code is not compiling (anymore ?)

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

fanf:
Great!

VinceMacBuche:
OK, merging this PR

PR #1068, opened 2016-03-25
Fixes #7524: Java OOM during Java's log migration

fanf:
https://www.rudder-project.org/redmine/issues/7524

There is two main point of that change: 

- make each migration be only by step of "max batch", so that we can't explose the memory consumption. I set up the batch size to 1000, because eventlog object use ~10kb, but can use to up to ~50kb for very big events (say, group update with thousand of nodes). So with that, we should keep the memory consumption below 100Mo for the migration

- make the process async so that the event log migration, which is not necessary for Rudder to work, is done in background. It can be stop and restart as many time as needed, so no problem if Rudder stop before the end of migration

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/migration/XmlEntityMigration.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/migration/XmlEntityMigration.scala
@@ -14,6 +14,8 @@ import java.sql.Timestamp
 import java.util.Calendar
 import com.normation.utils.XmlUtils
 import scala.xml.Node
+import java.sql.PreparedStatement
+import com.normation.rudder.domain.logger.MigrationLogger
 
 /**
  * specify from/to version
@@ -164,7 +166,7 @@ trait ControlXmlFileFormatMigration extends XmlFileFormatMigration {
           logger.info(s"Start migration of ${migrator.elementName} from format '${fromVersion}' to '${toVersion}'")
 
           migrator.process() match {
-            case Full(i) =>
+            case Full(MigrationProcessResult(i, nbBatches)) =>
               logger.info(s"Migration of ${migrator.elementName} fileFormat from '${fromVersion}' to '${toVersion}' done, ${i} EventLogs migrated")
               Full(MigrationSuccess(i))
 
@@ -214,8 +216,8 @@ trait ControlXmlFileFormatMigration extends XmlFileFormatMigration {
         previousMigrationController match {
           case None =>
             logger.info(s"The detected format ${detectedFileFormat} is no more supported, you will have to " +
-            		"use an installation of Rudder that understand it to do the migration. For information, " +
-            		"Rudder 2.6.x is the last major version which is able to import file format 1.0")
+                "use an installation of Rudder that understand it to do the migration. For information, " +
+                "Rudder 2.6.x is the last major version which is able to import file format 1.0")
             Full(MigrationVersionNotSupported)
 
           case Some(migrator) => migrator.migrate() match{
@@ -265,6 +267,11 @@ trait IndividualElementMigration[T <: MigrableEntity] {
   def migrate(element:T) : Box[T]
 }
 
+final case class MigrationProcessResult(
+    totalMigrated: Int
+  , nbBatches    : Int
+)
+
 /**
  * A trait that explain how to migrate one type of data in data base
  * (it delelegates the actual XML migration to someone else)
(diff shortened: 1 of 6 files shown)


rudder-core/src/main/scala/com/normation/rudder/migration/XmlEntityMigration.scala
```diff
@@ -312,20 +315,31 @@ trait BatchElementMigration[T <: MigrableEntity] extends XmlFileFormatMigration
...
+                      , save = save
+                      , successLogger = successLogger
+                    )
+      } yield {
+        migrated
+      }
+
+      res match {
+        case Full(k) if(k < 1) => Full(mig)
+        case Full(k) => recProcessOneBatch(MigrationProcessResult(mig.totalMigrated+k, mig.nbBatches+1))
```
ncharles: as long as you are migrating, you do migrate, am I correct ? 
shoudln't we have a logger.debug there, to know that batches are migrating in case of issue ?

rudder-core/src/main/scala/com/normation/rudder/migration/XmlEntityMigration.scala
```diff
@@ -364,6 +374,76 @@ trait BatchElementMigration[T <: MigrableEntity] extends XmlFileFormatMigration
...
+  override final def selectAllSqlRequest(batchSize: Int) = {
+    s"select id, eventType, data from (select id, eventType, data, ((xpath('/entry//@fileFormat',data))[1]::text) as version from eventlog) as T where version='${fromVersion}' limit ${batchSize}"
+  }
+
+  override final protected def save(logs:Seq[MigrationEventLog]) : Box[Seq[MigrationEventLog]] = {
+    val UPDATE_SQL = "UPDATE EventLog set eventType = ?, data = ? where id = ?"
+
+    val ilogs = logs match {
+      case x:IndexedSeq[_] => logs
+      case seq => seq.toIndexedSeq
```
ncharles: why to we need to convert the seq to idexedseq ?
I fail to see the reason :/

ncharles:
i have a question on this PR: i didn't see any code removed or modified that selected serialized elements in a specific version to migrate them, but I do see new code that select by version
Does it means that we:
a) selected all events and processed them all, even if correct, or
b) you forgot to clean that part in this PR ?

rudder-core/src/main/scala/com/normation/rudder/migration/XmlEntityMigration.scala
fanf: I'm not sure what you meant by "as long as you are migrating, you do migrate", but yes, if there is no problem, I continue. 

Threre is already debug logs at other points of the logic, but yes, I can check and see where to add other information. You were thinking to a kind of progress log ? "I did 11 batches", "I did 12 batches", etc ?

rudder-core/src/main/scala/com/normation/rudder/migration/XmlEntityMigration.scala
fanf: That's not a part of the code I modified, I just moved from an other class. 
The reason may be that we need to access the sed element by index after that point, and we want to be sure that it won't fail.

fanf:
We used to select all element and then filter by version on the scala side. Which was obviously not the best option memory wise. Now, we are selecting them by version, and by batches, so we are assure to always cap the memory used to the max of 1000xelement_size, where element_size is O(xml size of event).

fanf:
PR updated

FlorianHeigl:
500MB Ram is probably still in the wiggle room, could you run a test to see if that would make it considerably (say 50%) faster?
If yes, try that rather and give higher memory req. in the docs; that should be no biggie for anyone.

ncharles:
Migrating more eventlog at the same time will not increase dramatically the performance - taking 5000 logs in a batch rather than 1000 will only divide by 5 the number of queries, and they are not so long by themselves (everything else is: fetching the data from the DB, modifying data, etc).
Plus, this is an async process now, so it will not block the startup of Rudder.
@fanf may give more insight about time of migration with large dataset

fanf:
@FlorianHeigl : as @ncharles said, the process is for now extremly sequential. Having bigger batch does not implie much more performance. It limits the i/o fixe cost, which are rather limited because we already have a connection pool. Having batch is still interesting compared to fully streamed solution, because the fixe i/o costs DO exists non the less. In my test, batch of 1000 events seemed reasonnable. 

So the obvious alternative is to make the process parallel. With more ram, we could process several batches in parallel. The full migration process would be obviously be much shorter. 
It raise two questions, one on CPU and I/O load (do we want the migration to use more resource, perhaps impacting more soft real time action like user request), and one on the complexity of the migration process (today, it's dead simple. Here, the migration end is more complexe to assess, because parallel exec, and there is certainly many more things). I don't have answers to that, so I opened that ticket to track it: https://www.rudder-project.org/redmine/issues/8125

FlorianHeigl:
Don't get me wrong - if it looked efficient (in terms of spooling enough data into ram and not stuttering between processing and IO) then just stick with 1000.

PR #1075, opened 2016-04-06
Fixes #8072: Issue while trying to remove a technique

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8072

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
@@ -91,14 +91,14 @@ object TechniqueEditForm {
     (for {
       xml <- Templates("templates-hidden" :: "components" :: "ComponentTechniqueEditForm" :: Nil)
     } yield {
-      chooseTemplate("component", "popupRemoveForm", xml)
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
```diff
@@ -91,14 +91,14 @@ object TechniqueEditForm {
...
-      chooseTemplate("component", "popupRemoveForm", xml)
+      chooseTemplate("component", "popupremoveform", xml)
     }) openOr Nil
 
   private def popupDisactivateForm =
     (for {
       xml <- Templates("templates-hidden" :: "components" :: "ComponentTechniqueEditForm" :: Nil)
     } yield {
-      chooseTemplate("component", "popupDisactivateForm", xml)
+      chooseTemplate("component", "popupdisableform", xml)
```
ncharles: why the case doesn't match the one in html ??

VinceMacBuche: it's the html5 parser it ignore cases so you have to put it all downcase...

PR #1078, opened 2016-04-15
Fixes #8107: The default number of lines in table is too low

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8107


fanf:
Why targeting 3.0 for that one ?

VinceMacBuche:
That opeion is only valid for dattable 1.10, which is only avalaible in Rudder 3.0

fanf:
Ok, seems legit :)

PR #1079, opened 2016-04-19
Fixes #8195: Correct memory profile and efficiency of error report logger

fanf:
https://www.rudder-project.org/redmine/issues/8195

Some information: 
- a lot of modification in the ReportsJdbcRepository are cosmetic and just help see what are the differences between queries
- the previous code was doing 3 i/o (get directive, rule, technique) for each log to write, which can be made more efficient by just getting all the rules/directives/techniques. There are already in Rudder memory, that doesn't change the memory profile of the app, but it won a LOT of time (from > 15min to ~10s for writting 250 000 lines)
- the whole process is split in little batch of 10 000 non-compliant-reports. I didn't see much interest in making them bigger, since it could help only for some fixes cost like getting the connection from pool, which is clearly not a big problem. 

One question remains: the actual code has a kind of local cache for the last non-compliant-report id processed. It was like that, and I didn't changed it, but I'm not sure about the gain compared to just getting it from the DB. And it may lead to inconsistencies if the value in DB is updated from an other process (like an human doing tests, or wanting to produce again the non-compliant-report from a given id).

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
@@ -40,13 +40,15 @@ package com.normation.rudder.batch
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.logger.AllReportLogger
 import com.normation.rudder.domain.nodes.NodeInfo
+import com.normation.rudder.domain.policies.Rule
+import com.normation.rudder.domain.policies.RuleId
 import com.normation.rudder.domain.reports.Reports
+import com.normation.rudder.repository.FullActiveTechniqueCategory
 import com.normation.rudder.repository.ReportsRepository
 import com.normation.rudder.repository.RoDirectiveRepository
 import com.normation.rudder.repository.RoRuleRepository
 import com.normation.rudder.repository.RudderPropertiesRepository
 import com.normation.rudder.services.nodes.NodeInfoService
-import com.normation.utils.UuidRegex
 
 import net.liftweb.actor._
 import net.liftweb.common._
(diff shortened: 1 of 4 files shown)


rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -93,114 +96,166 @@ class AutomaticReportLogger(
...
         logger.error("Wrong message received by non compliant reports logger, do nothing")
     }
 
+    private[this] def logReportsBetween(lastId: Long, currentId: Long): Unit = {
+
+      val batchSize = 10000
+      //log from "from". Return the last logged id
+      def log(from: Long, allNodes: Map[NodeId, NodeInfo], rules: Map[RuleId, Rule],  directives: FullActiveTechniqueCategory): Box[Long] = {
+        for {
+          reports <- reportsRepository.getErrorReportsBeetween(from, currentId, batchSize, reportskind)
```
ncharles: you don't have any error handling there, is it expected ?
if the method getErrorReportsBeetween fails, we'll never know it's during the non compliant error reporting

ncharles: ha, this is handled below

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -93,114 +96,166 @@ class AutomaticReportLogger(
...
-     */
-    def logReport(report:Reports, allNodes: Map[NodeId, NodeInfo]) = {
-      val execDate = report.executionDate.toString("yyyy-MM-dd HH:mm:ssZ")
 
-      val rule = "%s [%s]".format(
-            report.ruleId.value
-          , ruleRepository.get(report.ruleId).map(_.name).openOr("Unknown rule")
-          )
+    def logReports(reports : Seq[(Long, Reports)], allNodes: Map[NodeId, NodeInfo], rules: Map[RuleId, Rule],  directives: FullActiveTechniqueCategory): Option[Long] = {
+      if(reports.isEmpty) None
```
ncharles: it would be much better to use { } for readability

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -293,12 +294,12 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
      }
   }
 
-  override def getHighestId : Box[Long] = {
-    val query = "select id from RudderSysEvents order by id desc limit 1"
+  override def getHighestErrorId(kinds: List[String]) : Box[Long] = {
```
ncharles: why calling it with Error in its name if you can pass any severity ?

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -93,114 +96,166 @@ class AutomaticReportLogger(
...
-        case Failure(message,_,_) =>
-          logger.error("could not fetch last id, don't log anything, wait next run, cause is %s".format(message))
-      }
+          case Full(lastId) =>
+            logger.trace("***** get current highest non-compliant log id")
+            val highest = reportsRepository.getHighestErrorId(reportskind)
+            logger.trace(s"***** highest non-compliant log id = ${highest} and last processed id = ${lastId}")
+            highest match {
+              case Full(currentId) if (currentId > lastId) =>
+                logReportsBetween(lastId, currentId)
```
ncharles: If I understand correctly, the lastId will only increment if we have repaired or error (or warn)
But there are many cases when we don't, and so in this case, we'll look, each time, for an increasing size of the DB over and over to see if we have reports.
We should always increment the lastId to the last report id checked (that was the raison d'etre of getHighestId

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
VinceMacBuche: HighestIdByKind ?

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
fanf: well, Postgres is doing the request for the last id of the event. This things are indexed, so even in very big DB, it should be very quick. 
And for now, I can't find of a solution that give us the real last processed id - the old solution would implie that in case of any problem before the end of the end of the process, you will do again all the processing of reports.

I will think to that again.

VinceMacBuche: I think I agree with Nicolas, and that the request to fetch error reports will be easier if we augment the last id more regurlarly

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -93,114 +96,166 @@ class AutomaticReportLogger(
     /*
      * List of all reports kind processed by the logger
      */
-    private[this] var Reportskind = List(Reports.LOG_REPAIRED,Reports.RESULT_ERROR,Reports.RESULT_REPAIRED,Reports.LOG_WARN)
+    private[this] val reportskind = List(Reports.LOG_REPAIRED,Reports.RESULT_ERROR,Reports.RESULT_REPAIRED,Reports.LOG_WARN)
```
VinceMacBuche: reportsKind ?

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
ncharles: Even if the query is indexed, you'll search for all error/repaired between last stored id and id + batch size, which can be at the bottom of the DB - so you'll teach Postgres to become efficient for old reports rather than the one we use in the compliance. Plus, you'll go over and over the same reports again and again, waiting for a new repaired.
I don't know also how i would behave if last id is so old it gets out of the DB (archived reports)

fanf: I'm not sure I understand you. I'm asking Postgres "tell me the id of the last non-complient report", whatever its position is. I'm not asking it to look at a particular place. 
Then, I'm querying non-compliant reports by batch until we reach that id - but I should never go on the same report twice (safe in case of a Rudder crash, were the last batch is not yet marked as finished. 

So could you explain why postgres will learn to look at old reports, or why the code will go over and over the same reports again and again ? 

If the id is not in DB anymore, postgres doesn't care.

VinceMacBuche: Your last id was 8042, for a looooong time you don't have repaired, so your last repaired id stay a 8042, but id still growing up (and not the highest is 1000000000045, you waited a long time)  ant next repaired is: 1000000000046

on next run, it will try to get report from 8042 to 1000000000046 by batch of 10k reports which can be long

but if you update the last id to the last id in the databse on each run you will never have a huge gap ( maybe you have 10M nodes, you will have a huge gap, but there will be other problems ...)

VinceMacBuche: Lastid here should be "last treated id" not "last error/repaired id treated"

fanf: I do understand that, but the actual request is: 
"give me 10k the non-compliant reports after 8042 [last processed id] (and not after 8042, because it's the highest non-compliant id). So the actual batch size in your case is 0. And postgres knows there is no repaired/error with the constraint due to the index. 

And if 1000000000045 DOES be a non-compliant report, that does change much: you get one report, update the index, and be done. 

Again, I do understand what you are saying, it would be better to look to non-compliant since a lower time, but that does not seems really bad as it is, or at least I don't understand why you insist at thinking that I'm going to process other reports than non-compliant one again and again. 
The fact that we don't look in the database for the last id seems much worse than that, in comparison.

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
fanf: yes

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
fanf: yes, seems better

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
fanf: Ok, so the problem is that some of the reports kind we want are not specifically indexed, so it may be long in tje descripbed case. It has nothing to do with the batch size or anything, but just to the fact that "select id from reports where id between 1 and 10000000000 and reporttype='xxxx' or reporttype='yyy' limit 10000" is slow when we don't have an index on, say, yyy.

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -87,120 +90,183 @@ class AutomaticReportLogger(
...
-              logger.trace("***** no reports to log")
-          }
 
-        case Failure(message,_,_) =>
-          logger.error("could not fetch last id, don't log anything, wait next run, cause is %s".format(message))
-      }
+          case Full(lastId) =>
+            logger.trace("***** get current highest non-compliant log id")
+            val highest = reportsRepository.getHighestId()
+            logger.trace(s"***** highest non-compliant log id = ${highest} and last processed id = ${lastId}")
```
ncharles: the trace message is invalid, should be
highest log id = ${highest} and last processed id = ${lastId}

ncharles: actually, since highest is a Box, how would it serialize ?

VinceMacBuche: Full(id) :(

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -87,120 +90,183 @@ class AutomaticReportLogger(
...
 
-    def logReports(reports : Seq[Reports]) = {
-      for {
-        allNodes <- nodeInfoService.getAll
-        report <- reports
-      } logReport(report, allNodes)
-    }
     /*
-     * Transform to log line and log a report with the appropriate kind
+     * This method handle the logic to process non-compliant reprots between the
```
ncharles: typo: reports

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
VinceMacBuche: I think the trace should be 3 lines below in the Full case on highet pattern matching

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -87,120 +90,183 @@ class AutomaticReportLogger(
...
+              Full(id)
+            }
+        }
+      }
 
-      val directive : String = "%s [%s]".format(
-            report.directiveId.value
-          , directiveRepository.getDirective(report.directiveId).map(_.name).openOr("Unknown directive")
-          )
+      val startAt = lastProcessedId+1
```
ncharles: this deserve a comment on why the +1

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -87,120 +90,183 @@ class AutomaticReportLogger(
...
-              logger.trace("***** no reports to log")
-          }
 
-        case Failure(message,_,_) =>
-          logger.error("could not fetch last id, don't log anything, wait next run, cause is %s".format(message))
-      }
+          case Full(lastId) =>
+            logger.trace("***** get current highest report id")
+            val highest = reportsRepository.getHighestId()
+            logger.trace(s"***** highest report id = ${highest} and last processed id = ${lastId}")
```
VinceMacBuche: Still same remark as before (removed by a PR updated)

I think the trace should be in Full case of the pattern matching below

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -57,25 +57,29 @@ import com.normation.rudder.reports.execution.AgentRun
 import com.normation.rudder.domain.reports.Reports
 import com.normation.rudder.reports.execution.AgentRunId
 import com.normation.rudder.reports.execution.AgentRun
+import net.sourceforge.sizeof.SizeOf
```
ncharles: are you sure you need that import ?

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -57,25 +57,29 @@ import com.normation.rudder.reports.execution.AgentRun
...
 class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsRepository with Loggable {
 
   val reportsTable = "ruddersysevents"
   val archiveTable = "archivedruddersysevents"
 
-  private[this] val baseQuery = "select executiondate, nodeid, ruleid, directiveid, serial, component, keyValue, executionTimeStamp, eventtype, policy, msg from RudderSysEvents where 1=1 ";
-  private[this] val baseArchivedQuery = "select executiondate, nodeid, ruleid, directiveid, serial, component, keyValue, executionTimeStamp, eventtype, policy, msg from archivedruddersysevents where 1=1 ";
-
   private[this] val reportsExecutionTable = "reportsexecution"
+  private[this] val common_reports_column = "executiondate, nodeid, ruleid, directiveid, serial, component, keyvalue, executiontimestamp, eventtype, policy, msg"
```
ncharles: this makes everything more readable, thank you !

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
fanf: I don't agree, in trace, I want to know why the proces fails, and getting a Full(id) is what I want. In fact, I want to get an "Empty" if it's the case. 
And it's trace, so it is ok to get code related message.

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
fanf: well, not anymore :) Good catch.

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
```diff
@@ -87,120 +90,185 @@ class AutomaticReportLogger(
...
+              Thread.sleep(100)
+              maxTry(tried+1)
+            }
+        }
+      }
+      maxTry(0)
+    }
 
-      AllReportLogger.FindLogger(report.severity)(reportLineTemplate.format(execDate,node,report.severity,rule,directive,technique,report.component,report.keyValue,report.message))
+    def logReports(reports : Seq[(Long, Reports)], allNodes: Map[NodeId, NodeInfo], rules: Map[RuleId, Rule],  directives: FullActiveTechniqueCategory): Option[Long] = {
```
ncharles: this is much more readable, thank you !

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
ncharles: i do agree, knowing why it fails is important !

ncharles:
Looks good to me, this is a much welcomed change.
Thank you !

rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportLogger.scala
VinceMacBuche: You have a logger.error with the failure below,

But that's only a trace logger ... so I don't think it need that much argument :)

VinceMacBuche:
Merging!

PR #1084, opened 2016-05-03
Fixes #7735: OOM in Rudder when there are too many repaired reports

fanf:
https://www.rudder-project.org/redmine/issues/7735

For information, with the 3 patches applied (+https://www.rudder-project.org/redmine/issues/8195), with > 1.2 Millions changes in the last 3 days (with one 6h interval at 1M changes), we are able to start a Rudder with -Xmx600m, have the non-compliant report be iteratively written, login, go to rule details, see graphes, click and see details on all interval, and everything is fast - with a top heap consomption at 290MB. 

So, there is still some cleaning to do, but the general architecture of the patch is ok. It is in 3 parts, I'm letting them in three commits to make them better understandable. Perhaps it could be split in 3 tickets, to, but they are highly coupled (at least the last two), so not sure. 

1/ Today, we reload compliance each time we switch from the Rule edit form tab to the compliance one. This is useless since we have nice "reload" button on compliance tab, it is an UX strangeness because it causes delay for switch, and don't let the user choose when data should change (difficult to understand, or crazy when you were trying to analized something and clicked on the tab by mistake), and it leads to OOM when  there is a lot of reports (because even if we succeeded in calculating one time, the second one, we have much less memory, since a big part is used). 

2/ That patches allows to only load the events details (display on the bottom grid) for the last interval in rule details (and load details of other on click on the corresponding graph interval). That also means that much less data are transfered to the browser, making user happier (end rule details faster to display). More over, a hard limit of max 10 000 event for a rule, for an interval, is introduced. It is not a good solution, but the correct one (pagination on each interval) is much harder to implement, with higher risks in a released version.

3/ With the third patch, the event cache only keep a COUNT of the number of events by rule by interval, which consume much less memory than keeping all events (of course). Also rewrote some other methods to consum less memory overall.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
@@ -437,10 +437,23 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
 
 
 
-  override def getChangeReports(notBefore: DateTime): Box[Seq[ResultRepairedReport]] = {
-    val query = s"${baseQuery} and eventtype='${Reports.RESULT_REPAIRED}' and executionTimeStamp > '${new Timestamp(notBefore.getMillis)}'::timestamp order by executionTimeStamp asc"
+  override def countChangeReports(startTime: DateTime, intervalInHour: Int): Box[Map[RuleId, Map[Interval, Int]]] = {
+    //be careful, extract from 'epoch' gives seconds, not millis
+    val mod = intervalInHour * 3600
+    val start = startTime.getMillis / 1000
+    val query = s"select ruleid, count(*) as number, ( extract('epoch' from executiontimestamp)::bigint - ${start})/${mod} as interval from ruddersysevents " +
+                s"where eventtype = 'result_repaired' and executionTimeStamp > '${new Timestamp(startTime.getMillis)}'::timestamp group by ruleid, interval;"
+
+    //that query will return interval number in the "interval" column. So interval=0 mean
+    //interval from startTime to startTime + intervalInHour hours, etc.
+    //=> little function to build an interval from its number
+    def interval(t:DateTime, int: Int)(num: Int) = new Interval(t.plusHours(num*int), t.plusHours((num+1)*int))
+
     try {
-      Full(jdbcTemplate.query(query,ReportsMapper).asScala.collect{case r:ResultRepairedReport => r})
+      val res = jdbcTemplate.query(query, CountChangesMapper(interval(startTime, intervalInHour))).asScala
+      //group by ruleId, and then interval
+      val groups = res.groupBy(_._1).mapValues( _.groupBy(_._3).mapValues(_.map( _._2).head)) //head non empty due to groupBy, and seq == 1 by query
+      Full(groups)
     } catch {
       case ex: Exception =>
         val error = Failure("Error when trying to retrieve change reports", Some(ex), Empty)
@@ -449,17 +462,34 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
     }
   }
 
+
   override def getChangeReportsOnInterval(lowestId: Long, highestId: Long): Box[Seq[ResultRepairedReport]] = {
     val query = s"${baseQuery} and eventtype='${Reports.RESULT_REPAIRED}' and id >= ${lowestId} and id <= ${highestId} order by executionTimeStamp asc"
-    try {
-      Full(jdbcTemplate.query(query,ReportsMapper).asScala.collect{case r:ResultRepairedReport => r})
-    } catch {
-      case ex: Exception =>
-        val error = Failure("Error when trying to retrieve change reports", Some(ex), Empty)
-        logger.error(error)
-        error
-    }
+    transformJavaList(jdbcTemplate.query(query, ReportsMapper))
   }
+
+  override def getChangeReportsByRuleOnInterval(ruleId: RuleId, interval: Interval, limit: Option[Int]): Box[Seq[ResultRepairedReport]] = {
+    val l = limit match {
+      case Some(i) if(i > 0) => s"limit ${i}"
+      case _                 => ""
+    }
+    val query = s"${baseQuery} and eventtype='${Reports.RESULT_REPAIRED}' and ruleid='${ruleId.value}' " +
+                s" and executionTimeStamp >  '${new Timestamp(interval.getStartMillis)}'::timestamp " +
+                s" and executionTimeStamp <= '${new Timestamp(interval.getEndMillis)  }'::timestamp order by executionTimeStamp asc ${l}"
+
+    transformJavaList(jdbcTemplate.query(query, ReportsMapper))
+  }
+
+  private[this] def transformJavaList(l: java.util.List[Reports]) = {
+      try {
+        Full(l.asScala.collect{case r:ResultRepairedReport => r})
+      } catch {
+        case ex: Exception =>
+          val error = Failure("Error when trying to retrieve change reports", Some(ex), Empty)
+          logger.error(error)
+          error
+      }
+    }
 
 
   override def getReportsByKindBeetween(lower: Long, upper: Long, limit: Int, kinds: List[String]) : Box[Seq[(Long,Reports)]] = {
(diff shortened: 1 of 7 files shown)


rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -478,6 +515,16 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
   }
 }
 
+final case class CoundChangesMapper(intMapper: Int => Interval) extends RowMapper[(RuleId, Int, Interval)] {
```
ncharles: it should be CountChangesMapper

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -449,18 +462,42 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
...
   override def getChangeReportsOnInterval(lowestId: Long, highestId: Long): Box[Seq[ResultRepairedReport]] = {
-    val query = s"${baseQuery} and eventtype='${Reports.RESULT_REPAIRED}' and id >= ${lowestId} and id <= ${highestId} order by executionTimeStamp asc"
-    try {
-      Full(jdbcTemplate.query(query,ReportsMapper).asScala.collect{case r:ResultRepairedReport => r})
-    } catch {
-      case ex: Exception =>
-        val error = Failure("Error when trying to retrieve change reports", Some(ex), Empty)
-        logger.error(error)
-        error
+    def theQueryOnly = {
```
ncharles: your naming skills equal mine :)
what about "repairedQuery" ?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -449,18 +462,42 @@ class ReportsJdbcRepository(jdbcTemplate : JdbcTemplate) extends ReportsReposito
...
+      val query = s"${baseQuery} and eventtype='${Reports.RESULT_REPAIRED}' and ruleid='${ruleId.value}' " +
+                  s" and executionTimeStamp >  '${new Timestamp(interval.getStartMillis)}'::timestamp " +
+                  s" and executionTimeStamp <= '${new Timestamp(interval.getEndMillis)  }'::timestamp order by executionTimeStamp asc ${l}"
+      jdbcTemplate.query(query,ReportsMapper)
+    }
+
+    transformJavaList(theQueryOnly)
+  }
+
+  private[this] def transformJavaList(l: java.util.List[Reports]) = {
```
ncharles: just a question: isn't this method effectively doubling the memory footprint ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -59,111 +60,105 @@ import net.liftweb.http.js.JE._
...
+  private[this] final def getInterval(since: DateTime, to: DateTime): List[Interval] = {
+    if(to.isBefore(since)) {
+      Nil
+    } else {
+      // We want our first interval limit to be 00:00 | 06:00 | 12:00 | 18:00
+
+      //utility that create an interval from the given date to date+6hours
+      def sixHours(t: DateTime): Interval = {
+        //6 hours in milliseconds
+        new Interval(t, new DateTime(t.getMillis + 6l * 3600 * 1000))
```
ncharles: why 61 ?

fanf: That's a "L", for long, with the horrible GitHub police.

fanf: So bad it isn't a site for coders who needs to have a nice coding police....

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
fanf: that was mesured, and even if it is the case (not exactly the case, not sur why), it's negligible compared to other hot spot

VinceMacBuche:
First commit is ok ! and is the easiest one, now let's take a look at the big ones ...

fanf:
PR updated

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
jooooooon: Police in English refers to the state-funded folk that keep order. I think you're talking about a font :)

fanf: yeah, that the consequences of the Place de la Republique proxomity...

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -277,57 +289,56 @@ class CachedNodeChangesServiceImpl(
...
-object NodeChanges {
 
-  // Filter changes on based on rules
-  def changesOnRule(ruleId : RuleId)(changes  : Map[Interval,Seq[ResultRepairedReport]]) = {
-    changes.mapValues(_.filter(_.ruleId == ruleId))
+  /**
+   * Get the changes for the given interval, without using the cache.
+   */
+  override def getChangesForInterval(ruleId: RuleId, interval: Interval, limit: Option[Int]): Box[Seq[ResultRepairedReport]] = {
+    for {
```
VinceMacBuche: Why not just 'changeService.getChangesForInterval(ruleId, interval, limit)' ?

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleCompliance.scala
```diff
@@ -115,137 +113,177 @@ class RuleCompliance (
...
+      <div class="section-title">Compliance by Node</div>
+    </div>
+    <div id="nodeCompliance">
+       <table id="nodeReportsGrid" cellspacing="0">  </table>
+    </div>
+    <div id="recentChangesSection" class="unfoldedSection" onclick="$('#recentChanges').toggle(400); $('#recentChangesSection').toggleClass('foldedSection');$('#recentChangesSection').toggleClass('unfoldedSection');">
+      <div class="section-title">Recent changes</div>
+    </div>
+    <div id="recentChanges">
+      {SHtml.button("Refresh", () => refresh())}
```
VinceMacBuche: Should this refresh all (like before) or only the recent changes graph ?

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleCompliance.scala
```diff
@@ -115,137 +113,177 @@ class RuleCompliance (
...
+        data.splice(0,0,'Recent changes')
+        var x = recentChanges.x
+        x.splice(0,0,'x')
+        var chart = c3.generate({
+          data: {
+                x: 'x'
+              , columns: [ x , data ]
+              , type: 'bar'
+              , onclick: function (d, element) {
+                  console.log(recentChanges.t[d.index]);
```
VinceMacBuche: Remove the console.log :)

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleCompliance.scala
```diff
@@ -115,137 +113,177 @@ class RuleCompliance (
...
+      <div class="section-title">Recent changes</div>
+    </div>
+    <div id="recentChanges">
+      {SHtml.button("Refresh", () => refresh())}
+      <div id="changesChart">  </div>
+      <hr class="spacer" />
+      <table id="changesGrid" cellspacing="0">  </table>
+    </div> ++
+    Script(After(0,JsRaw(s"""
+      function refresh() {${refresh().toJsCmd}};
```
VinceMacBuche: and this should be refreshCompliance no ?

VinceMacBuche: no I'm wrong here :)

VinceMacBuche: maybe not haaaa

VinceMacBuche: So we discussed with françois, we keep the behavior like before, refresh the 3 parts then the refresh call is ok :)

VinceMacBuche:
Apart the two minor fixes it's ok and can be merged (once it's one commit! )

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleCompliance.scala
fanf: I believe it's better to stay as near as possible to the previous user experience. If the user used to click somewhere to refresh everythink, we can't change that in a minor release.

rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
fanf: well, certainly there was some other comands at some point of the dev, that were removed in the end (but not the for/yield  structure).

fanf:
PR updated

PR #1088, opened 2016-05-06
Fixes #8246: Node configuration cache is invalidated too often

fanf:
https://www.rudder-project.org/redmine/issues/8246

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -441,8 +441,8 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
 
     (nodeIds.flatMap { nodeId:NodeId =>
       (for {
-        nodeInfo <- Box(allNodeInfos.get(nodeId)) ?~! s"Node with ID ${nodeId.value} was not found"
-        policyServer <- Box(allNodeInfos.get(nodeInfo.policyServerId)) ?~! s"Node with ID ${nodeId.value} was not found"
+        nodeInfo <- Box(allNodeInfos.get(nodeId)) ?~! s"Node with ID '${nodeId.value}' was not found in the list of known nodes"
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -439,10 +439,13 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
     , globalSystemVariables: Map[String, Variable]
   ): Map[NodeId, InterpolationContext] = {
 
+    println(s"**** AGAIN ${allNodeInfos.size} nodes in allNodeInfos, check exists:\n *** ${List("06da3556-5204-4bd7-b3b0-fa5e7bcfbbea","009e0af1-1a3c-4ee0-a91d-10b68a008f5d", "073239d7-c633-4b60-8477-57131839fd34").map( k => (k, allNodeInfos.isDefinedAt(NodeId(k)))).mkString("\n *** ")}")
```
matya: What is this line doing :o
Are the hard-coded nodeIds not just for debug?

VinceMacBuche: drunk @fanf is drunk ;) I think it's debug / dev logger that need to be removed ... or changed as logger.trace without hardcoded ids ...

VinceMacBuche:
apart the print it's ok to me and can be merged !

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
matya: Pay attention to the Balmer Peak!
https://xkcd.com/323/

fanf: It was debug. The nodes in the list had specific interesting properties, like "no policy server", "policy server id not a known node", "policy server a known node (which don't know it is a policy server)"

fanf:
PR updated

PR #1090, opened 2016-05-11
Fixes #8262: Prevent red color usage in small graphs

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/8262

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/homePage.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/homePage.js
@@ -110,7 +110,8 @@ function homePage (
 }
 
 function displayInventoryGraph (id,data) {
-
+  var colorPatternDonutCharts = ['rgb(31, 119, 180)', 'rgb(23, 190, 207)', 'rgb(255, 127, 14)', 'rgb(255, 224, 14)', 'rgb(227, 119, 194)', 'rgb(44, 160, 44)', 'rgb(255, 104, 105)', 'rgb(148, 103, 189)', 'rgb(140, 86, 75)', 'rgb(160, 160, 160)', 'rgb(146, 201, 74)', '#ffd203', 'rgb(132, 63, 152)'];
+    
   var smallHeight =  $(window).height() / 4 ;
   
   // The chart, without legend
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -131,6 +131,9 @@ function displayInventoryGraph (id,data) {
       , legend :  {
           show : false
         }
+      , color: {
+        pattern: ['rgb(31, 119, 180)', 'rgb(23, 190, 207)', 'rgb(255, 127, 14)', 'rgb(255, 224, 14)', 'rgb(227, 119, 194)', 'rgb(44, 160, 44)', 'rgb(255, 104, 105)', 'rgb(148, 103, 189)', 'rgb(140, 86, 75)', 'rgb(160, 160, 160)', 'rgb(146, 201, 74)', '#ffd203', 'rgb(132, 63, 152)']
```
VinceMacBuche: Since it in the same function, you should put the pattern in an array and put the variable here :)

rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -175,6 +178,9 @@ function displayInventoryGraph (id,data) {
            }
          }
        }
+      , color: {
+        pattern: ['rgb(31, 119, 180)', 'rgb(23, 190, 207)', 'rgb(255, 127, 14)', 'rgb(255, 224, 14)', 'rgb(227, 119, 194)', 'rgb(44, 160, 44)', 'rgb(255, 104, 105)', 'rgb(148, 103, 189)', 'rgb(140, 86, 75)', 'rgb(160, 160, 160)', 'rgb(146, 201, 74)', '#ffd203', 'rgb(132, 63, 152)']
```
VinceMacBuche: and use the variable here, so when you will have to modify the colors, you will have to modify only at one place :)

rudder-web/src/main/webapp/javascript/rudder/homePage.js
RaphaelGauthier: Hu yeah, of course... Why did not I think of that ! I will adjust that.

RaphaelGauthier:
PR updated

PR #1098, opened 2016-05-19
Fixes #7972: API call to get the list of node is REALLY slow with latest version

fanf:
https://www.rudder-project.org/redmine/issues/7972

It's a bigger PR that what I wanted initialy. So, the problem is that in API for nodes v6, we are systematicaly getting all fullInventories + all nodeinfo, whatever the details level, where in previous API version, we were relying only on nodeinfo. Moreover, when software where needed (and only on that case, what was good!), one call to get software was made by node.  
NodeInfoService is cached, and is quick, where full inventory is SOOOOO SLOW. And I don't even talk about softwares (that is even slower). 

So, the general solution is to make again defaults level rely only on the (cached) NodeInfoService. And here, we have 2 problems:
- NodeInfo does not exactly contains all info needed in defaults detail level (but almost), and
- NodeInfo are not available for pending and deleted nodes (but again, almost). 

So the correction: 
- make the actual query done (nodeinfo or full inventory or software, or a mix) depending of detailLevel, 
- extends nodeInfo to hold all the data needed in detaults detaillevel, 
- make more specific what data (attribute) are retrieve from entries for nodeInfo, to avoid having big things like softwareId and env var going thought the wires just to be ignored
- extends nodeinfo to work for pending/deleted node (but without the cached part. Only accepted nodeInfo are cached). 

Everything else is cosmetic and making the compiler happy. 
(in the cosmetic part, I made the field to always be return in the same order in json, it's saner for the user :)


fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/domain/RudderLDAPConstants.scala
```diff
@@ -168,7 +168,7 @@ object RudderLDAPConstants extends Loggable {
 
   OC += (OC_RUDDER_NODE,
     must = Set(A_NODE_UUID, A_NAME, A_IS_BROKEN, A_IS_SYSTEM),
-    may = Set(A_DESCRIPTION, A_SERIALIZED_AGENT_RUN_INTERVAL, A_SERIALIZED_HEARTBEAT_RUN_CONFIGURATION))
+    may = Set(A_DESCRIPTION, A_SERIALIZED_AGENT_RUN_INTERVAL, A_SERIALIZED_HEARTBEAT_RUN_CONFIGURATION, A_NODE_PROPERTY))
```
VinceMacBuche: And it was already in the ldap schema ?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -158,25 +237,25 @@ trait NodeInfoServiceCached extends NodeInfoService  with Loggable with CachedRe
...
    * attributes is the list of attributes needed in returned entries
    */
-  def getNodeInfoEntries(con: RoLDAPConnection, attributes: Seq[String]): Seq[LDAPEntry]
+  def getNodeInfoEntries(con: RoLDAPConnection, attributes: Seq[String], status: InventoryStatus): Seq[LDAPEntry]
 
   /*
    * Our cache
    */
-  private[this] var nodeCache = Option.empty[Map[NodeId, (LDAPNodeInfo, NodeInfo, Node)]]
+  private[this] var nodeCache = Option.empty[Map[NodeId, (LDAPNodeInfo, NodeInfo)]]
```
VinceMacBuche: Why do we keeep LDAPNodeInfo in cache ?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -282,6 +360,92 @@ trait NodeInfoServiceCached extends NodeInfoService  with Loggable with CachedRe
   }
 
   /**
+   * An utility method that gets data from backend for things that are
+   * node really nodes (pending or deleted).
+   */
+  private[this] def getNotAcceptedNodeDataFromBackend(status: InventoryStatus): Box[Map[NodeId, NodeInfo]] = {
```
VinceMacBuche: Almost all code here is the same as accepted Node :( I would be very happuy if they were merged ...

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -282,6 +360,92 @@ trait NodeInfoServiceCached extends NodeInfoService  with Loggable with CachedRe
...
+        }
+      }.toMap
+    }
+  }
+
+  override final def getPendingNodeInfos(): Box[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(PendingInventory)
+
+  override final def getDeletedNodeInfos(): Box[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(RemovedInventory)
+
+  private[this] def getNotAcceptedNodeInfo(nodeId: NodeId, status: InventoryStatus): Box[Option[NodeInfo]] = {
```
VinceMacBuche: Is that function used ?

rudder-web/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
```diff
@@ -154,12 +154,14 @@ object JsInitContextLinkUtil extends Loggable {
     nodeInfoService.getNodeInfo(id) match {
       case t: EmptyBox =>
         <span>Node {id.value}</span>
-      case Full(node) =>
```
VinceMacBuche: Since you are doing the same for emptybox and Full(None) Can you do:

case Full(some(node)) => 
case _ =>

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -120,18 +153,62 @@ trait NodeInfoService {
    * @return
    */
   def getAllSystemNodeIds() : Box[Seq[NodeId]]
+
+
+  /**
+   * Getting something like a nodeinfo for pending / deleted nodes
+   */
+  def getPendingNodeInfos(): Box[Map[NodeId, NodeInfo]]
```
VinceMacBuche: Would it be easier if we had "getAll( state : InventoryStatus)" and refactor some logic that is the same for the three states ?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -120,18 +153,62 @@ trait NodeInfoService {
...
+  def getPendingNodeInfo(nodeId: NodeId): Box[Option[NodeInfo]]
```
VinceMacBuche: and getNodeInfo(nodeId, state) ?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
VinceMacBuche: Ah yes ... dumb VinceMacBuche is dumb

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -282,6 +360,92 @@ trait NodeInfoServiceCached extends NodeInfoService  with Loggable with CachedRe
...
+  override final def getDeletedNodeInfos(): Box[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(RemovedInventory)
+
+  private[this] def getNotAcceptedNodeInfo(nodeId: NodeId, status: InventoryStatus): Box[Option[NodeInfo]] = {
+    val dit = status match {
+      case AcceptedInventory => inventoryDit
+      case PendingInventory  => pendingDit
+      case RemovedInventory  => removedDit
+    }
+
+    def sureGet(con: RoLDAPConnection, dn: DN): Box[Option[LDAPEntry]] = {
```
ncharles: i'm a bit unsure about this method (its name or its lack of comment)

VinceMacBuche: As i understand, wraps the empty into a None ?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: yes, it's a resynchro

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: I didn't find a way to do so, without completely broking the other method. As it was one of the most used code in Rudder, I thought it was a little risky. Do you see how it could be done?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: I'm not sure, do you see how?

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: it's a method to factor the common code for not accepted (i.e, deleted or pending) nodes

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
VinceMacBuche: I think it's the answer on my firt comment on the ldap schema !! (or i don't understand anything!)

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
VinceMacBuche: Ok ... I tried like 15 minutes, I see it's complicated and it's ok it's two seprated methods that looks a lot the same :)

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
VinceMacBuche: AS saud above I see how, but it's complicated and probably risky (but it would provide a cache for deleted/pending nodes, not sure we want it though ... )

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -282,6 +360,92 @@ trait NodeInfoServiceCached extends NodeInfoService  with Loggable with CachedRe
...
+        case Empty      => Full(None)
+        case Full(e)    => Full(Some(e))
+      }
+    }
+
+    for {
+      con          <- ldap
+      optNodeEntry <- sureGet(con, dit.NODES.NODE.dn(nodeId))
+      nodeInfo     <- (optNodeEntry match {
+                        case None            => Full(Option.empty[NodeInfo])
```
VinceMacBuche: You can replace Option.empty by None

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -83,7 +116,7 @@ trait NodeInfoService {
    * @param nodeId
    * @return
    */
-  def getNodeInfo(nodeId: NodeId) : Box[NodeInfo]
+  def getNodeInfo(nodeId: NodeId) : Box[Option[NodeInfo]]
```
VinceMacBuche: Why changing to box Option and not keeping the failure 'the Node {id} was not found' ?

VinceMacBuche: To me, it's adding lot of boilerplate handling of the None case and emptybox case to do the same thing with specifying an error message for None

VinceMacBuche: But i understand it allows to handle the 'not found' which may not be a Failure more elegantly ...

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: Does not typecheck

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: yes, very strange :/

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
VinceMacBuche: It does on my branch :(

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: For LDAPNodeInfo: we are still using it in LdapQueryProcessor.

rudder-web/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
fanf: done

fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: Ah, it does now... Perhaps I changed something else since the time it wasn't compilling.

rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -282,6 +360,92 @@ trait NodeInfoServiceCached extends NodeInfoService  with Loggable with CachedRe
...
+    }
+
+    for {
+      con          <- ldap
+      optNodeEntry <- sureGet(con, dit.NODES.NODE.dn(nodeId))
+      nodeInfo     <- (optNodeEntry match {
+                        case None            => Full(None)
+                        case Some(nodeEntry) =>
+                          nodeEntry.getAsDn(A_CONTAINER_DN) match {
+                            case None     => Full(Option.empty[NodeInfo])
```
VinceMacBuche: You can put Full(None) too here

fanf:
PR rebased

VinceMacBuche:
Good to me!

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService4.scala
```diff
@@ -49,13 +49,16 @@ import com.normation.rudder.web.rest.RestUtils._
 import net.liftweb.json.JArray
 import net.liftweb.json.JsonDSL._
 import com.normation.rudder.services.nodes.NodeInfoService
+import scalaz.Need
```
ncharles: why this scalaZ import ?

fanf: yes, why ?

fanf: Removed as of 8d47b45

ncharles: only you may know.
This breaks the build :(

PR #1099, opened 2016-05-19
Fixes #8051: Compliance is not correctly computed if we receive run agent right after generation

fanf:
https://www.rudder-project.org/redmine/issues/8051

So it seems that somehow, the calls to future { } lead to some king of deadlock, and the corresponding code was never executed. It may be linked to the fact that there was a future{} in an method called from a future{}, or not. 

At least, removing the future and making the order of execution stricter lead to a comprehensive behavior from the user. If the last run in "select * from reportsexecution order by date desc limit 5;" is marked as finished, just after - and BEFORE the next processing of runs - the compliance is reprocessed, and the debug output of ```<logger name="com.normation.rudder.services.reports.CachedReportingServiceImpl" level="debug" />``` and 
```
<logger name="explain_compliance.root" level="trace" additivity="false">
    <appender-ref ref="OPSLOG" />
    <appender-ref ref="STDOUT" />
  </logger>
```
are consitant.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
@@ -49,6 +49,8 @@ import ExecutionContext.Implicits.global
 import scala.util.Success
 import com.normation.rudder.services.reports.CachedFindRuleNodeStatusReports
 import com.normation.inventory.domain.NodeId
+import org.joda.time.Duration
+import org.joda.time.format.PeriodFormat
 
 /**
  * That service contains most of the logic to merge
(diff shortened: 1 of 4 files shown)


rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
```diff
@@ -110,19 +121,10 @@ class ReportsExecutionService (
...
-              // Save new executions
-              writeExecutions.updateExecutions(reportExec) match {
-                case Full(result) =>
-                  val executionTime = DateTime.now().getMillis() - startTime
-                  logger.debug(s"[${FindNewReportsExecution.SERVICE_NAME}] Task #${processId}: Finished analysis in ${executionTime} ms. Added or updated ${result.size} agent runs, up to SQL table ID ${maxReportId} (last run time was ${maxDate.toString(format)})")
-                  statusUpdateRepository.setExecutionStatus(maxReportId, maxDate)
-
-                case eb:EmptyBox => val fail = eb ?~! "could not save nodes executions"
-                  logger.error(s"Could not save execution of Nodes from report ID ${lastReportId} - date ${lastReportDate} to ${(lastReportDate plusDays(maxDays)).toString(format)}, cause is : ${fail.messageChain}")
-              }
```
ncharles: shoudln't we have a log there to measure the time of the hook ?

fanf: I can add one.

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
fanf: Added one in the hook methods.

fanf:
Some more information: even with 300 node runs imported at ounce (directly with SQL inserts), the hook are taking <500ms on my laptop. 
Note that tests weren't testing with merge reports, but all the I/O are done to that point (the merge should not be the most costly).

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -344,11 +344,9 @@ object ExecutionBatch extends Loggable {
               val expireTime = currentConfig.creation.plus(updateValidityTime(intervalInfo))
 
               if(expireTime.isBefore(now)) {
-                ComplianceDebugLogger.node(nodeId).trace(s"Run config for node ${nodeId.value}: NoReportInInterval")
```
ncharles: why removing the logs ?

ncharles:
This looks awesome, except the question about logger removed.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
ncharles: ha, there are duplicate of runType

PR #1101, opened 2016-05-23
Fixes #8363: Some agent runs are not seen in the update process

fanf:
https://www.rudder-project.org/redmine/issues/8363


fanf:
Well, in fact, not sur it is the problem. I believe I get it even with the patch. 
And I changed the code to use SprintTemplate, and I believe I get a missing run, to, but I'm not sure and I can't reproduce. 
So, there is something very fishy here. 
I don't think it's some thread madness, because we are in an agent which is garding us from these thing. But nonetheless, we don't log every update, even is there is no other part of the code which is doing modification on that part of the db :/

fanf:
So, I can reproduce the problem with JdbcTemplate. So nothing to do with squeryl.

PR #1102, opened 2016-05-24
Fixes #8364: Only invalidate cache for node with new completed runs

fanf:
https://www.rudder-project.org/redmine/issues/8364


FlorianHeigl:
<3

PR #1107, opened 2016-05-31
Fixes #8427: Parallel traversal of new promises directories causes rules.new to remain after generation

fanf:
https://www.rudder-project.org/redmine/issues/8427


ncharles:
Fix should be in 2.11, there is also there:
val folders = collection.mutable.Buffer[(NodeConfigurat
ion, String, String, String)]()
                         // Writing the policy
                         for (node <- allNodeConfigs.filterKeys(configToWrite.co
ntains(_)).values.toSeq.par) {

fanf:
PR replaced by https://github.com/Normation/rudder/pull/1108

PR #1111, opened 2016-06-02
Fixes #8424: When updating runs, hooks should really be async

fanf:
https://www.rudder-project.org/redmine/issues/8424

So. That PR set back the future {} in hook, and I was not able to reproduce the thread problems I experienced the last time. So perhaps the problem was with the future{} in future{}, but I wasn't able to reproduce it either. 

So I'm a little unconfortable with that proposed solution, but can't see a better. Perhaps it would be helpfull to build a war with the fixe (before commiting) and using it for some time to see what happen (for ex. in our labo).


ncharles:
hum, we should indeed probably run a test version

fanf:
PR replaced by https://github.com/Normation/rudder/pull/1127

PR #1113, opened 2016-06-02
Fixes #8118: When a node send reports with a wrong config_id it is never marked as unresponsive

fanf:
https://www.rudder-project.org/redmine/issues/8118


VinceMacBuche:
The problem was that we grouped by nodeId/RuleId, so if there wasn't any data in it would have no value ? (and this is treated in the new toCompliance function )

fanf:
Not only, but part:
- when grouping like that with the flatmap in the case of nodeId, we could get less nodeIds than ask. For the one not returned, no bar update generated;
- if the compliance sum was exactly 0, we didn't know what to generate for the bar, so nothing was generated. 
So, this is the cause of the spliting.

PR #1118, opened 2016-06-20
Fixes #8409: When a node send reports with an unknow configID, no expected reports are showned

fanf:
https://www.rudder-project.org/redmine/issues/8409


fanf:
This pull request is not implementing what we want. Closing it.

fanf:
PR rebased

fanf:
This PR now implements what we want: 
- process all kind of unexpected version correctly, 
- when the result is a full "missing / unexpected", only display expected reports (i.e missing one), in a new table without the "compliance" row.

PR #1120, opened 2016-06-24
Fixes #8437: Recent changes graph can overflow the page width

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/8437

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
@@ -227,10 +227,14 @@ class RuleEditForm(
             $$( "#editRuleZone" ).tabs('select', ${tab});"""
           )) &
           JsRaw(s"""
-
             $$("#editRuleZonePortlet").removeClass("nodisplay");
             ${Replace("details", new RuleCompliance(rule, rootRuleCategory).display).toJsCmd};
             scrollToElement("${idToScroll}", ".rudder_col");
+            $$("#editRuleZone").bind( "show", function(event, ui) {
+              if((ui.panel.id== 'ruleComplianceTab')&&(recentChart !== undefined)) {
+                recentChart.flush();
+              }
+            });
             """
           )
         )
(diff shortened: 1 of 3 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -227,10 +227,14 @@ class RuleEditForm(
             $$( "#editRuleZone" ).tabs('select', ${tab});"""
           )) &
           JsRaw(s"""
-
             $$("#editRuleZonePortlet").removeClass("nodisplay");
             ${Replace("details", new RuleCompliance(rule, rootRuleCategory).display).toJsCmd};
             scrollToElement("${idToScroll}", ".rudder_col");
+            $$("#editRuleZone").bind( "show", function(event, ui) {
```
VinceMacBuche: 'var recentChart' is not defined before

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -227,10 +227,14 @@ class RuleEditForm(
...
+              if(ui.panel.id== 'ruleComplianceTab') {
+                recentChart.flush();
```
VinceMacBuche: Before flushing, you should check that recentChart is not undefined, unless you will have a js error when calling flush on 'undefined'

RaphaelGauthier:
PR updated

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
VinceMacBuche: Ok, i didn't know that you could 'declare' global variables by doing directly "xxx =" without var before so ok !

rudder-web/src/main/scala/com/normation/rudder/web/components/RuleEditForm.scala
```diff
@@ -227,10 +227,14 @@ class RuleEditForm(
             $$( "#editRuleZone" ).tabs('select', ${tab});"""
           )) &
           JsRaw(s"""
-
             $$("#editRuleZonePortlet").removeClass("nodisplay");
             ${Replace("details", new RuleCompliance(rule, rootRuleCategory).display).toJsCmd};
             scrollToElement("${idToScroll}", ".rudder_col");
+            $$("#editRuleZone").bind( "show", function(event, ui) {
+              if((ui.panel.id== 'ruleComplianceTab')&&(typeof recentChart !== 'undefined')) {
```
VinceMacBuche: you should not use typeof and only do:

```
recentChart !== undefined
```

RaphaelGauthier:
PR updated

PR #1125, opened 2016-07-05
Fixes #8630: Refresh button on recent changes graph on Rule details does not refresh data

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8630


VinceMacBuche:
Commit modified

PR #1137, opened 2016-07-20
Fixes #8722: Missing pattern matching case

fanf:
https://www.rudder-project.org/redmine/issues/8722


rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
```diff
@@ -62,6 +62,7 @@ import com.normation.rudder.domain.eventlog.ModifyAgentRunStartMinuteEventType
 import com.normation.rudder.domain.eventlog.ModifyAgentRunSplaytimeEventType
 import com.normation.rudder.reports._
 import com.normation.rudder.domain.eventlog.ModifyRudderSyslogProtocolEventType
+import scala.language.implicitConversions
```
ncharles: why do you need it there ?

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestUtils.scala
```diff
@@ -94,7 +94,7 @@ object RestUtils extends Loggable {
    *   - Add a new line at the end and beginning of an array and indent one more level array data
    *   - space after colon
    */
-  def render(value: JValue): Document = {
+  def rendrrer(value: JValue): Document = {
```
ncharles: hum - there is something fishy there: you change a method name, but the comment is not updated
link should probably be 
https://github.com/lift/framework/blob/2.6.3-release/core/json/src/main/scala/net/liftweb/json/JsonAST.scala#L406

ncharles: i don't think the method name should be changed according to pasted link

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
fanf: In 2.11, it raises compiler warning without it, because it is "an advanced feature". Add the import let the compiler you didn't wrote "implicit" by mistake, I guess

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestUtils.scala
fanf: ah, it is an error, perhaps a focus that wasn't intended at that point. 
Strange that is didn't broke anything. Dead code ?

fanf:
PR rebased

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestUtils.scala
ncharles: maybe it is used only at runtime ?
could you at least update the comment on top of the method to link to the actual version of the method in lift 2.6.3?

fanf: the code is really taken from the given version, and I don't know if it changed or not. Changing the comment without changing the code accordingly would be false, and misleading if someone want to refactor it latter on.

fanf:
OK, merging this PR

PR #1141, opened 2016-07-22
Fixes #8629: Allows generation-time javascript eval in directive parameters

fanf:
https://www.rudder-project.org/redmine/issues/8629

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/appconfig/RudderWebProperty.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/appconfig/RudderWebProperty.scala
@@ -39,6 +39,10 @@ package com.normation.rudder.domain.appconfig
 
 import com.normation.utils.HashcodeCaching
 import java.util.regex.Pattern
+import net.liftweb.common.Full
+import net.liftweb.common.Failure
+import net.liftweb.common.Box
+import ca.mrvisser.sealerate
 
 case class RudderWebPropertyName(value:String) extends HashcodeCaching
 
@@ -54,3 +58,28 @@ case class RudderWebProperty(
   , value      : String
   , description: String
 )
+
+
+/**
+ * A little domain language for feature switches
+ * (just enabled/disabled with the parsing)
+ */
+sealed trait FeatureSwitch { def name: String }
+object FeatureSwitch {
+
+  final case object Enabled  extends FeatureSwitch { override val name = "enabled"  }
+  final case object Disabled extends FeatureSwitch { override val name = "disabled" }
+
+  final val all: Set[FeatureSwitch] = sealerate.values[FeatureSwitch]
+
+  def parse(value: String): Box[FeatureSwitch] = {
+    value match {
+            case null|"" => Failure("An empty or null string can not be parsed as a feature switch status")
+            case s => s.trim.toLowerCase match {
+              case Enabled.name  => Full(Enabled)
+              case Disabled.name => Full(Disabled)
+              case _             => Failure(s"Cannot parse the given value as a valid feature switch status: '${value}'. Authorised values are: '${all.map( _.name).mkString(", ")}'")
+            }
+          }
+  }
+}
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
@@ -1,0 +1,614 @@
+/*
+*************************************************************************************
+* Copyright 2016 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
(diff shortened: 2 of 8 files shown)


fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/domain/appconfig/RudderWebProperty.scala
```diff
@@ -53,4 +57,29 @@ case class RudderWebProperty(
...
+)
+
+
+/**
+ * A little domain language for feature switches
+ * (just enabled/disabled with the parsing)
+ */
+sealed trait FeatureSwitch { def name: String }
+object FeatureSwitch {
+
```
fanf: The feature switch from the new search bar should used that one (it's in rudder core, the other is in web, we don't need two)

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,604 @@
...
+           * For now, failing because it seems to be the safe bet.
+           */
+          Failure(s"Value '${v}' starts with the ${EVAL} keyword, but the 'parameter evaluation feature' "
+                  +"is disable. Please, either don't use the keyword or enable the feature")
+      }
+    }
+  }
+
+  final object SandboxedJsEngine {
+     val MAX_EVAL_DURATION = (5, TimeUnit.SECONDS)
```
fanf: The value is purelly arbitrary. We expects that a normal use case ends in tens of ms. But we don't want the user to have a whole generation fails because it scripts took 2 seconds for reasons. 
As it is something rather exceptionnal, and which will ends the Policy Generation, the value can be rather hight.

fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,614 @@
...
+  final val EVAL = "$evaljs "
+  final val EVAL_SIZE = EVAL.size
+
+  final object DisabledEngine extends JsEngine {
+    /*
+     * Eval does nothing on variable without the EVAL keyword, and
+     * fails on variable with the keyword.
+     */
+    def eval(variable: Variable, lib: JsRudderLibBinding): Box[Variable] = {
+      variable.values.find( _.startsWith(EVAL) ) match {
```
ncharles: Password fields start with the name of the hash algorithm:
plain:$evaljs rudder.password.sha512('secret', '11111')
aix-ssha512:{ssha1}10$FxDqBWMKU8aiIvGV$LWzO4RcL8wZYW654QSWjrvWv.Iq

So this "startsWith" fails :-1:

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,614 @@
...
+          case x: FilePermission     if( x.getActions == "read" && filePerms.contains(x.getName)    ) => // ok
+          case x: SecurityPermission if( securityPerms.exists( p => x.getName.startsWith(p) )       ) => // ok
+          case x: RuntimePermission  if( x.getName.startsWith("accessClassInPackage.jdk.nashorn")   ) => // ok
+          case x: RuntimePermission  if( x.getName.startsWith("accessClassInPackage.sun.security.") ) => // ok
+          case x: RuntimePermission  if( x.getName.startsWith("accessClassInPackage.sun.org.mozilla.javascript") ) => // ok
+          case x: RuntimePermission  if( x.getName.startsWith("accessClassInPackage.jdk.internal.org.objectweb.asm") ) => // ok
+          case x: RuntimePermission  if( runtimePerms.contains(x.getName)  ) => // ok
+          case x: ReflectPermission  if( reflectPerms.contains(x.getName)  ) => // ok
+          case x: PropertyPermission if( x.getActions == "read"            ) => // ok
+          case x: FilePermission if(x.getName.startsWith("/home/fanf") ) =>
```
ncharles: Forbid 'rudder-jsengine-1' access to ("java.io.FilePermission" "/home/nicolas/dev/3.2/rudder/rudder-web/target/classes/org/apache/commons/codec/digest/Sha2Crypt.class" "read")

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
fanf: I don't think we should change the "startWith" rules, but change password to add a 'if plain && startswith $eval, let it as it is", and add a special unserialize rule for that (i.e: if not algo:xxx, check is startswith $eval, if so plain)

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
fanf: that was only for debugging, to allows println to be done just a few line below (else, we are stuck in a checkPermission loop). Remove it :)

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
ncharles: actually, we authorize now everywhere the script, so it is checked by regexp

ncharles:
Replaced by https://github.com/Normation/rudder/pull/1144

PR #1144, opened 2016-07-28
Fixes #8629: Allows generation-time javascript eval in directive para…

ncharles:
…meters

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
@@ -1,0 +1,708 @@
+/*
+*************************************************************************************
+* Copyright 2016 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
+* distributed under the license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* Rudder is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+*
+*************************************************************************************
+*/
+
+package com.normation.rudder.services.policies
+
+import java.security.Permission
+import java.util.PropertyPermission
+import java.util.concurrent._
+import java.util.concurrent.Callable
+import java.util.concurrent.Executors
+import java.util.concurrent.TimeUnit
+import java.util.concurrent.TimeoutException
+import java.util.concurrent.atomic.AtomicInteger
+import java.io.File
+
+import scala.language.implicitConversions
+
+import com.normation.cfclerk.domain.HashAlgoConstraint._
+
+import com.normation.cfclerk.domain.Variable
+import com.normation.rudder.domain.appconfig.FeatureSwitch
+import com.normation.rudder.services.policies.HashOsType._
+import com.normation.rudder.services.policies.JsEngine._
+import com.normation.utils.Control. _
+
+import ca.mrvisser.sealerate
+import javax.script.Bindings
+import javax.script.ScriptEngine
+import javax.script.ScriptEngineManager
+import javax.script.ScriptException
+import net.liftweb.common.Box
+import net.liftweb.common.Empty
+import net.liftweb.common.Failure
+import net.liftweb.common.Full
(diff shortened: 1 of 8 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,708 @@
...
+ *   * auto automatically choose the encryption based on the node type
+ *   * unix generated Unix crypt password compatible hashes (Linux, BSD, ...)
+ *   * aix generates AIX password compatible hashes
+ */
+final class JsRudderLibImpl(
+  hashKind: HashOsType
+) {
+
+  ///// simple hash algorithms /////
+  private val hash = new JsLibHash()
```
VinceMacBuche: Why do you neet a private val if you have a getter ? I would only have a val

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,708 @@
...
+
+        def md5    (s: String, salt: String): String = super.aixMd5(s, salt)
+        def sha256 (s: String, salt: String): String = super.aixSha256(s, salt)
+        def sha512 (s: String, salt: String): String = super.aixSha512(s, salt)
+
+      }
+  }
+
+  // with the getter, it will be accessed with rudder.password...
+  def getPassword = password
```
VinceMacBuche: Same question as above for getter/ private val ...

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
ncharles: i suspect this is for visibility issue. With the getHash, it gives access to hash - but if hash is public, there might be some java magic  going on

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,708 @@
...
+      case sha:String if (sha=="sha512" || sha=="sha-512") => cryptSha512(s)
+      case _ => // unknown value, fail
+        throw new NoSuchAlgorithmException(s"Evaluating script 'unix(${algo}, ${s})' failed, as algorithm ${algo} is not recognized")
+    }
+  }
+
+  def unix(algo: String, s: String, salt: String): String = {
+    algo.toLowerCase match {
+      case "md5" => cryptMd5(s)
+      case sha:String if (sha=="sha256" || sha=="sha-256") => aixSha256(s, salt)
```
fanf: can be wrote case "sha256" | "sha-256" =>

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,708 @@
...
+      case _ => // unknown value, fail
+        throw new NoSuchAlgorithmException(s"Evaluating script 'unix(${algo}, ${s}, ${salt})' failed, as algorithm ${algo} is not recognized")
+    }
+  }
+
+  def aix(algo: String, s: String): String = {
+    algo.toLowerCase match {
+      case "md5" => aixMd5(s)
+      case sha:String if (sha=="sha256" || sha=="sha-256") => cryptSha256(s)
+      case sha:String if (sha=="sha512" || sha=="sha-512") => aixSha512(s)
```
fanf: idem, directly: case "sha-256" | "sha256" => ...

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
fanf: because Java conventions are used for the scripting engine

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
fanf: same answer

ncharles:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -0,0 +1,708 @@
...
+
+/*
+ * Our JsEngine.
+ */
+final object JsEngine {
+  // Several evals: one default and one JS (in the future, we may have several language)
+  final val DEFAULT_EVAL = "eval"
+  final val EVALJS = "evaljs"
+
+  final val default_pattern = """(?ms)(.*)\$\{eval\s+(.*)}(.*)""".r
```
fanf: Parsing ${evaljs ...} with regex won't work in intersting fashion, especially with rudder variables and js blocks. 
In all case, it will have a massive impact on generation time - we switch from a "startWith" method, dicidable in constant(EVALJS.size) time, to an unbound one - at least o(n), but with extrem backtracking bad case, see stackoferflow recent outage. 

Finally, I don't even think we should have {}. It add visual clutters for no reason, because really, we want that the whole input is a script or not. Supporting multiple script need to have a ROBUST input parser, understanding each language to know if the closing } is its or ours - something I don't even want to start coding. And if it is to have part of the input "as is", here you go : javascript is able to manage strings !

So to sum-up: 
- I don't thing we should have {} letting the user think it can have several eval in an input ; 
- I don't thing we should autorise $evaljs at an other place than the strict begining of the field.

fanf: (yes, the second statement mean that input fields need adaptation to handle that. But is it more than the password one ? And if really we can't specialise input field to handle $evaljs correctly, perhaps we should had a "authorised ignored starting string before $evaljs, declared for each field")

peckpeck: I'm OK with point 1 and 2

ncharles:
PR updated

ncharles:
i'm rebasing it

ncharles:
pr updated

VinceMacBuche:
OK, merging this PR

PR #1153, opened 2016-07-28
Fixes #8782: Fix details on script (remove $ from token and fix double password prefix)

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8782

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
@@ -85,8 +85,6 @@ final object HashOsType {
   def all = sealerate.values[HashOsType]
 }
 
-
-
 /*
  * Define implicit bytes from string methods
  */
@@ -307,7 +305,6 @@ object JsRudderLibBinding {
   }
 }
 
-
 /**
  * This class provides the Rhino (java 7) or Longhorn (Java 8 & up)
  * java script engine.
@@ -317,7 +314,6 @@ object JsRudderLibBinding {
  *
  */
 final object JsEngineProvider {
-
 
   /**
    * Initialize a new JsEngine with the correct bindings.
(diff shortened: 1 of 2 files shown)


VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -358,48 +354,54 @@ sealed trait JsEngine {
...
   final object DisabledEngine extends JsEngine {
     /*
      * Eval does nothing on variable without the EVAL keyword, and
      * fails on variable with the keyword.
      */
     def eval(variable: Variable, lib: JsRudderLibBinding): Box[Variable] = {
       val (default, js, _) = getEvaluatorTuple(variable)
-      variable.values.find( x => x.startsWith(default) || x.startsWith(js) ) match {
-        case None    => Full(variable)
+      variable.values.find( x => x.startsWith(default)) match {
```
ncharles: i think it's less readable to have two separated tests. Or maybe add a comment before (i guess it's for performance reason), and move the comme "Here, we nned to chose" just before this line

ncharles:
Appart from the comment part, this lloks great to me !

rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
VinceMacBuche: I find it less readable too, but it was not able to differentiate which EVAL was found, 

Séparate them allow a much more précise error message

Le 29 juillet 2016 07:53:43 GMT+02:00, Nicolas Charles <notifications@github.com> a écrit :
>>    final object DisabledEngine extends JsEngine {
>>      /*
>>       * Eval does nothing on variable without the EVAL keyword, and
>>       * fails on variable with the keyword.
>>       */
>>      def eval(variable: Variable, lib: JsRudderLibBinding):
>Box[Variable] = {
>>        val (default, js, _) = getEvaluatorTuple(variable)
>> -      variable.values.find( x => x.startsWith(default) ||
>x.startsWith(js) ) match {
>> -        case None    => Full(variable)
>> +      variable.values.find( x => x.startsWith(default)) match {
>
>i think it's less readable to have two separated tests. Or maybe add a
>comment before (i guess it's for performance reason), and move the
>comme "Here, we nned to chose" just before this line
>
>---
>You are receiving this because you authored the thread.
>https://github.com/Normation/rudder/pull/1153/files/150059137b7d15c19d9169ad39ce59aea863c83d#r72744578

-- 
Envoyé de mon appareil Android avec K-9 Mail. Veuillez excuser ma brièveté.

ncharles: having the "Here, we need to chose betwwen" in between masks that they are related and do the same

PR #1163, opened 2016-08-17
Fixes #8856: JS engine tests broken due to security exeception when using jdk 1.8

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8856


VinceMacBuche:
Commit modified

PR #1164, opened 2016-08-17
Fixes #8857: Some Javascript engine fails when accessing system values

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8857


VinceMacBuche:
OK, merging this PR

PR #1165, opened 2016-08-17
Fixes #8863: When no edition rights are given to Groups, you drag an drop them between categories

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8863


VinceMacBuche:
OK, merging this PR

PR #1169, opened 2016-08-19
Fixes #8874: Unable to save/edit a directive via rudder web ui when technique use an optionnal SELECT field

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8874

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/model/DirectiveFieldEditors.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/model/DirectiveFieldEditors.scala
@@ -108,7 +108,6 @@ class ReadOnlyTextField(val id:String) extends DirectiveField {
   def getDefaultValue = ""
 }
 
-
 /**
  * A textarea field, with a css class
  * "textareaField"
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/model/DirectiveFieldEditors.scala
```diff
@@ -176,10 +175,10 @@ class SelectField(val id: String, items: Seq[ValueLabel]) extends DirectiveField
     val valuesByLabel = items.map(_.reverse.tuple).toMap
     val labelsByValue = items.map(_.tuple).toMap
 
-    val defaultLabels = values.map(labelsByValue(_))
+    val defaultLabels = values.map(labelsByValue.getOrElse(_,""))
```
VinceMacBuche: Maybe we need a logger warning here...

I think invalid values should be handled before and should not arrive at this point

fanf: Not sure a warning will serve anything but spoil the logs. Perhaps in debug, if somebody is trying to understand a related problem. But agree, the model should not even authorise incorrect values to reach that point.

VinceMacBuche:
OK, merging this PR

PR #1175, opened 2016-08-23
Fixes #8492: Cannot access Technique Editor with role technique_all

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8492


VinceMacBuche:
OK, merging this PR

PR #1176, opened 2016-08-24
Fixes #8696: When there is an error in a directive form, the target rule is unchecked

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/8696


fanf:
OK, merging this PR

PR #1184, opened 2016-08-31
Fixes #2787: Use Doobie or Slick in place of SpringTemplate / squeryl / pur JDBC

fanf:
https://www.rudder-project.org/redmine/issues/2787

A new version which use Doobie in place of slick, formelly #1190 

Overall, it is a MUCH better experience than with Slick, globally without surprises.

We can see that there is a lot less boilerplate. Generally, the type are more well behaved than in Slick, i.e, there was less times than I just tried random thing to make the compiler happy.

One downside is for table with a lot of fields: here, we are doing them as in sql, with "... values(?,?,?,?,?,?,?,?,?,?)". Barelly readable. I think we need to find convention about how to deals with that.

Some convention should still be found regarding that case.

The slick version, for anyone interested, lives in https://github.com/fanf/rudder/tree/old_slick/arch_2787/use_doobie_or_slick_in_place_of_springtemplate_squeryl_pur_jdbc


jooooooon:
Didn't you mean to squash all these commits into the last one?

rudder-core/src/main/scala/com/normation/rudder/db/SlickSchema.scala
```diff
@@ -0,0 +1,676 @@
...
+    def nodeId      = column[String]("nodeid")
+    def date        = column[DateTime]("date")
+    def nodeConfigId= column[Option[String]]("nodeconfigid")
+    def isCompleted = column[Boolean]("complete")
+
+
+    /*
+     * WHAT IS THE GOAL OF THAT ? SHOULD WE SORT BY IT OR SOMETHING, SOMEWHERE ?
+     */
+    def insertionId = column[Long]("insertionid")
```
ncharles: when we getNodesLastRun, we sort by insertionId desc

rudder-core/src/main/scala/com/normation/rudder/db/SlickSchema.scala
```diff
@@ -0,0 +1,676 @@
...
+  }
+
+  //////////
+
+  final case class AgentRun(
+      nodeId      : String
+    , date        : DateTime
+    , nodeConfigId: Option[String]
+    , isCompleted : Boolean
+    , insertionId : Long // PURPOSE ?
```
ncharles: the goal is to have the id of the reports - rationnal is: latest we receive is the latest sent (network paquet ordering is more reliable than nodes times)

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala
```diff
@@ -210,66 +198,58 @@ class EventLogJdbcRepository(
...
-      Full(jdbcTemplate.query(select, EventLogReportsMapper).toSeq)
-    } match {
-      case Success(events) => events
-      case Catch(e) => val msg = s"could not find event log with request ${select} cause: ${e}"
-        logger.error(msg)
-        Failure(msg)
-    }
+
+    val q = s"""
+      select crid, eventtype, id, modificationid, principal, creationdate, causeid, severity, reason, data
```
ncharles: query was 
 "SELECT id, creationDate, modificationId, principal, eventType, severity, data, reason, causeId FROM EventLog

PR #1189, opened 2016-09-05
Fixes #8924: Policy mode API (Global, Directive, Node) 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8924

### DIFF
--- a/rudder-core/src/main/resources/ldap/rudder.schema
+++ b/rudder-core/src/main/resources/ldap/rudder.schema
@@ -316,6 +316,14 @@ attributetype ( RudderAttributes:230
   EQUALITY caseIgnoreMatch
   SUBSTR caseIgnoreSubstringsMatch
   SYNTAX 1.3.6.1.4.1.1466.115.121.1.15{4096} )
+  
+### Policy mode, common for Nodes/Directive at first, Surely Rules and groups later
+attributetype ( RudderAttributes:231
+  NAME 'policyMode'
+  DESC 'Policy mode (Enforce/Verify) for this element '
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15
+  EQUALITY caseIgnoreMatch
+  SUBSTR caseIgnoreSubstringsMatch )
 
 
 ### API principal and tokens 
@@ -416,7 +424,8 @@ objectclass ( RudderObjectClasses:1
   SUP top
   STRUCTURAL
   MUST ( nodeId $ cn $ isSystem $ isBroken)
-  MAY ( description $ serializedNodeProperty $ serializedAgentRunInterval $ serializedHeartbeatRunConfiguration ) )
+  MAY ( description $ serializedNodeProperty $ serializedAgentRunInterval $
+        serializedHeartbeatRunConfiguration $ policyMode ) )
   
 objectclass ( RudderObjectClasses:2
   NAME 'rudderPolicyServer' 
@@ -484,7 +493,7 @@ objectclass ( RudderObjectClasses:23
   STRUCTURAL
   MUST ( directiveId $ techniqueVersion )
   MAY ( cn $ description $ longDescription $ isEnabled $ isSystem $
-        directivePriority $ directiveVariable  ) )
+        directivePriority $ directiveVariable $ policyMode ) )
 
 ### rules ###
   
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
@@ -37,15 +37,17 @@
 
 package com.normation.rudder.domain.nodes
 
+import com.normation.inventory.domain.FullInventory
+
 import com.normation.inventory.domain.NodeId
-import com.normation.utils.HashcodeCaching
-import com.normation.rudder.reports.ReportingConfiguration
-import net.liftweb.json.JsonAST.JObject
-import org.joda.time.DateTime
+import com.normation.rudder.domain.policies.SimpleDiff
+import com.normation.rudder.domain.policies.PolicyMode
 import com.normation.rudder.reports.AgentRunInterval
 import com.normation.rudder.reports.HeartbeatConfiguration
-import com.normation.rudder.domain.policies.SimpleDiff
-import com.normation.inventory.domain.FullInventory
+import com.normation.rudder.reports.ReportingConfiguration
+import com.normation.utils.HashcodeCaching
+
+import org.joda.time.DateTime
 
 /**
  * The entry point for a REGISTERED node in Rudder.
@@ -63,6 +65,7 @@ case class Node(
   , creationDate              : DateTime
   , nodeReportingConfiguration: ReportingConfiguration
   , properties                : Seq[NodeProperty]
+  , policyMode                : Option[PolicyMode]
 ) extends HashcodeCaching
 
 case object Node {
@@ -77,6 +80,7 @@ case object Node {
       , inventory.node.inventoryDate.getOrElse(new DateTime(0))
       , ReportingConfiguration(None,None)
       , Seq()
+      , None
     )
   }
 }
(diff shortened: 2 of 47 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
```diff
@@ -64,61 +65,54 @@ case class DirectiveId(value:String) extends HashcodeCaching
...
+     * i.e: "SEC-042 Debian Etch"
+     * Can not be empty nor null.
+     */
+  , name : String
+    /**
+     * Short description, typically used as field description
+     * Can not be empty nor null.
+     */
+  , shortDescription : String
+  , policyMode : Option[PolicyMode]
```
fanf: could you adda a description like for other parameters here?

rudder-core/src/main/scala/com/normation/rudder/policyMode/PolicyMode.scala
```diff
@@ -0,0 +1,83 @@
...
+case object Verify extends PolicyMode {
+  val name = "verify"
+}
+
+case object Enforce extends PolicyMode {
+  val name = "enforce"
+}
+
+object PolicyModeName {
+  val allModes : List[PolicyMode] = Verify :: Enforce :: Nil
```
fanf: Could you rename that object to PolicyMode, and  put the two case obhect into it? So we have the nice convention:

```
sealed trait MyGADT
final object MyGADT {
  final object CASE1 { ... }
  final object CASE2 { ... }
  
  def allModes = ca.mrvisser.sealerate.values[MyGADT]
}


```

rudder-core/src/main/scala/com/normation/rudder/policyMode/PolicyMode.scala
```diff
@@ -0,0 +1,83 @@
...
+    if (globalValue.overridable) {
+      nodeMode match {
+        case Some(Verify) => Verify
+        case _ => directiveMode.getOrElse(globalValue.mode)
+      }
+    } else {
+      globalValue.mode
+    }
+  }
+}
```
fanf: Could you add an other method that allows to combine several policyMode and get the result? It is necessary for now, if directives for a same technique have different modes. I believe we could just have a variant with that signature:
```
def computeMode(globalValue : GlobalPolicyMode, nodeMode : Option[PolicyMode], directivesMode : Seq[Option[PolicyMode]])
```

In fact, I'm wondering if we really want to have the one you did for now, because we always need to look for the whole technique's directives status applied to that node, not just one in isolation.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/resources/ldap/rudder.schema
```diff
@@ -484,7 +493,7 @@ objectclass ( RudderObjectClasses:23
   STRUCTURAL
   MUST ( directiveId $ techniqueVersion )
   MAY ( cn $ description $ longDescription $ isEnabled $ isSystem $
-        directivePriority $ directiveVariable  ) )
+        directivePriority $ directiveVariable $ policyMode ) )
```
ncharles: you need to update also the 099-1-rudder.ldif file in test/resources/ldap-data/schema

rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
```diff
@@ -44,8 +44,9 @@ import com.normation.cfclerk.domain.TechniqueVersion
 import com.normation.utils.HashcodeCaching
 import com.normation.cfclerk.domain.SectionSpec
 import com.normation.cfclerk.domain.Technique
+import com.normation.rudder.policyMode.PolicyMode
 
-case class DirectiveId(value:String) extends HashcodeCaching
+case class DirectiveId(value : String) extends HashcodeCaching
```
ncharles: why the extra space ?

rudder-core/src/main/scala/com/normation/rudder/policyMode/PolicyMode.scala
```diff
@@ -0,0 +1,89 @@
...
+      case None =>
+         Failure(s"Unable to parse policy mode name '${value}'. was expecting ${allModes.map(_.name).mkString("'", "' or '", "'")}.")
+      case Some(mode) =>
+        Full(mode)
+    }
+  }
+}
+
+case class GlobalPolicyMode (
+    mode : PolicyMode
```
ncharles: could you align the parameters type ?

VinceMacBuche:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -558,6 +583,9 @@ class LDAPEntityMapper(
     entry +=! (A_PRIORITY, directive.priority.toString)
     entry +=! (A_IS_ENABLED, directive.isEnabled.toLDAPString)
     entry +=! (A_IS_SYSTEM, directive.isSystem.toLDAPString)
+    for {
+      mode <- directive.policyMode
+    }  entry +=! (A_POLICY_MODE, mode.name)
```
ncharles: why this construct ? (compared to a map, or a yield ?)

VinceMacBuche:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
VinceMacBuche: It's foreach syntax with a for 

I don't need the result, so don't need yield/flatmap or map

rudder-core/src/main/resources/ldap/rudder.schema
VinceMacBuche: good catch!

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPI8.scala
```diff
@@ -0,0 +1,112 @@
+/*
+*************************************************************************************
+* Copyright 2013 Normation SAS
```
ncharles: 2016, isn't it ?

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -0,0 +1,110 @@
+/*
+*************************************************************************************
+* Copyright 2013 Normation SAS
```
ncharles: 2016

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
ncharles: the readability is poor compared to 
directive.policyMode.map( x => entry +=! (A_POLICY_MODE, x.name))

VinceMacBuche: Could do it with a foreach, but i prefer the for way

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
```diff
@@ -102,27 +102,46 @@ sealed trait NodeDiff
...
+}
 
 /**
  * Diff on a change on agent run period
  */
-final case class ModifyNodeAgentRunDiff(
-    id         : NodeId
-  , modAgentRun: Option[SimpleDiff[Option[AgentRunInterval]]]
-) extends NodeDiff with HashcodeCaching
+object ModifyNodeAgentRunDiff{
```
ncharles: why replacing the case class by an object ?

rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
```diff
@@ -102,27 +102,46 @@ sealed trait NodeDiff
...
+ */
+final case class ModifyNodeDiff(
     id           : NodeId
+  , modHeartbeat : Option[SimpleDiff[Option[HeartbeatConfiguration]]]
+  , modAgentRun  : Option[SimpleDiff[Option[AgentRunInterval]]]
   , modProperties: Option[SimpleDiff[Seq[NodeProperty]]]
+  , modPolicyMode: Option[SimpleDiff[Option[PolicyMode]]]
 )
 
+object ModifyNodeDiff {
```
ncharles: why both a case class and an object ??

rudder-core/src/main/scala/com/normation/rudder/repository/EventLogRepository.scala
```diff
@@ -342,46 +340,18 @@ trait EventLogRepository {
...
-          , reason = reason
-        )
-    )
-  }
-  def saveModifyNodeProperties(
-      modId: ModificationId
-    , principal: EventActor
-    , modifyDiff: ModifyNodePropertiesDiff
-    , reason:Option[String]) = {
+  def saveModifyNode(
```
ncharles: i think i prefered the name NodeProperties, as modify node leads me to think we are changing the node, and not properties attached to it

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
```diff
@@ -60,78 +59,19 @@ class WoLDAPNodeRepository(
...
-
-  /**
-   * Update the node with the given ID with the given
-   * parameters.
-   *
-   * If the node is not in the repos, the method fails.
-   * If the node is a system one, the methods fails.
-   */
-  private[this] def update(nodeId: NodeId, updateNode: Node => Node, logAction: (Node, Node) => Box[EventLog]) : Box[Node] = {
+  def updateNode(node: Node, modId: ModificationId, actor:EventActor, reason:Option[String]) : Box[Node] = {
```
ncharles: again, here i tend to think the method change a node

rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogDetailsService.scala
```diff
@@ -967,13 +941,54 @@ class EventLogDetailsServiceImpl(
...
       ModifyNodePropertiesDiff(nodeId, details)
     }
-    getModifyNodeDetails(xml, "properties", extract, build)
+    getModifyNodeDetails(xml, "properties", extractNodeProperties(xml), build)
   }
 
+  def getModifyNodeDetails(xml:NodeSeq) : Box[ModifyNodeDiff] = {
+    for {
+      entry <- getEntryContent(xml)
+      node                    <- (entry \ "node").headOption ?~! ("Entry type is not node : " + entry)
```
ncharles: could you align the <- please ?

rudder-web/src/main/scala/com/normation/rudder/web/components/ShowNodeDetailsFromNode.scala
```diff
@@ -98,6 +98,17 @@ class ShowNodeDetailsFromNode(
   private[this] val nodeRepo             = RudderConfig.woNodeRepository
   private[this] val asyncDeploymentAgent = RudderConfig.asyncDeploymentAgent
   private[this] val configService = RudderConfig.configService
+  private[this] var nodeInfo = nodeInfoService.getNodeInfo(nodeId) match {
+    case Full(Some(info)) => info
+    case eb : EmptyBox =>
+      val e = eb ?~! s"Could not get Node '${nodeId.value}' details"
+      logger.error(e.messageChain)
+      throw new Exception(e.messageChain)
```
ncharles: i'm really not convinced by throwing an exception when loading the page. 
With this, we go from a safe-ish approach where the error The node with id '${nodeId.value}' was not found was displayed to user to an exception, that won't probably be of any help to user :/

rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
VinceMacBuche: The goal (discussed with @fanf) is to remove the old events (ModifyNodeAgentRun, ModifyNodeProperties ModifyNodeHeartbeat) by one and only one event that is 'ModifyNode'

I removed possibility to save/create those events but we need to be able to read previous events

VinceMacBuche: Those object are only used as syntax sugar when we are reading the old event (Could be replaced by ModifyNode(agentRun,None,None,None)

rudder-core/src/main/scala/com/normation/rudder/repository/EventLogRepository.scala
VinceMacBuche: We are changing the node now, not only 'properties'

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

PR #1192, opened 2016-09-09
Fixes #8957: Web API doesn't process uuid for newly added directive

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8957


VinceMacBuche:
OK, merging this PR

PR #1193, opened 2016-09-12
Fixes #7221: Set dry-run mode on each directive call

fanf:
https://www.rudder-project.org/redmine/issues/7221

The diff should be viewed with "?w=1".

The main change is the appearance of BuildBundleSequence.scala that contains all the code to actually write the bundle sequence related vars. It was previously in PrepareTemplateVariables.scala, but that was becoming far too complexe. 
The most interesting part in BuildBundleSequence is """prepareBundleVars""" and used data structure. The remaining is mostly implementation details.

Most of the remaining files are either test file (a full generation, so that's for big number of files), or just getting the [global|directive|node] policy mode and passing them around. 

It actually output the correct bundle sequence for the root node with two directives, with a default global mode to "verify / overridable", one directive (clock) set to "Enforce", the other to nothing (so default). 

It is also a great occasion to see what it looks like to set up BY HAND that very simple directives/rule config. It's really time to get a middle format (JSON \o/ or HOCON or YAML or whatever) for that.


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -0,0 +1,373 @@
...
+
+
+
+  /*
+   * For each techniques, build:
+   * - the promiser
+   * - the list of bundle included,
+   */
+  def buildTechniqueBundles(nodeId: NodeId)(t3: (Technique, List[BundleOrder], PolicyMode)): TechniqueBundles = {
+    // naming things to make them cleare
```
ncharles: clear

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -0,0 +1,373 @@
...
+
+  /*
+   * Each bundle must be preceded by the correct "set_dry_mode" mode,
+   * and we always end with a "set_dry_mode("false")".
+   * Also, promiser must be differents for all items so that cfengine
+   * doesn't try to avoid to do the set, so we are using the technique
+   * promiser.
+   */
+  implicit final class DryRunManagement(bundles: List[TechniqueBundles]) {
+    def addDryRunManagement: List[TechniqueBundles] = bundles match {
```
ncharles: if we have a global dry mode enabled, shouldn't we let the set_dry_mode("true") ? And also, is the global mode info computed within the TechniqueBundles (so that we cannot override a global mode by mistake)?

fanf: > if we have a global dry mode enabled, shouldn't we let the set_dry_mode("true") ?

Well, we discussed it internally, but: if we do that, it means that we want to alter the behavior for bundle outside user bundle, so necesserally system bundle (as in "used by Rudder to make Rudder works, not by the user to make his infra works") - by definition. The consensus seems to have been that "dry mode" is different to "rudder agent does nothing on that node", and "dry mode" was still implying that the rudder inner tools need to work as expected (and so, error should be corrected).

fanf: > s the global mode info computed within the TechniqueBundles (so that we cannot override a global mode by mistake)?

The mode is computed for each technique based on the 3 inputs: [global mode, node mode, directives for that technique mode]

fanf:
PR rebased

fanf:
PR rebased

PR #1194, opened 2016-09-15
Fixes #9057: Allow the use of node properties info in Directive parameters

fanf:
Don't merge yet: we need a guard for people saving data in json, then disabling the feature. Now, it is an exception because the unserializer is lost. 
==> in fact, it's OK. It's just in a case you come back to a rudder version that does not understand json in nodeproperties. 
=> that means that you can't have migration from more recent to older with JSON values

https://www.rudder-project.org/redmine/issues/9057

This pull request allows to use node properties un directive parameters. The implemented rules are: 
- the syntax is *exactly* the same than in ncf/technique, i.e: ${node.properties[key1][key2]} 
- when there is only one level of key, the left hand side can be a simple string (as it is for now). If there is several level of keys, then the left hand side must be a valid json data string. 
- a non matching property leads to a generation error. It means that if a node does not have the property """key1 = { "key2": "value" }""", the generation will fail (it is the case for any level of key, even the first one) 
- for only one level of key, the left hand side is taken as it is. 
- for level > 1 of keys, if the subkey exists, then the whole value is taken: if it is a string, then it is used. If it is an other json value, it is compact-rendered to a string
- everything is case sensitive  
- the feature is hidden behind a feature switch, 
- space and line breaks are **authorized**  in the middle of a ${} expression around separator chars like "." or "]"

- we can add a flag that tells to generation part "don't touch this, let it go to the node"
 ```${node.properties[key1][key2]|node} ```
- we can add a default value if the node is missing property. The default value can be other parameters, or string, or string with parameters. 

Some example of what is accepted:
```
${node.properties[key1][key2] | default = "foo" }
${node.properties[key1][key2] | default = """ {"foo"="bar"} """}
${node.properties[key1][key2] | default = ${node.properties[other] | node}}
${node.properties[key1][key2] | default = ${node.properties[other] | default="foo"}}
${node.properties[key1][key2] | default = ${rudder.param.key3}}
```

It does not make sens to have both "node" and "defaults", so for now we don't allow several options.

Also, there is (a lot of) tests to demonstrate the strange cases. 

The pull request also correct a problem in node properties REST API that was not correctly managing badly formatted json properties. Now, on that case, an error is returned to the user. 
It also add the support of real json Value, because before it, we only had "key=value" where "value" was a quoted string, understood as a quoted string, neither a json value, not even on the agent side. 

Finally, when https://www.rudder-project.org/redmine/issues/9118 will be merged, error are reported like:
```
⇨ Policy update error for process '285' at 2016-09-26 01:56:05
⇨ Cannot build target configuration node
⇨ Error with parameters expansion for node 'agent1.rudder.local' (e58d1658-7cae-4b10-aafa-7147b8d02095): When processing directive '15. Welcome message': On variable 'MOTD': Missing property '${node.properties[datacenter][Europe][France]}'
⇨ Error with parameters expansion for node 'agent2.rudder.local' (cc987140-c5c3-4ddb-992f-c837083ef025): When processing directive '15. Welcome message': On variable 'MOTD': Missing property '${node.properties[datacenter][Europe][France]}'
```

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
@@ -46,6 +46,8 @@ import com.normation.rudder.reports.AgentRunInterval
 import com.normation.rudder.reports.HeartbeatConfiguration
 import com.normation.rudder.domain.policies.SimpleDiff
 import com.normation.inventory.domain.FullInventory
+import net.liftweb.json.JsonAST.JValue
+import net.liftweb.json.JsonAST.JString
 
 /**
  * The entry point for a REGISTERED node in Rudder.
@@ -81,7 +83,12 @@ case object Node {
   }
 }
 
-case class NodeProperty(name: String, value: String)
+case class NodeProperty(name: String, value: JValue) {
+  def renderValue: String = value match {
+    case JString(s) => s
+    case v          => net.liftweb.json.compactRender(v)
+  }
+}
 
 /**
  * Node diff for event logs:
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -128,6 +128,7 @@ trait DeploymentService extends Loggable {
       agentRunStartMinute <- getAgentRunStartMinute() ?~! "Could not get agent run start time (minute)"
       agentRunStartHour   <- getAgentRunStartHour() ?~! "Could not get agent run start time (hour)"
       scriptEngineEnabled <- getScriptEngineEnabled() ?~! "Could not get if we should use the script engine to evaluate directive parameters"
+      nodePropFeature     <- getNodePropEnabled() ?~! "Could not get if we should allow ${node.properties[key]} in directive parameters"
       fetch6Time          =  System.currentTimeMillis
       _                   =  logger.trace(s"Fetched run infos in ${fetch4Time-fetch3Time}ms")
 
@@ -140,7 +141,7 @@ trait DeploymentService extends Loggable {
       _             =  logger.debug(s"Historization of names done in ${timeHistorize}ms, start to build rule values.")
 
       ruleValTime   =  System.currentTimeMillis
-      ruleVals      <- buildRuleVals(allRules, directiveLib, groupLib, allNodeInfos) ?~! "Cannot build Rule vals"
+      ruleVals      <- buildRuleVals(allRules, directiveLib, groupLib, allNodeInfos, nodePropFeature) ?~! "Cannot build Rule vals"
       timeRuleVal   =  (System.currentTimeMillis - ruleValTime)
       _             =  logger.debug(s"RuleVals built in ${timeRuleVal}ms, start to expand their values.")
 
@@ -161,6 +162,7 @@ trait DeploymentService extends Loggable {
                            , globalRunInterval
                            , globalComplianceMode
                            , scriptEngineEnabled
+                           , nodePropFeature
                          ) ?~! "Cannot build target configuration node"
       timeBuildConfig =  (System.currentTimeMillis - buildConfigTime)
       _               =  logger.debug(s"Node's target configuration built in ${timeBuildConfig}, start to update rule values.")
@@ -237,6 +239,7 @@ trait DeploymentService extends Loggable {
   def getAgentRunStartHour   : () => Box[Int]
   def getAgentRunStartMinute : () => Box[Int]
   def getScriptEngineEnabled : () => Box[FeatureSwitch]
+  def getNodePropEnabled     : () => Box[FeatureSwitch]
 
   /**
    * Find all modified rules.
@@ -259,7 +262,7 @@ trait DeploymentService extends Loggable {
    * rules.
    * These objects are a cache of all rules
    */
-  def buildRuleVals(rules: Seq[Rule], directiveLib: FullActiveTechniqueCategory, groupLib: FullNodeGroupCategory, allNodeInfos: Map[NodeId, NodeInfo]) : Box[Seq[RuleVal]]
+  def buildRuleVals(rules: Seq[Rule], directiveLib: FullActiveTechniqueCategory, groupLib: FullNodeGroupCategory, allNodeInfos: Map[NodeId, NodeInfo], nodePropFeature: FeatureSwitch) : Box[Seq[RuleVal]]
 
   /**
    * Compute all the global system variable
@@ -279,6 +282,7 @@ trait DeploymentService extends Loggable {
     , globalAgentRun       : AgentRunInterval
     , globalComplianceMode : ComplianceMode
     , scriptEngineEnabled  : FeatureSwitch
+    , nodePropFeature      : FeatureSwitch
   ) : Box[(Seq[NodeConfiguration])]
 
   /**
@@ -393,6 +397,7 @@ class DeploymentServiceImpl (
   , override val getAgentRunStartHour: () => Box[Int]
   , override val getAgentRunStartMinute: () => Box[Int]
   , override val getScriptEngineEnabled: () => Box[FeatureSwitch]
+  , override val getNodePropEnabled    : () => Box[FeatureSwitch]
 ) extends DeploymentService with
   DeploymentService_findDependantRules_bruteForce with
   DeploymentService_buildRuleVals with
@@ -450,14 +455,14 @@ trait DeploymentService_buildRuleVals extends DeploymentService {
    * rules.
    * These objects are a cache of all rules
    */
-  override def buildRuleVals(rules:Seq[Rule], directiveLib: FullActiveTechniqueCategory, groupLib: FullNodeGroupCategory, allNodeInfos: Map[NodeId, NodeInfo]) : Box[Seq[RuleVal]] = {
+  override def buildRuleVals(rules:Seq[Rule], directiveLib: FullActiveTechniqueCategory, groupLib: FullNodeGroupCategory, allNodeInfos: Map[NodeId, NodeInfo], nodePropFeature: FeatureSwitch) : Box[Seq[RuleVal]] = {
     val appliedRules = rules.filter(r => ruleApplicationStatusService.isApplied(r, groupLib, directiveLib, allNodeInfos) match {
       case _:AppliedStatus => true
       case _ => false
     })
 
     for {
-      rawRuleVals <- sequence(appliedRules) { rule => ruleValService.buildRuleVal(rule, directiveLib) } ?~! "Could not find configuration vals"
+      rawRuleVals <- bestEffort(appliedRules) { rule => ruleValService.buildRuleVal(rule, directiveLib, nodePropFeature) } ?~! "Could not find configuration vals"
     } yield rawRuleVals
   }
 
@@ -501,10 +506,10 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
    *   - to node info parameters: ok
    *   - to parameters : hello loops!
    */
-  private[this] def buildParams(parameters: Seq[GlobalParameter]): Box[Map[ParameterName, InterpolationContext => Box[String]]] = {
-    sequence(parameters) { param =>
+  private[this] def buildParams(parameters: Seq[GlobalParameter], nodePropFeature: FeatureSwitch): Box[Map[ParameterName, InterpolationContext => Box[String]]] = {
+    bestEffort(parameters) { param =>
       for {
-        p <- interpolatedValueCompiler.compile(param.value) ?~! s"Error when looking for interpolation variable in global parameter '${param.name}'"
+        p <- interpolatedValueCompiler.compile(param.value, nodePropFeature) ?~! s"Error when looking for interpolation variable in global parameter '${param.name}'"
       } yield {
         (param.name, p)
       }
@@ -578,10 +583,11 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
     , globalAgentRun       : AgentRunInterval
     , globalComplianceMode : ComplianceMode
     , scriptEngineEnabled  : FeatureSwitch
+    , nodePropFeature      : FeatureSwitch
   ) : Box[Seq[NodeConfiguration]] = {
 
 
-    val interpolatedParameters = buildParams(parameters) match {
+    val interpolatedParameters = buildParams(parameters, nodePropFeature) match {
       case Full(x) => x
       case eb: EmptyBox => return eb ?~! "Can not parsed global parameter (looking for interpolated variables)"
     }
@@ -667,12 +673,32 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
       )
     }
 
-    //1.3: build node config, binding ${rudder.parameters} parameters
+
+    /*
+     * Utility class that helps deduplicate same failures in a chain
+     * of failure when using bestEffort.
+     */
+    implicit class DedupFailure[T](box: Box[T]) {
+      def dedupFailures(failure: String, transform: String => String = identity) = {
+        box match { //dedup error messages
+            case Full(res)   => Full(res)
+            case eb:EmptyBox =>
+              val msg = eb match {
+                case Empty      => ""
+                case f: Failure => //here, dedup
+                  ": " + f.failureChain.map(m => transform(m.msg).trim).toSet.mkString("; ")
+              }
+              Failure(failure + msg)
+        }
+      }
+    }
+
+    //1.3: build node config, binding ${rudder./node.properties} parameters
     // open a scope for the JsEngine, because its init is long.
     JsEngineProvider.withNewEngine(scriptEngineEnabled) { jsEngine =>
       for {
-        nodeConfigs <- sequence(interpolationContexts.toSeq) { case (nodeId, context) =>
-          for {
+        nodeConfigs <- bestEffort(interpolationContexts.toSeq) { case (nodeId, context) =>
+          (for {
             drafts          <- Box(policyDraftByNode.get(nodeId)) ?~! "Promise generation algorithme error: cannot find back the configuration information for a node"
             /*
              * Clearly, here, we are evaluating parameters, and we are not using that just after in the
@@ -684,19 +710,19 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
              *  - by node
              *  + use them in variable expansion (the variable expansion should have a fully evaluated InterpolationContext)
              */
-            parameters      <- sequence(context.parameters.toSeq) { case (name, param) =>
+            parameters      <- bestEffort(context.parameters.toSeq) { case (name, param) =>
                                  for {
                                    p <- param(context)
                                  } yield {
                                    (name, p)
                                  }
                                }
-            cf3PolicyDrafts <- sequence(drafts) { draft =>
-                                  for {
+            cf3PolicyDrafts <- bestEffort(drafts) { draft =>
+                                  (for {
                                     //bind variables with interpolated context
                                     expandedVariables <- draft.variableMap(context)
                                     // And now, for each variable, eval - if needed - the result
-                                    evaluatedVars     <- sequence(expandedVariables.toSeq) { case (k, v) =>
+                                    evaluatedVars     <- bestEffort(expandedVariables.toSeq) { case (k, v) =>
                                                             //js lib is specific to the node os, bind here to not leak eval between vars
                                                             val jsLib = context.nodeInfo.osDetails.os match {
                                                              case AixOS => JsRudderLibBinding.Aix
@@ -717,7 +743,7 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
                                       , ruleOrder = draft.ruleOrder
                                       , directiveOrder = draft.directiveOrder
                                     )
-                                  }
+                                  }).dedupFailures(s"When processing directive '${draft.directiveOrder.value}'")
                                 }
           } yield {
             NodeConfiguration(
@@ -727,7 +753,10 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
               , parameters = parameters.map { case (k,v) => ParameterForConfiguration(k, v) }.toSet
               , isRootServer = context.nodeInfo.id == context.policyServerInfo.id
             )
-          }
+          }).dedupFailures(
+                s"Error with parameters expansion for node '${context.nodeInfo.hostname}' (${context.nodeInfo.id.value})"
+              , _.replaceAll("on node .*", "")
+            )
         }
       } yield {
         nodeConfigs
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
@@ -45,6 +45,9 @@ import com.normation.utils.Control._
 import net.liftweb.common.{Failure => FailedBox, _}
 import com.normation.rudder.domain.parameters.ParameterName
 import com.normation.inventory.domain.NodeInventory
+import com.normation.rudder.domain.nodes.NodeProperty
+import com.normation.rudder.domain.appconfig.FeatureSwitch
+import net.liftweb.json.JsonAST.JValue
 
 /**
  * A parser that handle parameterized value of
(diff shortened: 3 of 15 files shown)


ncharles:
This looks great - i'll add some questions/remarks on some specific points

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -242,6 +255,50 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+   * Get the node property value, or fails if it does not exists.
+   * If the path length is 1, only check that the property exists and
+   * returned the corresponding string value.
+   * If the path length is more than one, try to parse the string has a
+   * json value and access the remaining part as a json path.
+   */
+  def getNodeProperty(context: InterpolationContext, path: List[String]): Box[String] = {
+    val errmsg = s"Missing property '$${node.properties[${path.mkString("][")}]}' on node '${context.nodeInfo.hostname}' [${context.nodeInfo.id.value}]"
+    path match {
+      //we shoul not reach that case since we enforce at leat one match of [...] in the parser
```
ncharles: s/shoul/should

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -242,6 +255,50 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+      case Nil       => FailedBox(s"The syntaxe $${node.properties} is invalid, only $${node.properties[propertyname]} is accepted")
```
ncharles: "Syntax $${node.properties} is invalid"

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,31 +310,29 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
-  def emptyVar: Parser[List[CharSeq]] = """(?iums)(\s)*""".r ^^ { x => List(CharSeq(x)) }
+  def emptyVar   : Parser[List[CharSeq]] = """(?iums)(\s)*""".r ^^ { x => List(CharSeq(x)) }
 
-  def plainString: Parser[CharSeq] = ("""(?iums)((?!\Q${rudder.\E).)+""").r  ^^ { CharSeq(_) }
+  // plain string must not match our identifier, ${rudder.* and ${node.properties.*}
+  def plainString: Parser[CharSeq] = ("""(?iums)((?!(\Q${rudder.\E|\Q${node.properties\E)).)+""").r  ^^ { CharSeq(_) }
 
   //identifier for step in the path or param names
-  def propId: Parser[String] = """[\-_a-zA-Z0-9]+""".r
+  def propId     : Parser[String] = """[\-_a-zA-Z0-9]+""".r
```
ncharles: wouldn't you need some modifier ? i imagine that there may be multiline keys, and maybe be gready ? so (?m)

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -624,10 +624,28 @@ case class RestExtractorService (
...
 
-    Box(json.extractOpt[RestNode]) ?~! "Error when extracting node information"
+    for {
+      props <- json \ "properties" match {
+        case JArray(props) => Full(props)
+        case x             => Failure(s"""Error: the given parameter is not a JSON object with a 'properties' key""")
+      }
+      seq   <- sequence(props) { p =>
+                 p match {
+                   case JObject(JField("name", JString(nameValue)):: JField("value", valueValue) :: Nil) =>
```
ncharles: can you unset values with this method ? i'm not sure i understand how it would work

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: Greedy is perhaps needed, thanks. For the multiline, I think we must forbid it in an identifier itself. It must be authorized between token (and I will add a test to be sure of that), but in one, it seems strange. Have you an use case in mind ?

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
fanf: You're right, it would impose to use a value "", and I'm not sure it will work. I'm updating that.

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: So, greedy is not what we want, because propId are always (must be) separated by special chars, either "." or "][". 
For the multiline part, it is explicitly forbiden in the middle of ${}. I'm not remembering, nor finding, reason why it is so. I'm not sure we should forbid it around special char delimiters, i.e:
```
${node.properties[mypar
ameter]} //forbiden

${node.properties[myparameter]
[withsubkey]} //ok
```
Ideas ?

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
fanf: So no, it's ok, it works as advertised in http://www.rudder-project.org/rudder-api-doc/#api-Nodes-updateNodeProperties
You can remove properties by setting their value to "", which allow to only POST the subset of properties to update in the general case (i.e, it you have 1000 properties and just have one to update, you only give that one in the POST, and the other are not removed but let untouched).

fanf:
PR rebased

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
ncharles: Great, thank you for the confirmation.
One remark: it is so impossible to have key without values (like a tag). It's by design then?

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
ncharles: Internet seems to say that multiline keys are invalid in json : http://stackoverflow.com/questions/2392766/multiline-strings-in-json , so indeed there is no need for multiline for string, but your proposition around the [ ] will help readability
I have a question for the non greedy, i'm not sure of the behaviour of the parser, but if it is not greedy, isn't there a risk to match partial string ? (it seems not, but a test with keys data and datacenter would ensure that future refactoring don't break it)

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
fanf: Yes, it was restricted to key = value.

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -128,6 +128,7 @@ trait DeploymentService extends Loggable {
       agentRunStartMinute <- getAgentRunStartMinute() ?~! "Could not get agent run start time (minute)"
       agentRunStartHour   <- getAgentRunStartHour() ?~! "Could not get agent run start time (hour)"
       scriptEngineEnabled <- getScriptEngineEnabled() ?~! "Could not get if we should use the script engine to evaluate directive parameters"
+      nodePropFeature     <- getNodePropEnabled() ?~! "Could not get if we should allow ${node.properties[key]} in direcitve parameters"
```
ncharles: typo: directive

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -59,19 +61,26 @@ import com.normation.inventory.domain.NodeInventory
...
+ *
  * 2/ ${rudder.node.ACCESSOR}
  *    where:
  *    - "node" is a keyword ;
  *    - ACCESSOR is an accessor for that node, explained below.
  *    - the value can not contains other interpolation
  *
+ * 3/ ${node.properties[keyone][keytwo]
+ *     where:
+ *       - keyone, keytwo are path on json values
```
ncharles: wrong indentation here

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -242,6 +281,50 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+   * returned the corresponding string value.
+   * If the path length is more than one, try to parse the string has a
+   * json value and access the remaining part as a json path.
+   */
+  def getNodeProperty(context: InterpolationContext, path: List[String]): Box[String] = {
+    val errmsg = s"Missing property '$${node.properties[${path.mkString("][")}]}' on node '${context.nodeInfo.hostname}' [${context.nodeInfo.id.value}]"
+    path match {
+      //we should not reach that case since we enforce at leat one match of [...] in the parser
+      case Nil       => FailedBox(s"The syntax $${node.properties} is invalid, only $${node.properties[propertyname]} is accepted")
+      case h :: tail => context.nodeInfo.properties.find(p => p.name == h) match {
```
ncharles: just to be sure: this is supposed to be case-sensitive, right ?

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,31 +336,56 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
   //identifier for step in the path or param names
-  def propId: Parser[String] = """[\-_a-zA-Z0-9]+""".r
+  def propId      : Parser[String] = """[\-_a-zA-Z0-9]+""".r
+
+  // all interpolation
+  def interpol    : Parser[Interpolation] = rudderProp | nodeProp
 
   //an interpolated variable looks like: ${rudder.XXX}, or ${RuDder.xXx}
-  def interpol: Parser[Interpolation] = "${" ~> id("rudder") ~> "." ~> (nodeProp | parameter ) <~ "}"
+  def rudderProp  : Parser[Interpolation] = "${" ~> space ~> id("rudder") ~> space ~> id(".") ~> space ~>(nodeAccess | parameter ) <~ space <~ "}"
```
ncharles: As we now accept spaces in the variable, comments should be updated accordingly

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,31 +336,56 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+  def interpolOpt : Parser[DefaultValue] = interpol ^^ { x => DefaultValue(x::Nil) }
+  def emptySQuote : Parser[DefaultValue] = id("\"\"")         ^^   { _ => DefaultValue(CharSeq("")::Nil) }
+  def emptyTQuote : Parser[DefaultValue] = id("\"\"\"\"\"\"") ^^   { _ => DefaultValue(CharSeq("")::Nil) }
 
+  //string must be simple or triple quoted string
+  def sqString    : Parser[DefaultValue] = id("\"")  ~> rep1(sqplainStr | interpol )<~ id("\"") ^^ { case x => DefaultValue(x) }
+  def sqplainStr  : Parser[Token]        = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|")).)+""").r  ^^ { str => CharSeq(str) }
 
+  def tqString    : Parser[DefaultValue]   = id("\"\"\"")  ~> rep1(tqplainStr | interpol )<~ id("\"\"\"") ^^ { case x => DefaultValue(x) }
+  def tqplainStr  : Parser[Token]          = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|"""+"\"\"\""+ """)).)+""").r  ^^ { str => CharSeq(str) }
```
ncharles: this is clearly too complex for me :(

rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndParameterLookup.scala
```diff
@@ -501,6 +618,30 @@ class TestNodeAndParameterLookup extends Specification {
...
+      compare("""${node.properties[missing][key]|default= ${node.properties[datacenter][Europe]|default="some default"}}"""
+           , ("datacenter", json) :: Nil) must beEqualTo("""{"France":"Paris"}""")
+    }
+    "correctly return the default value, two level deep, if all keys are missing" in {
+      compare("""${node.properties[missing][key]|default= ${node.properties[missing_datacenter][Europe]|default="some default"}}"""
+           , ("datacenter", json) :: Nil) must beEqualTo("""some default""")
+    }
+    "correctly return the default value which is ${node.properties...}, if default has the 'node' optiopn" in {
+      compare("""${node.properties[missing][key]|default= ${node.properties[datacenter][Europe]|node}}"""
+           , ("datacenter", json) :: Nil) must beEqualTo("""${node.properties[datacenter][Europe]}""")
```
ncharles: Could you also add some tests with feature disabled, to check that it is working as expected

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -59,19 +61,26 @@ import com.normation.inventory.domain.NodeInventory
...
  *    - "node" is a keyword ;
  *    - ACCESSOR is an accessor for that node, explained below.
  *    - the value can not contains other interpolation
  *
+ * 3/ ${node.properties[keyone][keytwo]
+ *     where:
+ *       - keyone, keytwo are path on json values
+ *      - the return value is the string representation, in compact mode, of the resulting access
+ *      - if the key is not found, we raise an error.
+ *
```
ncharles: Comments should be updated to state that spaces are authorized in parameters, so that it won't be surprising

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: done

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: done

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: Sorry, I just completly mis transcript / understood. 
All reperition operators are greedy. We don't want there lazy counterparts, for eg ".?", "+?" etc because, yes, in that case we will get partial string. We don't risk getting too much string with greedy operators because we have clear delimiters out of greedy scope. 
"?m" is for multiline. We don't want line break in the middle of word, so we don't want that either (but "space" is matching any whitespace characters, including line breaks, so we get them where space are put (i.e around chars used for separators, like ".", "]", etc))

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: done

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
fanf: done

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: yes (supposed to be, and should be given the implementation)

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: Simplified by factoring out the main regex in a common "plainStringRegex". It helps understanding that most of the regex is to avoid "${rudder." and "${node.properties", and that optionnally we are adding " or """ in the avoided items.

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: done

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: Documentation completed also with the default value & "node" options.

rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndParameterLookup.scala
fanf: Test added

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndParameterLookup.scala
ncharles: Thank you so much !

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -59,19 +62,41 @@ import com.normation.inventory.domain.NodeInventory
...
+ *      - "node" : mean that the interpolation will be done on the node,
+ *        and so the parameter must be outputed as an equivalent string
+ *        parameter for the node (the same without "| node")
+ *      - default = XXX mean that if the properties is not found, "XXX" must
+ *        be used in place, with XXX one of:
+ *        - a double-quoted string, like: "default value",
+ *        - a triple-double quoted string, like: """ some "default" must be used""",
+ *        - an other parameter among the 3 described here.
+ *        Quoted string may contain parameters, like in:
+ *        ${node.properties[charge] | default = """Node "color" is: ${node.properties[color} | "blue" }""" }
```
ncharles: you are missing a ] before the } of color

rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
```diff
@@ -81,7 +83,12 @@ case object Node {
   }
 }
 
-case class NodeProperty(name: String, value: String)
+case class NodeProperty(name: String, value: JValue) {
+  def renderValue: String = value match {
```
VinceMacBuche: Why not : net.liftweb.json.compactRender(value) ?

fanf: Because it output surnemerary "" around the string if we don't special case it

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,28 +350,66 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
-  def parameter: Parser[Interpolation] = {
-    id("param") ~> "." ~> propId ^^ { p => Param(p) }
+  def parameter   : Parser[Interpolation] = { id("param") ~> space ~> id(".") ~> space ~> propId ^^ { p => Param(p) } }
+
+  //an interpolated variable looks like: ${rudder.XXX}, or ${RuDder.xXx}
+  def nodeProp    : Parser[Interpolation] = "${" ~> space ~> id("node") ~> space ~> id(".") ~> space ~> id("properties") ~> rep1( space ~> "[" ~> space ~>
+                                            propId <~ space <~ "]" ) ~ opt(propOption) <~ space <~ "}" ^^ { case ~(path, opt) => Property(path, opt) }
+
+  //here, the number of " must be strictly decreasing - ie. triple quote before
+  def propOption  : Parser[PropertyOption] = space ~> "|" ~> space ~>  ( onNodeOpt |  "default" ~> space ~> "=" ~> space ~>
```
VinceMacBuche: Maybe having a 'defaultOpt : Parser[DefaultValue]' so it's easier to add a new option ?

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,28 +350,66 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+  def interpolOpt : Parser[DefaultValue] = interpol ^^ { x => DefaultValue(x::Nil) }
+  def emptySQuote : Parser[DefaultValue] = id("\"\"")         ^^   { _ => DefaultValue(CharSeq("")::Nil) }
+  def emptyTQuote : Parser[DefaultValue] = id("\"\"\"\"\"\"") ^^   { _ => DefaultValue(CharSeq("")::Nil) }
+
+  //string must be simple or triple quoted string
+  def sqString    : Parser[DefaultValue] = id("\"")  ~> rep1(sqplainStr | interpol )<~ id("\"") ^^ { case x => DefaultValue(x) }
+  def sqplainStr  : Parser[Token]        = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|")).)+""").r  ^^ { str => CharSeq(str) }
+
+  def tqString    : Parser[DefaultValue]   = id("\"\"\"")  ~> rep1(tqplainStr | interpol )<~ id("\"\"\"") ^^ { case x => DefaultValue(x) }
+  def tqplainStr  : Parser[Token]          = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|"""+"\"\"\""+ """)).)+""").r  ^^ { str => CharSeq(str) }
```
VinceMacBuche: Maybe you could use planStringRegex function here 

planStringRegex(Some("\"\"\""))

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,28 +350,66 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+
+  def onNodeOpt   : Parser[InterpreteOnNode.type] = id("node") ^^^ InterpreteOnNode
+
+  def interpolOpt : Parser[DefaultValue] = interpol ^^ { x => DefaultValue(x::Nil) }
+  def emptySQuote : Parser[DefaultValue] = id("\"\"")         ^^   { _ => DefaultValue(CharSeq("")::Nil) }
+  def emptyTQuote : Parser[DefaultValue] = id("\"\"\"\"\"\"") ^^   { _ => DefaultValue(CharSeq("")::Nil) }
+
+  //string must be simple or triple quoted string
+  def sqString    : Parser[DefaultValue] = id("\"")  ~> rep1(sqplainStr | interpol )<~ id("\"") ^^ { case x => DefaultValue(x) }
+  def sqplainStr  : Parser[Token]        = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|")).)+""").r  ^^ { str => CharSeq(str) }
```
VinceMacBuche: same as below, why not use: planStringRegex function



planStringRegex(Some("\""))

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -253,28 +350,66 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+
+  def tqString    : Parser[DefaultValue]   = id("\"\"\"")  ~> rep1(tqplainStr | interpol )<~ id("\"\"\"") ^^ { case x => DefaultValue(x) }
+  def tqplainStr  : Parser[Token]          = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|"""+"\"\"\""+ """)).)+""").r  ^^ { str => CharSeq(str) }
+
+  // plain string must not match our identifier, ${rudder.* and ${node.properties.*}
+  // here we defined a function to build them, with the possibility to add one other string to avoid
+  def planStringRegex(optAvoid: Option[String]) = {
```
VinceMacBuche: And rename function to plainStringRegex

rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndParameterLookup.scala
```diff
@@ -104,8 +133,9 @@ class TestNodeAndParameterLookup extends Specification {
 
   val dangerVariable = InputVariableSpec("danger", "").toVariable(Seq("${rudder.param.danger}"))
 
-  val multilineParamVariable = InputVariableSpec("multiParam", "").toVariable(Seq("== ${rudder.\rparam.\nfoo} =="))
+  val invalidMultilineParamVariable = InputVariableSpec("multiParam", "").toVariable(Seq("== ${rudder.\rparam.\nfoo} =="))
```
VinceMacBuche: Why is it 'invalid' ?

rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndParameterLookup.scala
```diff
@@ -79,10 +105,11 @@ class TestNodeAndParameterLookup extends Specification {
   )
 
   def lookup(
-      variables: Seq[Variable]
-    , context: InterpolationContext
+      variables    : Seq[Variable]
+    , context      : InterpolationContext
+    , nodePropSwith: FeatureSwitch = FeatureSwitch.Enabled
```
VinceMacBuche: nodePropsSwitCh

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -667,12 +673,32 @@ trait DeploymentService_buildNodeConfigurations extends DeploymentService with L
       )
     }
 
-    //1.3: build node config, binding ${rudder.parameters} parameters
+
+    /*
+     * Utility class that helps deduplicate same failures in a chain
+     * of failure when using bestEffort.
+     */
+    implicit class DedupFailure[T](box: Box[T]) {
```
fanf: Yes

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: Lexing is such a mess that I can't really predict what it will look like. I prefer to keep it focused for now.

rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndParameterLookup.scala
fanf: In fact, it isn't used anymore, because it isn't anymore :)

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: Well. It was the intention. But for some reason, it doesn't work anymore when I'm using the mothod. And I don't know at all it's the case. 
So for now, I'm keeping the verbose, but working, solution :(

VinceMacBuche:
OK, merging this PR

PR #1208, opened 2016-09-27
Fixes #9141: When there is missing and received reports, we don't see the received one

fanf:
https://www.rudder-project.org/redmine/issues/9141

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -951,7 +951,7 @@ object ExecutionBatch extends Loggable {
     }
 
     // Regroup all None value into None component Value
-    // There should only one None report per component and
+    // There should only one None report per component
     val (noneValue, noneReports) = {
       if(valueKind.none.isEmpty) {
       (None, Seq())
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -1251,14 +1251,13 @@ object ExecutionBatch extends Loggable {
               /* Nothing was received at all for that component so : No Answer or Pending */
               case 0 if noUnexpectedReports =>
                 MessageStatusReport(noAnswerType, None) :: Nil
-              /* Reports were received for that component, but not for that key, that's a missing report */
-              case x if(x < cardinality) =>
-                (0 until cardinality).map( _ => MessageStatusReport(MissingReportType, None)).toList
+              case x if(x <= cardinality) =>
+                filteredReports.map { r => MessageStatusReport(ReportType(r), r.message) }.toList ++
+                /* We need to complete the list of correct with missing, if some are missing */
+                (0 until (cardinality-x)).map( i => MessageStatusReport(MissingReportType, s"[Missing report #${i+x}]")).toList
```
VinceMacBuche: x until cardinality would be clearer don't you think ?

fanf:
PR rebased

PR #1209, opened 2016-09-27
Fixes #9133: 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9133

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestQuicksearch.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/RestQuicksearch.scala
@@ -14,7 +14,10 @@ import net.liftweb.json.JsonAST._
 import net.liftweb.json.JsonDSL._
 import com.normation.rudder.services.quicksearch.FullQuickSearchService
 import com.normation.rudder.services.quicksearch.QuickSearchResult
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/RestQuicksearch.scala
```diff
@@ -52,7 +55,18 @@ class RestQuicksearch (
           toJsonError(None, e.messageChain)("quicksearch", false)
 
         case Full(results) =>
-          toJsonResponse(None, prepare(results, MAX_RES_BY_KIND))("quicksearch", false)
+          val filteredResults = {
+            import QuickSearchResultId._
+            results.filter { _.id match {
+              case QRNodeId(_)      => CurrentUser.checkRights(Read("node"))
+              case QRGroupId(_)     => CurrentUser.checkRights(Read("group"))
```
fanf: you could test with: _:QRNodeID etc. "isInstanceOf" is amazingly fast on the jvm. 
On the same idea, it would be quicker to have val nodeOK = CurrentUser.checkRights(Read("node")) etc; and to put everything in a dedicated method "filterResultByRights(results, user)

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1225, opened 2016-09-30
Fixes #9197: Quicksearch does not redirect to a rule when you are on the rule page. 

fanf:
https://www.rudder-project.org/redmine/issues/9197


fanf:
PR rebased

PR #1226, opened 2016-10-01
Fixes #8985: Add node key in managed_nodes system variable

fanf:
https://www.rudder-project.org/redmine/issues/8985

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
@@ -37,18 +37,23 @@
 
 package com.normation.rudder.domain.nodes
 
-import com.normation.inventory.domain.AgentType
-import org.joda.time.DateTime
-import com.normation.inventory.domain.NodeId
+import java.io.StringReader
+import java.security.KeyFactory
+import java.security.MessageDigest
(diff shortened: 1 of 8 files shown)


fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
```diff
@@ -93,4 +98,100 @@ final case class NodeInfo(
...
+   * ASN.1 that much.
+   *
+   * Oh, and Java Security API force a lot of casting.
+   * This is insane.
+   */
+  def getHash(key: PublicKey): Box[String] = {
+
+    //the parser able to read PEM files
+    val parser = new PEMParser(new StringReader(key.value))
+
```
ncharles: There are a lot of asInstanceOf - shouldn't you wrap it around a try to catch casting exception ?

fanf: They are already with tryo - which one did I missed ?

ncharles: huh - i didn't see any of them ... don't know why
It does look ok then

PR #1227, opened 2016-10-02
Fixes #9201: Add basic infrastructure to test REST API

fanf:
https://www.rudder-project.org/redmine/issues/9201

### DIFF
--- a/rudder-web/src/test/resources/api/api_v2.yml
+++ b/rudder-web/src/test/resources/api/api_v2.yml
@@ -1,0 +1,16 @@
+---
+description: Simple test of api/status - just "ok"
+method: GET
+url: /api/status
+response: 
+  type: text #optionnal, by default json
+  code: 200
+  content: 'OK'
+# example of a bad spec, because type "plop" is not known
+# ---
+# method: GET
+# url: /api/techniqueLibrary/reload
+# response:
+#  type: plop
+#  code: 200
+#  content: 'OK'
--- a/rudder-web/src/test/scala/com/normation/rudder/web/rest/RestTestSetUp.scala
+++ b/rudder-web/src/test/scala/com/normation/rudder/web/rest/RestTestSetUp.scala
@@ -1,0 +1,136 @@
+/*
+*************************************************************************************
+* Copyright 2016 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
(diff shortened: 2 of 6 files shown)


fanf:
PR rebased

rudder-web/src/test/scala/com/normation/rudder/web/rest/RestTestSetUp.scala
```diff
@@ -0,0 +1,136 @@
...
+      }) match {
+        case Full(x)      => tests(x)
+        case eb: EmptyBox => tests(eb)
+      }
+    }
+  }
+
+
+  def GET(path: String) = {
+    val mockReq = new MockHttpServletRequest("http://localhost:8080")
```
fanf: AFAIK, the server/port is only used to build the fake Request/Response, not to actually send anything on that URL. It could be anything, I believe.

rudder-web/src/test/resources/api/api_v2.yml
```diff
@@ -0,0 +1,15 @@
...
+url: /api/status
+response: 
+  type: text #optionnal, by default json
+  code: 200
+  content: 'OK'
+---
+method: GET
+url: /api/techniqueLibrary/reload
+response:
+  type: plop
```
fanf: Yes, I wanted to test parsing error. 
For now, the behaviour is to log:
```
[2016-11-15 11:33:32] ERROR com.normation.rudder.web.rest.TestRestFromFileDef - I wasn't able to run the 2th tests in file api/api_v1.yml <- The YAML document is empty
[2016-11-15 11:33:33] ERROR com.normation.rudder.web.rest.TestRestFromFileDef - I wasn't able to run the 1th tests in file api/api_v2.yml <- Unrecognized response type: 'plop'. Use 'json' or 'text'
```
And the test is not run. Would you find better to also fail the test? Effectively, it could be easier to correct all problems when we will change the format.

ncharles: yes, i'd rather enjoy a fail

fanf: Ok, it's better to have failing tests, but now I can't commit them because they will fails the test... not sur how to handle that.

fanf:
PR rebased

rudder-web/src/test/scala/com/normation/rudder/web/rest/RestTestSetUp.scala
ncharles: ha ok !

PR #1230, opened 2016-10-04
Fixes #9219: Searching on os type does not work

fanf:
https://www.rudder-project.org/redmine/issues/9219

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
@@ -286,6 +286,14 @@ object QSLdapBackend {
    * Mapping between attributes and their filter in the backend
    */
   object QSAttributeLdapFilter {
+    //
+    val NODE_POSTFIX = "Node"
+
+    //the list of value to avoid in OC for matching
+    val ocRudderTypes = Set("top", OC_RUDDER_NODE, OC_POLICY_SERVER_NODE, OC_NODE
+                            , OC_RULE, OC_PARAMETER, OC_RUDDER_NODE_GROUP)
+
+    // Find a substring in a LDAP attribute, i.e 'token' becomes "*token*"
     def sub(a: QSAttribute, token: String) = Some(SUB(a.ldapName, null, Array(token), null ))
 
     //use val to compile only one time these patterns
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
```diff
@@ -276,6 +276,12 @@ object QSLdapBackend {
    * Mapping between attributes and their filter in the backend
    */
   object QSAttributeLdapFilter {
+
+    //the list of value to avoid in OC for matching
+    val ocRudderTypes = Set("top", OC_RUDDER_NODE, OC_POLICY_SERVER_NODE, OC_NODE
+                            , OC_RULE, OC_PARAMETER, OC_RUDDER_NODE_GROUP)
```
ncharles: shouldn't you avoid also OC_SPECIAL_TARGET, OC_ACTIVE_TECHNIQUE_LIB_VERSION, OC_ABSTRACT_RULE_WITH_CF3POLICYDRAFT, OC_RULE_WITH_CF3POLICYDRAFT, OC_RULE_WITH_CF3POLICYDRAFT, OC_NODE_CONFIGURATION,OC_PROPERTY, or are they filtered differently ?

rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
```diff
@@ -404,25 +410,31 @@ object QSLdapBackend {
...
-          testers.find(f => f(token)).flatMap(findValue)
+          testers.find(f => f(token)).flatMap(findValue(a.getValues, _))
         }
 
         if(a.getName == A_OC) {
-          None
+          //here we must just avoid "top" and "kind" object class
+          //and also remove the "node" part for "linuxNode", etc
+          val values = a.getValues.flatMap(s =>
+            if(ocRudderTypes.contains(s)) None else Some(s.replace("Node", ""))
```
ncharles: i'm not really fond of the twice hardcoded "Node" - could you at least use a constant ?

rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
fanf: We only select object in OC_NODE, OC_RUDDER_NODE, OC_RULES, OC_PARAMETER, OC_RUDDER_GROUP (and their children), so we don't need to filter these one.

rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
fanf: why not

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
ncharles: ok, thanks for the clarification

PR #1235, opened 2016-10-05
Fixes #8669: Remove old node search and feature switch in Rudder 3.3

fanf:
https://www.rudder-project.org/redmine/issues/8669


rudder-core/src/main/resources/Migration/ldapMigration-3.1.x-3.1.14-3.2.7-disable-js-directive.ldif
```diff
@@ -1,39 +0,0 @@
-###############################################################################
```
VinceMacBuche: That migration should still be used !! (migratin from 3.1 to 4.0 should keep it disabled)

fanf:
PR rebased

PR #1238, opened 2016-10-07
Fixes #9252: System variable "MANAGED_NODES_KEY" is empty

ncharles:
https://www.rudder-project.org/redmine/issues/9252

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
@@ -157,13 +157,14 @@ object CFEngineKey {
    *
    * Oh, and Java Security API force a lot of casting.
    * This is insane.
+   *
+   * Caution: to use the complete key, with header and footer, you need to use key.key
    */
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
```diff
@@ -157,11 +157,13 @@ object CFEngineKey {
...
    * Oh, and Java Security API force a lot of casting.
    * This is insane.
+   *
+   * Caution: to use the complete key, with header and footer, you need to use key.key
    */
   def getHash(key: PublicKey): Box[String] = {
 
     //the parser able to read PEM files
-    val parser = new PEMParser(new StringReader(key.value))
+    val parser = new PEMParser(new StringReader(key.key))
```
fanf: Shouldn't we also add a test on null for parser with something like:
`
for {
   parser <- Box(Option(new PEMParser(new StringReader(key.key))))
...
`
 Or was the NPE elsewhere ?

ncharles:
Commit modified

PR #1243, opened 2016-10-07
Fixes #9263: Error on report batch: NULL for non option type & on eventlog

fanf:
https://www.rudder-project.org/redmine/issues/9263


ncharles:
awesome

PR #1245, opened 2016-10-07
WIP: only the new report type are added. No compliance mode done

fanf:
This is a WIP. 

Only new audit report type are added (compliant, nonCompliant, auditError, auditNotApplicable). 

Now, we need to add in the database a new column for expected reports for the policy mode, and most likelly a new column for the node for the same thing. 
Then, we need to adapt ExecutionBatch to take care of that in the merging of report. 
I'm almost sure that we want to a special top level message for the case where the node was expecting one report type and got an other.


fanf:
Bad ticket number.

PR #1248, opened 2016-10-08
Fixes #9273: Process audit reports and adapt compliance displaying for audit mode

fanf:
This can be merge, but it is know to have broken part:
- the current policy mode of (node, directives) is used to calculate the compliance (what is false each time you chagne policy mode for run before the change)
- it is not exactly clear how we report mixed audit/enforce, and the impact on "worst reports"
- the message and css is broken in the case mention above in compliance report. 

But it could be merge to allow tests.


rudder-core/src/main/scala/com/normation/rudder/domain/reports/ReportType.scala
```diff
@@ -49,66 +49,78 @@ sealed trait ReportType {
...
+      ( reportTypes :\ (EnforceNotApplicable : ReportType) ) {
+        case (_, Unexpected)          | (Unexpected, _)           => Unexpected
+        case (_, EnforceError)        | (EnforceError, _)         => EnforceError
+        case (_, EnforceRepaired)     | (EnforceRepaired, _)      => EnforceRepaired
+        case (_, Missing)             | (Missing, _)              => Missing
+        case (_, NoAnswer)            | (NoAnswer, _)             => NoAnswer
+        case (_, Disabled)            | (Disabled, _)             => Disabled
+        case (_, Pending)             | (Pending, _)              => Pending
+        case (_, EnforceSuccess)      | (EnforceSuccess, _)       => EnforceSuccess
+        case (_, EnforceNotApplicable)| (EnforceNotApplicable, _) => EnforceNotApplicable
```
ncharles: why isn't there audit reports there ?

fanf: I'm not sure we want to keep the worst method, which is used only in the status report for a node, when there is several identical values. I think in that case, we just want to display the habitual compliance bar OR one not aggregate the different lines if they are of several types. 

And even if we keep it, I'm not sur about the semantic of having Compliant AND Success. It looks like something we don't want at all, because it must be an agent error in some way (on one node, having a multi-value for ONE directive giving both enforce and audit result... bad). 

Well, so for now, I didn't touched it.

PR #1250, opened 2016-10-10
Fixes #9250: JS error on Rudder 4.0 that prevent compliance & software display

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9250


VinceMacBuche:
Commit modified

PR #1251, opened 2016-10-10
Fixes #9237: Web interface tries to load inexistant css angucomplete-alt.css fails

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9237


VinceMacBuche:
Commit modified

PR #1256, opened 2016-10-11
Fixes #9232: Add an API to list shared files properties

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9232

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/sharedFiles/SharedFilesAPI.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/sharedFiles/SharedFilesAPI.scala
@@ -1,0 +1,226 @@
+/*
+***************************o**********************************************************
+* Copyright 2016 Normation SwnloaAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
+* distributed under the license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* Rudder is distributed in the hope that it will be useful,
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

rudder-web/src/main/scala/com/normation/rudder/web/rest/sharedFiles/SharedFilesAPI.scala
```diff
@@ -0,0 +1,225 @@
...
+      if (Files.isRegularFile(path)) {
+        val data = Files.readAllBytes(path)
+        val headers =
+         ("Content-type" -> "application/octet-stream") ::
+         ("Content-length" -> data.length.toString) ::
+         ("Content-disposition" -> s"attachment; filename=${path.getFileName}") ::
+         Nil
+
+        Full(StreamingResponse(new ByteArrayInputStream(data),() => {}, data.length, headers, Nil, 200))
+      } else {
```
fanf: Here, you are reading the file into memory to then stream it to the network. You should side-step the memory part to avoid to consume huge quantity of RAM on ISO.
It's easy to get the file size without reading it with (new File(path)).length (perhaps there is an quivalent method on nio). 
And for the streaming part, a simple FileIntputStream should be ok:
val file = new File(path)
Full(StreamingResponse(new FileInputStream(file),() => {}, file.length, headers, Nil, 200))

VinceMacBuche:
Commit modified

rudder-web/src/main/scala/com/normation/rudder/web/rest/sharedFiles/SharedFilesAPI.scala
```diff
@@ -0,0 +1,225 @@
...
+          response
+        case eb : EmptyBox =>
+          val fail = eb ?~! s"An error occured while looking into directory"
+          logger.error(fail.messageChain)
+          errorResponse(fail.messageChain)
+      }
+
+    }
+  }
+  serve("secure" / "api" / s"sharedfile" prefix requestDispatch)
```
fanf: why the s"" for sharedFile ?

VinceMacBuche:
Commit modified

rudder-web/src/main/scala/com/normation/rudder/web/rest/sharedFiles/SharedFilesAPI.scala
```diff
@@ -0,0 +1,226 @@
...
+        Failure(s"File '${path}' is not a directory")
+      }
+    } else {
+      Failure(s"File '${path}' does not exist")
+    }
+  }
+
+  def fileContent(path : Path) : Box[LiftResponse] = {
+    if (Files.exists(path)) {
+      if (Files.isRegularFile(path)) {
```
fanf: I think we should check if the file is binary or text.

VinceMacBuche: That will be done in another PR

VinceMacBuche:
PR rebased

PR #1265, opened 2016-10-12
Fixes #9211: Add Directive field with a graphical File Explorer for shared files 

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9211


RaphaelGauthier:
PR updated

PR #1269, opened 2016-10-12
Fixes #9294: Quicksearch doesn't display results at all

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9294


fanf:
Oh thank you, it's so good to be able to use again for tests :)

PR #1271, opened 2016-10-12
Fixes #9346: Compilation error because RestSharedFolder not found

fanf:
https://www.rudder-project.org/redmine/issues/9346


fanf:
This is trivial and blocking the build, I'm merging it.

PR #1274, opened 2016-10-13
Fixes #9290: Add an API endpoint to call rudder-remote-run

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9290

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
@@ -57,6 +57,15 @@ import net.liftweb.json._
 import net.liftweb.json.JsonDSL._
 import com.normation.eventlog.EventActor
 import com.normation.rudder.domain.nodes.Node
+import java.io.InputStream
+import java.io.OutputStream
+import java.io.IOException
+import com.zaxxer.nuprocess.NuAbstractProcessHandler
+import java.nio.ByteBuffer
+import java.io.PipedInputStream
+import java.io.PipedOutputStream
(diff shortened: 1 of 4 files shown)


VinceMacBuche:
Commit modified

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -81,4 +90,66 @@ class NodeApiService8 (
...
+      while (read != -1) {
+        Arrays.fill(bytes,zero)
+        read = in.read(bytes)
+        out.write(bytes)
+        out.flush()
+      }
+
+    } catch {
+      case e : IOException =>
+        // should we log ? Should we end ?
```
fanf: You can't do that, or you will leak ressource. The best way to handle that is to let the exception throws here, and manages it above. As your method is only used locally, it's ok. So remove try/catch here.

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -81,4 +90,66 @@ class NodeApiService8 (
...
+        case Nil => Nil
+        case classes => "-D" :: classes.mkString(",") :: Nil
+      }
+      import scala.collection.JavaConversions._
+      val pb = new NuProcessBuilder("/usr/bin/rudder" :: action ::: classOption)
+      val handler = new RunHandler(out)
+      pb.setProcessListener(handler)
+      pb.start();
+
+      runResponse(in)
```
fanf: Here, add a try catch, and on exception, close all the open resources: in, out

VinceMacBuche:
Commit modified

PR #1277, opened 2016-10-15
Fixes #9315: Update DB schema of expected reports to store policy mode

fanf:
https://www.rudder-project.org/redmine/issues/9315

Work in progress. 

What is done: 
- addition of a new table, "nodeconfigurations" [ nodeId, nodeConfigId, begindate, enddate, json] with json containing two part: "node modes", i.e compliance mode, run agent definition, and politcy mode ; and "rule expected reports" (down to the values)
- the modes in the tables mean that we are going to have CORRECT information to calcul compliance based on the read config data at the moment the node knows them. 
- ExpectedReportsUpdateImpl is *much* simplified thanks to the new table, and the update process works
- update time is *much* improved (for ~100 nodes, from ~15s => 900ms)
- update the compliance calculus to use mode information from the table
- update ExecutionBatch#computeNodesRunInfo to use an Option[NodeExpectedReport], and return it for run types where it is needed; 
- use the NodeExpectedReport from the run type in ExecutionBatch#getNodeStatusReports in place of all the filttering and loops on rulenode (minimum so that it works)

What does not work:
- tests related to DB (i.e, it should not block a build if merged right now)

What remain to be done:
- all the tests :(
- clean up everything (in particular, the old expected reports tables and related repository methods)

And then, profit !


fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -321,224 +308,227 @@ object ExecutionBatch extends Loggable {
...
+        // Then analyse the consistancy of the result.
+        //
+        case Some(runInfos) =>
+//          ComplianceDebugLogger.node(nodeId).debug(s"Node run configuration: ${(nodeId, complianceMode, runInfos).toLog }")
+//
+//          val computed =    computeNodeRunInfo(
+//                    nodeId, optInfo, missingReportType
+//                  , intervalInfo, updateValidityTime(intervalInfo), runValidityTime(intervalInfo)
+//                  , now, run
+//              )
```
ncharles: extra comments

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -321,224 +308,227 @@ object ExecutionBatch extends Loggable {
...
-    }.toMap
-  }
+            //
+            // #2 : run without config ID (neither in it nor found)
+            //      no expected config for the run. Why so? At least, a recent config.
+            case ((AgentRunWithNodeConfig(AgentRunId(_, t), None, _, _)), Some(currentConfig)) =>
+              /*
+               * Here, we want to check two things:
+               * - does the run should have contain a config id ?
+               *   It should if the oldest config was created a long time ago,
```
ncharles: why do we search for the oldest ? I can't understand the logic there :(

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
```diff
@@ -68,20 +94,34 @@ case class RoReportsExecutionRepositoryImpl (
...
-            Full(nodeIds.map(n => (n, runs.get(n))).toMap)
-          case -\/(ex) => Failure(errorMSg + ": " + ex.getMessage, Full(ex), Empty)
-        }
+        (for {
+          runs <- sql"""select distinct on (r.nodeid)
+                         r.nodeid, r.date, r.nodeconfigid, r.complete, r.insertionid
+                       , c.nodeid, c.nodeconfigid, c.begindate, c.enddate, c.configuration
+                       from  reportsexecution r
+                       left outer join nodeconfigurations c
+                         on r.nodeId = c.nodeid and r.nodeconfigid = c.nodeconfigid
```
fanf: Yep, I'm adding it.

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -110,95 +107,115 @@ class FindExpectedReportsJdbcRepository(
...
+  }
 
-  /**
-   * This is the implementation of getNodeConfigIdInfos but without commiting the
-   * transaction so that it can be used in a sequence of queries.
+
+  override def findCurrentNodeIds(ruleId : RuleId) : Box[Set[NodeId]] = {
+    sql"""
+      select distinct nodeid from nodeconfigurations
+      where enddate is null and configuration like ${"%"+ruleId.value+"%"}
```
fanf: Yeah... And it's a temporary solution (for 4.0, I think) until we have a "compliance" table, with compliance result by

rudder-core/src/main/scala/com/normation/rudder/services/policies/ExpectedReportsUpdate.scala
```diff
@@ -132,7 +148,122 @@ class ExpectedReportsUpdateImpl(
...
+          configs
+        }
+    }
+  }
+
+  /*
+   * Save a nodeExpectedReport, closing the previous one is
+   * opened for that node (and a different nodeConfigId)
+   */
+  def doUpdateNodeExpectedReports(
```
fanf: we used to have an other "update...", and it was before names settle. I will do a cleaning pass in a next release that will handle that and other "NoButReallyNowDoTheActuallyInterestingThing" names :)

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: The logic is that only nodes with initial promises send reports without a config Id. So if a node is in that case, it is because it never got genererated promises. 
If the first generated promises for that node are beyond the grace period, it means that the run should have used theses promises, and we have a (DNS) problem because it didn't.

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
ncharles: Which cache ? I fail to see where it is cached :/

fanf: My bad, it's the other method that is cached in findDirectiveRuleStatusReportsByRule. 
But at worst, it will have to do full text search request in (Number of nodes) lines, which is in the order of magnitude of 1-10k, which should perform well enought. 
I'm added a logger to spy on that metric.

rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -140,7 +141,6 @@ CREATE SEQUENCE ruleSerialId START 1;
 -- that sequence is used for nodeJoinKey value
 CREATE SEQUENCE ruleVersionId START 1;
 
-
 -- Create the table for the reports information
 CREATE TABLE expectedReports (
```
VinceMacBuche: Should'nt we removed it from reportschema.sql ?

fanf: Yes clearly. It will go in a following PR where I'm adressing tests and cleaning code.

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -145,9 +206,327 @@ case class ComponentExpectedReport(
...
+        , agentRun( json \ "globalAgentRunInterval", None)
+        , policyMode( json \ "globalPolicyMode" \ "mode")
+        , json \ "globalPolicyMode" \ "overridable"
+        , policyMode(json \ "nodePolicyMode")
+      ) match {
+        case (Some(gcm), JInt(ghp), Some(gari), Some(gpm), JBool(gpo), npm) =>
+          try {
+            Full(NodeModeConfig(
+                GlobalComplianceMode(gcm, ghp.toValidInt)
+              , heartbeat( json \ "nodeHeartbeatPeriod")
```
VinceMacBuche: Maybe it could be out in the match above ... even if we match anything ... so we can have  an nhp varaible !!

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -145,9 +206,327 @@ case class ComponentExpectedReport(
...
+              , gari
+              , agentRun( json \ "nodeAgentRunInterval", Some(true) )
```
VinceMacBuche: Same here with a nari varaible ...

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
fanf: what is the advantage to do so?

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
```diff
@@ -131,7 +134,55 @@ class CachedReportsExecutionRepository(
...
+          cache.get(x.agentRunId.nodeId) match {
+            case Some(Some(AgentRunWithNodeConfig(_, Some((oldConfig, Some(expected))), _, _))) if(oldConfig == newConfig) =>
+                Right(AgentRunWithNodeConfig(x.agentRunId, Some((oldConfig, Some(expected))), x.isCompleted, x.insertionId))
+            case _ =>
+                Left(AgentRunWithNodeConfig(x.agentRunId, x.nodeConfigVersion.map(c => (c, None)), x.isCompleted, x.insertionId))
+          }
+      }
+    }
+    //now, get back the one in left, and query for node config
+    val nodeAndConfigs = runWithConfigs.collect { case Right(AgentRunWithNodeConfig(id, Some((config, None)), _, _)) => NodeAndConfigId(id.nodeId, config)}
```
VinceMacBuche: Here you get back the one in right, so it does not match the commentary above

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
```diff
@@ -84,6 +86,7 @@ trait WoReportsExecutionRepository {
 class CachedReportsExecutionRepository(
     readBackend : RoReportsExecutionRepository
   , writeBackend: WoReportsExecutionRepository
+  , fingConfigs : FindExpectedReportRepository
```
VinceMacBuche: findConfigs

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -260,15 +277,15 @@ class FindExpectedReportsJdbcRepository(
             // also, get policy mode from lib - if the directive is missing, we don't know what to do, but it's
             // not more false than if the value changed since expected reports were written.
             val mode = directivesLib.allDirectives.get(directiveId).flatMap( _._2.policyMode)
-            DirectiveExpectedReports(directiveId, mode, lines.map( x =>
-              ComponentExpectedReport(x.component, x.cardinality, x.componentsValues, x.unexpandedCptsValues)
-            ).distinct /* because we have the cardinality to deals with mutltiplicity */ )
+            DirectiveExpectedReports(directiveId, mode, false, lines.map( x =>
```
VinceMacBuche: What if we a System directive ? can't we get bak isSystem value when getting mode just above ?

VinceMacBuche:
To me it's ok and can be rebased!

fanf:
PR rebased

PR #1282, opened 2016-10-17
Fixes #9397: logback.xml doesn't contains most entries in 4.0

ncharles:
https://www.rudder-project.org/redmine/issues/9397

### DIFF
--- a/rudder-web/src/main/resources/logback.xml
+++ b/rudder-web/src/main/resources/logback.xml
@@ -31,21 +31,21 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
 
 -->
 
-<!-- 
+<!--
   The scan / scan period allows to change logs parameter
   and have them hot-reloaded in Rudder. It may have some
   performance impact, so you may want to remove these two
   attributes. But on the other hand, it allows to just
   change a debug level or add new loggers and have them
-  used immediately, without a server restart. 
+  used immediately, without a server restart.
  -->
 <configuration scan="true" scanPeriod="5 seconds">
-  <!-- 
+  <!--
     This is the default logging configuration file. It will be used if you
-    didn't specify the "logback.configurationFile" JVM option. 
+    didn't specify the "logback.configurationFile" JVM option.
     For example, to use a loggin configuration file in "/etc/rudder":
     java ... -Dlogback.configurationFile=/etc/rudder/logback.xml
-    
+
     Full information about the file format is available on the project
     web site: http://logback.qos.ch/manual/configuration.html#syntax
    -->
@@ -57,13 +57,14 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
    -->
 
 
+
   <!--
     Appender configuration - where&how to write logs in SLF4J speaking.
     ===================================================================
     Our default configuration : log on stdout appender so that our logs
     are managed by the container log system (and so, if Tomcat/Jetty/etc
     logs are stored in files and rotated, so are our log information).
-    
+
     Log format is:
     - date/time/thread of the log on 30 chars (fixed)
     - log level on 5 char (fixed)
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/resources/logback.xml
```diff
@@ -57,30 +57,31 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
...
     - log message follows
     - limit exception trace to 30 calls
-    
-    You should not have to modify that. 
+
+    You should not have to modify that.
   -->
   <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
     <encoder class="ch.qos.logback.classic.encoder.PatternLayoutEncoder">
       <Pattern>%d{[yyyy-MM-dd HH:mm:ss]} %-5level %logger - %msg%n%xEx{0}</Pattern>
```
fanf: I know that it has nothing to do with the PR, but I'm questionning the utility to set stack trace to 0. If there is a problem, it may be interesting to have the trace. Pb with scala being that setting it 5 or something small generally don't help at all (it's likelly a bunch a map/flatmap calls...)

rudder-web/src/main/resources/logback.xml
```diff
@@ -164,103 +165,240 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
...
+     The following logger allow to see information about
+     the promise generation process, like step of generation and
+     their timing, information about writing files, etc.
+   -->
+   <!--
+     Set to info to only have macro information, debug to have timing information about
+     each step (good default), trace to have more information.
+   -->
+   <!-- Rudder 3.1 and below -->
+  <logger name="com.normation.rudder.services.policies.DeploymentServiceImpl" level="debug" />
```
fanf: More on that... Debug should not be the "good default". I think it makes sense to always have timing information. Also, I think they should get their log file, in a line-by-line json format (json and not CSV so that we can add/remove inter-point with the level of debug), so that we can easely add it in the anon stats.

rudder-web/src/main/resources/logback.xml
```diff
@@ -164,103 +165,240 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
   -->
   <logger name="trace.ldif.in.file" level="off" />
 
+  <!--
+    Debug LDAP queries related to inventory search
+    ==============================================
+    Uncomment and set to "debug" or "trace" to have precise logs about
+    LDAP queries done by the node search engine
+  -->
+  <logger name="com.normation.rudder.services.queries" level="off" />
```
fanf: When we are at queries, perhaps we could add a logger for quicksearch:
```
  <!--
      Quicksearch
      ===========
   Uncomment and set to "debug" or "trace" to get debugging information about the 
    quicksearch: what is the parsed query, how many results come from each backend, etc.
   -->
  <logger name="com.normation.rudder.services.quicksearch" level="info" />

```

rudder-web/src/main/resources/logback.xml
fanf: (it could go in an other PR)

rudder-web/src/main/resources/logback.xml
ncharles: I do agree with you - no stacktrace means it's impossible to debug - so to debug we need to change the value, which defeat the concept of ease of use

rudder-web/src/main/resources/logback.xml
ncharles: Oh yes, but it should be in 3.1 then

rudder-web/src/main/resources/logback.xml
ncharles: i think we can remove this, as it doesn't exist anymore in this version
As for the debug, this is a topic for another PR :)

ncharles:
Commit modified

rudder-web/src/main/resources/logback.xml
fanf: Ah yes. And in fact, same for the exception log level.

PR #1284, opened 2016-10-18
Fixes #9402: Node properties update doesn't invalidate node configuration hash

fanf:
https://www.rudder-project.org/redmine/issues/9402


fanf:
PR rebased

PR #1298, opened 2016-10-18
Fixes #9438: The link in the Rudder interface should not use french Normation website link

amousset:
https://www.rudder-project.org/redmine/issues/9438


amousset:
OK, merging this PR

PR #1302, opened 2016-10-19
Fixes #9443: When clicking on the "edit group" on the Rule details, we don't get redirected to the said group

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9443


VinceMacBuche:
Commit modified

PR #1305, opened 2016-10-19
Fix typo in configuration.properties.sample

aanriot:
Hello,

Here is a quick typo fix.

Thanks.


peckpeck:
Hello,

Thank you for the fix !

We usually open an issue in redmine every time we fix something.
http://www.rudder-project.org/redmine/projects/rudder/issues/new
However, this is very a simple fix, so I'm going to merge it anyway.

PR #1306, opened 2016-10-19
Fixes #9458: Invalid display in compliance for non-compliant report: shows "Not applicable" when not compliance

ncharles:
https://www.rudder-project.org/redmine/issues/9458


rudder-core/src/main/scala/com/normation/rudder/domain/reports/ReportType.scala
```diff
@@ -104,13 +104,17 @@ object ReportType {
         case (_, BadPolicyMode)       | (BadPolicyMode, _)        => BadPolicyMode
```
fanf: You will have to change the start case above to AuditNotApplicable, else we will always get EnforceNotApplicable for them

ncharles: ha, i'm not sure I want to have by default the auditnotapplicable
Can i simply switch the last two lines ?

fanf: yes

ncharles:
Commit modified

PR #1308, opened 2016-10-19
Fixes #9465: Super slow query on dashboard loading

ncharles:
https://www.rudder-project.org/redmine/issues/9465

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
@@ -95,14 +95,26 @@ case class RoReportsExecutionRepositoryImpl (
         // to interpolate the actual value of in - sql""" """ is more
         // than just interpolation. Need to check perfomances.
         (for {
-          runs <- sql"""select distinct on (r.nodeid)
-                         r.nodeid, r.date, r.nodeconfigid, r.complete, r.insertionid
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
```diff
@@ -95,14 +95,29 @@ case class RoReportsExecutionRepositoryImpl (
...
-                       left outer join nodeconfigurations c
-                         on r.nodeId = c.nodeid and r.nodeconfigid = c.nodeconfigid
-                       where r.complete = true and r.nodeid in (${nodes : nodes.type})
-                       order by r.nodeid, r.insertionid desc
+          // Here to make the query faster, we distinct only on the reportexecution to get the last run
+          // but we need to get the matching last entry on nodeconfigurations.
+          // I didn't find any better solution than doing a distinct on the table
+          // However, I am surprised that we do a left outer join:
+          // we cannot do a search reports on a node that we don't know of
+          // so we will always have a matching nodeconfigurations (except after migrating from 3.x to 4.x)
```
fanf: no, we can have agent run without node configurations: 
- before first generation, 
- for deleted nodes (they are still sending runs)
- certainly for a lot of other cases. 
Perhaps amend the comment, I'm ok witht the query so you can self merge afterward.

ncharles: but we will never look for these runs (we are searching with a list of nodes in parameters)

ncharles:
Commit modified

PR #1314, opened 2016-10-20
Fixes #9477: Empty table rudderproperties

ncharles:
https://www.rudder-project.org/redmine/issues/9477


rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/RudderPropertiesRepository.scala
```diff
@@ -58,15 +58,14 @@ class RudderPropertiesRepositoryImpl(
 
   /**
    * Get the last report id processed by the non compliant report Logger.
-   * If there was no id processed, get the last report id in reports database
-   * and add it to the database.
+   * If there was no id processed, returns an Empty box so that the AutomaticReportLogger can update the value
    */
   def getReportLoggerLastId: Box[Long] = {
```
VinceMacBuche: We should have done Box[Option[Long]] :( 

I merge because it will work but we need to keep in mind this 

Thank you Nicolas!

PR #1340, opened 2016-10-25
Fixes #9530: Each policy generation lead to a new nodeconnfigid even without promise written

fanf:
https://www.rudder-project.org/redmine/issues/9530

There was several problems:

- the nodeconfigId was regenerated too many times, 
- the date used in the new hashcode was not consistent, 
- there was a missing case in run analysis when the runConfigId and expectedConfigId were not the same, 
-  there was a missing case in compare/merge when directives were added to an existing rule.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -88,6 +88,7 @@ import com.normation.rudder.domain.appconfig.FeatureSwitch
 import com.normation.inventory.domain.AixOS
 import com.normation.rudder.domain.reports.NodeModeConfig
 import com.normation.rudder.reports.HeartbeatConfiguration
+import org.joda.time.format.DateTimeFormatter
 
 /**
  * The main service which deploy modified rules and
@@ -107,6 +108,8 @@ trait PromiseGenerationService extends Loggable {
     logger.info("Start policy generation, checking updated rules")
 
     val initialTime = System.currentTimeMillis
+
+    val generationTime = new DateTime(initialTime)
     val rootNodeId = Constants.ROOT_POLICY_SERVER_ID
 
     val result = for {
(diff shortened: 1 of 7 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -853,42 +855,79 @@ trait PromiseGeneration_updateAndWriteRule extends PromiseGenerationService {
...
+     * The hash is directly the NodeConfigHash.hashCode, because we want it to be
+     * unique to a given generation and the "writtenDate" is part of NodeConfigurationHash.
+     * IE, even if we have the same nodeConfig than a
+     * previous one, but the expected node config was closed, we want to get a new
+     * node config id.
+     */
+    def hash(h: NodeConfigurationHash): String = {
+      //we always set date = 0, so we have the possibility to see with
+      //our eyes (and perhaps some SQL) two identicals node config diverging
+      //only by the date of generation
```
ncharles: I fail to understand the goal of that - we could also see with SQL two identical node config
Plus if we have a difference between database and what we receive from the node, it will be super weird

VinceMacBuche: If only the date is different between two config, 

date1-nodeConfigHash and date2-nodeConfigHash

if you do not set the same date when using hashcode, the hash will be different and you will have 

date1-nodeConfigHash1 and date2-nodeConfigHash2

so you can't know that they have only the date that is different

fanf: I don't get the comment :/
So let's try to eplain the intent in an other way: today the datetime is part of the object nodeconfigurationhash. 
But for real, that object is (the hash of important things about the node) + (writtendate). We should have two subobjects, one for each group, but I didn't refactored it because it would have mean a lot more big changes. 
The written date does not reflect the configuration by itself, just the time when it is generated. And with the proposed hash format, that information is already part of the hash. 
So I chose to only took the first group about the config part in NodeConfigHash to generate a hash, and the simplest way to do so was to set the writtenDate to a dummy value (epoch is good for that).
 
Finally, all of that is completely opaque for the node and the user. We could have chosen a pur random number, it would be the same. So not sure about what mean having a difference between database and what we receive from the node. 

Does it help ?

ncharles: Ok, this is completely not what the comment seems to say. From what I understood from the comment is that it helped to see duplicate nodeconfigid in SQL, so I gathered it would be the nodeconfigid written, hence loosing the interest of having a date in it

PR #1341, opened 2016-10-25
Fixes #9539: Two generation are started after changing the global policy mode

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9539


VinceMacBuche:
PR rebased

PR #1355, opened 2016-11-03
Fixes #9604: There is a request to fonts.googleapi.com

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9604


RaphaelGauthier:
PR updated

ncharles:
if we ship these licences, we need to include the OFL licence
http://scripts.sil.org/cms/scripts/page.php?site_id=nrsi&id=OFL_web

PR #1365, opened 2016-11-07
Fixes #9256: Filter rules by application in the directive page

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9256

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
@@ -36,13 +36,18 @@
 */
 
 var anOpen = [];
-
 var ruleCompliances = {};
 var recentChanges = {};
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
```diff
@@ -427,7 +425,7 @@ function createRuleTable(gridId, data, needCheckbox, needActions, needCompliance
          }
        })
       }
-    , "aaSorting": [[ sortingDefault, "asc" ]]
+    , "aaSorting": [[ 0, "desc" ]]
```
VinceMacBuche: That will sort by name in the reverse order :( maybe we should keep asc and reverse our sort order

RaphaelGauthier:
PR updated

PR #1367, opened 2016-11-08
Fixes #9642: Link to setting from node preference doesn't work

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9642


RaphaelGauthier:
PR updated

PR #1368, opened 2016-11-08
Fixes #9593: If i override compliance on a node to "audit", it is still shows "enforce" in its details

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9593

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
@@ -110,19 +110,19 @@ object ComputePolicyMode {
       case (None,None) =>
         (globalMode.mode, "This mode is the globally defined default. You can change it in the global <i><b>settings</b></i>.")
       case (Some(Enforce),Some(Enforce)) =>
-        (Enforce, "<b>Enforce</b> is overriden by both this <i><b>Node</b></i> and this <i><b>Directive</b></i> mode")
+        (Enforce, "<b>Enforce</b> is forced by both this <i><b>Node</b></i> and this <i><b>Directive</b></i> mode")
       case (Some(Enforce),None) =>
-        (Enforce, "<b>Enforce</b> is overriden by this <i><b>Node</b></i> mode")
+        (Enforce, "<b>Enforce</b> is forced by this <i><b>Node</b></i> mode")
       case (None,Some(Enforce)) =>
-        (Enforce, "<b>Enforce</b> is overriden by this <i><b>Directive</b></i> mode")
+        (Enforce, "<b>Enforce</b> is forced by this <i><b>Directive</b></i> mode")
       case (Some(Audit), Some(Audit)) =>
-        (Audit, "<b>Audit</b> mode is overriden by both this <i><b>Node</b></i> and this <i><b>Directive</b></i>")
+        (Audit, "<b>Audit</b> mode is forced by both this <i><b>Node</b></i> and this <i><b>Directive</b></i>")
       case (Some(Audit), None) =>
-        (Audit, "<b>Audit</b> mode is overriden by this <i><b>Node</b></i>")
+        (Audit, "<b>Audit</b> mode is forced by this <i><b>Node</b></i>")
       case (Some(Audit), Some(Enforce)) =>
         (Audit, "The <i><b>Directive</b></i> is configured to <b class='text-Enforce'>enforce</b> but is overriden to <b>audit</b> by this <i><b>Node</b></i>. ")
       case (None,Some(Audit))  =>
-        (Audit, "<b>Audit</b> mode is overriden by this <i><b>Directive</b></i>")
+        (Audit, "<b>Audit</b> mode is forced by this <i><b>Directive</b></i>")
       case (Some(Enforce),Some(Audit))  =>
         (Audit, "The <i><b>Node</b></i> is configured to <b class='text-Enforce'>enforce</b> but is overriden to <b>audit</b> by this <i><b>Directive</b></i>. ")
     }
@@ -135,12 +135,12 @@ object ComputePolicyMode {
       case PolicyModeOverrides.Always =>
         nodeMode match {
           case Some(Audit) =>
-            (Audit.name,"<b>Audit</b> mode is overriden by this <i><b>Node</b></i>")
+            (Audit.name,"<b>Audit</b> mode is forced by this <i><b>Node</b></i>")
           case nodeMode    =>
 
             val (defaultMode,expl) = nodeMode match {
               case Some(mode) =>
-                ( mode, s"<b>${mode.name}</b> mode is overriden by this <i><b>Node</b></i>")
+                ( mode, s"<b>${mode.name}</b> mode is forced by this <i><b>Node</b></i>")
               case None =>
                 ( globalMode.mode, "This mode is the globally defined default. You can change it in the global <i><b>settings</b></i>.")
             }
@@ -158,18 +158,18 @@ object ComputePolicyMode {
                     case Some(Enforce) =>
                      (Audit.name, "The <i><b>Node</b></i> is configured to <b class='text-Enforce'>enforce</b> but is overriden to <b>audit</b> by all <i><b>Directives</b></i>. ")
                     case Some(Audit) =>
-                     (Audit.name, "<b>Audit</b> mode is overriden by both this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
+                     (Audit.name, "<b>Audit</b> mode is forced by both this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
                     case None =>
-                     (Audit.name,"<b>Audit</b> mode is overriden by all <i><b>Directives</b></i>")
+                     (Audit.name,"<b>Audit</b> mode is forced by all <i><b>Directives</b></i>")
                   }
                 case Some(Enforce) =>
                   nodeMode match {
                     case Some(Audit) =>
                      (Audit.name, "All <i><b>directives</b></i> are configured to <b class='text-Enforce'>enforce</b> but is overriden to <b>audit</b> by the <i><b>Node</b></i>. ")
                     case Some(Enforce) =>
-                     (Enforce.name, "<b>Enforce</b> mode is overriden by both this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
+                     (Enforce.name, "<b>Enforce</b> mode is forced by both this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
                     case None =>
-                     (Enforce.name,"<b>Enforce</b> mode is overriden by all <i><b>Directives</b></i>")
+                     (Enforce.name,"<b>Enforce</b> mode is forced by all <i><b>Directives</b></i>")
                   }
 
               }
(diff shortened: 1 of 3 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,163 @@ object RulePolicyMode {
...
+  def directiveModeOnNode(nodeMode : Option[PolicyMode], globalMode : GlobalPolicyMode)(directiveMode : Option[PolicyMode]) = {
+   globalMode.overridable match {
+      case PolicyModeOverrides.Unoverridable =>
+        (globalMode.mode.name, s"""Rudder's global agent policy mode is set to <i><b>${globalMode.mode.name}</b></i> and is not overridable on a per-node or per-directive basis. Please check your Settings or contact your Rudder administrator.""")
+      case PolicyModeOverrides.Always =>
+        val (mode,expl) = matchMode(nodeMode,globalMode,directiveMode)
+        (mode.name,expl)
+    }
+  }
+  def matchMode(nodeMode : Option[PolicyMode], globalMode : GlobalPolicyMode, directiveMode : Option[PolicyMode]) : (PolicyMode,String) = {
```
ncharles: You should add a coment there, saying that it cannot be called with a globalmode not overridable (and probably enforce it in the code)

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,163 @@ object RulePolicyMode {
...
+        (globalMode.mode.name, s"""Rudder's global agent policy mode is set to <i><b>${globalMode.mode.name}</b></i> and is not overridable on a per-node or per-directive basis. Please check your Settings or contact your Rudder administrator.""")
+      case PolicyModeOverrides.Always =>
+        nodeMode match {
+          case Some(Audit) =>
+            (Audit.name,"<b>Audit</b> mode is overriden by this <i><b>Node</b></i>")
+          case nodeMode    =>
+
+            val (defaultMode,expl) = nodeMode match {
+              case Some(mode) =>
+                ( mode, s"<b>${mode.name}</b> mode is overriden by this <i><b>Node</b></i>")
```
ncharles: isn't it ${globalMode.mode.name} mode is overriden by this Node ?

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,163 @@ object RulePolicyMode {
...
+              case Nil => default
+              // We only have one element!!
+              case mode :: Nil => mode match {
+                case None => default
+                case Some(Audit) =>
+                  nodeMode match {
+                    case Some(Enforce) =>
+                     (Audit.name, "The <i><b>Node</b></i> is configured to <b class='text-Enforce'>enforce</b> but is overriden to <b>audit</b> by all <i><b>Directives</b></i>. ")
+                    case Some(Audit) =>
+                     (Audit.name, "<b>Audit</b> mode is overriden by both this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
```
ncharles: why overriden ? If node and directive are both audit, then result is audit.
Plus, if i'm correct, this case is caught by line 137: if nodeMode is audit, then it is audit

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,163 @@ object RulePolicyMode {
...
+
+              }
+
+              case modes  =>
+                // Here we now need to check if global mode is the same that all selected mode for directives
+                // So we replace any "None" (use global mode) by global mode and we will decide after what happens
+
+                directivesMode.map(_.getOrElse(defaultMode)).toList match {
+                  case Nil => default
+                  case mode :: Nil => (mode.name, s"<b>${mode.name}</b> mode is overriden by this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
```
ncharles: why the logic is not the same as in previous section? 
And if node is enforce and you have defaultMode audit (overridable), and the directives are Audit + Default audit (overridable), then the text will be incorrect, as the Audit (overridable) will not by overriden by the node (because it is enforce).

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,163 @@ object RulePolicyMode {
...
+              }
+              // Directive applied on No Nodes, that should not happen ... we should use NonEmptyList here ... , fallback to globalMode
+              case Nil => default
+              case modes =>
+                // Here we now need to check if global mode is the same that all selected mode for directives
+                // So we replace any "None" (use global mode) by global mode and we will decide after what happens
+
+                nodeModes.map(_.getOrElse(defaultMode)).toList match {
+                  case Nil => default
+                  case mode :: Nil => (mode.name, s"<b>${mode.name}</b> mode is overriden by this <i><b>Node</b></i> and all <i><b>Directives</b></i>")
```
ncharles: same remark here.

VinceMacBuche:
To answer your question, case with one directive is more complex because i can be more precise than with N Directives. going deeper into explanation with N directive will be a huge mess and will not give much more information for user. 

The best thing for them is to look deeper in compliance to understand it (and look for 1 directive case)

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
VinceMacBuche: Yeah I need to comment much more this not so easy piece of code ...  That was my intention before merging it but wanted some feedback early from you ;)

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
VinceMacBuche: In fact 'overriden' is not the word I should have used but 'forced' instead ie:

> Audit mode is forced by this Node

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
VinceMacBuche: Same remark as above should be 'forced' and not 'overriden'

VinceMacBuche: And You're right about line 137, but i need to treat the case Some(Audit) since i have an Option[PolicyMode] (scala compiler / pattern matching does not authorize me here to have an Option[Enforce] ...) 

I wanted to add a comment here on my commenting phase ...

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
VinceMacBuche: defaultMode may be NodeMode or globalMode (see line 141) 

If node mode is Enforce (or audit, but treated above ...) , then defaultMode will be the nodeMode and not the globalMode

If not mode is not overriding any mode, then defaultMode will be global mode 

logic is simpler here because it would be very complicated to explain precisely what makes it this (globally it's that Directive / global mode and node mode are the same)

ncharles: But if defaultMode is NodeMode(Enforce), but Directive is Audit, then it will tell that the mode is forced by Node, which is false

VinceMacBuche:
@ncharles  I added three more commits

=> First one improved message and treat node in rule compliance (was not done before)
=> Second one factorize
-> Thied one add comments

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,162 @@ object RulePolicyMode {
...
+  }
+
+  // The goal of that function is to compute the effective mode applied by a combination of a element of one type and a set of element of another type
+  // and explain why it's result. (ie : Compute the policy mode applied on a RUle depending on mode of One Node and all Directive from that Rule
+
+  // It uses:
+  // * The global defined mode
+  // * A unique policy mode of one element of type A as base (generally a Directive or a Node)
+  // * multiples Policy modes of a set elements of type B that we will look into it to understand what's going on
+  private[this] def generic(
```
ncharles: i don't want to nag, but the name is really non-informative. Could you call it genericModeComputataton, or something like that ?

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,162 @@ object RulePolicyMode {
...
+      uniqueMode : Option[PolicyMode]
+    , uniqueKind : String
```
ncharles: please, can you explain in the comments what are uniqueKind and multipleKind ? this is not straightforward

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,166 @@ object RulePolicyMode {
...
+                }
+            }
+        }
+    }
+  }
+
+  // Used to compute mode applied on a Rule depending on the Node state and directive from the Rule
+  // Used in compliance table in node page
+  def ruleModeOnNode(nodeMode : Option[PolicyMode], globalMode : GlobalPolicyMode)( directivesMode : Set[Option[PolicyMode]]) : (String,String) = {
+    val mixed = "This Rule has at least one Directive that will <b class='text-Enforce'>enforces</b> configurations, and at least one that will <b class='text-Audit'>audits</b> them."
```
ncharles: invalid english: either:
"that will enforce"
or
"that enforces"

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -95,6 +95,166 @@ object RulePolicyMode {
...
+  // Used in Node compliance table in Rule page
+  def nodeModeOnRule(nodeMode : Option[PolicyMode], globalMode : GlobalPolicyMode)( directivesMode : Set[Option[PolicyMode]]) : (String,String) = {
+    val mixed = "This Node applies at least one Directive that will <b class='text-Enforce'>enforces</b> configurations, and at least one that will <b class='text-Audit'>audits</b> them."
+    genericComputeMode(nodeMode, "Node", directivesMode, "Directives", globalMode, mixed)
+  }
+
+  // Used to computed mode applied on a Directive depending on it's mode and and all Nodes appliying it
+  // Used in Node compliance table in Rule page
+  def directiveModeOnRule(nodeModes : Set[Option[PolicyMode]], globalMode : GlobalPolicyMode)( directiveMode : Option[PolicyMode]) = {
+    val mixed = "This Directive is applied on at least one Node that will <b class='text-Enforce'>enforces</b> configurations, and at least one that will <b class='text-Audit'>audits</b> them."
```
ncharles: same

rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
ncharles: same with audits

PR #1384, opened 2016-11-21
Fixes #9645: Add a table for node compliance

fanf:
https://www.rudder-project.org/redmine/issues/9645

This PR add the storage of the compliance by node by run in a new Postgres Table. 

The only interesting things are: 
- when we call the store action: in ReportsExecutionService, just after having updated the node compliance cache (we use the result, all calculus are done)
- the database structure. 

### column

There is a good description on the table declaration of field. Some more "what/why":

- nodeId, runTimestamp, nodeConfigId: what we are interested in to do queries. We would much likely want to get the last line by runTimestamp for a nodeId. 
- endOfLife: date after which the compliance is not valide anymore. If no new runs, it means the node doesn't send reports anymore. Perhaps the name is not that good, and "endDate" should be used. 
- runAnalysis: information about the run, the same that are used in the node details screen, on the colored header (but without information already in the 3 first columns). It does not contains the "expected node configuration", because 1/ it's huge, 2/ we can get it by looking to run status, and if the run had a configId, find back the expected config from theire 3/ we almost have all the information from the compliance appart in certain (hill) cases
- summary: compliance summary at the node level - to query only that on the future for node list or in debug
- details: the whole compliance details down to the rule/directive/component/value/message. It should be sufficient to any use case in the future (ie no information destroyed, everything is kept)

**Question**: should we add a dedicated column for nodeConfigId ? It seems rather useful, even if it may be null in pathological cases. 

### Compliance Values in percent - aggregated needed ?

In Scala, we store compliance level with a count of each type of reports (and of course we have the total with the sum of each kind). That allows to manipulate ints, which is better than double for rounding error :)
A chose to store the compliance level in percent + the total number of report because if data are not normalized, it will be impossible to do queries in the future like "give me all the directive with success < 50%". 

And **I'm wondering** if we should not go even further, and store in addition to the completely detailed version also aggregated result for success (ie report success + report compliant + repaired + n/a), non-compliant, error, etc - i.e all the top level status

### it is by node by run, not by node-rule by run or aggregated on several runs. 

Aggregation is much more complexe to achieve on a node level. It needs a "Delta" logic (ie what change between a snapshot point and following runs) to be efficient. All that should be done in a different store or different module. 

So why not just use a by run by node/rule/directive(/component) table? Because it would lead to much more redondancies in the first columns between lines, and many, many more lines. 
As an example, we have ~300 runs by node by day. Addind rule and directives, we add (low estimate) a x20 number of lines. Adding componenent, and it's (low estimation) a x100.  

And it felt more logical / consistent with other part of Rudder to have a by-run-by-node result.  

### disk size

The produce json for 12 directives, including systems (so some with a large number of components), is around 75kB. 

The corresponding stored json seems to amortize toward 2.5kB - Postgres is compressing like hell :)
With that value, we get ~700MB of data by day for 1000 nodes. 

Using the JSON type format doesn't change AT ALL the disk size (certainly because postgres store json as text, only checking it's valid json). 
Using JSONB would improve the space taken, but it's only on Postgres 9.5 and up.

### DIFF
--- a/rudder-core/src/main/resources/reportsSchema.sql
+++ b/rudder-core/src/main/resources/reportsSchema.sql
@@ -169,6 +169,10 @@ CREATE TABLE nodeConfigurations (
 -- here, I'm using text but with valid JSON in it, because for now we can't impose postgres > 9.2, 
 -- and interesting function are on 9.3/9.4.  we will be able to migrate with:
 -- ALTER TABLE table1 ALTER COLUMN col1 TYPE JSON USING col1::JSON;
+-- or if version == 9.2, we need to do a several steps script like: 
+-- https://github.com/airblade/paper_trail/issues/600#issuecomment-136279154
+-- and then garbage collect space with:  
+-- vacuum full nodecompliance;
 
 , configuration     text NOT NULL CHECK (configuration <> '' )
 
@@ -191,6 +195,66 @@ CREATE TABLE archivedNodeConfigurations (
 , endDate           timestamp with time zone
 , configuration     text NOT NULL CHECK (configuration <> '' )
 , PRIMARY KEY (nodeId, nodeConfigId, beginDate)
+);
+
+/* 
+ *************************************************************************************
+ * The following tables stores "node compliance", i.e all the interesting information
+ * about what was the compliance of a node FOR A GIVEN RUN. 
+ * That table *only* store information for runs, and does not track (non exaustively):
+ * - when the node expected configuration is updated - only a new run will check,
+ * - node not sending runs - only the fact that we don't have data can be observed
+ * - if a node is deleted
+ *************************************************************************************
+ */
+
+-- Create the table for the node compliance
+CREATE TABLE nodeCompliance (
+  nodeId            text NOT NULL CHECK (nodeId <> '')  
+, runTimestamp      timestamp with time zone NOT NULL
+
+-- endOfList is the date until which the compliance information
+-- are relevant. After that date/time, the node must have sent
+-- a more recent run, this one is not valide anymore. 
+, endOfLife         timestamp with time zone
+
+-- all information about the run and what lead to that compliance:
+-- the run config version, the awaited config version, etc
+-- It's JSON (but in a string, cf explanation in nodeConfigurations table)
+-- and has such, it must not be empty (at least '{}')
+
+, runAnalysis       text NOT NULL CHECK (runAnalysis <> '' )
+
+-- node compliance summary (i.e, no details by rule etc), in percent
+-- that JSON, again
+
+, summary           text NOT NULL CHECK (summary <> '' )
+
+-- the actual compliance with all details
+-- Again, JSON
+
+, details  text NOT NULL CHECK (details <> '' )
+
+-- Primary key is given by a run timestamp and the node id. We could
+-- have duplicate if node clock change, but it would need to have 
+-- exact same timestamp down to the millis, quite improbable.
+
+, PRIMARY KEY (nodeId, runTimestamp)
+);
+
+CREATE INDEX nodeCompliance_nodeId ON nodeCoCompliance (nodeId);
+CREATE INDEX nodeCompliance_runTimestamp ON nodeCoCompliance (runTimestamp);
+CREATE INDEX nodeCompliance_endOfLife ON nodeCoCompliance (endOfLife);
+
+-- Create the table for the archived node compliance
+CREATE TABLE archivedNodeCompliance (
+  nodeId            text NOT NULL CHECK (nodeId <> '')  
+, runTimestamp      timestamp with time zone NOT NULL
+, endOfLife         timestamp with time zone
+, runAnalysis       text NOT NULL CHECK (runAnalysis <> '' )
+, summary           text NOT NULL CHECK (summary <> '' )
+, details  text NOT NULL CHECK (details <> '' )
+, PRIMARY KEY (nodeId, runTimestamp)
 );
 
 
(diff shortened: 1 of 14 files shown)


fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -340,3 +340,140 @@ object MessageStatusReport {
...
+  implicit class AggregatedStatusReportToJs(x: AggregatedStatusReport) {
+    import ComplianceLevelSerialisation._
+
+    def toJValue(): JValue = {
+
+      //here, I'm not sure that we want compliance or
+      //compliance percents. Having a normalized value
+      //seems far better for queries in the futur.
+      //but in that case, we should also keep the total
+      //number of events to be able to rebuild raw data
```
ncharles: Do we need the raw data really ? what would be the benefits?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ComplianceRepository.scala
```diff
@@ -0,0 +1,133 @@
...
+        case Some((runTime, expected)) =>
+          Some(RunCompliance.from(runTime, x.expirationDateTime, r))
+      }
+
+      case x:NoExpectedReport =>
+        // here, the expiration date has not much meaning, since we don't have
+        // information on that node configuration (and so the node has most likelly no
+        // idea whatsoever of any config, even global). Take default values,
+        // ie 5min for run + 5min for grace
+        Some(RunCompliance.from(x.lastRunDateTime, x.lastRunDateTime.plusMinutes(10), r))
```
ncharles: does it make sense to store something if we have no expected reports ? how can we really compute compliance ?

rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
fanf: Sometimes, we want the numbers to build new metrics, not only the percent, especially on aggreagation metrics.  
For ex, it allows to get the number of changes for a run. Or the number of non-compliance.

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ComplianceRepository.scala
fanf: Today, we do compule compliance on that case. Not sure how (un)successful it is, but we need to just be exaustive on the error cases we can encounter.

fanf:
>     what happens when the webapp didn't run for a long time - will it catch up all reports, and store them on table ? (with the risk that the reports massively lags to show up) Or will it skip the old ones?

It is exactly the same process than the one used to process runs. Same risks, same behavior. 
We could implement better tactics on that. 
For now, it doesn't have any impact on compliance displaying, because it is not use. On the long run, we will need to better think to the run problem. 

>  we should have something to clean these compliance, why not include it directly ?
I want to add it within the same process as the expected node configuration clean up, which is not done. So I need to do it in 4.0, then modify that PR.

rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
ncharles: Ha, make sense

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ComplianceRepository.scala
ncharles: ok

rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -193,6 +197,66 @@ CREATE TABLE archivedNodeConfigurations (
...
+ * - when the node expected configuration is updated - only a new run will check,
+ * - node not sending runs - only the fact that we don't have data can be observed
+ * - if a node is deleted
+ *************************************************************************************
+ */
+
+-- Create the table for the node compliance
+CREATE TABLE nodeCompliance (
+  nodeId            text NOT NULL CHECK (nodeId <> '')  
+, runTimestamp      timestamp with time zone NOT NULL
```
ncharles: i do agree with your remark that nodeconfigid should be in a dedicated column, I'm pretty sure it will be useful in many cases

ncharles:
This is great - is you can add the nodeconfigid in the table, this would be great !

rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -155,26 +181,122 @@ object ComplianceLevel {
...
+      , JE.Num(compliance.pc.nonCompliant)        // 11
+      , JE.Num(compliance.pc.auditError)          // 12
+      , JE.Num(compliance.pc.badPolicyMode)       // 13
     )
+
+    def toJson(): JObject = {
+      import compliance._
+      toJObject(pending, success, repaired, error, unexpected, missing, noAnswer
+                 , notApplicable, reportsDisabled, compliant, auditNotApplicable
+                 , nonCompliant, auditError, badPolicyMode
```
ncharles: i'm pretty sure we'l have bugs because of that one day

fanf: I have no good alternative. At least, it's an object, i.e attribute are named, and we can see in the json if the numbers are ok.

ncharles: Yes - it just a sad remark :(

PR #1385, opened 2016-11-22
Fixes #9703: Rudder v.3.2.9 web UI \"Content of the file\" field design was degraded.

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9703

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService2.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService2.scala
@@ -155,7 +155,7 @@ case class NodeApiService2 (
 
   def deleteNode(req : Req, ids: Seq[NodeId]) = {
     implicit val prettify = restExtractor.extractPrettify(req.params)
-    implicit val action = "changePendingNodeStatus"
+    implicit val action = "deleteNode"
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/webapp/style/rudder-bootstrap.css
```diff
@@ -438,7 +438,12 @@
     resize: vertical;
     border-radius: 4px !important;
 }
-
+table .tw-bs textarea.form-control,.tw-bs table textarea.form-control{
```
VinceMacBuche: maybe the second selector is enough ? 

Should we also do it on text field ?

RaphaelGauthier:
PR updated

ncharles:
Shouldn't it be targeted to 3.1 ?
I just compared 3.1 and 4.0, they have the same font

VinceMacBuche:
@ncharles You are absolutely right, i changed it in Redmine

@RaphaelGauthier you can retarget this PR using "rudder-dev retarget" when you are on the good branch

RaphaelGauthier:
PR updated

VinceMacBuche:
@RaphaelGauthier It seems it does not ahev recreated a new PR ... can you do 'rudder-dev commit VME -f" ?

PR #1388, opened 2016-11-23
Fixes #9725: Data source Scala model

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9725

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
@@ -1,0 +1,71 @@
+/*
+*************************************************************************************
+* Copyright 2016 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -0,0 +1,66 @@
...
+
+sealed trait DataSourceType {
+  def name : String
+}
+
+case object ByNodeSourceType extends DataSourceType {
+  val name = "byNode"
+}
+
+case object GlobalSourceType extends DataSourceType {
```
fanf: This should be renamed to "AllNodeSourceType"

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -0,0 +1,66 @@
...
+  val name = "global"
```
fanf: Here, we should use a case class with two more parameters: 

- identifyingAttributePath: the Json Path that identifies the attribute used to know what node the json is for; 
- rudderNodeProperty: the "${node.property[foo][bar]}" or "${rudder.node.hostname}" to compare identifyingAttributePath against to.

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -0,0 +1,66 @@
...
+}
+
+case class DataSourceName(value : String) extends AnyVal
+
+case class DataSource (
+    name       : DataSourceName
+  , description: String
+  , sourceType : DataSourceType
+  , url        : String
+  , headers    : Map[String,String]
```
fanf: perhaps we should add a "httpMethod", even if we only authorize GET for now.

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -0,0 +1,66 @@
...
+  , path       : String
+  , frequency  : Int
```
VinceMacBuche: Do you think changing type to "joda.time.Seconds" would be interesting ?

VinceMacBuche:
Commit modified

PR #1392, opened 2016-11-26
Fixes #9724: Query HTTP datasource

fanf:
https://www.rudder-project.org/redmine/issues/9724

Work in progress about getting node dataset from Rest service. 
Current status: mostly everything work for "node by node" request. 

It can be merged as it is (it's fully functionnal) and other PRs may be done for missing parts.

## What is added by that PR 

- able to GET dataset from a datasource (getting info from repos, updating nodes in repos) + test on how it behave with 1000 nodes and timeout / errors for some nodes,
- fetch data with scheduling (or not) + triggers (on new node, on generation, on user request for one / all nodes)
- api to trigger update for one / all nodes
- logging (action and requests timing)
- repository facade which manage the update of live scheduler when datasources are modified

## Missing 
- the "one request for all node" mode. 
- the status updates on database (for now, we only have reports via logs)
- the actual database serialisation/desirialisation (current repository is only in memory)

## Open question

- do we want to start a generation after an update? I think so (in all case, the generation will see if things were changed since it's last run)
- do we want to prevent two consecutives non-interactive updates at less than "period" appart? The use case in mind is if we start generation after update, and we have "trigger data fetching on generation", at each periodic update we are going to have a new fetch just after [note also that the fetching on generation must not trigger a generation, else its a never ending game :) ]
- validate endpoints for API update (currently ..../api/latest/nodes/fetchData and .../nodes/xxx-xxx/fetchData)

## Some more  #details

The fetch logic is working, ie. 
- from a datasource and node information/contexte, 
- compile the URL to fetch, 
- fetch data, 
- select sub data, 
- create a node property. 

From that, we have a whole service for a node, with I/O for param&result:
- get datasource and node information from repositories, 
- fetch, 
- in case of success, update node property and node dataset info, 
- in all case, update datasource last updated. 

## Framework info

- I'm using https://github.com/jayway/JsonPath/ for selecting sub-json. It kind of works, but none of the seemingly interesting option were actually interesting. And I HATE JAVA, mutable, no data representation of the logic, cast.... ARG;
- for client, I'm using for now https://github.com/scalaj/scalaj-http. Extremlly simple,  but fully synchrone. It seems to not be a problem for now, but perhaps we are going to want a much more efficient (async and pooled) client that efficiently manage a ton of outgoing connexion (because one connexion by node can become huge)
- for tests, I'm using http4s (http://http4s.org/). It seems really, really nice but support for Java 7 is EOL, so I'm a little bother to add it for something else than dependency. That constraint is becoming a real mess. 
- monix (https://monix.io) for async computing and scheduling, because it's light and fast and really easy to use.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
@@ -38,6 +38,26 @@
 package com.normation.rudder.datasources
 
 import net.liftweb.common.Box
+import ch.qos.logback.core.db.DataSourceConnectionSource
+import com.normation.eventlog.EventActor
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.nodes.NodeInfo
+import org.joda.time.DateTime
+import net.liftweb.common.EmptyBox
+import net.liftweb.common.Full
+import com.normation.rudder.domain.parameters.Parameter
+import com.normation.utils.StringUuidGenerator
+import com.normation.eventlog.ModificationId
+import com.normation.rudder.domain.eventlog._
+import scala.concurrent.duration._
+import net.liftweb.common.Failure
+
+
+final case class PartialNodeUpdate(
+    nodes        : Map[NodeId, NodeInfo] //the node to update
+  , policyServers: Map[NodeId, NodeInfo] //there policy servers
+  , parameters   : Set[Parameter]
+)
 
 trait DataSourceRepository {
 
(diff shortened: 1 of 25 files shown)


fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -48,4 +68,212 @@ trait DataSourceRepository {
...
+ * It doesn't deal at all with the serialisation / deserialisation of data source
+ * in data base.
+ */
+class DataSourceRepoImpl(
+    backend: DataSourceRepository
+  , fetch  : QueryDataSourceService
+  , uuidGen: StringUuidGenerator
+) extends DataSourceRepository with DataSourceUpdateCallbacks {
+
+  private[this] var datasources = Map[DataSourceId, DataSourceScheduler]()
```
VinceMacBuche: It should be initialisated with all datasources from backend

VinceMacBuche: Can it be considered as a cache ? (and get informations for getAll / get in instead of going directly to backend)

VinceMacBuche:
PR replaced by https://github.com/Normation/rudder/pull/1415

PR #1393, opened 2016-11-29
Fixes #9742: API allows to create duplicate UUIDs at group creation

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9742

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
@@ -145,7 +145,6 @@ class RoLDAPNodeGroupRepository(
     }
   }
 
-
   def getGroupsByCategory(includeSystem:Boolean = false) : Box[SortedMap[List[NodeGroupCategoryId], CategoryAndNodeGroup]] = {
     groupLibMutex.readLock { for {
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
```diff
@@ -757,8 +747,10 @@ class WoLDAPNodeGroupRepository(
   override def create(nodeGroup: NodeGroup, into: NodeGroupCategoryId, modId: ModificationId, actor:EventActor, reason:Option[String]): Box[AddNodeGroupDiff] = {
     for {
       con           <- ldap
-      exists        <- if (nodeGroupExists(con, nodeGroup.name)) Failure("Cannot create a group with name %s : there is already a group with the same name".format(nodeGroup.name))
-                       else Full(Unit)
+      exists        <- if (checkNodeGroupExists(con, nodeGroup))
+                         Failure(s"Cannot create a group '${nodeGroup.name}': a group with the same id (${nodeGroup.id.value}) or name already exists".format(nodeGroup.name))
```
VinceMacBuche: You're right, format is the old way to build string with values, i removed "format" calls from ost string but forget at least one ... thanks !

VinceMacBuche:
Commit modified

PR #1396, opened 2016-11-29
Fixes #9743: When applying same directive in different rule, reporting is broken

fanf:
https://www.rudder-project.org/redmine/issues/97434

So. I can't reproduce it even with the data. 
The test show that the computed compliance is what is awaited. Not sure what to do with that further more. Merge the PR, mark the problem as "can't reproduce" ?


rudder-core/src/test/scala/com/normation/rudder/services/reports/ComplianceTest.scala
```diff
@@ -0,0 +1,208 @@
...
+        (reports, config)
+      }).openOrThrowException("the test should not throw an exception")
+
+      val runTime = reports.head.executionTimestamp
+      val runinfo = ComputeCompliance(runTime, config, runTime)
+      val status = ExecutionBatch.getNodeStatusReports(config.nodeId, runinfo, reports)
+      println(status.forNode)
+      println(status.runInfo)
+      println(status.statusInfo)
+      println(status.compliance)
```
ncharles: i think you want to remove these println

rudder-core/src/test/scala/com/normation/rudder/services/reports/ComplianceTest.scala
```diff
@@ -0,0 +1,208 @@
...
+ * the corresponding reports, compute the resulting node status, and compare it
+ * to an expected one.
+ *
+ */
+
+
+@RunWith(classOf[JUnitRunner])
+class ComplianceTest extends Specification {
+
+  val dateParser = ((new DateTimeFormatterBuilder())
```
ncharles: these utility methods may be useful in other tests. Would it make sense to put them in a separate file to be used generically ?

fanf:
PR rebased

rudder-core/src/test/scala/com/normation/rudder/services/reports/ComplianceTest.scala
```diff
@@ -0,0 +1,225 @@
+/*
+*************************************************************************************
+* Copyright 2011 Normation SAS
```
ncharles: 2016

fanf:
PR rebased

fanf:
OK, merging this PR

fanf:
OK, merging this PR

fanf:
OK, merging this PR

fanf:
OK, merging this PR

PR #1398, opened 2016-12-01
Fixes #9700: API Rest to define Data sources

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9700

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
@@ -38,34 +38,106 @@
 package com.normation.rudder.datasources
 
 import org.joda.time.DateTime
-import org.joda.time.Seconds
+import net.liftweb.common._
+import scala.concurrent.duration.FiniteDuration
+import com.normation.inventory.domain.NodeId
+
+object DataSource {
+  val defaultDuration = FiniteDuration(5,"minutes")
+}
 
 sealed trait DataSourceType {
   def name : String
 }
 
-case object ByNodeSourceType extends DataSourceType {
+final case class HttpDataSourceType (
+    url            : String
+  , headers        : Map[String,String]
+  , httpMethod     : String
+  , sslCheck       : Boolean
+  , path           : String
+  , requestMode    : HttpRequestMode
+  , requestTimeOut : FiniteDuration
+) extends DataSourceType {
+  val name = HttpDataSourceType.name
+}
+
+object HttpDataSourceType{
+  val name = "http"
+}
+
+sealed trait HttpRequestMode {
+  def name : String
+}
+final case object OneRequestByNode extends HttpRequestMode {
   val name = "byNode"
 }
 
-case class AllNodesSourceType(
+final case class OneRequestAllNodes(
     matchingPath  : String
   , nodeAttribute : String
-) extends DataSourceType {
-  val name = "allNodes"
+) extends HttpRequestMode {
+  val name = OneRequestAllNodes.name
 }
 
-case class DataSourceName(value : String) extends AnyVal
+object OneRequestAllNodes {
+  val name = "allNodes"
+}
+final case class DataSourceName(value : String) extends AnyVal
+final case class DataSourceId  (value : String) extends AnyVal
 
-case class DataSource (
-    name       : DataSourceName
-  , description: String
-  , sourceType : DataSourceType
-  , url        : String
-  , headers    : Map[String,String]
-  , httpMethod : String
-  , path       : String
-  , frequency  : Seconds
-  , lastUpdate : Option[DateTime]
-  , enabled    : Boolean
+sealed trait DataSourceSchedule {
+  def duration : FiniteDuration
+}
+
+final case class NoSchedule(
+  savedDuration : FiniteDuration
+) extends DataSourceSchedule {
+  val duration = savedDuration
+}
+
+final case class Scheduled(
+  duration : FiniteDuration
+) extends DataSourceSchedule
+
+final case class DataSourceRunParameters (
+    schedule     : DataSourceSchedule
+  , onGeneration : Boolean
+  , onNewNode    : Boolean
 )
+
+final case class DataSourceStatus (
+    lastRunDate : Option[DateTime]
+  , nodesStatus : Map[NodeId, DataSourceUpdateStatus]
+)
+
+sealed trait DataSourceUpdateStatus{
+  def state       : String
+  def lastRunDate : DateTime
+}
+
+final case class DataSourceUpdateSuccess(
+  lastRunDate : DateTime
+) extends DataSourceUpdateStatus {
+  val state = "success"
+}
+
+final case class DataSourceUpdateFailure(
+    lastRunDate     : DateTime
+  , message         : String
+  , lastSuccessDate : Option[String]
+) extends DataSourceUpdateStatus {
+  val state = "failure"
+}
+
+final case class DataSource (
+    id            : DataSourceId
+  , name          : DataSourceName
+  , sourceType    : DataSourceType
+  , runParam      : DataSourceRunParameters
+  , description   : String
+  , enabled       : Boolean
+  , updateTimeOut : FiniteDuration
+) {
+  val scope = "all"
+}
--- a/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
@@ -35,41 +35,17 @@
 *************************************************************************************
 */
 
-package com.normation.rudder.web.rest
+package com.normation.rudder.datasources
 
-import org.junit._
-import org.junit.Assert._
-import org.junit.runner._
-import org.junit.runner.RunWith
-import org.specs2.mutable._
-import org.specs2.runner._
-
-import RestTestSetUp._
-import net.liftweb.common.Full
-import net.liftweb.http.PlainTextResponse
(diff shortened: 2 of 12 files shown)


rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -39,11 +39,21 @@ package com.normation.rudder.datasources
...
 import org.joda.time.Seconds
+import net.liftweb.common._
 
 sealed trait DataSourceType {
   def name : String
 }
 
+object DataSourceType {
+  def apply(name : String) : Box[DataSourceType] = {
+    if (name == ByNodeSourceType.name) {
```
fanf: Not sure I understand why only ByNode?
The generic way for that is to add a "val all = ca.mrvisser.sealerate.values[DataSourceType]"
And then use a "all.find(name.toLowerCase == _.name.toLowercase) match {case None => Failure etc"

To make all work, you also need to put you case object into DataSourceType.
And they should be mark "final", because scala is stupid.

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -52,9 +62,12 @@ case class AllNodesSourceType(
...
   , nodeAttribute : String
 ) extends DataSourceType {
-  val name = "allNodes"
+  val name = AllNodesSourceType.name
 }
 
+object AllNodesSourceType {
+  val name = "allNodes"
+}
 case class DataSourceName(value : String) extends AnyVal
```
fanf: should be final

rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApi.scala
```diff
@@ -0,0 +1,102 @@
...
+      enab = enabled.getOrElse(false)
+
+    } yield {
+      DataSource(
+          name
+        , desc
+        , sourceType
+        , u
+        , head
+        , "get"
```
fanf: shouldn't it be GET ? (not sure, but the convention is more on GET. We can always .toLowerCase when needed)

rudder-web/src/test/scala/com/normation/rudder/web/rest/RestStatusTest.scala
```diff
@@ -47,9 +47,55 @@ import org.specs2.runner._
 import RestTestSetUp._
 import net.liftweb.common.Full
 import net.liftweb.http.PlainTextResponse
```
fanf: I think it should go into its own file

rudder-web/src/test/scala/com/normation/rudder/web/rest/RestTestSetUp.scala
```diff
@@ -81,10 +84,34 @@ object RestTestSetUp {
...
+  val restExtractorService =
+  RestExtractorService (
+      null //roRuleRepository
+    , null //roDirectiveRepository
+    , null //roNodeGroupRepository
+    , null //techniqueRepository
+    , null //queryParser
+    , null //userPropertyService
+    , null //workflowService
+  )
```
fanf: Something is rotten in the Realm of dependency injection.

fanf: not that we can do much with current set up.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
VinceMacBuche: I focused on the "By Node" since we want this at least to work. That's a mistake and we should be able to build all DataSourceType with apply

rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApi.scala
VinceMacBuche: Alright, can we modify it in first version ?

rudder-web/src/test/scala/com/normation/rudder/web/rest/RestStatusTest.scala
VinceMacBuche: Done!

VinceMacBuche:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -39,33 +39,57 @@ package com.normation.rudder.datasources
 
 import org.joda.time.DateTime
 import org.joda.time.Seconds
+import net.liftweb.common._
 
 sealed trait DataSourceType {
   def name : String
```
fanf: And (because that's the last comment ;) I believe we also need to expose a "datasource update" timeout, which is a global timeout for the update (can be infinite) 
IE, for ex "I don't want the task to ever take more than 5 minutes, because it's a problem on my set up that need correction" or "I don't care, just get the data even if it means wainting 20 minutes on each generation".

fanf: (and so, there is no sane value we can choose for the user)

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSource.scala
```diff
@@ -39,33 +39,73 @@ package com.normation.rudder.datasources
...
+  , onNewNode    : Boolean
 )
+
+final case class DataSource (
+    id            : DataSourceId
+  , name          : DataSourceName
+  , sourceType    : DataSourceType
+  , runParam      : DataSourceRunParameters
+  , description   : String
+  , lastUpdate    : Option[DateTime]
```
VinceMacBuche: Still miss to do a "status" object here

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -0,0 +1,51 @@
...
+*************************************************************************************
+*/
+
+package com.normation.rudder.datasources
+
+import net.liftweb.common.Box
+
+trait DataSourceRepository {
+
+  def getAll : Box[Map[DataSourceId,DataSource]]
```
VinceMacBuche: Id is the key and cannot be changed, so i need to change all api method to be based on id more than name ?

VinceMacBuche: I think so :)

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestDataSerializer.scala
```diff
@@ -586,4 +590,36 @@ case class RestDataSerializerImpl (
...
+    ~ ( "id"        -> source.id.value  )
+    ~ ( "description" -> source.description )
+    ~ ( "type" -> (
+        ( "name" -> source.sourceType.name )
+        ~ { source.sourceType match {
+          case HttpDataSourceType(url,headers,method,path,mode,timeOut) =>
+            ( ( "url"        -> url     )
+            ~ ( "headers"    -> headers )
+            ~ ( "path"       -> path    )
+            ~ ( "requestTimeout"  -> timeOut.length )
```
fanf: length is not what we want: 
- it is the number of Unit (but not the unit)
- it is not defined for not finite duration. 

So either we use here a ".toStrign", which seems to be consistant with the unapply of the different duration (finite or not), and it means that we will have things like "102 ms" and "Inf" in the JSON, or we always transform to something (ms, likelly), and have a norm for infinite (Long.[Min|Max]Value, most likelly), and we have some logic to handle these special number in both way.

rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApi.scala
```diff
@@ -0,0 +1,184 @@
...
+import com.normation.rudder.datasources.OneRequestAllNodes
+import com.normation.rudder.datasources.OneRequestByNode
+import org.joda.time.Seconds
+import scala.concurrent.duration.Duration
+
+trait DataSourceApi extends RestAPI {
+  val kind = "datasources"
+}
+/*
+sealed trait RestRequestMode[T <: HttpRequestMode]{
```
fanf: no need to keep that

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -897,11 +1013,89 @@ case class RestExtractorService (
...
+                  )
+                }
+            }
+          }
+
+          for {
+            url <- extractOneValueJson(obj, "url")(boxedIdentity)
+            path <- extractOneValueJson(obj, "path")(boxedIdentity)
+            method <- extractOneValueJson(obj, "requestMethod")(boxedIdentity)
+            timeout <- extractJsonBigInt(obj, "requestTimeout")(d => tryo{Duration(d.toLong, TimeUnit.SECONDS )})
```
fanf: here, it will have to be 
```
extractOneValueJson(obj, "requestTimeout")(d => tryo(Duration(d)) )
```

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1401, opened 2016-12-05
Fixes #9701: Store data sources in backend

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9701

This PR consists in two things:

Adding a backend for datasources, easy part

Having one json serialization that will work both for backend and rest api. Most complicated part was that it has not the same needs (backend data must be complete, rest API may be partial)
I used some Monad magic from Scalaz (but I think i need an implicit somewere to make it much more readable) 

Two implementation on json serialization, one with monad Id (backend, fails if a data is missing), one with Option (allow to have partial data)

enjoy!

### DIFF
--- a/rudder-core/src/main/resources/reportsSchema.sql
+++ b/rudder-core/src/main/resources/reportsSchema.sql
@@ -456,6 +456,16 @@ start_minute int,
 startTime timestamp with time zone default now(),
 endTime timestamp with time zone
 );
+
+-- Create the table for the node configuration
+CREATE TABLE dataSources (
+  id            text PRIMARY KEY NOT NULL CHECK (id <> '')  
+
+-- data source properties and status are valid json, until we can use postgres 9.2 keep text type (more details in configuration details)
+
+, properties text NOT NULL CHECK (properties <> '' )
+);
+
 /*
  *************************************************************************************
  * end
--- a/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
@@ -51,6 +51,14 @@ import com.normation.eventlog.ModificationId
 import com.normation.rudder.domain.eventlog._
 import scala.concurrent.duration._
 import net.liftweb.common.Failure
+import scalaz.concurrent.Task
+import doobie.imports._
+import doobie.contrib.postgresql.pgtypes._
+import net.liftweb.common.Loggable
+import com.normation.rudder.db.Doobie
+import com.normation.rudder.db.Doobie._
+import scalaz.{Failure => _, _}, Scalaz._
+import net.liftweb.common.Empty
 
 final case class PartialNodeUpdate(
     nodes        : Map[NodeId, NodeInfo] //the node to update
@@ -66,7 +74,7 @@ trait DataSourceRepository {
 
   def save(source : DataSource) : Box[DataSource]
 
-  def delete(id : DataSourceId) : Box[DataSource]
+  def delete(id : DataSourceId) : Box[DataSourceId]
 }
 
 /*
@@ -111,14 +119,9 @@ class MemoryDataSourceRepository extends DataSourceRepository {
     Full(source)
   }
 
-  def delete(id : DataSourceId) : Box[DataSource] = synchronized {
-    sources.get(id) match {
-      case Some(source) =>
-        sources = sources - (id)
-        Full(source)
-      case None =>
-        Failure(s"Data source '${id}' does not exists, and thus can't be deleted")
-    }
+  def delete(id : DataSourceId) : Box[DataSourceId] = synchronized {
+     sources = sources - (id)
+     Full(id)
   }
 }
 /**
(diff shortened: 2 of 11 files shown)


VinceMacBuche:
PR rebased

VinceMacBuche:
PR rebased

rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -456,6 +456,18 @@ start_minute int,
...
 );
+
+-- Create the table for the node configuration
+CREATE TABLE dataSources (
+  id            text NOT NULL CHECK (id <> '')  
+
+-- data source properties and status are valid json, until we can use postgres 9.2 keep text type (more details in configuration details)
+
+, properties     text 
+, status     text
```
fanf: I believe they must not be empty, else it can't be changed to json latter on. So, nullable, but with a constrain for empty string

fanf:
Nice!

fanf:
It will need the added getAllDataSources (without statuse). And I'm not a fan a using lift auto sterilization, but it's good enough for now. A little test to check update of status of only one node and it can be merge.

rudder-core/src/main/resources/reportsSchema.sql
VinceMacBuche: Ok, I didn't know about that constraint, You added the constraint on configuration i'll keep the same

VinceMacBuche:
This PR is not finished, lift serialization/unserialization does not work ( does not serialize datetime, can't unserialize datasource type without typehints ...) 

I suggest we use the same serialization than Rest API for now, Even if we know that they can change and drift in the future

VinceMacBuche:
(I forgot to rudder-dev wip ... :( )

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

VinceMacBuche:
@fanf @ncharles Enjoy that pull request !!

VinceMacBuche:
ps: it works, need more testing though

VinceMacBuche:
And maybe some names are poor choices

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -275,5 +281,67 @@ class DataSourceRepoImpl(
 
 }
 
+class PGDataSourceRepository(
```
fanf: Could you use DataSourceJdbcRepository to match other repo naming scheme?

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -275,5 +281,67 @@ class DataSourceRepoImpl(
...
+                        case n => throw new RuntimeException(s"Expected 0 or 1 change, not ${n} for ${source.id}")
+                      }
+    } yield {
+      result
+    }
+
+    sql.attempt.transact(xa).run match {
+      case \/-(x)  => Full(source)
+      case -\/(ex) => Failure(s"could not update lastId from database, cause is: ${ex.getMessage}", Full(ex), Empty)
+    }
```
fanf: There is an auto transform to avoid that boilerplate, you could use:
```
sql.map(_ => source).attempt.transact(xa).run
```
The implicit def is in Doobie object (```xorToBox```)

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -275,5 +281,67 @@ class DataSourceRepoImpl(
...
+
+    /*
+    (for {
+       _ <- Update[DataSource](insert).run(source).attempt.transact(xa).run
+    } yield {
+      source
+    })*/
```
fanf: can be removed

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -275,5 +281,67 @@ class DataSourceRepoImpl(
...
+  }
 
+  override def delete(sourceId : DataSourceId): Box[DataSource] = {
+    val query = sql"""delete from datasources where id = ${sourceId}"""
+    for {
+      source <-  get(sourceId).flatMap { _ match {
+        case None => Failure("no source")
```
fanf: If I understand correctly, here, trying to delete a source which doesn't exists will lead to an error. I think it should not be the case so that we have an idempotent operation for deletion (ie: deleting a non existing datasource is a success with invariant: "after deletion, the datasource does not exist"

rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
```diff
@@ -179,6 +180,20 @@ object Doobie {
...
 
+  implicit val DataSourceComposite: Composite[RudderDataSource] = {
+    import com.normation.rudder.repository.json.CompleteJson._
+    import com.normation.rudder.datasources.DataSourceJsonSerializer._
+    import net.liftweb.json.compactRender
+    import net.liftweb.json.parse
+    Composite[(DataSourceId,String)].xmap(
+        tuple => extractDataSource(tuple._1,parse(tuple._2)) match {
+          case Full(s) => s
+          case eb  => throw new RuntimeException("oops")
```
fanf: Could you change the message to log what is the cause of the error ("error when deserialising blabla") and perhaps even the faulty json (not sure if it is already logged in extractDataSource)

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

PR #1404, opened 2016-12-13
Fixes #9802: Error when displaying Rules when a rule is applying a directive that does not exists anymore

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9802

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
@@ -542,7 +542,10 @@ class RuleGrid(
     val t0 = System.currentTimeMillis
 
     val nodes = groupsLib.getNodeIds(line.rule.targets, nodesInfo)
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -542,7 +542,7 @@ class RuleGrid(
     val t0 = System.currentTimeMillis
 
     val nodes = groupsLib.getNodeIds(line.rule.targets, nodesInfo)
-    val (policyMode,explanation) = ComputePolicyMode.ruleMode(globalMode, line.rule.directiveIds.map(directiveLib.allDirectives(_)).map(_._2))
+    val (policyMode,explanation) = ComputePolicyMode.ruleMode(globalMode, line.rule.directiveIds.map(directiveLib.allDirectives.get(_)).flatMap(_.map(_._2)))
```
fanf: Could you add a comment sating what is the will here? 
Something like:
"We don't fail on missing directive. Most likelly a deleted directive link wasn't updated correctly. We should assure consistency on all rudder config elem, but until that (big) change, it seems to be the most resilient solution".

VinceMacBuche:
Commit modified

PR #1408, opened 2016-12-26
Fixes #9853: Incomplete logging in debug_timming

ncharles:
https://www.rudder-project.org/redmine/issues/9853

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -48,6 +48,7 @@ import scala.collection.mutable.Buffer
 import com.normation.rudder.domain.logger.ReportLogger
 import com.normation.rudder.domain.logger.ComplianceDebugLogger
 import com.normation.rudder.domain.logger.ComplianceDebugLogger._
+import com.normation.rudder.domain.logger.TimingDebugLogger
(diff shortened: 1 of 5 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -761,6 +771,8 @@ object ExecutionBatch extends Loggable {
 
                                      ((directive.directiveId, component.componentName), (policyMode, missingReportStatus, component))
                                    }).toMap
+                                   val t2 = System.currentTimeMillis
+                                   TimingDebugLogger.trace(s"Compliance: mergeCompareByRule - prepare data: ${t2-t1}ms")
```
ncharles: this is too verbose

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -947,27 +976,30 @@ object ExecutionBatch extends Loggable {
    */
   private[reports] def checkExpectedComponentWithReports(
       expectedComponent: ComponentExpectedReport
-    , filteredReports  : Seq[Reports]
+    , filteredReports  : Seq[ResultReports]
```
ncharles: by design, we pass only resultreports to it

PR #1410, opened 2016-12-28
Fixes #9734: Store tags on Directives/Rules

ncharles:


ncharles:
Commit modified

ncharles:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/domain/policies/Tags.scala
```diff
@@ -0,0 +1,90 @@
...
+final case class TagName( name : String )
+
+final case class TagValue( value : String )
+
+final case class Tag( tagName : TagName, tagValue : TagValue )
+
+/**
+ * We can have multiple Tags with same name - unicity is really on TagName + TagValue
+ */
+final case class Tags(
```
VinceMacBuche: Do we need that class ? what is the advantage of having a 'Tags' instead of a 'Set[Tags]' directly in our objects ?

VinceMacBuche:
Apart the question on the Tags type, that can be merged !

rudder-core/src/main/scala/com/normation/rudder/domain/policies/Tags.scala
ncharles: I don't know what would be the advantage. It is more verbose, and the only possible advantage might be that one day Tags would be different, and so it wouild encapsulate change; but this is dubious
Should I drop it ?

VinceMacBuche: I would have removed it yes!

ncharles: Actually, having the Tags class let me define in one location only the rules on unicities - so if we want to change in the future the rules on tags, it'll be easier
so i favour keeping the Tags class

PR #1409, opened 2016-12-26
Fixes #9857: ReportingServiceImpl moves around a lot of data and is fairly slow

ncharles:
https://www.rudder-project.org/redmine/issues/9857

PR #1411, opened 2016-12-28
Fixes #9864: getNodeStatusReports is called for each node, and is slow to compute PolicyMode Inconsistency

ncharles:
https://www.rudder-project.org/redmine/issues/9864


ncharles:
This must be merged after https://github.com/Normation/rudder/pull/1409 to have any benefits at all

ncharles:
Gain is not extraordinary, but is significant still:
- getNodeStatusReports is about 50% faster
- buildNodeStatusReports is about 20% faster
- batch Store Run Agent is a little bit faster (not sure if significant at all)

This should be further tested to ensure we gain something

ncharles:
This is not relevant, as https://github.com/Normation/rudder/pull/1413/ fixes it in a more meaningful way

PR #1412, opened 2016-12-28
Fixes #9865: Move hooks to /opt/rudder/etc/

amousset:
https://www.rudder-project.org/redmine/issues/9865


rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
```diff
@@ -212,7 +212,7 @@ class WriteSystemTechniqueTest extends Specification with Loggable with ContentM
     , logNodeConfig
     , prepareTemplateVariable
     , new FillTemplatesService()
-    , "/var/rudder/hooks.d"
+    , "/opt/rudder/etc/hooks.d"
```
amousset: Really unsure about this one, no idea what it does.

PR #1413, opened 2016-12-29
Fixes #9869: Compliance on all reports is always computed, even when not necessary

ncharles:
https://www.rudder-project.org/redmine/issues/9869


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1416, opened 2017-01-02
Fixes #9879: Update datasource status when a data source data are updated

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9879

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApiService.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApiService.scala
@@ -58,7 +58,7 @@ class DataSourceApiService(
   def getSources() : Box[JValue] = {
     for {
       sources <- dataSourceRepo.getAll
-      data = sources.values.map(restDataSerializer.serializeDataSource(_))
+      data = sources.values.map(s => restDataSerializer.serializeDataSource(s))
     } yield {
       data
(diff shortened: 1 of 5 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApiService.scala
```diff
@@ -58,7 +58,7 @@ class DataSourceApiService(
   def getSources() : Box[JValue] = {
     for {
       sources <- dataSourceRepo.getAll
-      data = sources.values.map(restDataSerializer.serializeDataSource(_))
+      data = sources.values.map(s => restDataSerializer.serializeDataSource(s._1))
```
fanf: Status may be huge. I believe we need a "getAllDataSources"  without the status part here (and corresponding implementation of course)

fanf:
Cool!

rudder-web/src/main/scala/com/normation/rudder/web/rest/datasource/DataSourceApiService.scala
VinceMacBuche: I'm ok with you making an update

VinceMacBuche:
Commit modified

peckpeck:
This code should go to the plugin repo.
It may serve as a reference in a future implementation.

PR #1418, opened 2017-01-02
Fixes #9884: Typo in root of policy libray tree name

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9884

### DIFF
--- a/rudder-core/src/main/resources/ldap/bootstrap.ldif
+++ b/rudder-core/src/main/resources/ldap/bootstrap.ldif
@@ -163,7 +163,7 @@ isSystem: TRUE
 dn: techniqueCategoryId=Active Techniques,ou=Rudder,cn=rudder-configuration
 objectClass: techniqueCategory
 objectClass: top
-cn: Root of active techniques's library
+cn: Active Techniques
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/resources/ldap/bootstrap.ldif
```diff
@@ -163,7 +163,7 @@ isSystem: TRUE
 dn: techniqueCategoryId=Active Techniques,ou=Rudder,cn=rudder-configuration
 objectClass: techniqueCategory
 objectClass: top
-cn: Root of active techniques's library
+cn: Root of active techniques library
```
jooooooon: Could we take advantage of this change to make it simple "Active Techniques"?

VinceMacBuche:
Commit modified

rudder-core/src/main/resources/ldap/bootstrap.ldif
VinceMacBuche: Simpler, shorter, go let's go (only concern would be about the T of techniqes to be capitalized or not) I guess it should always be capitalized

PR #1424, opened 2017-01-05
Fixes #9899: Change format of time duration in datasource Rest API

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9899


rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -1102,12 +1099,13 @@ case class RestExtractorService (
             }
           }
 
+          def HttpDataSourceParameters (obj : JObject) : Box[HttpDataSourceType] = {
```
ncharles: could you correct indentation on this method ?

VinceMacBuche:
Commit modified

PR #1426, opened 2017-01-06
Fixes #9910: Bugs on datasource initilization and update

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9910

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
@@ -52,7 +52,6 @@ import com.normation.rudder.domain.eventlog._
 import scala.concurrent.duration._
 import net.liftweb.common.Failure
 
-
 final case class PartialNodeUpdate(
     nodes        : Map[NodeId, NodeInfo] //the node to update
   , policyServers: Map[NodeId, NodeInfo] //there policy servers
@@ -92,6 +91,11 @@ trait DataSourceUpdateCallbacks {
    * all resources on them.
    */
   def startAll(): Unit
+
+  /*
+   * Define all datasource scheduler from data sources in backend
+   */
+  def initialize(): Unit
 }
 
 class MemoryDataSourceRepository extends DataSourceRepository {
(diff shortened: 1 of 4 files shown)


rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
```diff
@@ -397,6 +397,11 @@ object RudderConfig extends Loggable {
   val historizationRepository : HistorizationRepository =  historizationJdbcRepository
   val roApiAccountRepository : RoApiAccountRepository = roLDAPApiAccountRepository
   val woApiAccountRepository : WoApiAccountRepository = woLDAPApiAccountRepository
+  lazy val dataSourceRepository = new DataSourceRepoImpl(
```
ncharles: this lazy val is in the middle of non lazy val - could you move it with others lazy val ?

rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceRepository.scala
```diff
@@ -133,6 +133,16 @@ class DataSourceRepoImpl(
 
   private[this] var datasources = Map[DataSourceId, DataSourceScheduler]()
 
+  // Initialize data sources scheduler, with all sources present in backend
+  getAll match {
```
fanf: I think it should not be done at class instanciation (but we have to document that the start is not automatic). Not having it hiden in the class instanciation allows to split:

- error from instanciation (most likelly code problem, and hard to debug: "no such class error" most of the time) from error due to boot (most likelly connection pb)
- let the user choose when to trigger datasource fetch (in a future version of Rudder, we may want to boot completly and as fast as possible before starting datasources schedules)

fanf: And so, in Rudder the solution is to use a ```BootstrapChecks``` (that will give nice init log for free)

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1429, opened 2017-01-09
Fixes #9835: When we receive reports from node with old reports, caches are not used for compliance computation

ncharles:
https://www.rudder-project.org/redmine/issues/9835


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1431, opened 2017-01-10
Fixes #9931: correct tests in UpdateHttpDatasetTest

fanf:
http://www.rudder-project.org/redmine/issues/9931


fanf:
Self merging that one as it is only tests, and block the nightly build.

PR #1433, opened 2017-01-10
Fixes #9934: Clean tags data model 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9934

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Tags.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Tags.scala
@@ -43,16 +43,23 @@ import com.normation.rudder.repository.json.JsonExctractorUtils
 
 /**
  * Tags that apply on Rules and Directives
- * We do not warranty unicity of tags name, only of tagsname + tagvalue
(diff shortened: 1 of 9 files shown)


VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/domain/policies/Tags.scala
```diff
@@ -44,11 +44,15 @@ import com.normation.utils.Control._
  * Tags that apply on Rules and Directives
  * We do not warranty unicity of tags name, only of tagsname + tagvalue
  */
-final case class TagName( name : String )
 
-final case class TagValue( value : String )
+object Tag {
+  final implicit class TagName ( val value : String )
```
fanf: I think we really need them to be implicit *case* class, else the equals/hashcode won't be what we want at all

VinceMacBuche:
Commit modified

VinceMacBuche:
@fanf case class back, going back to implicits;

VinceMacBuche:
PR rebased

PR #1443, opened 2017-01-11
Fixes #9958: Nodeslist file is empty on a root server with one node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9958


rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -393,6 +393,10 @@ class SystemVariableServiceImpl(
         , varManagedNodesKey
         , varReportsDBname
         , varReportsDBuser
+        , varSubNodesName
+        , varSubNodesId
+        , varSubNodesServer
+        , varSubNodesKeyhash
```
ncharles: good catch :)

PR #1446, opened 2017-01-13
Fixes #9995: Forbid modification of properties set by datasource

fanf:
https://www.rudder-project.org/redmine/issues/9995


rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -67,23 +67,34 @@ import java.io.PipedOutputStream
 import com.zaxxer.nuprocess.NuProcessBuilder
 import java.util.Arrays
 import com.normation.rudder.domain.nodes.NodeInfo
+import com.normation.rudder.datasources.DataSourceRepository
 
 class NodeApiService8 (
```
VinceMacBuche: Needs to be done in NodeApiService5, maybe we can always use nodeAPI8Service and remove NodeApi5Service

fanf:
PR rebased

PR #1458, opened 2017-01-24
Fixes #10064: Remove dependecies inherited from parent-pom dependencyManagement in rudder-core

fanf:
https://www.rudder-project.org/redmine/issues/10064


pom.xml
```diff
@@ -58,9 +58,7 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
     <!-- fill-in rule templates -->
   </modules>
 
-  <dependencies>
-
-  </dependencies>
```
ncharles: this was really useful :)

PR #1460, opened 2017-01-24
Fixes #10017: Improve data sources UI

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10017

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
@@ -59,7 +59,7 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
   $scope.datasources = [];
   // Selected data source
   $scope.selectedDatasource;
-
+  $scope.treeId = "#datasources-tree";
   /* Get data sources */
   $scope.getDataSources = function(){
     return $http.get(contextPath + '/secure/api/latest/datasources').then(function(response){
@@ -68,12 +68,14 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
       $scope.reverse = false;
       $scope.datasources =  orderBy(res, $scope.propertyName, $scope.reverse);
       for(var i=0 ; i<$scope.datasources.length ; i++){
-        $scope.datasources[i].newHeader = {"key":"","value":""};
+        $scope.datasources[i].newHeader = {"name":"","value":""};
+        $scope.datasources[i].newParam  = {"name":"","value":""};
         $scope.datasources[i].modifiedTimes = {
           'schedule'       : timeConvert($scope.datasources[i].runParameters.schedule.duration)
         , 'updateTimeout'  : timeConvert($scope.datasources[i].updateTimeout)
         , 'requestTimeout' : timeConvert($scope.datasources[i].type.parameters.requestTimeout)
-        }
+        };
+        //This will need to be changed when Bug #10554 will be fixed
       }
     });
   }
@@ -91,9 +93,6 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
     }
     return false;
   }
-  $scope.updateKey = function(header, index, key){
-    renameKey(header, Object.keys(header)[index], key)
-  }
   // DATASOURCES
   $scope.createNewDatasource = function(){
     if($scope.forms.datasourceForm){
@@ -106,13 +105,14 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
       "type"        :{
         "name"       :"http",
         "parameters" :{
-          "url"     :"",
-          "headers" :{},
+          "checkSsl":true,
+          "headers" :[],
+          "params" :[],
           "path"    :"",
-          "checkSsl":false,
-          "requestTimeout":5,
+          "requestTimeout":30,
           "requestMethod":"GET",
-          "requestMode":{"name":"byNode"}
+          "requestMode":{"name":"byNode"},
+          "url"     :""
         }
 	  },
 	  "runParameters":{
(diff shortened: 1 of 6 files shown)


RaphaelGauthier:
Commit modified

RaphaelGauthier:
PR rebased

rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
```diff
@@ -127,13 +132,47 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
...
+        r = r.replace(new RegExp("ç", 'g'),"c");
+        r = r.replace(new RegExp("[èéêë]", 'g'),"e");
+        r = r.replace(new RegExp("[ìíîï]", 'g'),"i");
+        r = r.replace(new RegExp("ñ", 'g'),"n");
+        r = r.replace(new RegExp("[òóôõö]", 'g'),"o");
+        r = r.replace(new RegExp("œ", 'g'),"oe");
+        r = r.replace(new RegExp("[ùúûü]", 'g'),"u");
+        r = r.replace(new RegExp("[ýÿ]", 'g'),"y");
+        r = r.replace(new RegExp("\\W", 'g'),"_");
+        $scope.selectedDatasource.id = r;
```
fanf: Isn't their a function in JS to map UTF-8 space to ascii? (+take care of ascii space, at least this one is enumerable). 
Because, we could do nothing, it would be least surprising than here (for ex, what about the "ğ" ?
And why limit key space? It could be full UTF-8 chars (trimed), no?

fanf: OK, js is just a ridiculus language. But that requirement should appear in a lot of place in our code, no? 
If so, having an external function like: https://github.com/backbone-paginator/backbone.paginator/blob/a579796a30e583c4dfa09e0a86e4abd21e0b5b56/plugins/diacritic.js would be better.

RaphaelGauthier: I don't find any Javascript native function to map UTF-8 space to ASCII.. But maybe we can use a more complete solution like this ?
https://github.com/Kepro/angular-remove-diacritics

fanf: @VinceMacBuche what do you think?

RaphaelGauthier: Oops I didn't see your previous answer. The Angular Service I linked is precisely based on this function :)

RaphaelGauthier: So I think is the best solution we have in our case...

fanf: Some more thinking: there is no other keys in Rudder that are not UTF-8 ; 
Other node properties can be UTF-8 ;
It is not enforced by API (so that we can add data sources with utf-8 id in all case);.
So I'm really not sure that we want to enforce that only in the web UI (I'm pretty sure of the opposite). 
@jooooooon @VinceMacBuche what do you think?

VinceMacBuche: I agree with @fanf, did some research too, and found the same library ... we should not do ourselves something someone else already did !!

VinceMacBuche:
apart utf8 remparks this is all good

rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
VinceMacBuche: if some validation needs to be done on keys/id we need to do them in backend (prevent invalid characters, being valid identifier ...)

checking on UI side gives a better UX but validating on backend side seems more important

RaphaelGauthier:
PR rebased

RaphaelGauthier:
Commit modified

rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
VinceMacBuche: Policy generation fails when you use a property based on utf8 characters 


error_properties_utf8


Indeed cfengine arrays does only support cfengine valid ([a-zA-Z_]+) keys, we should valideate properties keys likweise

VinceMacBuche: Replacing  invalid key characters (not "[\-_a-zA-Z0-9]")  by _, is a better solution

RaphaelGauthier:
Commit modified

rudder-web/src/main/webapp/secure/administration/dataSourceManagement.html
```diff
@@ -23,168 +25,209 @@ <h4 class="no-margin">
...
             <span ng-if="selectedDatasource.name.length>0">{{selectedDatasource.name}}</span>
             <span ng-if="!selectedDatasource.name && selectedDatasource.isNew">New data source</span>
             <span ng-if="!selectedDatasource.name && !selectedDatasource.isNew" style="font-style:italic; color:#777;">Unnamed</span>
           </div>
+          <!-- <h3 class="section-title">Data source information</h3> -->
           <form class="bloc-body show-details" name="forms.datasourceForm">
             <div class="form-group">
               <label for="name">Name</label>
-              <input type="text" class="form-control input-sm" id="name" ng-value="selectedDatasource.name" name="datasourceName" required ng-model="selectedDatasource.name" >
+              <input type="text" class="form-control" id="name" ng-value="selectedDatasource.name" name="datasourceName" required ng-model="selectedDatasource.name" ng-change="updateKeyName(selectedDatasource.name)">
```
VinceMacBuche: You may want this to be trimmed (using ng-trim)

RaphaelGauthier: By default, this option is activated. ng-trim allows us to set it to false if necessary but it is not our case :)

VinceMacBuche: Ok :D

RaphaelGauthier:
PR rebased

rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
```diff
@@ -125,15 +125,36 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
...
 	  "updateTimeout":5,
 	  "enabled":false,
-	  "newHeader":{"key":"","value":""},
+	  "newHeader":{"name":"","value":""},
 	  "modifiedTimes":{
-	    "schedule":{"day":0,"hour":0,"minute":5},
-	    "updateTimeout":{"day":0,"hour":0,"minute":5},
-	    "requestTimeout":{"day":0,"hour":0,"minute":5}
+	    "schedule":{"second":0,"minute":5},
+	    "updateTimeout":{"second":0,"minute":5},
```
fanf: Should be 30s

rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
```diff
@@ -125,15 +125,36 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
...
+	    "requestTimeout":{"second":0,"minute":5}
```
fanf: Same here: 30s

rudder-web/src/main/webapp/javascript/rudder/angular/dataSource.js
```diff
@@ -125,15 +125,36 @@ app.controller("datasourceCtrl", ['$scope', '$timeout', 'orderByFilter','$http',
 	  },
 	  "updateTimeout":5,
 	  "enabled":false,
-	  "newHeader":{"key":"","value":""},
+	  "newHeader":{"name":"","value":""},
 	  "modifiedTimes":{
-	    "schedule":{"day":0,"hour":0,"minute":5},
```
fanf: Here, by default, every hour is OK.

fanf: 6h, even

RaphaelGauthier:
Commit modified

PR #1461, opened 2017-01-24
Fixes #9714: Remote run API should use relay API

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9714

### DIFF
--- a/rudder-web/src/main/resources/configuration.properties.sample
+++ b/rudder-web/src/main/resources/configuration.properties.sample
@@ -68,6 +68,11 @@ rudder.dir.lock=/var/rudder/lock/
 #
 rudder.endpoint.cmdb=http://localhost:8080/endpoint/upload/
 
+#
+# Location of the relay api used by rudder webapp
+# It's the base url of relay api, Rudder will manage to call the correct url from that base
+#
+rudder.server.relay.api=https://localhost/rudder/relay-api
 
 #
 # The port used by the rsyslog server on the Rudder root server. 
--- a/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
@@ -68,13 +68,15 @@ import com.zaxxer.nuprocess.NuProcessBuilder
 import java.util.Arrays
 import com.normation.rudder.domain.nodes.NodeInfo
 import com.normation.rudder.datasources.DataSourceRepository
+import scalaj.http.Http
 
 class NodeApiService8 (
-    nodeRepository : WoNodeRepository
-  , nodeInfoService: NodeInfoService
-  , uuidGen        : StringUuidGenerator
-  , datasourceRepo : DataSourceRepository
-  , asyncRegenerate: AsyncDeploymentAgent
+    nodeRepository  : WoNodeRepository
+  , nodeInfoService : NodeInfoService
+  , uuidGen         : StringUuidGenerator
+  , datasourceRepo  : DataSourceRepository
+  , asyncRegenerate : AsyncDeploymentAgent
+  , relayApiEndpoint: String
 ) extends Loggable {
 
   def updateRestNode(nodeId: NodeId, restNode: RestNode, actor : EventActor, reason : Option[String]) : Box[Node] = {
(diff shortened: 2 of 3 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -138,54 +128,32 @@ class NodeApiService8 (
...
-        case None => // Nothing to do for now
-      }
-    }
-
-    // On exit will be called at the end of the process, or if any error/exception occurs, so we can close our outputstream here
-    override def onExit( statusCode : Int) = {
-      out.map(_.close())
-    }
+  def remoteRunRequest(nodeId: NodeId, classes : List[String], keepOutput : Boolean, asynchronous : Boolean) = {
+    val url = s"https://localhost/rudder/relay-api/remote-run/nodes/${nodeId.value}"
```
fanf: This should be configured. I'm sure there will be case where relay-api won't be on the same machine than the web app (or behind an other root)

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -138,54 +128,32 @@ class NodeApiService8 (
...
-             case e : Throwable =>
-               out.close()
-               Failure(s"An error occured when applying policy on Node '${node.id.value}', cause is: ${e.getMessage}")
-           }
-    } yield {
-      runResponse(in)
-    }
+       httpResponse <- tryo { request.execute { runResponse } }
+       response <- if (httpResponse.isSuccess ) {
+         Full(httpResponse.body)
```
fanf: I don't think this is streamed - is it?

VinceMacBuche: request.execute allow to get data from the response as a stream ( inputStrem -> T ) we transform it using "runResponse" to an "OutputStream => Unit" which will read from the inputstream to put into the api output stream response and stream it


so yeah, body is not the "response data" but "how you want to get the response data" ... and it's quite misleading

fanf: ok, better saw it without the diff view

VinceMacBuche:
Commit modified

rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -68,6 +68,10 @@ rudder.dir.lock=/var/rudder/lock/
 #
 rudder.endpoint.cmdb=http://localhost:8080/endpoint/upload/
 
+#
+# Location of the relay api used by rudder webapp
+#
+rudder.server.relay.api=localhost
```
fanf: Here, we need to parametrize the whole address - nothing tell us what would be configured in multi-server in the future. 
I don't see any simple way to have: ```https://localhost/rudder/relay-api/remote-run/nodes/${nodeId}```, but we can use: ```https://localhost/rudder/relay-api/remote-run/nodes/%s" with a comment stating that "%s will be replaced by the node id".

rudder-web/src/main/scala/com/normation/rudder/web/rest/node/NodeAPIService8.scala
```diff
@@ -138,54 +129,32 @@ class NodeApiService8 (
...
-        case None => // Nothing to do for now
-      }
-    }
-
-    // On exit will be called at the end of the process, or if any error/exception occurs, so we can close our outputstream here
-    override def onExit( statusCode : Int) = {
-      out.map(_.close())
-    }
+  def remoteRunRequest(nodeId: NodeId, classes : List[String], keepOutput : Boolean, asynchronous : Boolean) = {
+    val url = s"https://${relayApiEndpoint}/rudder/relay-api/remote-run/nodes/${nodeId.value}"
```
fanf: and so here just use ```relayApiEndpoint.format(nodeId.value)```

rudder-web/src/main/resources/configuration.properties.sample
VinceMacBuche: Why not relay api endoit and just : https://localhost/rudder/relay-api

fanf: Yes.

VinceMacBuche:
Commit modified

PR #1463, opened 2017-01-25
Fixes #10054: Change headers AND params in datasource api to be an array

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10054


rudder-core/src/main/scala/com/normation/rudder/datasources/DataSourceSerializer.scala
```diff
@@ -61,8 +61,12 @@ object DataSourceJsonSerializer{
             source.sourceType match {
               case DataSourceType.HTTP(url,headers,method,params,checkSsl,path,mode,timeOut) =>
                 ( ( "url"            -> url     )
-                ~ ( "headers"        -> headers )
-                ~ ( "params"         -> params  )
+                ~ ( "headers"        -> headers.map{
+                  case (name,value) => ("name" -> name) ~ ("value" -> value)
```
fanf: Why name/value in place of key/value as written in the related ticket?

PR #1470, opened 2017-01-31
Fixes #10097: Configured Hicari-CP is not the java8 version

fanf:
https://www.rudder-project.org/redmine/issues/10097


rudder-core/pom.xml
```diff
@@ -87,7 +87,7 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
     <dependency>
         <groupId>org.bouncycastle</groupId>
         <artifactId>bcpkix-jdk15on</artifactId>
-        <version>1.55</version>
+        <version>${bcpkix-jdk15on-version}</version>
```
ncharles: where is this defined ?

ncharles: ha, in parent-pom PR

PR #1475, opened 2017-02-03
Fixes #10127: Switch to Scala 2.12 / Lift 3.0 

fanf:
https://www.rudder-project.org/redmine/issues/10127

This seems huge, because there is a lot of files touched, but change are splitted in these categories:

- changing pretty(render( and compact(render( to prettyRender / comptactRender
- changing template tag name fromr ```some:thingUsed``` to ```some-thingused```, because html5 parser, 
- using a replacement for the no-more-existing chooseTemplate: ChooseTemplate (which also remove quite some redondant code)
- switching from ```bind("prefix", xml, elem1 -> res)``` to ```(prefix-elem1 #> res)(xml)```
- lift ajax call are now under ```http://.../lift/ajax``` in place of ```http://.../liftAjax```  
- method ".is" is removed in form field, only ".get" remains
- our json render method was based on deprecated objects that are now gone. Either we go with the new default one, or we will need to do some more transformation. 

Less immediate: Lift 3.0 adds security rules against xss. Our code is not very friendly to them, and for now they are disabled (see code in Boot.scala). 

Also, some Json code need adaptation because JField isn't a JValue anymore (perhaps it was the code the least immediate - some thinking was needed).

Finaly, there was some variance pb with scalaz monoid instance in JsonExtractor.


fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
```diff
@@ -218,18 +222,18 @@ class WriteSystemTechniquesTest extends Specification with Loggable with Content
...
-      FileUtils.deleteDirectory(abstractRoot)
-    }
-  }
-  override def after: Unit = {
-    if(System.getProperty("tests.clean.tmp") != "false") {
-      logger.info("Deleting directory " + rootGeneratedPromisesDir.getAbsoluteFile)
-      FileUtils.deleteDirectory(rootGeneratedPromisesDir)
-    }
-  }
+//  override def afterAll(): Unit = {
```
VinceMacBuche: Why do you keep it commented ?

VinceMacBuche: Not a problem, I'll merge it we will remove it in another PR

VinceMacBuche: OK I read the rest of the file :) Nice way to handle a failing test !!

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateCategoryOrGroupPopup.scala
```diff
@@ -135,27 +133,34 @@ class CreateCategoryOrGroupPopup(
...
+            <div>
+              <h4 class="col-lg-12 col-sm-12 col-xs-12 audit-title">Change Audit Log</h4>
+              {f.toForm_!}
+            </div>
+            } }
+        & "item-cancel" #> ( SHtml.ajaxButton("Cancel", { () => closePopup() }) % ("tabindex","6") % ("class","btn btn-default") )
+        & "item-save" #> ( SHtml.ajaxSubmit("Create", onSubmit _) % ("id","createCOGSaveButton") % ("tabindex","5") % ("class","btn btn-success") )
+    )(popupTemplate)
+  ) ++ Script(OnLoad(initJs))
+  println("*"*10)
```
ncharles: nasty println

VinceMacBuche: Nice catch

PR #1477, opened 2017-02-06
Fixes #10145: We need to produce tests-jar for plugin to depend on

fanf:
https://www.rudder-project.org/redmine/issues/10145


fanf:
PR rebased

PR #1495, opened 2017-02-13
Fixes #10188: Spurious println introduced in parent ticket

ncharles:
https://www.rudder-project.org/redmine/issues/10188


fanf:
OK, merging this PR

PR #1510, opened 2017-02-17
Fixes #9859: Broken directive migration in Rudder 4.0.2

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/9859

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
@@ -178,7 +178,15 @@ class DirectiveEditForm(
   val displayDeprecationWarning = technique.deprecrationInfo match {
     case Some(info) =>
       ( "#deprecation-message *" #> info.message &
-        "#migrate-button *" #> migrateButton(fullActiveTechnique.techniques.keys.max,"Migrate now!","deprecation-migration")
+        "#migrate-button *" #> {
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
```diff
@@ -178,7 +178,15 @@ class DirectiveEditForm(
   val displayDeprecationWarning = technique.deprecrationInfo match {
     case Some(info) =>
       ( "#deprecation-message *" #> info.message &
-        "#migrate-button *" #> migrateButton(fullActiveTechnique.techniques.keys.max,"Migrate now!","deprecation-migration")
+        "#migrate-button *" #> {
+            val lastTechniqueVersion = fullActiveTechnique.techniques.keys.max
```
fanf: can't it fail with no such element ?

VinceMacBuche: Is it possible that a technique may have no technique version ? 

Here a directive has already been selected meaning that we have already built a directive based on a existing version of the technique.

I guess it's impossible in a normal usage of Rudder, but i think it's still possible by manipulating ldap to have no version and still one in the technique

VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #1516, opened 2017-02-21
Fixes #10245: Align all graphs in dashboard

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10245


fanf:
I'm closing that one until we have a better understanding about what is needed.

PR #1519, opened 2017-02-21
Fixes #10107: version is not properly set in web interface

fanf:
https://www.rudder-project.org/redmine/issues/10107


fanf:
PR rebased

PR #1520, opened 2017-02-21
Fixes #8171: Change graph lib to more efficent one

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/8171


rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -327,8 +327,8 @@ object NodeChanges {
     }
 
     JsObj(
-        ("x" -> JsArray(data.map(a => Str(a._1))))
-      , ("y" -> JsArray(data.map(a => Num(a._2))))
+        ("labels" -> JsArray(data.map(a => a._1)))
+      , ("values" -> JsArray(data.map(a => Num(a._2+Random.nextInt(500)))))
```
ncharles: Why the random ? Isn't it a leftover ?

VinceMacBuche:
Commit modified

fanf:
There is also the following problems, but I don't know where they come from exactly:
- on the home page, the bottom graphes are bigger than before the change, and they don't fit the screen even in HD resolution, 
- when the "display change graphe" option is set to "do not display", we get infiny loading image for all compliance bar, and no numbers in place of graphe.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf:
I'm still getting the wheel in place of compliance bar and no number ofr changes when the "display graph" setting option is disable.

VinceMacBuche:
Commit modified

VinceMacBuche:
@fanf also remarked that dashboard legends did not have the same behavior

There still some work on this, I'll put my wip here and chnage issue to "in progress", maybe closing this PR too

fanf:
So, the idea is to only do part of the work here, and in a second step add the different call-backs and interaction between graphe and legend. 
Here, we want to *only* have:
- a legend out of the graphe element, so that the size of the legend does NOT change the size of the graphe, 
- aspect/ration matching the ones before the change, 
- the correction in compliance barre when "display graphe" option is set to "don't show".

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1522, opened 2017-02-22
Fixes #10250: When trying to create a new API account that has the same name as an existing acocunt, nothing happen

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10250

### DIFF
--- a/rudder-web/src/main/webapp/secure/administration/apiManagement.html
+++ b/rudder-web/src/main/webapp/secure/administration/apiManagement.html
@@ -39,7 +39,7 @@
 </head>
 
 
-<div class="portlet" ng-app="accountManagement" id="accountManagement" ng-cloak="" >
+<div class="portlet" id="accountManagement" ng-cloak="" >
   <div class="portlet-content">
     <div class="inner-portlet">
       <div class="page-title">Configure allowed accounts</div>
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/webapp/secure/administration/apiManagement.html
```diff
@@ -118,7 +118,15 @@ <h1 id="title" style="color: #000; wrap-word:break-word;">{{formTitle(myNewAccou
                       <div class="simplemodal-bottom">
                         <div class="popupButton" style="width: 98%;border-top: 1px solid #9A9A9A;margin-top: 10px;">
                           <span>
-                            <input type="submit" class="btn-primary" style="float:right;" value="Save" ng-disabled="newAccount.$pristine || newAccount.$invalid" ng-click="checkAndSaveAccount(myNewAccount,myNewAccount.index,newAccount);newAccount.$setPristine();" />
+                            <button
```
VinceMacBuche: Why changing from input to button fix it ? 

We should put the check in the ng-disable i think ... but htat for another time :)

RaphaelGauthier: Oops, this was just a test. I forgot to replace it...

VinceMacBuche: Then i don't understand what is fixing the issue here :(

RaphaelGauthier:
Commit modified

rudder-web/src/main/webapp/secure/administration/apiManagement.html
RaphaelGauthier: The angular app was started twice in a row :
1. with the ng-app directive (<div class="portlet" ng-app="accountManagement" [...]>)
2. with : angular.bootstrap(document.getElementById('accountManagement'), ['accountManagement']);

This caused a JS error which was the origin of this bug

RaphaelGauthier: I just removed the ngApp directive!

RaphaelGauthier:
OK, merging this PR

PR #1524, opened 2017-02-22
Fixes #10266: When trying to create a new API account that has the same name as an existing acocunt, nothing happen (in 4.0)

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10266


RaphaelGauthier:
OK, merging this PR

PR #1528, opened 2017-02-22
Fixes #10259: Log for failed login attempt is not (correctly) reported

fanf:
https://www.rudder-project.org/redmine/issues/10259

### DIFF
--- a/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -88,6 +88,9 @@ import org.springframework.ldap.core.DirContextOperations
 import java.util.Collection
 import org.xml.sax.SAXParseException
 import com.normation.rudder.web.services.UserSessionLogEvent
+import org.springframework.security.web.authentication.SimpleUrlAuthenticationFailureHandler
+import org.springframework.security.web.authentication.AuthenticationFailureHandler
+import org.springframework.security.authentication.BadCredentialsException
 
 /**
  * Spring configuration for user authentication.
@@ -107,6 +110,7 @@ class AppConfigAuth extends Loggable {
 
   ///////////// FOR WEB INTERFACE /////////////
 
+  @Bean def rudderWebAuthenticationFailureHandler: AuthenticationFailureHandler = new RudderUrlAuthenticationFailureHandler("/index.html?login_error=true")
 
   @Bean def demoAuthenticationProvider : AuthenticationProvider = {
     try {
@@ -143,7 +147,7 @@ class AppConfigAuth extends Loggable {
 
   @Bean def restAuthenticationEntryPoint = new AuthenticationEntryPoint() {
     override def commence(request: HttpServletRequest, response: HttpServletResponse, ex: AuthenticationException) : Unit = {
-        response.sendError( HttpServletResponse.SC_UNAUTHORIZED, "Unauthorized" )
+      response.sendError( HttpServletResponse.SC_UNAUTHORIZED, "Unauthorized" )
     }
   }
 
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -166,6 +170,58 @@ class AppConfigAuth extends Loggable {
...
+  override def onAuthenticationFailure(request: HttpServletRequest, response: HttpServletResponse, exception: AuthenticationException): Unit = {
+    LogFailedLogin.warn(exception, request)
+    super.onAuthenticationFailure(request, response, exception)
+  }
+}
+
+object LogFailedLogin {
+
+  def warn(ex: AuthenticationException, request: HttpServletRequest): Unit = {
+    ApplicationLogger.warn(s"Login authentication failed for user '${getUser(ex)}' from IP '${getRemoteAddr(request)}': ${ex.getMessage}")
```
ncharles: I think it would be useful to know why URL it tries to use with getRequestURL(), so that we may detect if parameters are passed in query, or connection tries to by reused over network by taking over a jsessionid, or something else

rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -166,6 +170,58 @@ class AppConfigAuth extends Loggable {
...
+    // Some interesting header, get from https://gist.github.com/nioe/11477264
+    // the list seems to be roughtly correctly ordered - perhaps we don't want one, but all?
+    val headers = List("X-Forwarded-For", "Proxy-Client-IP", "WL-Proxy-Client-IP", "HTTP_CLIENT_IP", "HTTP_X_FORWARDED_FOR")
+
+    //utility to display correctly the pair header:value
+    def show(t2: Tuple2[String, String]) = t2._1+":"+t2._2
+
+    // always returns the ip addr as seen from the server, and the list of other perhaps interesting info
+    (
+      request.getRemoteAddr ::
```
VinceMacBuche: What format will this be ? Maybe it would be interessing to have the same format than headers

VinceMacBuche: headers are "headerName:headerValue"

VinceMacBuche: COuld be named "remoteAdrress" / "remoteIP"

fanf: it's a simple string of the IP. Not sure for the format, we have in the message "IP: xxx.xxx.xxx.xxx|header:foo|header:bar". That would feel strange to have "remoteIP" again preprended to that. +

VinceMacBuche: I find strange that one the ip does not show its origin, let's keep it this way, we will change it when we got feedback :)

rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
fanf: I looked, and we don't get very much interesting information with that:
```
[2017-02-24 17:53:33] WARN  application - Login authentication failed for user 'admin' from IP '0:0:0:0:0:0:0:1': Bad credentials on url: http://localhost:8082/rudder-web/j_spring_security_check
```
I'm really not sure we want to add it.

ncharles: Ha - i see, this is because is it redirected, so it is indeed non interesting.
Does the getQueryString returns something useful ?

fanf: not at all

ncharles: ok, so let's forget it

fanf:
OK, merging this PR

PR #1533, opened 2017-02-24
Fixes #10275: When we add/remove/update a tag to a Directive/Rule, we get an empty event log

fanf:
https://www.rudder-project.org/redmine/issues/10275


rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -161,7 +161,7 @@ class DirectiveUnserialisationImpl extends DirectiveUnserialisation {
       priority              <- (directive \ "priority").headOption.flatMap(s => tryo { s.text.toInt } ) ?~! ("Missing or bad attribute 'priority' in entry type directive : " + xml)
       isSystem              <- (directive \ "isSystem").headOption.flatMap(s => tryo { s.text.toBoolean } ) ?~! ("Missing attribute 'isSystem' in entry type directive : " + xml)
       policyMode            =  (directive \ "policyMode").headOption.flatMap(s => PolicyMode.parse(s.text) )
-      directiveIds          =  (directive \ "directiveIds" \ "id" ).map( n => DirectiveId( n.text ) ).toSet
```
VinceMacBuche: funny ...  should be empty quite often ...

fanf: Yeah, not sur what was the intent of  that code, since directiveIds is never used... Perhaps some debug code that get kept, no idea.

VinceMacBuche: I think it was a copy paste from Rule diff

PR #1536, opened 2017-02-27
Fixes #10295: Log an error when Rest API fails 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10295


rudder-web/src/main/scala/com/normation/rudder/web/rest/RestUtils.scala
```diff
@@ -171,6 +171,12 @@ object RestUtils extends Loggable {
   }
 
   private[this] def effectiveResponse (id:Option[String], message:JValue, status:HttpStatus, action : String , prettify : Boolean) : LiftResponse = {
+    status match {
+      case _:RestError =>
+        // Log any error
+        logger.error(compactRender(message))
```
VinceMacBuche: Maybe we should use the ApplicationLogger ? or even create a new one ?

VinceMacBuche:
OK, merging this PR

PR #1538, opened 2017-03-01
Fixes #10301: Remove "rights" in node property

fanf:
https://www.rudder-project.org/redmine/issues/10301

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
@@ -101,30 +101,6 @@ case object Node {
  */
 final case class NodePropertyProvider(value: String)
 
-sealed trait NodePropertyRights {
-  def value: String
-}
-
-object NodePropertyRights {
-
-  final object ReadWrite extends NodePropertyRights {
-    override val value = "read-write"
-  }
-  final object ReadOnly extends NodePropertyRights {
-    override val value = "read-only"
-  }
-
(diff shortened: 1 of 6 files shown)


rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
```diff
@@ -220,12 +198,12 @@ object CompareProperties {
...
-                         case Some(oldProp@NodeProperty(name, value, provider, rights)) => rights match {
-                             case Some(ReadOnly) if(!same(newProp.provider, provider)) =>
-                               Failure(s"You are trying to update the following property which is owned by an other provider with 'read-only' mode set: '${name}'")
-                             case _ =>
+                         case Some(oldProp@NodeProperty(name, value, provider)) =>
+                             if(same(newProp.provider, provider)) {
                                Full(updateOrRemoveProp(newProp))
-                           }
+                             } else {
+                               Failure(s"You are trying to update the following property which is owned by an other provider, '${provider.map(_.value).getOrElse("")}': '${name}'")
```
VinceMacBuche: Maybe it would be interesting to state here which provider you are  we only display the current provider )

VinceMacBuche: and state "undefined" or "unknown" instead of empty value in the getOrElse

VinceMacBuche:
That will also need a PR in plugin datasource which will not compile (it set rights)

fanf:
Yep, I'm on it

fanf:
PR rebased

PR #1540, opened 2017-03-01
Fixes #10307: Clone group popup is broken

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10307


VinceMacBuche:
Can you try on 4.0 if it happens too ?

RaphaelGauthier:
I tried but I didn't have this bug in 4.0.

PR #1568, opened 2017-03-08
Fixes #4519: Make static JS/CSS resources load faster

fanf:
https://www.rudder-project.org/redmine/issues/4519

Now, all static resources are accessed throught a path which contains Rudder release number, for example: 
```
Before: https://.../rudder/images/ajax-loader.gif
After: https://.../rudder/cache-4.1.0-rc2-SNAPSHOT/images/ajax-loader.gif
```

So now, the path to resources change with each Rudder version, and we can safelly cache then quite agressively. 
So now, these resources are served by the scala web application with the following headers: 
```
Cache-Control: max-age=31556926, public
Pragma: 
Expires: [current time + 6 month]
```
The empty pragme is here to not have jetty add a "pragma: no-cache", which it seems to do on my dev env (not sure on a prod one).


rudder-web/pom.xml
```diff
@@ -55,11 +55,20 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
   </repositories>
 
   <build>
+    <resources>
```
VinceMacBuche: After this ressources is added i don't have any other ressources in my build (/target/classes) I don't know if this is ok when the webapp is built or not.

If we keep it we will need to provide much more information in the jetty config on devellepment environement

fanf:
PR rebased

VinceMacBuche:
That is totally awesome!

It must have been a very very boring work, but so interesting

PR #1571, opened 2017-03-09
Fixes #10386: Bad Rudder logo in src/main/webapp/images

fanf:
https://www.rudder-project.org/redmine/issues/10386


fanf:
PR rebased

amousset:
Where is it used? We are moving from a small logo to a very big one.

VinceMacBuche:
I don't think it's used anymore

fanf:
It' used in the doc.

fanf:
I'm ok to remove it completelly and add it in the doc directly (or use in the doc an other already present)

amousset:
Which doc? All images should be embedded in the rudder-doc folder.

fanf:
The first, main image of old doc (in html) is taken from rudder repository.

amousset:
This doc is not embedded nor linked anywhere starting from 4.0. Anyway, we should probably remove it and use an icon directly in the rudder-doc repo.

fanf:
OK, I'm removing the logo from rudder repos 4.0 and up. It will most likelly breaks the doc build, and so doc "html" target should be removed or the image embeded.

fanf:
PR rebased

amousset:
It is also in the doc repo, I don't think this one is used anywhere https://github.com/Normation/rudder-doc/blob/master/style/html/logoRudder.jpg

PR #1575, opened 2017-03-10
Fixes #10328: No error message in group creation popup if no name is set

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10328


rudder-core/pom.xml
```diff
@@ -53,7 +53,7 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
      <plugin>
        <groupId>org.apache.maven.plugins</groupId>
        <artifactId>maven-jar-plugin</artifactId>
-       <version>3.0.2</version>
+       <version>3.0.2</version><!--$NO-MVN-MAN-VER$-->
```
fanf: We should remove it completly - I'm opening a ticket for that.

PR #1580, opened 2017-03-13
Fixes #10407: Cannot use slashes in quick search

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10407


VinceMacBuche:
OK, merging this PR

PR #1587, opened 2017-03-14
Fixes #10229: Find a way to display "deprecated" techniques which still contain directives in the directives tree

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10229


RaphaelGauthier:
Commit modified

PR #1594, opened 2017-03-16
Fixes #10340: In the rule table, Recent changes column change size during page loading

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10340


ncharles:
here is what it does
image

ncharles:
ha, if it's there at page loading (not editing css manually), it works
image

PR #1597, opened 2017-03-17
Fixes #10446: Policy validation fails

ncharles:
https://www.rudder-project.org/redmine/issues/10446


rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
```diff
@@ -141,9 +142,9 @@ class UpdateDynamicGroups(
                 val error = (e?~! "Error when trying to get the list of dynamic group to update")
 
             }
-          case _:StartProcessing if(!onePending) => onePending = true
```
ncharles: this prevented from detecting that dynamic group was pending

ncharles:
Replaced by https://github.com/Normation/rudder/pull/1601

PR #1601, opened 2017-03-21
Fixes #10446: Policy validation fails

ncharles:


ncharles:
https://www.rudder-project.org/redmine/issues/10446

PR #1602, opened 2017-03-21
Fixes #10451: A running Policy generation is not displayed on the Status dropdown

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10451

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
@@ -83,14 +83,16 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
     case d:DeploymentStatus => deploymentStatus = d ; reRender()
   }
 
-
   private[this] def displayTime(label: String, time: DateTime): NodeSeq = {
     val t = time.toString("yyyy-MM-dd HH:mm:ss")
     val d = DateFormaterService.getFormatedPeriod(time, DateTime.now)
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
```diff
@@ -132,6 +132,12 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
       <li><div id="deployment-end">{displayTime("Ended at ", end)}</div></li>
       <li class="dropdown-header" style="border-bottom-style: dotted; border-color: gray;">{durationText} {DateFormaterService.getFormatedPeriod(start,end)}</li>
     }
+    def loadingStatement(start : DateTime) = {
+      <li class="dropdown-header">Policies building...</li>
+      <li>{displayTime("Started at ", start)}</li>
```
fanf: I believe we can't use the "start" method here, because with it, we have "Started at blablabla -> 1s ago". But the "1s ago" never changes, and so it is extremelly misleading. 
So we need to only display the part with the date.

RaphaelGauthier:
Commit modified

fanf:
I think you misunderstood. We need to keep the start DATE, but not the "↳", because the DURATION won't be updated, and even after a couple of minute, we have "↳ less than 1 s ago", which is misleading because people will thought they can use that to measure the duration, but actually not. 

So please, keep the line with the "Started at 2017-03-24 12:36:38"

RaphaelGauthier:
Commit modified

PR #1604, opened 2017-03-21
Fixes #10452: GUI breaks completely with IE 11

amousset:
https://www.rudder-project.org/redmine/issues/10452

### DIFF
--- a/rudder-web/src/main/resources/rudder-apache-common.conf
+++ b/rudder-web/src/main/resources/rudder-apache-common.conf
@@ -4,6 +4,11 @@
   Header add X-Chrome-Exponential-Throttling "disable"
 </LocationMatch>
 
+# Force using latest mode in IE.
+<LocationMatch "^/(?!rudder/api)">
+  Header add X-UA-Compatible "IE=Edge"
+</LocationMatch>
+
 DocumentRoot /var/www
 
 # Prevent the apache logs from beeing cluttered by 404 errors


rudder-web/src/main/resources/rudder-apache-common.conf
```diff
@@ -4,6 +4,9 @@
   Header add X-Chrome-Exponential-Throttling "disable"
 </LocationMatch>
 
+# Force using latest mode in IE.
+Header add X-UA-Compatible "IE=Edge"
```
peckpeck: Shouldn't we add it only in interface specific urls, like 2 lines above ?

amousset:
Commit modified

amousset:
Indeed, I was testing the cpondition.

rudder-web/src/main/resources/rudder-apache-common.conf
```diff
@@ -4,6 +4,11 @@
   Header add X-Chrome-Exponential-Throttling "disable"
 </LocationMatch>
 
+# Force using latest mode in IE.
+<LocationMatch "^/(?!rudder/api)">
```
peckpeck: I still think this should match the location above.
By changing one or the other

amousset: The other one is specific to Rudder "autoreload" on error feature. We could exclude the API (it is done in 4.1), but there is not reason to apply it on the manual or ncf-builder.

peckpeck:
OK, merging this PR

PR #1605, opened 2017-03-21
Fixes #10456: Several core features don't work anymore with more than 1000 nodes

fanf:
https://www.rudder-project.org/redmine/issues/10456

PR #1606, opened 2017-03-22
Fixes #10456: Several core features don't work anymore with more than 1000 nodes

ncharles:
https://www.rudder-project.org/redmine/issues/10456

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
@@ -84,29 +84,40 @@ case class RoReportsExecutionRepositoryImpl (
       }
     }
 
-    nodeIds.map( _.value).toList.toNel match {
-      case None => Full(Map())
-      case Some(nodes) =>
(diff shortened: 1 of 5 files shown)


ncharles:
replaces https://github.com/Normation/rudder/pull/1605

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
```diff
@@ -84,29 +84,40 @@ case class RoReportsExecutionRepositoryImpl (
...
           // I didn't find any better solution than doing a distinct on the table
-          runs <- (fr"""select r.nodeid, r.date, r.nodeconfigid, r.complete, r.insertionid,
-                         c.nodeid, c.nodeconfigid, c.begindate, c.enddate, c.configuration from
-                         (
-                           select distinct on (nodeid)
-                              nodeid, date, nodeconfigid, complete, insertionid
-                              from reportsexecution
-                              where complete = true
-                              and """ ++ Fragments.in(fr"nodeid", nodes) ++
-                      fr"""order by nodeid, insertionid desc
```
ncharles: this order has been forgoteen - it makes everything faster, but it's uncommon to have the right result

PR #1613, opened 2017-03-27
Fixes #10513: Directive column of pending nodes table is not wide enough.

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10513


RaphaelGauthier:
OK, merging this PR

PR #1615, opened 2017-03-28
Fixes #10520: Dataource can not override an existing property

fanf:
https://www.rudder-project.org/redmine/issues/10520


fanf:
PR rebased

PR #1618, opened 2017-03-29
Fixes #10526: Group type order  is different between popup and group page

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10526


RaphaelGauthier:
OK, merging this PR

PR #1620, opened 2017-03-30
Fixes #10525: A magnifier icon appears in the filter fild after displaying directives that would be applied in pending nodes page

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10525


RaphaelGauthier:
OK, merging this PR

PR #1626, opened 2017-04-07
Fixes #10570: Regex search on node properties is broken: returns all node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10570

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -88,7 +88,6 @@ trait SpecialComparator extends BaseComparator
 case object Regex extends SpecialComparator { override val id = "regex" }
 case object NotRegex extends SpecialComparator { override val id = "notRegex" }
 
-
 sealed trait KeyValueComparator extends BaseComparator
 case object HasKey extends KeyValueComparator { override val id = "hasKey" }
 
@@ -138,8 +137,8 @@ sealed trait CriterionType  extends ComparatorList {
   //transform the given value to its LDAP string value
   def toLDAP(value:String) : Box[String]
 
-  def buildRegex(attribute:String,value:String):RegexFilter = RegexFilter(attribute,value)
-  def buildNotRegex(attribute:String,value:String):NotRegexFilter = NotRegexFilter(attribute,value)
+  def buildRegex(attribute:String,value:String): Box[RegexFilter] = Full(RegexFilter(attribute,value))
+  def buildNotRegex(attribute:String,value:String): Box[NotRegexFilter] = Full(NotRegexFilter(attribute,value))
 
   //build the ldap filter for given attribute name and comparator
   def buildFilter(attributeName:String,comparator:CriterionComparator,value:String) : Filter =
@@ -341,7 +340,6 @@ case object OstypeComparator extends CriterionType {
     }
   }
 
-
   override def toForm(value: String, func: String => Any, attrs: (String, String)*) : Elem =
     SHtml.select(
       (osTypes map (e => (e,e))).toSeq,
@@ -422,7 +420,7 @@ case object EditorComparator extends CriterionType {
   override def toLDAP(value:String) = Full(value)
 }
 
-case class JsonComparator(key:String, splitter:String = "", numericvalue:Boolean = false) extends TStringComparator {
+case class JsonComparator(key:String, splitter:String = "", numericvalue:Boolean = false) extends TStringComparator with Loggable {
   override val comparators = HasKey +: BaseComparators.comparators
 
   def splitJson(attribute:String, value:String) = {
@@ -441,7 +439,7 @@ case class JsonComparator(key:String, splitter:String = "", numericvalue:Boolean
           s""""${attval._1}":"${attval._2}""""
       ) )
     } else {
-      Failure("not enough argument")
+      Failure(s"Could not  split attribute '${attribute}'  of value '${value}' with splitter '${splitter}'")
     }
   }
 
@@ -453,14 +451,22 @@ case class JsonComparator(key:String, splitter:String = "", numericvalue:Boolean
     }
   }
 
-  override def buildRegex(attribute:String,value:String) : RegexFilter = {
-    val regexp = ".*%s.*".format(splitJson(attribute,value).getOrElse(Nil).mkString(".*"))
-    RegexFilter(key,regexp)
+  override def buildRegex(attribute:String,value:String) : Box[RegexFilter] = {
+    for {
+      splitted <- splitJson(attribute,value)
+      regexp = s".*${splitted.mkString(".*")}.*"
+    } yield {
+      RegexFilter(key,regexp)
+    }
   }
 
-  override def buildNotRegex(attribute:String,value:String) : NotRegexFilter = {
-    val regexp = ".*%s.*".format(splitJson(attribute,value).getOrElse(Nil).mkString(".*"))
-    NotRegexFilter(key,regexp)
+  override def buildNotRegex(attribute:String,value:String) : Box[NotRegexFilter] = {
+    for {
+      splitted <- splitJson(attribute,value)
+      regexp = s".*${splitted.mkString(".*")}.*"
+    } yield {
+      NotRegexFilter(key,regexp)
+    }
   }
 
   override def buildFilter(attributeName:String, comparator:CriterionComparator,value:String) : Filter = {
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -730,38 +711,36 @@ class InternalLDAPQueryProcessor(
...
+                case _ => Full(LDAPFilter(a.buildFilter(comp,value)))
+              }
+              filter.map((objectDnTypes(objectType),tpe, _))
+            } else {
+              Failure(s"The object type '${objectType}' is not know in criteria ${crit}")
+            }
+          }
+        // group 'filter'(_3) by objectType (_1), then by LDAPObjectType (_2)
+        }).map(_.groupBy(_._1).mapValues(_.groupBy(_._2).mapValues(_.map(_._3).toSet)))
+    } else {
```
ncharles: indentation is broken

rudder-core/src/test/scala/com/normation/rudder/domain/queries/CmdbQueryTest.scala
```diff
@@ -43,11 +41,11 @@ class JsonQueryTest extends Specification {
 
   "JsonComparator " should {
     "Convert a search request to a valid json request on one attribute " in {
-      JsonComparator("attribute").buildFilter("key",Equals,"value") must beEqualTo(SUB("attribute",null,("\"key\":\"value\""::Nil).toArray,null))
+      JsonComparator("attribute").buildFilter("key",Equals,"value") must beEqualTo(SUB("attribute",null,("\"key\":[\"{]value[\"}]"::Nil).toArray,null))
     }
 
     "Convert a search request to a valid json request on 2 attributes" in {
-      JsonComparator("attribute","=").buildFilter("key.key2",Equals,"value=value2") must beEqualTo(SUB("attribute",null,("\"key\":\"value\""::"\"key2\":\"value2\""::Nil).toArray,null))
+      JsonComparator("attribute","=").buildFilter("key.key2",Equals,"value=value2") must beEqualTo(SUB("attribute",null,("\"key\":[\"{]value[\"}]"::"\"key2\":[\"{]value2[\"}]"::Nil).toArray,null))
```
ncharles: could you add a test about the corrected issue ?

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -730,38 +711,36 @@ class InternalLDAPQueryProcessor(
...
-      }
-    }.groupBy { case (objectType, v) => objectDnTypes(objectType) }
-
-
-    val nodeFilters = groupedSetFilter.get(QueryNodeDn).flatMap( _.get("node").map( _.toSet) )
+    for {
+      groupedSetFilter <- groupFilterByLdapRequest()
+    } yield {
+        val nodeFilters = groupedSetFilter.get(QueryNodeDn).flatMap( _.get("node").map( _.toSet) )
     val otherFilters = groupedSetFilter.flatMap { case(dn, map) =>
```
ncharles: indentation is broken

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -730,38 +711,36 @@ class InternalLDAPQueryProcessor(
...
-    //validate that for each object type in criteria, we know it
-    query.criteria foreach { cl =>
-      if(!objectTypes.isDefinedAt(cl.objectType.objectType))
-        return Failure("The object type '%s' is not know in criteria %s".format(cl.objectType.objectType,cl))
+    def groupFilterByLdapRequest() = {
+      //validate that we knows the requested object type
+      if (objectTypes.isDefinedAt(query.returnType.value)) {
+        (sequence(query.criteria) {
+           case crit@CriterionLine(ot,a,comp,value) => {
+            val objectType = ot.objectType
```
ncharles: i'm not 100% sure, but it seems the semantic changed compared to what was defined before:
* we had criterai grouped by objectType, and Filter generated for each group of object class
* now for each criteria we create the filter, map the object type and group by dnObjectType , and then two mapValues (which are super hard to comprehend) that group by resulting LDAPObjectType and then by filter
So it seems we have an extra reordering.

VinceMacBuche:
Mostly broken about the indentation, You don't go with the back of the spoon like french says (or even you don't go with a dead hand )

But you're right

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
VinceMacBuche: The first "group by" was risky because we were making a group by then a map that changes the keys which could make a loss of data ( Would only keep the last thing with the same 'tpe', which does not appear now but it could and it would be veryyyy surprising) 

Here I compute all data we need in one go before grouping them, we produce the same data that we would produce

ncharles: ha ok, so it's a change for better :)

VinceMacBuche:
Commit modified

VinceMacBuche:
Wait before merge, closing for now

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -453,14 +451,23 @@ case class JsonComparator(key:String, splitter:String = "", numericvalue:Boolean
     }
   }
 
-  override def buildRegex(attribute:String,value:String) : RegexFilter = {
-    val regexp = ".*%s.*".format(splitJson(attribute,value).getOrElse(Nil).mkString(".*"))
-    RegexFilter(key,regexp)
+  override def buildRegex(attribute:String,value:String) : Box[RegexFilter] = {
+    for {
+      splitted <- splitJson(attribute,value)
+      regexp = s".*${splitted.mkString(".*")}.*"
```
ncharles: regex changed from .*%s.* to .*
is it normal?

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -453,14 +451,23 @@ case class JsonComparator(key:String, splitter:String = "", numericvalue:Boolean
...
+    } yield {
+      RegexFilter(key,regexp)
+    }
   }
 
-  override def buildNotRegex(attribute:String,value:String) : NotRegexFilter = {
+  override def buildNotRegex(attribute:String,value:String) : Box[NotRegexFilter] = {
     val regexp = ".*%s.*".format(splitJson(attribute,value).getOrElse(Nil).mkString(".*"))
```
ncharles: this regexp is not used anymore, isn't it ?

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -1,6 +1,6 @@
 /*
-*************************************************************************************
-* Copyright 2011 Normation SAS
+****************o*********************************************************************
+* Copyright 2011 deNormation SAS
```
ncharles: typo ?

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
VinceMacBuche: it did not change the regex (it was ".*%s.*")

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
VinceMacBuche: oops

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
VinceMacBuche: indeed, mistery is why i kept it !!

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
ncharles: where did the %s went ?

VinceMacBuche: %s was used used by the format function with "splitted.mkString(".*")"

Now it is replaced by ${splitted.mkString(".*")}

ncharles: ok, i completly forgot we used to replace %s

VinceMacBuche:
Commit modified

PR #1628, opened 2017-04-13
Fixes #10569: Adding a json node property by UI make it a string

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10569

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/angular/nodeProperties.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/angular/nodeProperties.js
@@ -134,8 +134,16 @@ app.controller('nodePropertiesCtrl', function ($scope, $http, $compile) {
     function checkNameUnicity(property, index, array) {
       return property.name == $scope.newProperty.name;
     }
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -636,6 +636,7 @@ case class RestExtractorService (
   private[this] def extractNodeProperty(json : JValue) : Box[NodeProperty] = {
     ( (json \ "name"), (json \ "value") ) match {
       case ( JString(nameValue), value ) =>
+        logger.info(value)
```
ncharles: it should either be debug, or not there, isn't it?

VinceMacBuche:
Commit modified

ncharles:
I cannot add node properties with this change :(

ncharles:
ha, even with branch 4.1 I cannot

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1632, opened 2017-04-19
Fixes #10552: On some resolution, top bar is on two line and we can't see the whole directive tree

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10552


RaphaelGauthier:
OK, merging this PR

PR #1627, opened 2017-04-13
Fixes #10578: Cannot add node property when audit message is mandatory

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/10578

PR #1639, opened 2017-04-25
Fixes #10638:  Cannot change policy mode when audit message is mandatory (4.0)

fanf:
https://www.rudder-project.org/redmine/issues/10638


ncharles:
this is very much like https://github.com/Normation/rudder/pull/1627/files :)

PR #1641, opened 2017-04-25
Fixes #5718: In reports list, if the directive contains < >, then the part in < > doesn't appear in the reports

ncharles:
https://www.rudder-project.org/redmine/issues/5718


fanf:
OK, merging this PR

PR #1646, opened 2017-04-26
Fixes #10567: Infinite "rudder is loading" page if rudder-init didn't run

fanf:
https://www.rudder-project.org/redmine/issues/10567


ncharles:
so much weirdness

PR #1649, opened 2017-04-27
Fixes #8017: Adding an authorized network when "common-root" directive is missing seems to work, but actually fails

fanf:
https://www.rudder-project.org/redmine/issues/8017

### DIFF
--- a/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckDIT.scala
+++ b/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckDIT.scala
@@ -44,6 +44,8 @@ import com.normation.inventory.ldap.core.InventoryDit
 import com.normation.ldap.sdk.LDAPConnectionProvider
 import javax.servlet.UnavailableException
 import com.normation.ldap.sdk.RwLDAPConnection
+import com.unboundid.ldap.sdk.DN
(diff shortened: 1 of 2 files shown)


rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckDIT.scala
```diff
@@ -141,6 +143,37 @@ class CheckDIT(
       case Full(_) => //ok
     }
 
+    //check Root Policy Server entries (from init-policy-server.ldif)
+    val dns = List(
+        "nodeId=root,ou=Nodes,cn=rudder-configuration"
```
ncharles: we have properties in config file for these
We should either use the properties, or either say they should never touched/used

fanf: I don't think so... What properties are you talking about?

fanf: oh, rudder base?

fanf: Well, the config file is lying, because rudder-init does care at all of these properties.

fanf:
OK, merging this PR

PR #1650, opened 2017-04-27
Fixes #10660: We have no way to know that a Technique is disabled

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/10660


RaphaelGauthier:
OK, merging this PR

PR #1662, opened 2017-06-12
Fixes #9302: Time since last generation is limited to hours

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/9302

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/components/DateFormaterService.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/components/DateFormaterService.scala
@@ -40,6 +40,7 @@ package com.normation.rudder.web.components
 import org.joda.time.DateTime
 import org.joda.time.format.PeriodFormatterBuilder
 import org.joda.time.Duration
+import org.joda.time.chrono.ISOChronology
 
 object DateFormaterService {
 
@@ -66,7 +67,7 @@ object DateFormaterService {
 
   def formatPeriod(duration:Duration) : String = {
     if(duration.getMillis < 1000) "less than 1 s"
-    else periodFormatter.print(duration.toPeriod)
+    else periodFormatter.print(duration.toPeriod(ISOChronology.getInstanceUTC))
   }
 
   def getFormatedPeriod(start : DateTime, end : DateTime) : String = {


rudder-web/src/main/scala/com/normation/rudder/web/components/DateFormaterService.scala
```diff
@@ -64,12 +64,9 @@ object DateFormaterService {
...
-  def formatPeriod(duration:Duration) : String = {
-    if(duration.getMillis < 1000) "less than 1 s"
-    else periodFormatter.print(duration.toPeriod)
-  }
-
   def getFormatedPeriod(start : DateTime, end : DateTime) : String = {
-    formatPeriod(new Duration(start,end))
+    val duration =  new Duration(start,end)
+    if(duration.getMillis < 1000) "less than 1 s"
+    else periodFormatter.print(duration.toPeriodFrom(start))
```
VinceMacBuche: in else do:  <pre>{
   val period = duration.toPeriod(PeriodType.YearMonthDayTime)
  periodFormatter.print(period)
}</pre>

fanf: no, it should already be done, see the def of periodformatter l64 and after:
```
   private[this] val periodFormatter = new PeriodFormatterBuilder().
     appendDays().
     appendSuffix(" day", " days").
appendSeparator(", ")....
```

fanf: In fact, I don't understand why days are not display now.

VinceMacBuche: The bug is in toPeriod, which create by default a periodType "Time" which is only "hours, minutes, seconds, millis"

VinceMacBuche: Hmm not sure in fact, documentation is not ... very clear

fanf: So, after some error & try, the correct code is: 

```
import org.joda.time.chrono.ISOChronology
....
def formatPeriod(duration:Duration) : String = {
if(duration.getMillis < 1000) "less than 1 s"
else periodFormatter.print(duration.toPeriod(ISOChronology.getInstanceUTC))
}
```

RaphaelGauthier:
Commit modified

PR #1664, opened 2017-06-14
Fixes #10908: When hooks in policy-generation-node-ready or policy-generation-node-finished fails, there is no error in the generation

fanf:
https://www.rudder-project.org/redmine/issues/10908

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
@@ -267,6 +267,9 @@ class Cf3PromisesFileWriterServiceImpl(
       }})
     implicit val scheduler = Scheduler.io(executionModel = ExecutionModel.AlwaysAsyncExecution)
 
+    //interpret HookReturnCode as a Box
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
```diff
@@ -267,6 +268,9 @@ class Cf3PromisesFileWriterServiceImpl(
       }})
     implicit val scheduler = Scheduler.io(executionModel = ExecutionModel.AlwaysAsyncExecution)
 
+    //interpret HookReturnCode as a Bow
```
ncharles: as a Box

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PromisesFileWriterService.scala
```diff
@@ -84,6 +84,7 @@ import scala.language.postfixOps
 import com.normation.rudder.hooks.RunHooks
 import com.normation.rudder.hooks.HookEnvPairs
 import com.normation.rudder.hooks.HooksLogger
+import com.normation.rudder.hooks.HooksImplicits
```
ncharles: this is so clever !

fanf:
PR rebased

PR #1673, opened 2017-06-21
Fixes #10927: Create system variables for apache authentication

ncharles:
https://www.rudder-project.org/redmine/issues/10927


rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -375,10 +389,33 @@ class SystemVariableServiceImpl(
...
+      val varSubNodesId      = systemVariableSpecService.get("SUB_NODES_ID"     ).toVariable(subNodesList.map(_._2.id.value))
+      val varSubNodesServer  = systemVariableSpecService.get("SUB_NODES_SERVER" ).toVariable(subNodesList.map(_._2.policyServerId.value))
+      val varSubNodesKeyhash = systemVariableSpecService.get("SUB_NODES_KEYHASH").toVariable(subNodesList.map(n => s"sha256:${n._2.sha256KeyHash}"))
+
+
+      // Construct the system variables for nodes with certificates
+      val nodesWithCertificate = nodesAgentWithCertificate.flatMap {case (agent, node) =>
+        agent.securityToken match {
+          // A certificat, we return it
+          case cert:Certificate => Some(node -> cert)
```
fanf: you should return here (cert, cert.cert) because parsing cert is expensive, so better doing only one time

ncharles:
Commit modified

PR #1695, opened 2017-07-04
Fixes #11072: Move out dsc code into the plugin part

fanf:
https://www.rudder-project.org/redmine/issues/11072


ncharles:
This looks good.

PR #1707, opened 2017-07-27
Fixes #11171: Error when trying to remove a node

ncharles:
https://www.rudder-project.org/redmine/issues/11171


ncharles:
Commit modified

PR #1713, opened 2017-08-08
Fixes #11217: Ipv6 constraint NOT PORTED to Rudder 4.0 and up

fanf:
https://www.rudder-project.org/redmine/issues/11217


fanf:
OK, merging this PR

PR #1714, opened 2017-08-08
Fixes #11223: /var/log/webapp/compliance/non-compliant-reports does not contain non-compliant (audit) report

fanf:
https://www.rudder-project.org/redmine/issues/11223


fanf:
OK, merging this PR

PR #1721, opened 2017-08-17
Fixes #11270: "no report" explanation is not correct in node details

fanf:
https://www.rudder-project.org/redmine/issues/11270


fanf:
OK, merging this PR

PR #1726, opened 2017-08-25
Fixes #11291: Highlight DSC compatible technique versions in the available versions table.

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11291


rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/DirectiveManagement.scala
```diff
@@ -355,16 +356,21 @@ class DirectiveManagement extends DispatchSnippet with Loggable {
       val isDeprecated = t.deprecrationInfo.isDefined
       val deprecationMessage = t.deprecrationInfo.map(_.message).getOrElse("")
       val acceptationDate = DateFormaterService.getFormatedDate(timeStamp)
+      val dscSupport = t.agentConfigs.forall { _.agentType match{
```
VinceMacBuche: Why forall ?  this means that all agentTYpe should be dsc, use exists instead!!

RaphaelGauthier:
Commit modified

PR #1728, opened 2017-08-25
Fixes #11307: There is no "node-post-deletion" hook directory but a "node-pre-deletion" one

fanf:
https://www.rudder-project.org/redmine/issues/11307


fanf:
OK, merging this PR

PR #1729, opened 2017-08-28
Fixes #11296: Add license files and checking utilities in Rudder

fanf:
https://www.rudder-project.org/redmine/issues/11296

This is a wip, but appart from more tests, all the basics infra are in place. I PR'ed it because it's simpler to discuss. 

So, almost everything is done in License.scala file. It's where we define all the data types, and basic logic for signing/cheching signature of license files. 

I attempted to be rigorous about data type definition, but even with that, I have nothing to ensure that digest and keyid are effectively in hex-encoded, or that date are in iso-8601. The fault to invisible .toString(). 

Note that we are expecting a RSA keypair in pkcs8 format. Anything else will fail, and particulary the default openssl generated keypair (because by default openssl writes pkcs1). 

Version.scala is just the factored-out part of Rudder plugin for license comparison (hence it is ordered now). We need to decide if we want to allows totally generic prefix for version (for now, they are limited to rudder major.minor because of the semantic of plugins). When everything is set up, we will need to remove Version from plugins package (but we need to keep signature to compatibilty of existing plugin). 

The cli is boring as only a cli can be. It parsed args, check if args are consistant, and call method defined elsewhere. Example of use are in the CLI test, but basically you have two subcommands:

```
license sign -k private.pem -o license.sign license.info
license check -k public.pem license.sign
``` 

Ho, and I hate bouncy castle (lake of) documentation.

### DIFF
--- a/rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
+++ b/rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
@@ -1,0 +1,213 @@
+/*
+*************************************************************************************
+* Copyright 2017 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
+* distributed under the license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* Rudder is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+*
+*************************************************************************************
+*/
+
+package com.normation.license.cli
+
+import scopt.OptionParser
+import ca.mrvisser.sealerate.values
+import com.normation.license.MaybeLicenseError._
+import com.normation.license.LicenseSigner
+import com.normation.license.LicenseReader
+import com.normation.license.RudderLicense
+import com.normation.license.RSAKeyManagement
+import com.normation.license.LicenseChecker
+import com.normation.license.LicenseError
+
+/**
+ * The configuration object for our CLI.
+ * The basic process is to take one file in input for the definition of variables, one set of files as template to change.
+ *
+ * By default, the files are generated with the same name in the current folder.
+ *
+ * - add the possibility to take directories as input (but a good shell can do it, so not very important)
+ */
+
+
+// there is no good way to define a command as the sum of sub commands
+// see: https://github.com/scopt/scopt/issues/134
+// we will use product + convention ("convention" == "do the right things with the chainsaw")
+
+final case class CheckConfig(
+    pubkeyPath     : String = "rudder.pub"
+  , licenseInPath  : String = "license.in"
+  , algorithm      : String = "SHA512WithRSA"
+)
+
+final case class SignConfig(
+    privkeyPath    : String = "rudder.pub"
+  , licenseOutPath : String = "license.sign"
+  , licenseInPath  : String = "license.in"
+  , algorithm      : String = "SHA512WithRSA"
+)
+
+sealed trait CliMode
+final object CliMode {
+  final case object Check extends CliMode
+  final case object Sign  extends CliMode
+
+  val all = values[CliMode]
+}
+
+final case class Config(
+    mode       : CliMode     = CliMode.Check
+  , checkConfig: CheckConfig = CheckConfig()
+  , signConfig : SignConfig  = SignConfig()
+)
+
+object LicenseCli {
+
+  //read version
+
+  val checkParser = new OptionParser[CheckConfig]("Check a Rudder related license") {
+    head("rudder-license-cli-check", "...")
+
+    opt[String]("key").abbr("k").action( (x, c) =>
+      c.copy(pubkeyPath = x) ).text("path toward public key to use to check signature")
+    arg[String]("input license file").action( (x, c) =>
+      c.copy(licenseInPath = x) ).text("path of license file to check.")
+
+  }
+
+  val signParser = new OptionParser[SignConfig]("Sign a Rudder related license") {
+    head("rudder-license-cli-check", "...")
+    opt[String]("key").abbr("k").action( (x, c) =>
+      c.copy(privkeyPath = x) ).text("path toward private key to use for signature")
+    opt[String]("output").abbr("o").action( (x, c) =>
+      c.copy(licenseOutPath = x) ).text("file name for the signed license")
+    arg[String]("input license file").action( (x, c) =>
+      c.copy(licenseInPath = x) ).text("path of license file to sign")
+  }
+
+  val parser = new OptionParser[Config]("Rudder License Manager") {
+    head("rudder-license-cli", "4.1.x")
+
+    def cfgSign(c: Config, f: SignConfig => SignConfig): Config = {
+      c.copy(signConfig = f(c.signConfig))
+    }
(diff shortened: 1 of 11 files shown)


fanf:
PR rebased

fanf:
I'm not sure that code should be in Rudder. Perhaps it should go in rudder-tools, or rudder-tools-private as a standalone library that can be used but is not directly linked by Rudder.

rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
```diff
@@ -0,0 +1,177 @@
...
+    cmd("sign").action( (_, c) => c.copy(mode = CliMode.Sign) ).
+      text("Write a new signed license file from the input license information.").
+      children(
+          opt[String]("key").abbr("k").action( (x, c) =>
+            c.copy(keyPath = x) ).text("path toward private key to use for signature")
+        , opt[String]("output").abbr("o").action( (x, c) =>
+            c.copy(licenseOutPath = x) ).text("file name for the signed license")
+        , arg[String]("input license file").action( (x, c) =>
+            c.copy(licenseInPath = x) ).text("path of license file to sign")
+//        , checkConfig( c =>
```
VinceMacBuche: Should remove this

rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
```diff
@@ -0,0 +1,177 @@
...
+  , licenseInPath  : String = "license.in"
+  , mode           : CliMode = CliMode.Sign // either "sign" or "check"
+  , algorithm      : String = "SHA512WithRSA"
+)
+
+object LicenseCli {
+
+
+  val parser = new OptionParser[Config]("Rudder License Manager") {
+    head("rudder-license-cli", "4.1.x")
```
VinceMacBuche: Version should be more precise no ? like we do for version number in the webapp ? or maybe we don't care ab out the x ...

rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
```diff
@@ -0,0 +1,177 @@
...
+
+sealed trait CliMode
+final object CliMode {
+  final case object Check extends CliMode
+  final case object Sign  extends CliMode
+
+  val all = values[CliMode]
+}
+
+final case class Config(
```
VinceMacBuche: So you define a value for "LicenseOutPath" even if you don't need it (ie Check) 

Should we not make it more demand on the CliMode ? either by integrating parameter into cliMode or changing the class model (make config a Trait with a generic CliMode, and two case class defining CheckConfig and SignConfig ? 

Maybe make licenceOutPath Optionnal ?

rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
```diff
@@ -0,0 +1,177 @@
...
+      written    <- LicenseSigner.writeLicenseFile(licenseOutPath, signed)
+    } yield {
+      signed
+    }
+  }
+
+  /*
+   * check logic
+   */
+  def chech(licenseInPath: String, publicKeyPath: String, algorithm: String): Maybe[RudderLicense.ValidSignature] = {
```
VinceMacBuche: Should be "check" not "chech" :)

rudder-license/src/main/scala/com/normation/license/License.scala
```diff
@@ -0,0 +1,724 @@
...
+ * maxversion=99.0
+ * startdate=2017-06-01T00:00:00Z
+ * enddate=2018-05-31T00:00:00Z
+ * ----
+ */
+
+
+/**
+ * A structure that allows to keep a field key name to
+ * allow parsing/reading them (serialization is such a
```
VinceMacBuche: 👍

rudder-license/src/main/scala/com/normation/license/License.scala
```diff
@@ -0,0 +1,724 @@
...
+  }
+}
+
+/*
+ * An helper object that holds ISO date reader/writer
+ */
+final object ISODate {
+  import MaybeLicenseError._
+
+  val pattern = "yyyy-MM-dd'T'HH:mm:ssZZ" // because no way to get back from formatter to pattern...
```
VinceMacBuche: 🤦‍♂️

rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
fanf: I'm two-sided on that. I think there is value to provide a tool that works for the whole 4.1 release cycle. But we can't really ensure that. 
Well, yes, I'm just going to use Rudder exact version, at least it will be precise.

rudder-license-cli/src/main/scala/com/normation/license/cli/LicenseCli.scala
fanf: It seems to be the way the cli arg parser is designed to work. It's not very satisfying, as you remarked. I will try&test to see if there is better model accepted by the lib.

rudder-license/src/main/scala/com/normation/license/License.scala
fanf: I think I should use the pattern as an entry point and don't use the provided ISO formatter at all, so that we can't have a desynchronisation between the two of them (there is no perf cost, JodaTime is caching the thing internally).

rudder-license/src/main/scala/com/normation/license/License.scala
```diff
@@ -0,0 +1,724 @@
...
+   * (SubjectPublicKeyInfo/PrivateKeyInfo, keyFactory.generatePublic/generatePrivate...)
+   * to be more work to factorize than duplicate :(
+   */
+
+  def readPKCS8PublicKey(path: String): Maybe[RSAPublicKey] = {
+    for {
+                    // the parser able to read PEM files
+                    // Parser may be null if the key is invalid
+      parser     <- {
+                      val p = new PEMParser(Files.newBufferedReader(Paths.get(path)))
```
VinceMacBuche: No need to check if the path exists ?

rudder-license/src/main/scala/com/normation/license/License.scala
fanf: Just so that you cry: this is the quantity of NO boilerplate brought by circe+doobie (ok, in a toy project, but still, it's amazing! Not ONE line of parsing): https://github.com/fanf/scala-pet-store/blob/master/src/main/scala/io/github/pauljamescleary/petstore/endpoint/PetEndpoints.scala
(the whole project is interesting for learning)

fanf: OK, the person must learn how to align things, but well...

rudder-license/src/main/scala/com/normation/license/License.scala
VinceMacBuche: And provide a better error message if does not exists!

rudder-license/src/main/scala/com/normation/license/License.scala
```diff
@@ -0,0 +1,724 @@
...
+      rsaPubkey
+    }
+  }
+
+  def readPKCS8PrivateKey(path: String): Maybe[RSAPrivateKey] = {
+    for {
+                     // the parser able to read PEM files
+                     // Parser may be null if the key is invalid
+      parser      <- {
+                       val p = new PEMParser(Files.newBufferedReader(Paths.get(path)))
```
VinceMacBuche: Same remark than public key function on the path existence

rudder-license/src/main/scala/com/normation/license/License.scala
fanf: you're right

fanf: Ah sorry, in fact it is actually checked:
```
{
....
} withError { ex => LicenseError.Crypto(s"Error when trying to read the private key at '${path}'. The key may be invalid (expecting PKCS8 format): ${ex.getMessage}")
}
```

rudder-license/src/main/scala/com/normation/license/License.scala
```diff
@@ -0,0 +1,724 @@
...
+  /**
+   * We can write any kind of license, but it may not be very smart to not write valide license file.
+   */
+  def writeLicenseFile(path: String, license: RudderLicense): Maybe[Unit] = {
+    // base64 encode to an utf-8 string
+    def encodeHex(bytes: Array[Byte]) =   new String(Hex.encode(bytes), "UTF-8")
+    // only errors here are I/O error, not very interesting to discriminate them: group everything in a try/catch
+    // digest and keyId are base64 encoded
+    val toWrite = s"""${license.fileFormat.header}
+        |algorithm=${license.signature.algorithm.value}
```
VinceMacBuche: can the key be the value from the object ? license.signature.algorithm.name ?

rudder-license/src/main/scala/com/normation/license/License.scala
VinceMacBuche: I could make a Scala project to learn people to align things, and they can learn me some category theory / functionnal magic

fanf:
PR rebased

fanf:
I think we really need to put that elsewhere. It makes no sense at all to have Rudder verisonning for it, and not very much sense to have it in Rudder oss at all. 
Closing it for now.

PR #1732, opened 2017-09-04
Fixes #11323: Fold automatically Group categories if there are too many of them (4.1)

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11323


rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/Groups.scala
```diff
@@ -287,6 +287,7 @@ class Groups extends StatefulSnippet with SpringExtendableSnippet[Groups] with L
 
         <div id={htmlId_groupTree}>Error: {e.msg}</div>
       case Full(lib) =>
+      val foldCategories = lib.allCategories.size > 11
```
ncharles: Could you add the comment that is in branch 4.2?
//We want to fold categories if there are more than 11 (10 + 1 taking into account the hidden root node)

RaphaelGauthier:
Commit modified

PR #1733, opened 2017-09-05
Fixes #11183: Group "All nodes known by Rudder directly connected to the XXXX server" excludes DSC nodes

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11183


VinceMacBuche:
Commit modified

PR #1736, opened 2017-09-15
Fixes #11326: chars are not escaped correctly in dsc techniques

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11326

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -802,6 +802,7 @@ trait PromiseGeneration_buildNodeConfigurations extends PromiseGenerationService
                                    (name, p)
                                  }
                                }
+            agent           <- Box(context.nodeInfo.agentsName.headOption) ?~! "Should not be Empty"
             cf3PolicyDrafts <- bestEffort(drafts) { draft =>
                                   (for {
                                     //bind variables with interpolated context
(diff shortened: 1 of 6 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PolicyDraft.scala
```diff
@@ -240,16 +243,19 @@ case class ParameterEntry(
...
  */
 object ParameterEntry {
-  def escapeString(x: String) : String = {
+  def escapeString(x: String, agentType: AgentType) : String = {
     if (x == null)
       x
     else
-      x.replaceAll("""\\""", """\\\\""").replaceAll(""""""", """\\"""")
+      agentType match {
+      case AgentType.Dsc => x.replaceAll("`", "``").replaceAll("\"", "`\"")
```
ncharles: can you correct the indentation ? :)

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -802,6 +802,7 @@ trait PromiseGeneration_buildNodeConfigurations extends PromiseGenerationService
                                    (name, p)
                                  }
                                }
+            agent           <- Box(context.nodeInfo.agentsName.headOption) ?~! "Should not be Empty"
```
fanf: "Should not be Empty" => "The agent type is not defined. Please, report it, it is most likelly a bug".

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PolicyDraft.scala
ncharles: why is it \" that we escape, rather than simply " ?

fanf:
Appart for the 2 little remarks, it's ok. You can correct and label for merge on your own.

VinceMacBuche:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/Cf3PolicyDraft.scala
VinceMacBuche: " is escaped because it's a double quote string, so to contains a " it needs to be escaped

VinceMacBuche:
Closing, wrong target branch

VinceMacBuche:
replaced by #1737

PR #1737, opened 2017-09-17
Fixes #11326: chars are not escaped correctly in dsc techniques

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11326


VinceMacBuche:
This was validated in #1736, self merging it

PR #1746, opened 2017-09-26
Fixes #11431: Update generation process to make it plugable by agenttype AND by os

fanf:
https://www.rudder-project.org/redmine/issues/11431


ncharles:
\o/

PR #1752, opened 2017-10-04
Fixes #11168: Search believe that CFEngine agents with "dsc" in their keys are also DSC agent 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11168

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -406,11 +406,11 @@ case object OsNameComparator extends CriterionType {
 case object AgentComparator extends CriterionType {
 
   val ANY_CFENGINE = "cfengine"
-  /*
-   * We are using 'oldShortName' because we are matching on substring,
-   * and that name was used in the past.
-   */
-  val agentTypes = ( (ANY_CFENGINE, "Any CFEngine based agent") :: AgentType.allValues.toList.map(a => (a.oldShortName, a.displayName))).sortBy( _._2 )
+  val (cfeTypes, cfeAgents) = ((ANY_CFENGINE, "Any CFEngine based agent"),(ANY_CFENGINE, AgentType.CfeCommunity :: AgentType.CfeEnterprise :: Nil))
+  val allAgents = AgentType.allValues.toList
+
+  val agentTypes = ( cfeTypes  :: allAgents.map(a => (a.oldShortName, (a.displayName)))).sortBy( _._2 )
+  val agentMap   = ( cfeAgents :: allAgents.map(a => (a.id, a :: Nil))).toMap
 
   override def comparators = Seq(Equals, NotEquals)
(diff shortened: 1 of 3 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

ncharles:
could you add a test with the offending key, to ensure correct behaviour ?

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -410,28 +410,35 @@ case object AgentComparator extends CriterionType {
    * We are using 'oldShortName' because we are matching on substring,
    * and that name was used in the past.
    */
-  val agentTypes = ( (ANY_CFENGINE, "Any CFEngine based agent") :: AgentType.allValues.toList.map(a => (a.oldShortName, a.displayName))).sortBy( _._2 )
+  val agentTypes = ( (ANY_CFENGINE, "Any CFEngine based agent") :: AgentType.allValues.toList.map(a => (a.id, a.displayName))).sortBy( _._2 )
 
+  val agentMap = AgentType.allValues.toList.map(a => (a.id, a :: Nil)).toMap
```
fanf: I believe you didn't do what you intented here: agentMap does not contain ANY_CFENGINE, I believe you wanted to do something like: 
- delete agentTypes
- val agentMap = ( (ANY_CFENGINE, "Any CFEngine based agent") :: AgentType.allValues.toList.map(a => (a.id, a :: Nil))).toMap

VinceMacBuche: Was on first version then eclipse havoc ... adding it back

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -406,32 +406,38 @@ case object OsNameComparator extends CriterionType {
...
   override def comparators = Seq(Equals, NotEquals)
   override protected def validateSubCase(v:String,comparator:CriterionComparator) = {
     if(null == v || v.length == 0) Failure("Empty string not allowed") else Full(v)
   }
   override def toLDAP(value:String) = Full(value)
 
-  private[this] def filterAgent(agent: String) = SUB(A_AGENTS_NAME, null, Array(agent), null)
+  private[this] def filterAgent(agent: AgentType) =
+      SUB(A_AGENTS_NAME, null, Array(s""""agentType":"${agent.id}""""), null) ::
+      EQ(A_AGENTS_NAME, agent.oldShortName) ::
```
fanf: Could you add a comment here so that we remember why we need to keep old short name here?

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #1754, opened 2017-10-06
Fixes #11244: Groups are not reloaded after restoring from archive

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11244

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
@@ -56,17 +56,13 @@ case object DelayedUpdate
 //a container to hold the list of dynamic group to update
 case class GroupsToUpdate(ids:Seq[NodeGroupId]) extends HashcodeCaching
 
-
 sealed trait DynamicGroupUpdaterStates //states into wich the updater process can be
 //the process is idle
-case object IdleUdater extends DynamicGroupUpdaterStates
+case object IdleGroupUpdater extends DynamicGroupUpdaterStates
 //an update is currently running for the given nodes
 case class StartDynamicUpdate(id:Long, modId:ModificationId, started: DateTime, groupIds:GroupsToUpdate) extends DynamicGroupUpdaterStates with HashcodeCaching
 //the process gave a result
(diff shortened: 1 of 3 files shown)


VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
```diff
@@ -174,76 +169,72 @@ class UpdateDynamicGroups(
...
+        // If one update is pending, immediately start a new group update
+        // We should no deploy during this run, but we still need to compute if we need a deployment for this group update
+        // Deployment will be launched when the pending group update will be finished (and even if it does not need any deployment)
+        val delayDeployment = onePending
 
-        // If one update is pending, immediatly start one other
-        // It should not clear the needDeployement
+        // Maybe should be done at the end, when we know it's ok to start deployment ...
         if(onePending) {
-          onePending = false
```
ncharles: i'm pretty sure this will break https://www.rudder-project.org/redmine/issues/10446

VinceMacBuche: Kept the logic of retrigger deployment, which prevents policy generation if there is a pending update 

Removed local var never reassigned in scope replaced by a val

Removed a needdeployment set to false, which only make you lose a deployment if last update does not change anything 

Even if needdeployment is true, it will be prevented by retrigger deployment or the updater not being idle

Can you point me something that would break it ? I will happily look into it 

I'll also try on Monday

fanf: Could you share the test results? The PR seems ok.

VinceMacBuche: I tested to accepted a hundred of Nodes with 4-5  dynamic groups with regexp.
Right All Nodes are accepted and generation starts, it appears that all groups contains all nodes they should contain (if running search, no more nodes appear, same when running group update)

And only one generation is started (after all Nodes accepted)

fanf:
OK, merging this PR

PR #1757, opened 2017-10-13
Fixes #11449: Allow variable as directive parameters in regexp validated inputs 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11449

### DIFF
--- a/rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
+++ b/rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
@@ -53,14 +53,14 @@ case class RegexConstraint(pattern: String, errorMsg: String) extends HashcodeCa
 
   val compiled = Pattern.compile(pattern)
 
+  val variablePattern = Pattern.compile(".*?\\$\\{.*?\\}.*?")
+
   /* throw a ConstraintException if the value doesn't match the pattern */
-  def check(varValue: String, varName: String) : Box[String ]=
-    if(!compiled.matcher(varValue).matches) {
-      Failure("%s%s".format(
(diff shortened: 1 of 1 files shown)


VinceMacBuche:
Close for now, this needs to be tested

rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
```diff
@@ -53,14 +53,14 @@ case class RegexConstraint(pattern: String, errorMsg: String) extends HashcodeCa
 
   val compiled = Pattern.compile(pattern)
 
+  val variablePattern = Pattern.compile(".*\\$\\{.*\\}.*")
```
ncharles: i fear it may be a bit too eager

ncharles: well, actually, i can't see why it should be an issue

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
VinceMacBuche: Yes we don't want to analyze it deeply do it does not need to be perfect matches for cases i tested ... (one var, two vars, var inside a var ... )

VinceMacBuche: I updated the regexp so it's less eager (added ? after * so it matches as less possible)

VinceMacBuche:
Commit modified

PR #1760, opened 2017-10-18
Fixes #5203: There is no way to remove a Directive based on a Technique that no longer exists

fanf:
See https://www.rudder-project.org/redmine/issues/5203#note-47 for screenshot of the proposed solution.


VinceMacBuche:
OK, merging this PR

PR #1767, opened 2017-10-26
Fixes #11667: Remove remaining use of spring-jdbc

fanf:
https://www.rudder-project.org/redmine/issues/11667

No more spring-jdbc \o/

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
@@ -52,7 +52,6 @@ import com.normation.rudder.domain.policies.RuleId
 import com.normation.rudder.domain.policies.DirectiveId
 import com.normation.inventory.domain.NodeId
 import net.liftweb.common._
-import scala.language.implicitConversions
 import com.normation.rudder.services.reports.RunAndConfigInfo
 import org.slf4j.LoggerFactory
 import doobie.util.log.ExecFailure
@@ -118,10 +117,14 @@ object Doobie {
     case \/-(a) => Full(a)
   }
   implicit class XorToBox[A](res: \/[Throwable, A]) {
-    def box = res match {
-      case -\/(e) => Failure(e.getMessage, Full(e), Empty)
-      case \/-(a) => Full(a)
-    }
+    def box = xorToBox(res)
+  }
+  implicit def xorBoxToBox[A](res: \/[Throwable, Box[A]]): Box[A] = res match {
+    case -\/(e) => Failure(e.getMessage, Full(e), Empty)
+    case \/-(a) => a
+  }
+  implicit class XorBoxToBox[A](res: \/[Throwable, Box[A]]) {
+    def box = xorBoxToBox(res)
   }
 
   /*
--- a/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ChangeRequestJdbcRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ChangeRequestJdbcRepository.scala
@@ -41,105 +41,65 @@ import net.liftweb.common.Loggable
 import com.normation.rudder.repository.RoChangeRequestRepository
 import org.joda.time.DateTime
 import net.liftweb.common._
-import java.sql.ResultSet
-import java.sql.Connection
-import java.sql.PreparedStatement
-import java.sql.Timestamp
-import org.springframework.jdbc.core.JdbcTemplate
 import com.normation.rudder.domain.workflows.ChangeRequest
-import org.springframework.jdbc.core.RowMapper
 import com.normation.rudder.domain.workflows.ConfigurationChangeRequest
 import com.normation.rudder.domain.workflows.ChangeRequestId
 import com.normation.rudder.domain.workflows.ChangeRequestInfo
-import scala.util.{Try, Failure => Catch, Success}
-import scala.collection.JavaConversions._
-import com.normation.rudder.domain.policies.DirectiveId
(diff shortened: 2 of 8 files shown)


fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
```diff
@@ -123,6 +122,16 @@ object Doobie {
...
   }
+  implicit def xorBoxToBox[A](res: \/[Throwable, Box[A]]): Box[A] = res match {
+    case -\/(e) => Failure(e.getMessage, Full(e), Empty)
+    case \/-(a) => a
+  }
+  implicit class XorBoxToBox[A](res: \/[Throwable, Box[A]]) {
+    def box = res match {
+      case -\/(e) => Failure(e.getMessage, Full(e), Empty)
+      case \/-(a) => a
+    }
```
ncharles: both are surprinsignly similar - is there a reason to have both ?

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ChangeRequestJdbcRepository.scala
```diff
@@ -41,109 +41,65 @@ import net.liftweb.common.Loggable
...
-    } match {
-      case Success(x) => x match {
-        case Seq() => Full(None)
-        case Seq(res) => Full(res.headOption)
-        case _ => Failure(s"Too many change request have the same id ${changeRequestId.value}")
-      }
-      case Catch(error) => Failure(error.toString())
-    }
+    val q = Query[ChangeRequestId, Box[ChangeRequest]](SELECT_SQL + " where id = ?", None).toQuery0(changeRequestId)
+    q.option.map(_.map(_.toOption)).transact(xa).unsafePerformSync
```
ncharles: this is so much neater

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ChangeRequestJdbcRepository.scala
```diff
@@ -180,89 +136,67 @@ class RoChangeRequestJdbcRepository(
...
+  private[this] def getChangeRequestsByXpathContent(xpath:String, shouldEquals:String, errorMessage:String, onlyPending:Boolean): Box[Vector[ChangeRequest]] = {
+
+    val q = {
+      if (onlyPending) {
+        (Fragment.const(s"""${SELECT_SQL_JOIN_WORKFLOW} where cast( xpath('${xpath}', content) as character varying[])""") ++ sql""" = ${Array(shouldEquals)} and state like 'Pending%'""").query[Box[ChangeRequest]]
+      } else {
+        (Fragment.const(s"""${SELECT_SQL} where cast( xpath('${xpath}', content) as character varying[])""") ++ sql""" = ${Array(shouldEquals)}""").query[Box[ChangeRequest]]
+      }
     }
+    println(q.sql)
```
ncharles: extra println here

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/WorkflowJdbcRepository.scala
```diff
@@ -38,166 +38,73 @@ package com.normation.rudder.repository.jdbc
...
+        exists  <- sql"""select state from workflow where id = ${crId}""".query[WorkflowNodeId].option
+        created <- exists match {
+                     case Some(s) =>
+                       if(s == from) {
+                         sql"""update workflow set state = ${state} where id = ${crId}""".update.run.attempt
+                       } else {
+                         (-\/(s"Cannot change status of ChangeRequest '${crId.value}': it has the status '${s.value}' "+
+                              s"but we were expecting '${from.value}'. Perhaps someone else changed it concurently?").pure[ConnectionIO])
+                       }
+                     case None    => (-\/(s"Cannot start a workflow for Change Request id ${crId.value}, "+
```
ncharles: wrong text: Cannot change status of workflow for Change Request id, as it is not part of any workflow yet

rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/WorkflowJdbcRepository.scala
```diff
@@ -38,166 +38,73 @@ package com.normation.rudder.repository.jdbc
...
+      } yield {
+        state
       }
-    } match {
-      case Success(x) => x
-      case Catch(error) =>
-        logger.error(s"Error when updating status of change request ${crId.value} : ${error.toString()}")
-        Failure(error.toString())
     }
+    process.attempt.transact(xa).unsafePerformSync
```
ncharles: changes in this file are awesome !

rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
VinceMacBuche: box method is never used, are you sure you need to define the implicit class ?

If so, Iwould do "def box = xorBoxToBox(res)"  to prevent code duplication

VinceMacBuche:
apart question on the implicit class, this can be merged (and in the worst case it's an unused utility ...)

rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
fanf: Both are similar but there is a Box more in param for the second. I need both, and I could express one with the other at the price of a class instance+deconstruction more (for nothing). I can express classes from methods.

fanf:
PR rebased

PR #1774, opened 2017-11-06
Arch 11699/clean up dead code

fanf:
https://www.rudder-project.org/redmine/issues/11699


rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -781,7 +781,7 @@ class GitTechniqueReader(
    * need to keep track of there relative id
    */
   private[this] def getTechniquePath(techniqueInfos:TechniquesInfo) : Set[(TechniquePath, TechniqueId)] = {
-    var set = scala.collection.mutable.Set[(TechniquePath, TechniqueId)]()
+    val set = scala.collection.mutable.Set[(TechniquePath, TechniqueId)]()
```
VinceMacBuche: funny a var mutable set ...

rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -171,7 +171,7 @@ trait TStringComparator extends CriterionType {
       comparator match {
         case Regex | NotRegex =>
           try {
-            val ok = java.util.regex.Pattern.compile(v) //yes, ok is not used, side effect are fabulous
+            val _ = java.util.regex.Pattern.compile(v) //yes, "_" is not used, side effect are fabulous
```
VinceMacBuche: Why affecting it then ? and not just 

`java.util.regex.Pattern.compile(v)`

fanf:
OK, merging this PR

PR #1775, opened 2017-11-07
Fixes #11705: Remove the use of serial in rules

fanf:
https://www.rudder-project.org/redmine/issues/11705


VinceMacBuche:
Awesome PR, just the modification on Technique that should not be here

rudder-core/src/main/scala/com/normation/cfclerk/domain/Technique.scala
```diff
@@ -77,6 +77,42 @@ final case class AgentConfig(
   , bundlesequence : Seq[BundleName]
 )
 
+
+/**
+ * A type that tells if the Technique supports directive by directive
+ * generation or not.
+ */
+sealed trait TechniqueGenerationMode {
```
VinceMacBuche: That should be done in the PR that add the new generation mode

rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
```diff
@@ -56,6 +56,7 @@ import com.normation.rudder.services.reports.RunAndConfigInfo
 import org.slf4j.LoggerFactory
 import doobie.util.log.ExecFailure
 import doobie.util.log.ProcessingFailure
+import scala.language.implicitConversions
```
ncharles: this seems unused

ncharles:
This is awesome :)

rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
fanf: it's necessary for using the keyword without warning with -Xlint compiler flag

fanf:
I believe we can close that one since #1780 is there, andsuperseed that PR.

PR #1776, opened 2017-11-07
Fixes #11706: Missing test resources in rudder-rest

fanf:
https://www.rudder-project.org/redmine/issues/11706


fanf:
Self merging as it is a no brainer

PR #1779, opened 2017-11-15
Fixes #11731: Add a new reason why rudder might not be loaded

peckpeck:
https://www.rudder-project.org/redmine/issues/11731


peckpeck:
OK, merging this PR

PR #1780, opened 2017-11-15
Fixes #10625: Don't merge directive from same technique on generation

fanf:
https://www.rudder-project.org/redmine/issues/10625

So, there is a lot of things going on, but it's globally a HUGE simplification of the generation process - and just the first step, some more will be done with the node by node generation. 
The state are bad because I added tests with all the generated files, what also can mean you can ignore what is added in "expected-share" (or start to look at them to get a general filling of what will be the result of multi-policy generation). 

The mains things remaining to do before merge: 
- perhaps find a clearer way to manage unicity of target policy directory name. Here, I need to patch path for bundle "input file", template dest file, resource dest file... Quite tedious and error prone. 
- clean up expected reports.
But these two point can go in a latter refactoring for node by node generation.

The main novelty is a big cleaning in data structure for generation. Now they are grouped in file policies/DataStructure.scala, and the naming make sense regarding the step of the generation process (we starts with ParsedPolicyDraft where (tada!) directive parameters are parsed but not bound yet, then we have BoundPolicyDraft where (tada!) directive parameter are expanded, and then we have Policy where all the sanity checking (technique unicity, merge directive etc) is done. 
That also mean that there is NO MORE CF3POLICYDRAFTCONTAINER !!!!! It tooks 5 years, but I finaly removed it :)

To start, interesting files are: 
- rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
- rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala (in particular method "buildPolicy")


rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -0,0 +1,469 @@
+/*
+*************************************************************************************
+* Copyright 2011 Normation SAS
```
ncharles: 2017

VinceMacBuche:
You should rebase your PR @fanf and make it one commit

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

PR #1781, opened 2017-11-17
Fixes #10599: Impossible to search or build groups based on JSON values in node properties

fanf:
https://www.rudder-project.org/redmine/issues/10599
This PR allows to corretly process JSON values stored in LDAP for the three kind of object needed to be search for in group: Process, Environment variables, and Node properties. 
The first commit split the common JsonComparator in three specialized one, one for each case, which allows to simplify the simple case, and make a special case for node properties. 

The strategie to correctly handle node properties is to make precise filter for eq/not eq for both quoted and quoted values. It means that we are now extremelly dependant of serialization format (we already were, now it is consumed). 

And for regex case, we had a pre-regex JSON parsing which allows to filter out other field than name/value so that we are sure to not return false positive with provider case. 

That last technique can be adoted in the future to create a real JSON query in place of regex.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -56,9 +56,7 @@ import net.liftweb.json._
 import JsonDSL._
 import com.normation.exceptions.TechnicalException
 import com.normation.utils.HashcodeCaching
-import com.normation.rudder.services.queries.RegexFilter
-import net.liftweb.common.EmptyBox
-import com.normation.rudder.services.queries.NotRegexFilter
+import com.normation.rudder.services.queries._
 
 sealed trait CriterionComparator {
   val id:String
@@ -137,8 +135,8 @@ sealed trait CriterionType  extends ComparatorList {
   //transform the given value to its LDAP string value
   def toLDAP(value:String) : Box[String]
 
-  def buildRegex(attribute:String,value:String): Box[RegexFilter] = Full(RegexFilter(attribute,value))
-  def buildNotRegex(attribute:String,value:String): Box[NotRegexFilter] = Full(NotRegexFilter(attribute,value))
+  def buildRegex(attribute:String,value:String): Box[RegexFilter] = Full(SimpleRegexFilter(attribute,value))
+  def buildNotRegex(attribute:String,value:String): Box[NotRegexFilter] = Full(SimpleNotRegexFilter(attribute,value))
 
   //build the ldap filter for given attribute name and comparator
   def buildFilter(attributeName:String,comparator:CriterionComparator,value:String) : Filter =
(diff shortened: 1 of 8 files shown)


rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -436,74 +436,194 @@ case object EditorComparator extends CriterionType {
...
+ *
+ * JsonFixedKeyComparator("my_serialized_data", "key2", false) => will search for value in "value2"
+ *
+ * Used for "process" attribute
+ */
+case class JsonFixedKeyComparator(ldapAttr:String, jsonKey: String, numericvalue: Boolean) extends TStringComparator with Loggable {
+  override val comparators = BaseComparators.comparators
+
+  def format(attribute:String, value:String) = {
+    val v = if (numericvalue) value else s""""$value""""
```
VinceMacBuche: is numeric value treating boolean and  maybe other json type (array ? object ?)

fanf: it's a very special use case, dedicated to process, and we don't have anything else than numeric. But I could call it "quoteValue", it would make more sense, you are right

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/test/resources/logback-test.xml
```diff
@@ -43,7 +43,7 @@
   </logger>
   -->
   
-  <logger name="com.normation.rudder.services.queries" level="info" /> 
+  <logger name="com.normation.rudder.services.queries" level="trace" />
```
VinceMacBuche: I think you want to go back to info!

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -594,8 +597,10 @@ class InternalLDAPQueryProcessor(
     val start = (Set[Filter](), Set[SpecialFilter]())
     Full((start /: filters) {
       case (  (ldapFilters,specials), LDAPFilter(f) ) => (ldapFilters + f, specials)
-      case (  (ldapFilters,specials), r:RegexFilter ) => (ldapFilters, specials + r)
-      case (  (ldapFilters,specials), r:NotRegexFilter ) => (ldapFilters, specials + r)
+      case (  (ldapFilters,specials), r:RegexFilter                ) => (ldapFilters, specials + r)
```
VinceMacBuche: SHould be SimpleFiler or else NodePropertyRegexFilter will not be reached

VinceMacBuche: or reverse order of march to have RegexFilter at the end

VinceMacBuche: Or since they seems to be the same, remove NodePropertyRegexFilter

rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -67,13 +68,17 @@ sealed trait ExtendedFilter
 final case class LDAPFilter(f:Filter) extends ExtendedFilter with HashcodeCaching
 
 //special ones
-sealed trait SpecialFilter extends ExtendedFilter
-final case class RegexFilter(attributeName:String, regex:String) extends SpecialFilter with HashcodeCaching
-final case class NotRegexFilter(attributeName:String, regex:String) extends SpecialFilter with HashcodeCaching
+sealed trait SpecialFilter  extends ExtendedFilter
+sealed trait RegexFilter    extends SpecialFilter
```
VinceMacBuche: Should the trait have attributeName and regex as def ? overriden by the two case classe to refactor some code ?

fanf: we don't use that, and using it would mean even more refactoring I think. So I'm for letting things as they are, and wait for the addition of a JSON-query feature to see what need to be speficialized and how.

fanf:
PR rebased

fanf:
OK, merging this PR

PR #1784, opened 2017-11-19
Fixes #11749: Add ncf logic to create file into Rudder instead of ncf

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11749

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
@@ -1,0 +1,85 @@
+/*
+*************************************************************************************
+* Copyright 2017 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
+* distributed under the license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* Rudder is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+*
+*************************************************************************************
+*/
+
+package com.normation.rudder.ncf
+
+import com.normation.inventory.domain.Version
+import com.normation.inventory.domain.AgentType
+
+object NcfId {
+  implicit class DscTechniqueEncoder(id : NcfId) {
+    val validDscName = {
+      id.value.split("_").map(_.capitalize).mkString("-")
+    }
+  }
+}
+
+sealed trait NcfId {
+  def value : String
+}
+final case class BundleName (value : String) extends NcfId
+
+final case class ParameterId (value : String) extends NcfId
+
+final case class Technique(
+    bundleName  : BundleName
+  , name        : String
+  , methodCalls : Seq[MethodCall]
+  , version     : Version
+  , description : String
+)
+
+final case class MethodCall(
+    methodId   : BundleName
+  , parameters : Map[ParameterId,String]
+  , condition  : String
+)
+
+final case class GenericMethod(
+    id             : BundleName
+  , name           : String
+  , parameters     : Seq[Parameter]
+  , classParameter : ParameterId
+  , classPrefix    : String
+  , agentSupport   : Seq[AgentType]
+  , description    : String
+)
+
+final case class Parameter(
+    id          : ParameterId
+  , description : String
+)
--- a/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -1,0 +1,383 @@
+/*
+*************************************************************************************
+* Copyright 2017 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
+* distributed under the license agreement of your choice.
+* A "Related Module" means a set of sources files including their
+* documentation that, without modification of the Source Code, enables
+* supplementary functions or services in addition to those offered by
+* the Software.
+*
+* Rudder is distributed in the hope that it will be useful,
+* but WITHOUT ANY WARRANTY; without even the implied warranty of
+* MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+* GNU General Public License for more details.
+*
+* You should have received a copy of the GNU General Public License
+* along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+*
+*************************************************************************************
+*/
+
+package com.normation.rudder.ncf
+
+import net.liftweb.common.Loggable
+import java.nio.file.Paths
+import java.nio.file.Files
+import scala.xml.NodeSeq
+import java.nio.charset.StandardCharsets
+import com.normation.inventory.domain.AgentType
+import net.liftweb.common.Box
+import com.normation.rudder.repository.xml.GitArchiverUtils
+import com.normation.cfclerk.services.GitRepositoryProvider
+import java.io.File
+import com.normation.rudder.repository.xml.RudderPrettyPrinter
+import com.normation.rudder.repository.GitModificationRepository
+import com.normation.rudder.repository.xml.GitArchiverFullCommitUtils
+import com.normation.eventlog.ModificationId
+import org.eclipse.jgit.lib.PersonIdent
+import com.normation.rudder.repository.GitPath
+import net.liftweb.common.Full
+import com.normation.eventlog.EventActor
+import com.normation.rudder.services.user.PersonIdentService
+import scala.xml.{ Node => XmlNode }
+import net.liftweb.common.EmptyBox
+import org.apache.commons.io.FileUtils
+import com.normation.cfclerk.services.UpdateTechniqueLibrary
+import net.liftweb.common.Failure
+import net.liftweb.common.Empty
+
+trait NcfError {
+  def message : String
+  def exception : Option[Throwable]
+}
+
+object ResultHelper {
+  type Result[T] = Either[NcfError,T]
+
+  def execute[T](f : => T )( errorCatcher: Throwable => NcfError ) : Result[T] = {
+    try {
+      Right(f)
+    } catch {
+      case e: Throwable => Left(errorCatcher(e))
+    }
+  }
+
+  def sequence[T,U](seq : Seq[T])( f : T => Result[U]) : Result[Seq[U]] = {
+    ((Right(Seq()) : Result[Seq[U]]) /: seq) {
+      case (e @ Left(_), _) => e
+      case (Right(res), value) =>f(value).map( res :+ _)
+    }
+  }
+
+  implicit def resultToBox[T] (res : Result[T]) : Box[T] = {
+    res match {
+      case Right(r) => Full(r)
+      case Left(e) => Failure(e.message,e.exception,Empty)
+    }
+  }
+}
+
+final case class IOError(message : String, exception : Option[Throwable]) extends NcfError
+final case class TechniqueUpdateError(message : String, exception : Option[Throwable]) extends NcfError
+final case class MethodNotFound(message : String, exception : Option[Throwable]) extends NcfError
+
+class TechniqueWriter (
+    archiver         : TechniqueArchiver
+  , techLibUpdate    : UpdateTechniqueLibrary
+  , xmlPrettyPrinter : RudderPrettyPrinter
+  , basePath         : String
+) extends Loggable {
+
+  import ResultHelper._
+  private[this] var agentSpecific = new ClassicTechniqueWriter :: new DSCTechniqueWriter(basePath) :: Nil
+
+  def techniqueMetadataContent(technique : Technique, methods: Map[BundleName, GenericMethod]) : Result[XmlNode] = {
+
+    def reportingValuePerMethod (methodId: BundleName, calls :Seq[MethodCall]) : Result[Seq[XmlNode]] = {
+      methods.get(methodId) match {
+        case None => Left(MethodNotFound(s"Could not generate reporting section for method '${methodId.value}' in Technique '${technique.bundleName.value}', because we could not find the method",None))
+        case Some(method) =>
+          for {
+            spec <- sequence(calls.map(_.parameters.get(method.classParameter))) {
+              case None => Left(MethodNotFound(s"Could not find reporting values '${method.classParameter.value}' method '${methodId.value}' in Technique '${technique.bundleName.value}'",None))
+              case Some(value) => Right (<VALUE>{value}</VALUE>)
+            }
+          } yield {
+            <SECTION component="true" multivalued="true" name={method.name}>
+              <REPORTKEYS>
+                {spec}
+              </REPORTKEYS>
+            </SECTION>
+          }
+      }
+    }
+    // Regroup method calls from which we expect a reporting
+    // We filter those starting by _, which are internal methods
+    val expectedReportingMethodsWithValues =
+      for {
+        (name, methodCalls) <- technique.methodCalls.groupBy(_.methodId).toList.sortBy(_._1.value)
+        if ! name.value.startsWith("_")
+      } yield {
+        (name, methodCalls)
+      }
+
+    for {
+      reportingSection     <- sequence(expectedReportingMethodsWithValues)((reportingValuePerMethod _).tupled)
+      agentSpecificSection <- sequence(agentSpecific)(_.agentMetadata(technique, methods))
+    } yield {
+      <TECHNIQUE name={technique.name}>
+        <DESCRIPTION>{technique.description}</DESCRIPTION>
+        {agentSpecificSection}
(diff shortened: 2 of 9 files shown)


ncharles:
@VinceMacBuche could you had tests to ensure that it works as expected ? Transcribing the existing tests would probably be enough

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
```diff
@@ -0,0 +1,78 @@
...
+*************************************************************************************
+*/
+
+package com.normation.rudder.ncf
+
+import com.normation.inventory.domain.Version
+import com.normation.inventory.domain.AgentType
+
+case class BundleName (value : String) {
+  val toDsc = value.split("_").map(_.capitalize).mkString("-")
```
fanf: I think that should be externalized in an obiect ("Encoder?) with a more generic name about what is the goal of the method (toHumanName ?)

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,271 @@
...
+import com.normation.eventlog.EventActor
+import com.normation.rudder.services.user.PersonIdentService
+
+class TechniqueWriter(
+  archiver : TechniqueArchiver
+) extends Loggable {
+
+  private[this] var agentSpecific = new ClassicTechniqueWriter :: new DSCTechniqueWriter :: Nil
+
+  def commonMetadata(technique : Technique, methods: Map[BundleName, GenericMethod]) = {
```
fanf: please add types on public sig. It helps both the reader and the compiler speed.

fanf: (and it's interesting to know it's an XMLNode, not a nodeseq, because transformation on simple node are simpler)

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,271 @@
...
+            </REPORTKEYS>
+          </SECTION>
+
+        }
+      }
+      </SECTIONS>
+    </TECHNIQUE>
+  }
+
+  def writeAll(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, commiter : EventActor) : Box[Unit] = {
```
fanf: I believe we would gain a lot here with a dedicated error type and custom cases, ie:
sealed trait MetadataGenError extends Exception  //still wondering if extending exception is the correct thing to do, but we need at least a string description and most of the time, it is handy to be able to capture a stack trace from elsewhere)
case object IOError(msg: String, ex: Throwable) extends MetadataGenError ...

writeTechnique(...): Either[MetadataGenError, Unit] = { ... }

And we will handle toBox at the API level. 

It will be more understandable for a reader (and even in logs), and we won't as easely miss cases where the reason of the error get lost. For ex, if an exception happen in writeTechnique below, I think it can quite hard to know what caused the "writeAll" error. If it's a "WriteTechniqueIOError(msg)", quite obvious.

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,271 @@
...
+    </AGENT>
+  }
+
+}
+
+class DSCTechniqueWriter extends AgentSpecificWriter{
+
+  val genericParams = "-reportId $reportId -techniqueName $techniqueName -auditOnly:$auditOnly"
+
+  def writeTechniqueFile(technique : Technique, methods : Map[BundleName, GenericMethod] ) = {
```
fanf: indent doesn't seem right below ?

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,271 @@
...
+
+  def methodCall(call : MethodCall) = {
+    methods.get(call.methodId) match {
+      case Some(method) if method.agentSupport.contains(AgentType.Dsc) =>
+        call.condition match {
+          case "any" =>
+            writeCall(method, call)
+          case _ =>
+
+         s"""
```
fanf: would you mind use "|" + .stripMargin ? I find it quite hard to read here. Moreover, you are adding a bunch of spaces in the last line, I'm almost sure it will cause troubles when testing.

rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
```diff
@@ -869,17 +873,20 @@ object RudderConfig extends Loggable {
...
 
+  val apiV9 = ncfAPI :: apiV8
+
   val apis = {
     Map (
-        //Rudder 3.0
+        //Rudder 3.0D
         ( ApiVersion(5,true) -> apiV5 )
         //Rudder 3.1
       , ( ApiVersion(6,true) -> apiV6 )
```
fanf: we forgot to remove that in 4.2, don't we?

rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
```diff
@@ -869,17 +873,20 @@ object RudderConfig extends Loggable {
...
         //Rudder 3.2
       , ( ApiVersion(7,false) -> apiV7 )
```
fanf: we forgot to put v7 deprecated, don't we?

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -990,4 +957,63 @@ case class RestExtractorService (
 
   def extractId[T] (req : Req)(fun : String => Full[T])  = extractString("id")(req)(fun)
 
+  def extractNcfTechnique (json : JValue, methods: Map[BundleName, GenericMethod]) : Box[NcfTechnique] = {
```
fanf: below things make me sad :(

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
VinceMacBuche: What to you think of the way i choose ?

VinceMacBuche: Regrouping the case clases into a trait, then adding an implciit class to have the encoder method when needed

I think i can regroup BundleName and ParameterId into a single case class ... but having two separate case class allow to have different semantic on the classe (which here is interesting, since i don't want to use a BundleName as a Key for my parameters ... ) but they have the same transformations and conditions

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
VinceMacBuche: Not done yet, in next commit

rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
VinceMacBuche: You done it in another PR

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
VinceMacBuche: It's not that sad and was quite quick to do ... But yes it can be not done at all with other frameworks ...

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
VinceMacBuche: Yes, totally ok with this, jsut i did not know the type i would return so wanted to add it at the end ...

I think i should encapsulate this in an error type

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,272 @@
+/*||
```
ncharles: why a || there ?

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,272 @@
...
+  def writeTechniqueFile(technique : Technique, methods : Map[BundleName, GenericMethod] ) : Box[Option[String]]
+
+  def metadataContent(technique : Technique, methods : Map[BundleName, GenericMethod] ) : NodeSeq
+}
+
+class ClassicTechniqueWriter extends AgentSpecificTechniqueWriter {
+
+  def writeTechniqueFile(technique : Technique, methods : Map[BundleName, GenericMethod]) = Full(None)
+  def metadataContent(technique : Technique, methods : Map[BundleName, GenericMethod]) = {
+    val completeReporting = technique.methodCalls.exists(m => m.condition != "any" || m.condition != "cfengine-community" )
```
ncharles: this name implies that the reporting is complete, which is the opposite of what it needs: the reporting need to be completed.
what about "val extraReporting" ?

ncharles: or incompleteReporting

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,319 @@
...
+import com.normation.eventlog.ModificationId
+import org.eclipse.jgit.lib.PersonIdent
+import com.normation.rudder.repository.GitPath
+import net.liftweb.common.Full
+import com.normation.eventlog.EventActor
+import com.normation.rudder.services.user.PersonIdentService
+import scala.xml.{ Node => XmlNode }
+import net.liftweb.common.EmptyBox
+
+trait Error {
```
fanf: ncfError please

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,319 @@
...
+
+    val path = s"/var/rudder/configuration-repository/${metadataPath}"
+    for {
+      content <- techniqueMetadataContent(technique, methods)
+      _ <- execute{
+             val filePath = Paths.get(path)
+             if (!Files.exists(filePath))
+               Files.createFile(filePath)
+             scala.xml.XML.save(path, content)
+           } {case e => IOError("",Some(e))}
```
fanf: never ever an empty message. It is what we will always give to the user, and need to explain the "business" pb.

rudder-core/src/main/scala/com/normation/rudder/ncf/package.scala
```diff
@@ -0,0 +1,70 @@
...
+*/
+
+package com.normation.rudder
+
+import net.liftweb.common.Box
+import net.liftweb.common.Full
+import net.liftweb.common.Failure
+import net.liftweb.common.Empty
+
+package object ncf {
```
fanf: please, don't put that in a package object. 
We need to put it in a Real object, like "NcfBoxBinding" or something. And we will import it where needed. 
Whithout that, it brings havok to the implicit scope. 
And package object are a mess, and will be removed in Scala 2.14.

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
VinceMacBuche: Just one i missed :(

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,373 @@
...
+      for {
+        (name, methodCalls) <- technique.methodCalls.groupBy(_.methodId).toList.sortBy(_._1.value)
+        if ! name.value.startsWith("_")
+      } yield {
+        (name, methodCalls)
+      }
+
+     for {
+       reportingSection     <- sequence(expectedReportingMethodsWithValues)((reportingValuePerMethod _).tupled)
+       agentSpecificSection <- sequence(agentSpecific)(_.agentMetadata(technique, methods))
```
fanf: align please

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,373 @@
...
+       </TECHNIQUE>
+     }
+  }
+
+  // Write and commit all techniques files
+  def writeAll(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, committer : EventActor) : Result[Seq[String]] = {
+    for {
+      agentFiles <- writeAgentFiles(technique, methods, modId, committer)
+      res <- writeMetadata(technique, methods, modId, committer)
+      _ <- techLibUpdate.update(modId, committer, Some(s"Update Technique library after creating files for ncf Technique ${technique.name}")) match {
```
fanf: align please

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,373 @@
...
+      res +: agentFiles
+    }
+  }
+
+  def writeAgentFiles(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, commiter : EventActor) : Result[Seq[String]] = {
+    for {
+      // Create/update agent files, filter None
+      files  <- sequence(agentSpecific)(_.writeAgentFile(technique, methods)).map(_.flatten)
+      _ <- sequence(files)(archiver.commitFile(technique, _, modId, commiter , s"Commiting Technique '${technique.bundleName.value}' file for agent " ))
+    } yield {
```
fanf: align please

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,373 @@
...
+  }
+
+  def writeMetadata(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, commiter : EventActor) : Result[String] = {
+
+    val metadataPath = s"techniques/ncf_techniques/${technique.bundleName.value}/${technique.version.value}/metadata.xml"
+
+    val path = s"${basePath}/${metadataPath}"
+    for {
+      content <- techniqueMetadataContent(technique, methods)
+      formattedContent = xmlPrettyPrinter.format(content)
```
fanf: align please

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,373 @@
...
+            |
+            |}""".stripMargin('|')
+
+      techPath = techniquePath(technique)
+      path <- execute (
+                Paths.get(s"${basePath}/${techPath}")
+              ) ( e =>
+                IOError(s"Could not find dsc Technique '${technique.name}' in path ${basePath}/${techPath}", Some(e))
+              )
+      _ <-    execute{
```
fanf: align please

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -0,0 +1,373 @@
...
+  Loggable with
+  GitArchiverUtils with TechniqueArchiver{
+
+  import ResultHelper._
+  override val encoding : String = "UTF-8"
+
+  def commitFile(technique : Technique, gitPath : String, modId: ModificationId, commiter:  EventActor, msg : String) : Result[Unit] = {
+    (for {
+      ident <- personIdentservice.getPersonIdentOrDefault(commiter.name)
+      commit  <- commitAddFile(modId,ident, gitPath, msg)
```
fanf: align please

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -990,4 +957,63 @@ case class RestExtractorService (
...
+      NcfTechnique(bundleName, name, calls, new Version(version), description)
+    }
+  }
+
+  def extractMethodCall (json : JValue, methods: Map[BundleName, GenericMethod]) : Box[MethodCall] = {
+
+    for {
+      methodId  <- CompleteJson.extractJsonString(json, "method_name", s => Full(BundleName(s)))
+      condition <- CompleteJson.extractJsonString(json, "class_context")
+      parameterValues  <- CompleteJson.extractJsonArray(json \ "args"){
```
fanf: align please

rudder-web/src/main/scala/com/normation/rudder/web/rest/RestExtractorService.scala
```diff
@@ -990,4 +957,63 @@ case class RestExtractorService (
...
+  }
+
+  def extractGenericMethod (json : JValue) : Box[List[GenericMethod]] = {
+    CompleteJson.extractJsonArray(json \ "methods") { json =>
+      for {
+        bundleName  <- CompleteJson.extractJsonString(json, "bundle_name", s => Full(BundleName(s)))
+        description <- CompleteJson.extractJsonString(json, "description")
+        name        <- CompleteJson.extractJsonString(json, "name")
+        classPrefix <- CompleteJson.extractJsonString(json, "class_prefix")
+        classParameter  <- CompleteJson.extractJsonString(json, "class_parameter", a => Full(ParameterId(a)))
```
fanf: align please

rudder-web/src/main/scala/com/normation/rudder/web/rest/ncf/NcfApi9.scala
```diff
@@ -0,0 +1,125 @@
...
+  }
+
+  def requestDispatch(apiVersion: ApiVersion) : PartialFunction[Req, () => Box[LiftResponse]] = {
+    case Nil JsonPost body -> req => {
+      val modId = ModificationId(uuidGen.newUuid)
+      val actor = RestUtils.getActor(req)
+      val response = for {
+        json <- req.json ?~! "No JSON data sent"
+        methods <- restExtractor.extractGenericMethod(json)
+        methodMap = methods.map(m => (m.id,m)).toMap
```
fanf: align please

rudder-web/src/main/scala/com/normation/rudder/web/rest/ncf/NcfApi9.scala
```diff
@@ -0,0 +1,125 @@
...
+      actionResp(Full(wrapper), req, "Could not update ncf technique")("UpdateTechnique")
+    }
+
+    case Nil JsonPut body -> req => {
+      val modId = ModificationId(uuidGen.newUuid)
+      val actor = RestUtils.getActor(req)
+      val response = for {
+        json <- req.json ?~! "No JSON data sent"
+        methods <- restExtractor.extractGenericMethod(json)
+        methodMap = methods.map(m => (m.id,m)).toMap
```
fanf: align please

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
OK, merging this PR

PR #1789, opened 2017-11-22
Fixes #11760: Move all API v1 into their own package

fanf:
https://www.rudder-project.org/redmine/issues/11760


fanf:
PR rebased

PR #1792, opened 2017-11-23
Fixes #10719: Remove the skipidentify variable and settings page entry

ncharles:
https://www.rudder-project.org/redmine/issues/10719

### DIFF
--- a/rudder-rest/src/main/scala/com/normation/rudder/rest/settings/SettingsApi.scala
+++ b/rudder-rest/src/main/scala/com/normation/rudder/rest/settings/SettingsApi.scala
@@ -72,7 +72,21 @@ import com.normation.rudder.AuthorizationType
 
 trait SettingsApi extends RestAPI {
   val kind = "settings"
-  def userService : UserService
+  implicit def userService : UserService
+
+  def restExtractor        : RestExtractorService
+  def configService        : ReadConfigService with UpdateConfigService
+  def asyncDeploymentAgent : AsyncDeploymentAgent
+  def uuidGen              : StringUuidGenerator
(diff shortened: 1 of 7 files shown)


rudder-rest/src/main/scala/com/normation/rudder/rest/settings/SettingsApi.scala
```diff
@@ -521,12 +495,88 @@ extends SettingsApi with Loggable {
...
+class SettingsAPI8(
+    override val restExtractor        : RestExtractorService
+  , override val configService        : ReadConfigService with UpdateConfigService
+  , override val asyncDeploymentAgent : AsyncDeploymentAgent
+  , override val uuidGen              : StringUuidGenerator
+) ( override implicit val userService : UserService )
+extends SettingsApi with Loggable {
+
+
+  val allSettings =
```
VinceMacBuche: I would have put the list in the SettingsAPI in a val (baseSettings) and remove or add skipidentify

I would go for baseSettings to be api v9 list 

in v9 allSettings = baseSettings
in v8 allSettings = baseSettings :+ RestUseReverseDNS

VinceMacBuche:
Apart the small refactoring remark, this is great ! thanks Nicolas !

rudder-rest/src/main/scala/com/normation/rudder/rest/settings/SettingsApi.scala
ncharles: Make sense ! Thank you

ncharles:
Commit modified

PR #1794, opened 2017-11-24
Fixes #11777: Lift-webkit has a dependency toward rhino (which conflicts with JS param eval)

fanf:
https://www.rudder-project.org/redmine/issues/11777


fanf:
OK, merging this PR

PR #1795, opened 2017-11-24
Fixes #11445: Node properties syntax is not the same for Windows agent

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11445


VinceMacBuche:
PR rebased

PR #1799, opened 2017-11-24
Fixes #11785: There's an "UTF-8"s in place of "UTF-8" in a method parameter

fanf:
https://www.rudder-project.org/redmine/issues/11785


fanf:
OK, merging this PR

PR #1800, opened 2017-11-24
Fixes #11788: Revert exclusion of rhino - webapp fails to start

fanf:
https://www.rudder-project.org/redmine/issues/11788


fanf:
OK, merging this PR

PR #1803, opened 2017-11-28
Fixes #1411: Node (hostname,policyserver,...) modification should trigger promises regeneration

fanf:
https://www.rudder-project.org/redmine/issues/1411

Be careful to notice that only a SUBSET of inventories are observed, ie only node properties, node main information (what is in node summary), machine type. 

Other inventory information are not important aspect for policy generation (appart for groups of course, but that is covered in dynamic group update scheduled job). So it should be fine.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
@@ -1,6 +1,6 @@
 /*
 *************************************************************************************
-* Copyright 2012 Normation SAS
+* Copyright 2017 Normation SAS
 *************************************************************************************
 *
 * This file is part of Rudder.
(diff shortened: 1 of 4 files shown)


fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
```diff
@@ -0,0 +1,75 @@
...
+import net.liftweb.common.Loggable
+import scala.concurrent.duration._
+
+
+/**
+ * A naive scheduler which checks every N seconds if inventories are updated.
+ * If so, I will trigger a promise generation.
+ */
+class CheckInventoryUpdate(
+    nodeInfoCacheImple  : NodeInfoServiceCachedImpl
```
ncharles: could it be nodeInfoCacheImpl ? (without the trailing e)

rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
```diff
@@ -0,0 +1,75 @@
...
+  , asyncDeploymentAgent: AsyncDeploymentAgent
+  , uuidGen             : StringUuidGenerator
+  , updateInterval      : FiniteDuration
+) extends Loggable {
+
+  //start batch
+  if(updateInterval < 1.second) {
+    logger.info(s"Disable automatic check for node inventories (update interval less than 1s)")
+  } else {
+    logger.trace("***** starting Technique Library Update batch *****")
```
ncharles: could you also add the update interval ?

fanf: I could. I could also add a log trace to tell what happens (and allows user to know if it runs).

fanf:
PR rebased

fanf:
OK, merging this PR

PR #1804, opened 2017-11-30
Fixes #11809: Make node states behavior and name clearer 

fanf:
https://www.rudder-project.org/redmine/issues/11809

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
@@ -173,7 +173,7 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
 
       for {
         // disabled nodes are ignored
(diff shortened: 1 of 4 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -173,7 +173,7 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
 
       for {
         // disabled nodes are ignored
```
VinceMacBuche: Comment not accurate!

(but "ignored Nodes are ignored" may be not interesting ...)

fanf: yeah... not sure what I can use in place of that.

fanf: So that's ok for you?

PR #1806, opened 2017-12-01
Fixes #11771: Add a flag to regenerate all techniques on rudder webapp restart

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11771

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/api/ApiAccountRepository.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/api/ApiAccountRepository.scala
@@ -56,6 +56,8 @@ import com.normation.rudder.repository.EventLogRepository
 import com.normation.rudder.services.user.PersonIdentService
 import com.normation.eventlog.ModificationId
 import com.normation.eventlog.EventActor
+import org.joda.time.DateTime
+import com.normation.utils.StringUuidGenerator
 
 /**
  * A repository to retrieve API Accounts
@@ -65,12 +67,13 @@ trait RoApiAccountRepository {
   /**
    * Retrieve all API Account
    */
-  def getAll(): Box[Seq[ApiAccount]]
+  def getAll: Box[Seq[ApiAccount]]
 
   def getByToken(token: ApiToken): Box[Option[ApiAccount]]
 
   def getById(id : ApiAccountId) : Box[Option[ApiAccount]]
 
+  def getSystemAccount : ApiAccount
 }
 
 /**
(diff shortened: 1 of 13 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckNcfTechniqueUpdate.scala
```diff
@@ -0,0 +1,150 @@
...
+    def updateNcfTechniques : Unit = {
+      import net.liftweb.json.parse
+      import com.normation.utils.Control.sequence
+      ( for {
+        // First get generic methods
+        methodsResponse    <- tryo { methodsRequest.asString } ?~! "An error occurred while fetching generic methods from ncf API"
+        methodsResult      <- if(methodsResponse.isSuccess) {
+                                Full(methodsResponse.body)
+                              } else {
+                                scheduler.scheduleOnce(5.seconds)(updateNcfTechniques)
```
fanf: I think you should not do that here, because it mixes the "when I do that" logic with the "what is done". 
And I'm not sure it is the only case for which we want to re schedule an inventory. 
So, what I propose: 
- add a:
  sealed trait NcfTechniqueUpgradeError (msg)
  object  NcfTechniqueUpgradeError { case class NcfApiRequest(msg), WriteTechnique(msg) }

- change Box => Either[NcfTechniqueUpgradeError, T]
- decide reschedule based on the kind of error.  (plus perhaps limit the number of retry ? Not sure it makes sense, though, because without the migration, it is just broken)
+ in the case of non recoverable error, make a big error message about what is broken, and what the user must do to make the situation work again.

rudder-core/src/main/scala/com/normation/rudder/api/ApiAccountRepository.scala
```diff
@@ -65,11 +67,11 @@ trait RoApiAccountRepository {
   /**
    * Retrieve all API Account
    */
```
fanf: I don't think the API should have that boolean. In generale, when we did that, it was error prone and a pain to use and upgrade. 
More over, AFAIK, there is only one place where we need to check for system token (in rest auth). And perhaps we want to do a special log for that one, to more specificaly trace the use of that token (what client used it for ex - we will to do that kind of log at some point). 

So, I would prefer to let the API as it is for existing method, + add a "getSystemToken" (ou Tokens, not sure we plan to add more, but for now, singular seems sufficient) and update the authentication method to call both.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #1811, opened 2017-12-05
Fixes #11833: rudder_reporting is added in metadata.xml if all condition are any

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11833


fanf:
OK, merging this PR

PR #1815, opened 2017-12-11
Fixes #11837: "Agent policy mode" should be "Node policy mode" 

fanf:
https://www.rudder-project.org/redmine/issues/11837


fanf:
OK, merging this PR

PR #1818, opened 2017-12-11
Fixes #7281: Compliance keeps on loading while node is not online and "last seen" is not "never"

fanf:
https://www.rudder-project.org/redmine/issues/7281


fanf:
OK, merging this PR

PR #1820, opened 2017-12-17
Fixes #5665: Adapt Rudder to allows to have ncf technique with parameter

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/5665

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
@@ -55,7 +55,6 @@ import com.normation.rudder.services.policies.Policy
 import com.normation.rudder.services.policies.ParameterEntry
 import com.normation.cfclerk.domain.TechniqueGenerationMode
 
-
 /**
  * This file groups together everything related to building the bundle sequence and
  * related system variables.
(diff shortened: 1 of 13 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -259,7 +256,15 @@ class BuildBundleSequence(
       case Some(cfg) =>
         val techniqueBundles = cfg.bundlesequence.map { bundleName =>
           if(bundleName.value.trim.size > 0) {
-            List(Bundle(ReportId(policy.id.getReportId), bundleName))
+            val vars = if (policy.technique.generationMode == TechniqueGenerationMode.MultipleDirectivesWithParameters)
```
ncharles: could you add { } around the for, to improve readability ? I had trouble parsing that the for what linked to the if

ncharles:
It seems to me it only changes the cfengine part. Isn't there a change to make also for dsc ?

VinceMacBuche:
Commit modified

fanf:
This seems good until we don't have more complexe parameters for ncf technique (i.e: arrays, multi-valued section, etc). 
Here, we have the limitation that the parameters are written directly in the bundle sequence, and so we can't pass more complexe things than strings. 

What we want to do in the next iteration is to have all directives generated from technique from technique editor to put there parameters in a "parameters.json" file, with roughtly that format: 

```
{
  "directiveid (or reportid?) 1": { "paramName1": "simple param" , "paramName2": { "structured"} etc
, "directiveid (or reportid?) 2": { "paramName1": "simple param" , "paramName2": { "structured"} etc
}
```
And then only give the directiveid (or reportid) as parameter of the bundle in the bundle sequence.

VinceMacBuche:
OK, merging this PR

PR #1821, opened 2017-12-20
Fixes #10664: Configuration file of Rudder mention that we may change "LDAP DIT configuration", which is wrong

ncharles:
https://www.rudder-project.org/redmine/issues/10664


ncharles:
OK, merging this PR

PR #1822, opened 2018-01-02
Fixes #5627: Readonly API or otherwise restrictable API Accounts

fanf:
https://www.rudder-project.org/redmine/issues/5627

Merge possible :)

Some important points: 
ApiAccount object gets a "kind" (system, user, public), an optionnal expiration time, and ACL. 
There is now a clean separation between an API definition (named an `EndpointSchema`), i.e the part were you tell "One can do a GET on resource "/rules" to have the list of rules"), the binding of these API to endpoints (ie "I want to map GET /rules to path "/api/7/rules" and "/secure/api/rules" etc) and the actual implementation, for Lift, about what is done for that endpoint (a `LiftApiModule`).

All the API definition are in `EndpointDefinition.scala`. 
All the binding / rooting code is in `ApiDatastructure`, especially `BuildHandler#buildApi`. That method is the whole abstract program that handle what is the logic for an API in Rudder (connect endpoint to correct path given version, given a request check token authz, write response, etc). Moreover, that part is completly independant from Lift, which should ease the pain when we will finaly migrate away from it. 

Lift specific implementation are in `com.normation.rudder.rest.lift`.
The (trivial) logic for Authz is in ApiAuthorizations.scala. Trivial, because once we have ACL in API Token, we just check if the token ACL matches the API canonical path. 

This PR also refactor quite a lot of things related to Rudder user roles / rights. It was needed to be able to map ACL <-> user roles. The big changes are: 
- rudder user now keep its roles (it was only keeping the linked authorization)
- all roles and authorisations are Algeabric Data Types in places of strings
- I added a role "userAccount" to handle the possibility to get/create/delete a personal API token
- the mapping between role and API ACL is in `RoleApiMapping`.

I also cleaned-up things: internal / v1 api are in `com.normation.rudder.rest.internal`, `com.normation.rudder.rest.data` is for RestObjects. In the future, we will want to add a phase in the abstract program which, for a given endpoint, extract the query parameters, and then only give that to the specific implementation. That will allow to abstract away json/non json request, and let the implementation only deal with the actual logic (and it will simplify tests a lots). 

Finally, just because it's possible now, I created a com.normation.rudder.rest.display.ApiPrinter which is an implementation of the abstract API handling program but only print what the request is. 
And I update `/api/info` endpoint, which get the list of all available endpoints, and `/api/info/{name}` detailed information about the corresponding endpoint (all supported version and path, etc)


ncharles:
I'm not sure if it is expected, as the UI is not finished yet, but when I create a token, it hasn't any ACL defined:
apiAcl: {"acl":[]}

ncharles:
This seems to work as expected :+1:

rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -0,0 +1,224 @@
...
+  }
+  // only the root is given, and the path ends with "**". It can even be onlyl "**"
+  final case class Root(segments: List[AclPathSegment])              extends AclPath {
+    def parts = NonEmptyList.ofInitLast(segments, AclPathSegment.DoubleWildcard)
+  }
+
+
+  // parse a path to an acl path.
+  // we don't accept empty string and ignore empty subpart,
+  // and "**" must be the last segment.
```
VinceMacBuche: This is not checked by the method, should we check it ?

rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -0,0 +1,224 @@
...
+  def values = ca.mrvisser.sealerate.values[HttpAction]
+
+  def parse(action: String): Either[String, HttpAction] = {
+    val lower = action.toLowerCase()
+    values.find( _.name == lower ).toRight(s"Action '${action}' is not recognized as a supported HTTP action")
+  }
+}
+
+/*
+ * An authorization controle is done in a path (so we are managing
```
VinceMacBuche: control

rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -0,0 +1,224 @@
...
+ *
+ */
+sealed trait AclPath {
+  def value: String = parts.toList.map( _.value ).mkString("/")
+  def parts: NonEmptyList[AclPathSegment]
+}
+
+final object AclPath {
+
+ // the full path is enumarated. At least one segment must be given ("/" is not possible
```
VinceMacBuche: enumerated

rudder-rest/src/main/scala/com/normation/rudder/web/services/Translator.scala
```diff
@@ -129,9 +129,10 @@ class Translators {
     }
   }
 
+
   def get[T](implicit m: Manifest[T]): Option[Translator[T]] = {
     reg.get(m) match {
-      case Some(t: Translator[T]) => Some[Translator[T]](t)
+      case Some(t: Translator[T @unchecked]) => Some[Translator[T]](t)
```
VinceMacBuche: One less warning ;)

fanf:
PR rebased

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
fanf: done

fanf:
OK, merging this PR

PR #1823, opened 2018-01-03
Fixes #11914: Add Rudder reporting context before calling technique bundle

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/11914


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -146,8 +152,10 @@ object BuildBundleSequence {
     , isSystem               : Boolean
     , providesExpectedReports: Boolean
     , policyMode             : PolicyMode
+    , bundleContext          : Option[BundleContext]
```
fanf: Why an option ? We always have that information. Is there some case where we don't want to use it?

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -118,6 +120,10 @@ object BuildBundleSequence {
   // The rudder id is ruleid@@directiveid@@0
   final case class ReportId(value: String)
```
fanf: Isn't reportId the same thing that BundleContext ?

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
VinceMacBuche: I don't think we want to define context for all Bundles ( We only want it for the main Bundle, not the pre and the post ones) 

But maybe we could add the bundle in the right place without having to set this ... I'll look if i find a better way

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
VinceMacBuche: The identifier produced at the end is the report id, But i wanted to build it agent side by sending ruleId and directiveid separately so we can use then more easily in the technique ('maybe add debug / verbose info ...

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -354,8 +356,7 @@ object CfengineBundleVariables extends AgentFormatBundleVariables {
         val newMain = tb.main match {
           case Nil => Nil
           // in that case, we are using the same ReportId for report and bundle
-          case firstBundle :: tail => (Bundle(firstBundle.id, BundleName(s"""current_technique_report_info("${firstBundle.name.value}")"""), Seq())
-                                       :: firstBundle :: tail)
+          case firstBundle :: tail =>  firstBundle :: tail
```
ncharles: i fail to see what's the benefit of this pattern matching

VinceMacBuche: You are totally right, just forgot to clean after removing what need to be removed

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #1825, opened 2018-01-10
Fixes #11943: Display of node state could be improved

ncharles:
https://www.rudder-project.org/redmine/issues/11943


ncharles:
this is a completly different behaviour as from node policy mode, which is done in full CSS

fanf:
OK, merging this PR

PR #1829, opened 2018-01-15
Fixes #11969: Remove "non-uniry def overrides unary def"

fanf:
https://www.rudder-project.org/redmine/issues/11969


fanf:
OK, merging this PR

PR #1834, opened 2018-01-16
Fixes #11858: Add common pre- and post- agent run action triggered by technique logic

fanf:
https://www.rudder-project.org/redmine/issues/11858


rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
```diff
@@ -312,13 +320,54 @@ class TechniqueParser(
...
+                        // for value, look first for <PARAMETER ... value="pvalue"/> and then <PARAMETER>pvalue</PARAMETER>
+              pvalue <- opt((p \ "@value").text) orElse opt(p.text)
+            } yield {
+              RunHook.Parameter(pname, pvalue)
+            }
+         ))
+      }
+    }
+
+    if(xml.label != RUN_HOOKS) throw new ParsingException(s"Error: in technique '${id.toString()}', tried to parse a <${RUN_HOOKS}> xml, but actually got: ${xml}")
```
VinceMacBuche: Do you think those checks are usefull if before calling that function we do "xml \ RUN_HOOKS"

The same goes for all other labels ...

VinceMacBuche: I hope one day we will have time to remove theose exception and put the result in an Either[ParsingError,Result]

fanf: They are, because without them the function is not autonomous. And they were added because at some point, we bang our head on a bug to discover that the node wasn't the correct anymore following a refactoring...
They may exist better way (less boilerplate) to do so with some sugar syntax

VinceMacBuche: Developper life is not the same if we cannot bang our head on bugs ... ;) (So great, keep theose checks !!!)

rudder-core/src/main/scala/com/normation/cfclerk/domain/Technique.scala
```diff
@@ -69,11 +69,44 @@ case class TechniqueId(name: TechniqueName, version: TechniqueVersion) extends O
...
+
+}
+
+
+/*
+ * A run hook is an (agent specific) action that should
+ * be run only one time per node per run.
+ */
+final case class RunHook(
+    name      : String // the kind of hook. The bundle method to call can be derived from that
```
ncharles: saying in the comment "the kind of hook" is quite misleading, as there is a Kind
could it be: "type of hook", or "name of the hook: the bundle method to call can be derived from that"

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -183,18 +187,14 @@ class BuildBundleSequence(
...
-          val inputs = cfg.templates.collect { case template if(template.included) => InputFile(template.outPath, p.technique.isSystem) } ++
-                       cfg.files.collect { case file if(file.included) => InputFile(file.outPath, p.technique.isSystem) }
-          //must replace RudderUniqueID in the paths
-          p.technique.generationMode match {
-            case TechniqueGenerationMode.MultipleDirectives =>
-              inputs.map(i => i.copy(path = Policy.makeUniqueDest(i.path, p)))
-            case _ =>
-              inputs
-          }
+      val inputs = p.technique.agentConfig.templates.collect { case template if(template.included) => InputFile(template.outPath, p.technique.isSystem) } ++
```
ncharles: why not doing it by agentType as before ?

fanf: I added a "PolicyTechnique" object which is a specialization of a Technique for one agent (on a node). I now build that oject when creating the node configuration, which is the best place to check if all techniques have an agent target matching the node one. And after that, I avoid all further agentType checking (and I have only one available).

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -1029,30 +1051,83 @@ object BuildNodeConfiguration extends Loggable {
...
+     *
+     * If at least one of kind, name, parameter is not the same between two hooks, the are considered different.
+     *
+     * The merge must keep both the List[Policy] order AND in a Policy, the List[RunHook] order. So we
+     * avoid groupBy
+     */
+    // utility class to store reportId + RunHook
+    case class BoundHook(id: PolicyId, mode: PolicyMode, hook: RunHook)
+
+    def recMerge(currentHook: BoundHook, remaining: List[BoundHook]): List[NodeRunHook] = {
```
ncharles: this looks overly complicated for the task
couldn't you do a list of all BoundHook, and then group them by hook.kind; hook.name; hook.parameters  ?

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
ncharles: That's very true ! Thank you

rudder-core/src/test/resources/configuration-repository/expected-share/node-cfe-with-two-directives/rules/cfengine-community/rudder-directives.cf
```diff
@@ -35,6 +35,7 @@ body file control {
 bundle agent rudder_directives {
 
   methods:
+      "pre-run-hook"                                  usebundle => do_run_hook("package","","{"parameters":{"package":"vim"},"reports":[{"id":"208716db-2675-43b9-ab57-bfbab84346aa@@16d86a56-93ef-49aa-86b7-0d10102e4ea9@@0","mode":"enforce"}]}");
```
ncharles: the quoting here looks dodgy
"{"parameters":{"package":"vim"},"reports":[{"id":"208716db-2675-43b9-ab57-bfbab84346aa@@16d86a56-93ef-49aa-86b7-0d10102e4ea9@@0","mode":"enforce"}]}"

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
fanf: no, I need to keep the same order, groupBy use a map (a set actually), and it would loose the order (and it was harder to try to restor it after merge)

rudder-core/src/test/resources/configuration-repository/expected-share/node-cfe-with-two-directives/rules/cfengine-community/rudder-directives.cf
fanf: what is the correct quoting?

ncharles: i guess replacing the inner double quotes by simple quotes (or the other way round)
so 
"{'parameters':{'package':'vim'},'reports':[{'id':'208716db-2675-43b9-ab57-bfbab84346aa@@16d86a56-93ef-49aa-86b7-0d10102e4ea9@@0','mode':'enforce'}]}"
or
'{"parameters":{"package":"vim"},"reports":[{"id":"208716db-2675-43b9-ab57-bfbab84346aa@@16d86a56-93ef-49aa-86b7-0d10102e4ea9@@0","mode":"enforce"}]}'

rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
ncharles: you are totally right

fanf:
PR rebased

rudder-core/src/test/resources/configuration-repository/techniques/ncf_techniques/Create_file/1.0/metadata.xml
```diff
@@ -12,6 +12,19 @@
...
   </FILES>
+  <!-- 
+   here run hooks are the same for each agent for the sake of test simplicity, but in reallity, they
+   will most likely be different.
+   -->
+  <RUNHOOKS>
+    <PRE name="package">
+      <PARAMETER name="package" value="vim"/>
+    </PRE>
+    <POST name="servive" if="debian">
```
VinceMacBuche: May be should be service

rudder-core/src/test/resources/configuration-repository/expected-share/node-cfe-with-two-directives/rules/cfengine-community/rudder-directives.cf
```diff
@@ -43,6 +44,7 @@ bundle agent rudder_directives {
       "60-rule-technique-std-lib/Package management." usebundle => set_dry_run_mode("false");
       "60-rule-technique-std-lib/Package management." usebundle => package_management;
       "remove_dry_run_mode"                           usebundle => set_dry_run_mode("false");
+      "post-run-hook"                                 usebundle => do_run_hook("servive","",'{"parameters":{"service":"some value","a post command":"/something/that/is/complicated \"with\" 'all sort of quote'"},"reports":[{"id":"208716db-2675-43b9-ab57-bfbab84346aa@@16d86a56-93ef-49aa-86b7-0d10102e4ea9@@0","mode":"enforce"}]}');
```
ncharles: why the empty "" ?

rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
```diff
@@ -312,13 +320,54 @@ class TechniqueParser(
...
+    </RUNHOOKS>
+  */
+  def parseRunHooks(id: TechniqueId, xml: Node): List[RunHook] = {
+    def parseOneHook(xml: Node, kind: RunHook.Kind): RunHook = {
+      def opt(s: String) = if(s == null || s == "") None else Some(s)
+
+      (xml \ "@name").text match {
+        case null | "" => throw new ParsingException(s"Error: in technique '${id.toString()}', tried to parse a run hooks, but attribute 'name' is missing in: ${xml}")
+        case name      =>
+          RunHook(name, kind, (xml \ "@condition").text, (xml \\ "PARAMETER").toList.flatMap(p =>
```
ncharles: i don't see the "if" parsed here, but rather a condition
so documentation is invalid, it should not be "if" but "condition"

PR #1844, opened 2018-01-26
Fixes #4670: Allows inventories to be augmented by the user

fanf:
https://www.rudder-project.org/redmine/issues/4670

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
@@ -47,7 +47,7 @@ import com.normation.inventory.ldap.core.LDAPConstants._
 import com.normation.ldap.sdk._
 import com.normation.cfclerk.domain._
 import com.normation.rudder.domain.RudderLDAPConstants._
-import com.normation.rudder.domain.{NodeDit,RudderDit}
+import com.normation.rudder.domain.{NodeDit, RudderDit}
 import com.normation.rudder.domain.nodes.Node
 import com.normation.rudder.domain.nodes.JsonSerialisation._
 import com.normation.rudder.domain.policies._
@@ -81,7 +81,7 @@ import com.normation.rudder.services.queries._
 import com.normation.utils.Control._
 import com.unboundid.ldap.sdk.DN
 import net.liftweb.common._
-import net.liftweb.common.Box.{ tryo => _, _ }
+import net.liftweb.common.Box.{tryo => _, _}
 import net.liftweb.json._
 import com.normation.rudder.reports._
 import com.normation.inventory.ldap.core.InventoryMapper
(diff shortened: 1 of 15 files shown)


rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -314,11 +315,33 @@ class LDAPEntityMapper(
           )
         }
       }
+      // custom properties mapped as NodeProperties
+      if(node.id.value == "root") {
+        println(inventoryEntry)
+      }
```
ncharles: you left debugging info there

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -314,11 +315,33 @@ class LDAPEntityMapper(
...
+      val customProperties = inventoryEntry.valuesFor(A_CUSTOM_PROPERTY).flatMap { json =>
+        import inventoryMapper.CustomPropertiesSerialization._
+        println("*** custom property:" + json)
```
ncharles: it should probably be a logger.trace

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -314,11 +315,33 @@ class LDAPEntityMapper(
...
+        json.toCustomProperty match {
+          case Left(ex)  =>
+            logger.error(s"Error when trying to deserialize Node Custom Property, it will be ignored: ${ex.getMessage}", ex)
+            None
+          case Right(cs) =>
+            Some(NodeProperty(cs.name, cs.value, Some(NodeProperty.customPropertyProvider)))
+        }
+      }
+      if(customProperties.nonEmpty) {
+        println("*"*30+customProperties)
```
ncharles: same here

rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -314,11 +315,33 @@ class LDAPEntityMapper(
...
+      }
+      val customPropertyNames = customProperties.map( _.name ).toSet
 
       // fetch the inventory datetime of the object
       val dateTime = inventoryEntry.getAsGTime(A_INVENTORY_DATE) map(_.dateTime) getOrElse(DateTime.now)
       NodeInfo(
-          node
+          node.copy(properties =
+            // filter out properties overriden by inventory property
+            node.properties.filterNot(x => customPropertyNames.contains(x.name) ) ++ customProperties
```
ncharles: i'm not sure I get the expected behaviour there ; does it completly replaces node properties from, say, datasource ?

fanf: it would, but I'm not sure it's what we want. Perhaps I could only override properties for default provider. 
And in all cases, I should add a log "warn" (or info?) to tell that a property is overriden

ncharles: ok, I was under the impression that we would keep the different type of properties (those from datasources, those from the user, and those from the node), with the possibility to use either of them (or to have a clear explaination of what wins in each case, but still show all the data)

fanf:
PR rebased

fanf:
OK, merging this PR

PR #1847, opened 2018-02-01
Fixes #12079: Add information about technique name/id to directive context

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12079


VinceMacBuche:
Commit modified

PR #1849, opened 2018-02-07
Fixes #12093: Make Rudder UI pluggable (login + common layout)

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12093


rudder-web/src/main/scala/bootstrap/liftweb/AppConfig.scala
```diff
@@ -1753,6 +1753,7 @@ object RudderConfig extends Loggable {
   }
 
  val aggregateReportScheduler = new FindNewReportsExecution(executionService,RUDDER_REPORTS_EXECUTION_INTERVAL)
+
```
ncharles: this is a great change :)

PR #1855, opened 2018-02-15
Fixes #12106: Implement general parameters for policy server policy copy for nodes

ncharles:
https://www.rudder-project.org/redmine/issues/12106


ncharles:
Commit modified

ncharles:
https://github.com/Normation/rudder/pull/1857

PR #1856, opened 2018-02-15
Fixes #5931: Api does not fail on bad JSON

fanf:
https://www.rudder-project.org/redmine/issues/5931

### DIFF
--- a/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -148,6 +148,22 @@ class Boot extends Loggable {
     })
 
     ////////// CACHE INVALIDATION FOR RESOURCES //////////
+    // fails on invalid JSON body because it's unsufferable
+    LiftRules.statelessDispatch.append {
+      case req: Req if(   req.json_?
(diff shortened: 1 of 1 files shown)


rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -148,6 +148,22 @@ class Boot extends Loggable {
...
+                       && req.requestType != GetRequest
+                       && req.requestType != HeadRequest
+                       && req.requestType != OptionsRequest
+                       && req.json != null
+                       && req.json.isEmpty
+        ) =>
+        () => Full(
+          JsonResponse(net.liftweb.json.parse(
+            """{"result": "error"
+              |, "errorDetails": "The request has a JSON content type but the body is not valid JSON"}""".stripMargin
```
VinceMacBuche: Should we try to display the content ??

VinceMacBuche: Let's merge it, better having a v1 :)

fanf: I don't think so, the user just sent it and we can't even try to tell him where it's not ok

PR #1857, opened 2018-02-16
Fixes #12106: Implement general parameters for policy server policy c…

ncharles:
…opy for nodes


ncharles:
https://www.rudder-project.org/redmine/issues/12106

replace PR https://github.com/Normation/rudder/pull/1855

rudder-web/src/main/webapp/secure/administration/policyServerManagement.html
```diff
@@ -551,7 +615,12 @@ <h3>Protocol</h3>
       </div>
     </div>
   </div>
+<<<<<<< HEAD
```
fanf: there is a merge problem here

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
```diff
@@ -356,6 +374,11 @@ class LDAPBasedConfigService(configFile: Config, repos: ConfigRepository, workfl
       SyslogTCP
     case _ => SyslogUDP
   }
+  private[this] implicit def toRelaySynchronisationMethod(p: RudderWebProperty): RelaySynchronizationMethod = p.value match {
+    case ClassicSynchronization.value  => ClassicSynchronization
+    case RsyncSynchronization.value    => RsyncSynchronization
+    case DisabledSynchronization.value => DisabledSynchronization
```
fanf: Heren the match is not exhaustive. If the value is not recognized, it will fails. 
I thing we want to have "classic" as the default, so use:
```
  case RsyncSynchronization.value    => RsyncSynchronization
  case DisabledSynchronization.value => DisabledSynchronization
  case _                             => ClassicSynchronization
```

rudder-core/src/test/scala/com/normation/rudder/services/policies/write/WriteSystemTechniquesTest.scala
```diff
@@ -188,6 +193,10 @@ object TestSystemData {
     //denybadclocks and skipIdentify are runtime properties
     , getDenyBadClocks         = () => Full(true)
     , getSkipIdentify          = () => Full(false)
+      // relay synchronisation method
+    , getSyncMethod            = () => Full(ClassicSynchronization)
+    , getSyncPromises          = () => Full(false)
```
fanf: Does it makes sense (from the feature behavior) to have separated items for sync method / sync promises / sync shared file ? The two last seems to have a meaning only for syncMethod=rsync, no ? (really asking, I don't have any idea from the ticket description about what should be the behavior)

ncharles: for the moment they have, but it could change in the future - to have rsync for one, and classic for another for instance
I didn't want to go the json path for something that were bound to change , and I prefered to stricly match the form and the system variable

rudder-web/src/main/scala/com/normation/rudder/appconfig/ConfigService.scala
ncharles: you are totally right

fanf:
Ho, you also need to update the API for settings to handle the new properties

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1860, opened 2018-02-21
Fixes #12144: Extract run analysis function 

fanf:
https://www.rudder-project.org/redmine/issues/12144

This PR is actually doing very little, but in a big number of lines: 

- the main goal is to extract ExecutionBatch#computeNodeRunAndConfigInfo from ExecutionBatch#computeNodesRunInfo (i.e the case where we DO have a run, because it will be the one to use in historisation)
- for that, we need to change how logging of RunAndConfigInfo (else, we would get logs for both before and after expiration date all the time, what is not good). So I added for all RunAndConfigInfo a description message and a standard name
- then, I had to change a billions place to add these informations in pattern matching or to adapt logs
- the remaining changes are making test pass again, and they actually find a bug for Pending case (where we can have a run without config, which was merged with the case for no run - not a big deal since that case should be extremelly rare, but still)


fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -282,15 +322,183 @@ object ExecutionBatch extends Loggable {
...
+  /*
+   * How long time a run is valid before receiving any report (but not after an update)
    */
-  private[this] def runType(traceMessage: String, runType: RunAndConfigInfo)(implicit nodeId: NodeId): RunAndConfigInfo = {
-    val msg = if(traceMessage.trim.size == 0) "" else ": " + traceMessage
-    ComplianceDebugLogger.node(nodeId).trace(s"Run config for node ${nodeId.value}: ${runType.logName} ${msg}")
-    runType
+  def runValidityDuration(runIntervalInfo: ResolvedAgentRunInterval, complianceMode: ComplianceMode) = complianceMode.mode match {
+    case ChangesOnly =>
+      //expires after run*heartbeat period - we need an other run before that.
```
ncharles: actually, it's a bit more complex than that:
- if runInterval did change, then it's indeed run*heartbeat
- else if run interval changed, we need to compute based on last run time, and expected changes in between
what is good is that we can't change the frequency when webinterface is down, so it's never during a catchup that we have changing frequencies

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -282,15 +322,183 @@ object ExecutionBatch extends Loggable {
...
+   * Method that, given a "run and config info" object, calculates what is the run state
+   * given node config info.
+   *
+   * The method returns either an error (no expected reports, bad config id, no reports at all...)
+   * or a datastructure with the Correct state and its validity interval, and the next state after the
+   * interval (before interval does not make sense, the begining of the interval is the run execution time)
+   */
+  def computeNodeRunAndConfigInfo(
+      runInfos         : AgentRunWithNodeConfig
+    , currentNodeConfig: Option[NodeExpectedReports]
```
ncharles: here, the currentNodeConfig is not the current one, but the one we are computed
so maybe use simply nodeConfig (it had to think for 5 minutes to come up with such a lousy name)

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -282,15 +322,183 @@ object ExecutionBatch extends Loggable {
...
+         *   If the first generated promises for that node are beyond the grace period, it means that
+         *   the run should have used theses promises, and we have a (DNS) problem because it didn't.
+         *
+         * - else, we look at the most recent
+         *   config and decide between pending / no answer
+         *
+         * Note: we must use value from current expected config for modes,
+         *       because we don't have anything else really.
+         */
+        val oldestConfigId = nodeConfigIdInfos.getOrElse(Seq(currentConfig.configInfo)).minBy( _.creation.getMillis)
```
ncharles: why not using the one from the nodeConfig timestamp ?

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -354,23 +542,23 @@ object ExecutionBatch extends Loggable {
               nodeConfigIdInfos.getOrElse(nodeId, None) match {
                 case None =>
                   //ok, it's a node without any config (so without runs, of course). Perhaps a new node ?
-                  runType(s"nodeId has no configuration ID version, perhaps it's a new Node?", NoRunNoExpectedReport)
+                  NoRunNoExpectedReport(s"nodeId has no configuration ID version, perhaps it's a new Node?")
```
ncharles: this is so much more readable ! thank you

rudder-core/src/test/scala/com/normation/rudder/repository/jdbc/ReportingServiceTest.scala
```diff
@@ -344,93 +374,101 @@ class ReportingServiceTest extends DBCommon with BoxSpecMatcher {
...
+      expectedInBase must beEqualTo(expected)
 
     }
 
+    "contains the node config ids" in {
+      val configs = findExpected.getNodeConfigIdInfos(nodeIds.toSet).
+        openOrThrowException("Test failed: the box should be full").values.flatten.map(_.map(_.configId)).flatten
+
+      configs must containTheSameElementsAs(allConfigs.map( _.nodeConfigId ))
+    }
```
ncharles: thank you for repairing the tests !

rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -380,47 +380,45 @@ object NodeStatusReportSerialization {
...
-        ( ( "type"             -> "ComputeCompliance" )
-        ~ ( "expectedConfigId" -> e.nodeConfigId.value       )
-        ~ ( "runConfigId"      -> e.nodeConfigId.value       )
-        )
-    }
+    (
+      ( "type" -> runInfo.name )
+      ~ ((runInfo match {
+        case NoRunNoExpectedReport(_) =>
+          Nil
```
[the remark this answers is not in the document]
fanf: Acutally, we still serialize it, but I factored out the "type" part just above, and that case doesn't provide any other usefull information. 

I think we are serializing it just as a trace to know that we tried to calculate compliance at that point, and it didn't work.

rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
```diff
@@ -221,11 +222,23 @@ class ReportsExecutionService (
             logger.error("Root exception was: ", ex)
           }
         case Full(x) => //youhou
-          //save compliance in DB
           complianceRepos.saveRunCompliance(x.values.toList)
           logger.trace("Cache for compliance updates after new run received")
       }
     }
+
+    // TODO: save compliance for each run
```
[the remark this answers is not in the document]
fanf: Yes, that's the interesting part :) It's not trivial, because it can be heavy, so we will need some queue logic with it's own "done until here" index, or some oracle to know what runs are not calculated (a little more effecient than "select all finish run not in compliance table", I fear)

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -133,10 +140,23 @@ sealed trait ExpiringStatus extends RunAndConfigInfo {
     def expirationDateTime: DateTime
 }
 
+/**
+ * We also define two broad kinds of run status: OK (for Pending and
+ * ComputeCompliance) or not OK
+ */
+sealed trait RunKind
+object RunKind {
+  trait Valid   extends RunKind
```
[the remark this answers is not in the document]
fanf: The name is qualified, it will be: `RunKind.Valid` so I thought it was clear. (or `RunType.Valid`)

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -133,10 +140,23 @@ sealed trait ExpiringStatus extends RunAndConfigInfo {
     def expirationDateTime: DateTime
 }
 
+/**
+ * We also define two broad kinds of run status: OK (for Pending and
+ * ComputeCompliance) or not OK
+ */
+sealed trait RunKind
```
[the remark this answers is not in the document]
fanf: we aleady have run status in a lot of place. I can use RunType

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: I think it's OK, because `ResolvedAgentRunInterval` take into account the real run interval (i.e, the one corresponding to the run config id): if the run interval changed, we need to wait for a run which knows about the change to take it into account.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: Hummm, I just move that part around, but you are right, the meaning of "current" (in the sense of "the last one available") is not at all the same now. I will look to it.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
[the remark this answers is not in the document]
ncharles: ha, of course, you are totally right

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: I'm not sure about the question, I'm trying to reformulate it. So you are wondering why, when I have a run without a configId, I'm looking for the oldest configuration, and not the configuration that was generated around the run timestamp? 
The reason is that the question I'm trying to answer is: "is it legit for that node to send a run without a configId?". And it is legit ONLY if the node never updated its promises and run initial promises, and so I need to know when was the first policy generation done for that node, to see if it is ok that the node is not yet sending reports for it. 

If I didn't understood your question, don't hesitate to comment further.

ncharles: ok, you answered a question i didn't know I wanted to ask :)
So, my real question is that we use oldest config id, and report on it that we have a UnexpectedNoVersion, but really, we have a UnexpectedNoVersion with the configId of the timestamp of the run we are considering.  So if (oldestExpiration.isBefore(t)) (which is the point i was missing, and thank you for your answer) the you should Left(UnexpectedNoVersion(...  with the configId of the time of the run, so that we know what was expected on the node at this time, and what was missing

fanf: That makes sense

fanf:
So. 

There is still quite some work in that PR, because I didn't understood something until now: the notion if `Pending` does make sense ONLY for a given observer, at some point in time. 

When we are calculating compliance validity interval (like we want to do in the extracted method), we can't have a `Pending` state. We have some valid result in some interval, and some non-valid result elsewhere. A run is either correct regarding the policy it reports on, or it is not (because of expiration time, bad config id or whatever, but never because a policy generation was started after it). 

So the whole `Pending` notion is only relevant as an interpretation layer on top of the compliance interval, for a given observer, at some point in time. Typically, `I` want to have the interpretation of the compliance intervalle for a given node, for `time == now`, and knowing that the node got a new policy generation 2min ago. 

So, yes, some more work need to be done.

peckpeck:
Need more work that initially foreseen, we need a new PR

PR #1861, opened 2018-02-23
Fixes #12148: On Settings page, in branch 4.2, current value for Synchronization method from Relay servers is not selected

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/12148

### DIFF
--- a/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/PropertiesManagement.scala
+++ b/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/PropertiesManagement.scala
@@ -555,10 +555,15 @@ class PropertiesManagement extends DispatchSnippet with Loggable {
               def radioHtml(label:String) : NodeSeq = {
                 val inputId = label+"-id"
                 val ajaxCall = SHtml.ajaxCall(Str(""), _ => setRelaySyncMethodJs(label))._2.toJsCmd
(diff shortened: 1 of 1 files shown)


RaphaelGauthier:
OK, merging this PR

rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/PropertiesManagement.scala
```diff
@@ -555,10 +555,15 @@ class PropertiesManagement extends DispatchSnippet with Loggable {
               def radioHtml(label:String) : NodeSeq = {
                 val inputId = label+"-id"
                 val ajaxCall = SHtml.ajaxCall(Str(""), _ => setRelaySyncMethodJs(label))._2.toJsCmd
+                val checked = relaySyncMethod.value == label
+                var inputCheck = checked match {
```
VinceMacBuche: That should be a val, you should never do a var !

VinceMacBuche: val inputCheck = if (relaySyncMethod.value == label) {
  <input id={inputId} type="radio" name="relaySync" onclick={ajaxCall} checked=""/>
} else {
  <input id={inputId} type="radio" name="relaySync" onclick={ajaxCall}/>
}

PR #1865, opened 2018-03-01
Fixes #12174:  Rename class to condition in the interface

amousset:
https://www.rudder-project.org/redmine/issues/12174


amousset:
OK, merging this PR

PR #1870, opened 2018-03-12
Fixes #12209: metadata.xml not generated when saving a technique from editor

fanf:
https://www.rudder-project.org/redmine/issues/12209

The main correction is the addition of a RudderAccount data type, which can `User` or `Api` and is know in `getCurrentUser`. This allows to choose the behiavior for when the ACL plugin is disabled based on the actual account type, because now we can discriminate between a User (who need full ACL when accessing internal API) and an API token link to a user (which can be disable if the plugin is so). These check are done in `ApiAuthorization.scala`.

All the remaining changes are:
- update in logback.xml to add api logger entry,
- repercussion of `UserService / User` move and update,
- imports cleaning (because of previous point, I had to clean a lot of imports in all cases),
- and the build of a system to enforce that all endpoint schema implementation actually match the schema definition. We had 3 cases where it wasn't the case (ncf create technique, group details, update parameters), and so obviously, the result wasn't the one expected for them.

### DIFF
--- a/rudder-rest/src/main/scala/com/normation/rudder/rest/ApiAuthorization.scala
+++ b/rudder-rest/src/main/scala/com/normation/rudder/rest/ApiAuthorization.scala
@@ -41,11 +41,14 @@ import com.normation.rudder.api.AclPath
 import com.normation.rudder.api.AclPathSegment
 import com.normation.rudder.api.HttpAction
 import com.normation.rudder.api.ApiAclElement
-import com.normation.rudder.api.{ ApiAuthorization => ApiAuthz }
-import com.normation.rudder.service.user.UserService
+import com.normation.rudder.api.{ApiAuthorization => ApiAuthz}
+import com.normation.rudder.UserService
 import com.normation.eventlog.EventActor
 import cats.implicits._
 import cats.data._
+import com.normation.rudder.RudderAccount
+import com.normation.rudder.api.ApiAccount
+import com.normation.rudder.api.ApiAccountKind
 
 
 /*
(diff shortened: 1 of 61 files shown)


fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

rudder-rest/src/main/scala/com/normation/rudder/rest/ApiAuthorization.scala
```diff
@@ -103,33 +106,59 @@ class AclApiAuthorization(logger: Log, userService: UserService, aclEnabled: ()
...
+                 */
+                // without plugin, api account linked to user are disabled
+                case (false, _, RudderAccount.Api(ApiAccount(_, ApiAccountKind.User, _, _, _, _, _, _))) =>
+                  logger.warn(s"API account linked to a user account '${user.actor.name}' is disabled because the API Authorization plugin is disabled.")
+                  None //token link to user account is a plugin only feature
+
+                // without plugin but ACL configured, standard api account are change to RO to avoid unwanted mod
+                case (false, ApiAuthz.ACL(acl), RudderAccount.Api(ApiAccount(_, _:ApiAccountKind.PublicApi, _, _, _, _, _, _))) =>
+                  logger.info(s"API account '${user.actor.name}' has ACL authorization but no plugin allows to interpret them. Revert to R0 rights.")
+                  checkACL(acl.filter( _.actions == HttpAction.GET), path, endpoint.schema.action)
```
VinceMacBuche: So we will filter acl to allow only authorized get from acl list ? I thought we would fall back to checkRO so acl feature was only available to plugin owners ?

fanf: hum, I thought I didn't commited it, not sure what happened. I'm changing it back.

fanf:
PR rebased

fanf:
OK, merging this PR

PR #1872, opened 2018-03-14
Fixes #12216: Unathenticated API v1 are not accessible anymore

fanf:
https://www.rudder-project.org/redmine/issues/12216

We were using an User account in place of an API account, so that went throught the User authentication pipeline which forbid un authenticated request.


fanf:
OK, merging this PR

PR #1873, opened 2018-03-15
Fixes #12225: Do not generate and copy rudder_expected_reports.csv anymore

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12225

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/AgentSpecificLogic.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/AgentSpecificLogic.scala
@@ -75,12 +75,10 @@ trait AgentFormatBundleVariables {
   ) : BundleSequenceVariables
 }
 
-
 // does that implementation knows something about the current agent type
 trait AgentSpecificGenerationHandle {
   def handle(agentType: AgentType, os: OsDetails): Boolean
 }
-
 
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/AgentSpecificLogic.scala
```diff
@@ -161,7 +155,7 @@ object CFEngineAgentSpecificGeneration extends AgentSpecificGeneration {
   }
 
   override def write(cfg: AgentNodeWritableConfiguration): Box[List[AgentSpecificFile]] = {
-    writeExpectedReportsCsv(cfg.paths, cfg.expectedReportsCsv, GENEREATED_CSV_FILENAME)
+    Full(Nil)
```
fanf: Could you had a comment referencing the ticket, something like: 
# We believe that expectedReports.csv is not used anymore on agent but still for expecter report generation.
# This is the first step to check that our hypothesis is ok before removing totally: https://www.rudder-project.org/redmine/issues/12225

VinceMacBuche: SHould i write iit this big ? :)

VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #1874, opened 2018-03-19
Fixes #12202: Missing information/bad UX in API authorization UI 

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12202

### DIFF
--- a/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
+++ b/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
@@ -174,7 +174,6 @@ class RestApiAccounts (
     }
   }
 
-
 }
 
 case class RestApiAccount(
--- a/rudder-web/src/main/webapp/javascript/rudder/angular/apiAccount.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/angular/apiAccount.js
@@ -55,7 +55,10 @@ accountManagement.directive('validEmpty', function() {
   };
 } );
     
-accountManagement.controller('AccountCtrl', function ($scope, $http, DTOptionsBuilder) {
+accountManagement.controller('AccountCtrl', function ($scope, $http, DTOptionsBuilder, DTColumnDefBuilder) {
+
+  $scope.aclPlugin = false;
+  $scope.authFilter = undefined;
   $scope.getAccounts = function() {
     $http.get(apiPath).
     then(function (response) {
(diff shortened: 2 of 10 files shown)


VinceMacBuche:
Commit modified

rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
```diff
@@ -198,6 +197,7 @@ case class RestApiAccount(
       case x => x
     }
 
+    println(expiration)
```
fanf: please remove it

rudder-web/src/main/webapp/javascript/rudder/angular/apiAccount.js
```diff
@@ -91,23 +94,45 @@ accountManagement.controller('AccountCtrl', function ($scope, $http, DTOptionsBu
       $('#oldAccountPopup').bsModal('hide');
   }
 
-
-$scope.addAccount = function() {
-  var newAccount = { id : "", name : "", token: undefined, enabled : true, description : ""}
+  $scope.filterAccount = function (value,index,array) {
+	console.log($scope.authFilter)
```
fanf: you should remove all the console.log

rudder-web/src/main/webapp/javascript/rudder/angular/apiAccount.js
```diff
@@ -91,23 +94,45 @@ accountManagement.controller('AccountCtrl', function ($scope, $http, DTOptionsBu
...
+      return true;
+    } else {
+      return value.authorizationType === $scope.authFilter;
+    }
+  }
+  $scope.addAccount = function() {
+  var now = new Date();
+  now.setMonth(now.getMonth()+1);
+  var now = now.getFullYear()+"-"+(now.getMonth()+1)+"-"+now.getDate()+" "+now.getHours()+":"+now.getMinutes();
+  console.log(now);
```
fanf: also here

rudder-web/src/main/webapp/javascript/rudder/angular/apiAccount.js
```diff
@@ -137,6 +162,7 @@ $scope.popupDeletion = function(account, action, actionName) {
    }
  }
  $scope.saveAccount = function(account,isPopup) {
+console.log(account);
```
fanf: this one

VinceMacBuche:
Commit modified

fanf:
OK to merge when the log are removed. It will still need woks because of the error on save which imply double save. 
Also, we don't get any message telling us that we can't enable an user account without the plugin, and we will need something in the UI to differenciate between the different kind of token

PR #1875, opened 2018-03-27
Fixes #12101: Escaping issue in Users Technique Parameter

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12101


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -116,11 +116,15 @@ object BuildBundleSequence {
...
+  sealed trait BundleParam {
+    def quote: String
+  }
   final object BundleParam {
-
     final case class SimpleQuote(value: String) extends BundleParam { def quote = "'"+value+"'" }
-    final case class DoubleQuote(value: String) extends BundleParam { def quote = "\""+value+"\"" }
+    final case class DoubleQuote(value: String) extends BundleParam {
+      val escapedValue = ParameterEntry.escapeString(value, AgentType.CfeCommunity)
+      def quote = "\""+escapedValue+"\""
```
ncharles: shouldn't it be done for DSC also ?

VinceMacBuche: Should be done in DSC plugin

PR #1876, opened 2018-03-28
Fixes #12002: JS error when starting Rudder 4.3 with jetty 9

fanf:
https://www.rudder-project.org/redmine/issues/12002


fanf:
PR rebased

ncharles:
looks good to me - can you squash the commits ?

fanf:
I would prefer to let them seperated, they are doing two different things

PR #1877, opened 2018-03-28
Fixes #12291: Identify in technique metadata if techniques should enable or disable ncf reporting

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12291


VinceMacBuche:
Commit modified

PR #1879, opened 2018-03-28
Fixes #12289: Generate dsc technique parameter

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12289


VinceMacBuche:
Commit modified

ncharles:
Were is the actual call to the technique, with the parameter value set (and correctly escaped) ?

VinceMacBuche:
In the plugin

PR #1880, opened 2018-03-29
Fixes #12310: Add name of the parameter to bundle parameter when generation bundle sequence

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12310


VinceMacBuche:
OK, merging this PR

PR #1881, opened 2018-03-29
Fixes #12314: Typo in agent build error "reource""

amousset:
https://www.rudder-project.org/redmine/issues/12314


amousset:
Commit modified

PR #1882, opened 2018-03-29
Fixes #12060: Incomplete "Rules to be applied" summary when accepting a new nodes when using groups of groups

fanf:
https://www.rudder-project.org/redmine/issues/12060

Implements algorithm described in above ticket (with a test!)

I also added a "NodeLogger" / "NodeLogger.Pending" / "NodeLogger.Pending.Policies" to be able to log (in trace) the behavior of the algo in case we need to debug in prod. 
And I changed the use of logger in NewNodeManager to use NodeLogger.Pending (nothing interesing in that file)

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
@@ -140,7 +140,7 @@ class UpdateDynamicGroups(
             dynGroupService.getAllDynGroups match {
               case Full(groupIds) =>
                 updateId = updateId + 1
-                LAUpdateDyngroup ! StartDynamicUpdate(updateId, ModificationId(uuidGen.newUuid), DateTime.now, GroupsToUpdate(groupIds))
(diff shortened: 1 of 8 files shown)


VinceMacBuche:
What happens if a dynamic group depends on a static group ? 

As far as I remember, static groups are not part of the algorithm and will be ignored (empty set) when  computing the nodes list of the dynamic group ...

I still need to read the whole algorithm, but i prefere to put the idea here before i forget to tell

VinceMacBuche:
I think it should be possible to add them (static groups) in the DONE list when initiailizing the algorithm

VinceMacBuche:
However, if the case is not treated, I think it could be merged and done in another PR

fanf:
Static group are never in TODO not DONE, because they never are in the init set of groups - only dyn groups are processed. So we have a dyngroup remaining in BLOCKED, its dep is set to 0, the rest follow in the common algo. 
There is a static group in the test.

VinceMacBuche:
In fact we are ok, because the static group cannot contain the nodes we want to check because the node is not accepted and therefore cannot be in static groups ...

VinceMacBuche:
Static groups do not enter in the algorithm, they are not fetched by the ldap query

PR #1885, opened 2018-04-03
Fixes #12343: LDAP query checker must succeed when criteria is empty

fanf:
https://www.rudder-project.org/redmine/issues/12343

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -40,7 +40,7 @@ package com.normation.rudder.services.queries
 import com.normation.rudder.repository.ldap.LDAPEntityMapper
 import com.normation.rudder.domain.nodes.NodeInfo
 import com.normation.rudder.domain.queries._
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -205,13 +205,17 @@ class PendingNodesLDAPQueryChecker(
 ) extends QueryChecker {
 
   override def check(query:Query, limitToNodeIds:Seq[NodeId]) : Box[Seq[NodeId]] = {
-    for {
-      entries <- checker.internalQueryProcessor(query, Seq("1.1"), Some(limitToNodeIds))
-      ids <- sequence(entries) { entry =>
-        checker.ldapMapper.nodeDn2OptNodeId(entry.dn) ?~! "Can not get node ID from dn %s".format(entry.dn)
+    if(query.criteria.isEmpty) {
+      Full(Seq.empty[NodeId])
```
ncharles: shouldn't it log a warning or something ?

fanf: I can add a log.debug, yes

fanf:
PR rebased

PR #1887, opened 2018-04-04
Fixes #12344: RudderUniqueId is not replaced when using OUTPATH in metadata, causing no report in File Permissions

fanf:
https://www.rudder-project.org/redmine/issues/12344


rudder-core/src/main/scala/com/normation/rudder/services/policies/RuleValService.scala
```diff
@@ -135,33 +135,33 @@ class RuleValServiceImpl(
         Full(None)
       case Some((fullActiveTechnique, directive)) =>
         for {
-          policyPackage <- Box(fullActiveTechnique.techniques.get(directive.techniqueVersion)) ?~! s"Version '${directive.techniqueVersion}' of technique '${fullActiveTechnique.techniqueName}' is not available for directive '${directive.name}' [${directive.id.value}]"
```
ncharles: policyPackage ! that's an old name ! :old_key:

PR #1894, opened 2018-04-09
Fixes #12324: Root server can be disable with node lifecycle "ignore" "set empty policies"

fanf:
https://www.rudder-project.org/redmine/issues/12324


fanf:
OK, merging this PR

PR #1896, opened 2018-04-10
Fixes #12390: Doubled "\" in GM call

fanf:
https://www.rudder-project.org/redmine/issues/12390

This is an API breaking change for Windows DSC plugin. 

So, the main problem was that we escaped value set in Directive only for a CFEngine Agent. 
The pull request change two majors things.  

First, it changes where the escape logic is in the code: 

- now, each agent has its own "escape" method bundled in its "agent specific generation" (where the agent-specific policy writing was stored). 
- this change make the fact that we needed an "agent register" apparent. So that `AgentRegister` is now a standalone class, with utility method to find what agent matches an `AgentType`/`osDetails` and does things with it. 

Secondly, it changes the types of values in string template. Until now, we were allowing values of type `Any`. But we don't actually use anything but string, and I can't apply string escape logic on `Any`. So now, we only deal with strings (which will certainly allow to simplify things in the long term - because yes, `String` is better than `Any`, especially when we don't use other things).

### DIFF
--- a/rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
+++ b/rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
@@ -43,29 +43,53 @@ import net.liftweb.common._
 
 class ConstraintException(val msg: String) extends Exception(msg)
 
-trait VTypeWithRegex {
-  def regex:Option[RegexConstraint]
-}
 /**
  * A constraint about the type of a variable
  */
 sealed trait VTypeConstraint {
   def name: String
 
-  // Escape a String to be CFengine compliant
-  // a \ will be escaped to \\
-  // a " will be escaped to \"
-  // The parameter may be null (for some legacy reason), and it should be checked
-  def escapeString(x : String) : String = {
-    if (x == null)
-      x
-    else
-      x.replaceAll("""\\""", """\\\\""").replaceAll(""""""","""\\"""")
-  }
+  /* type of the element to use in string template.
+   * we only support "string" (almost everything), and "boolean" which
+   * are used in &IF( expressions.
+   */
+  type STTYPE
 
-  //check if the value is compatible with that type constrain.
-  //return a Failure on error, the checked value on success.
-  def getTypedValue(value:String, forField:String) : Box[Any] = Full(escapeString(value))
+  /*
+   * This method does three things:
+   * - it checks if the input value is correct reguarding the field constraint
+   * - it formats the string given its internal representation.
+   * - it uses the provided escape method to escape need chars in value.
+   *
+   * The escape function is a parameter because it can change from one agent to the next.
+   */
+  def getFormatedValidated(value: String, forField: String, escapeString: String => String) : Box[STTYPE]
+}
+
+/*
+ * When we want to have a string put in string template
+ */
+sealed trait STString extends VTypeConstraint {
+  override type STTYPE = String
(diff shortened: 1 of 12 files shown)


rudder-core/src/test/resources/configuration-repository/expected-share/root-default-install/common/1.0/cf-served.cf
```diff
@@ -59,13 +59,13 @@ bundle server access_rules
 
     any::
           "/var/rudder/share/root/"
-            maproot    => { host2ip("server.rudder.local"), string_downcase(escape("server.rudder.local"))  },
-            admit      => { host2ip("server.rudder.local"), string_downcase(escape("server.rudder.local"))  },
+            maproot    => { string_downcase(escape("server.rudder.local"))  },
```
fanf: That change should not be linked with the modification at hand, but I'm not sure why I need it. I'm wondering if we didn't change something elsewhere.

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
```diff
@@ -104,7 +99,7 @@ object VTypeConstraint {
 sealed trait StringVType extends VTypeConstraint with VTypeWithRegex {
   def regex: Option[RegexConstraint]
 
-  override def getTypedValue(value:String, forField:String) : Box[Any] = regex match {
```
VinceMacBuche: The bow Any was ... interesting ...

rudder-core/src/test/resources/configuration-repository/expected-share/root-default-install/common/1.0/cf-served.cf
VinceMacBuche: strange, if not something in your PR then tests must failed but they passs ...

fanf: I know, and I don't know what to think about it. @ncharles, do you have an idea?

rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
```diff
@@ -192,12 +187,12 @@ final case object LinuxDerivedPasswordVType extends DerivedPasswordVType { overr
 
 case object BooleanVType extends VTypeConstraint {
   override val name = "boolean"
-  override def getTypedValue(value:String, forField:String) : Box[Any] = {
-      try {
+  override def getFormatedValidated(value:String, forField:String, escapeString: String => String) : Box[String] = {
```
VinceMacBuche: This is the cause of the issue in templates, the value was previous a boolean, not a string, then when used in conditional in string template, if considers truthy anything that has a value or check the boolean value if it's a boolean:

> IF actions test the presence or absence of an attribute unless the object is a Boolean/bool, in which case it tests the attribute for true/false. The only operator allowed is "not" and means either "not present" or "not true". For example, "$if(!member)$...$endif$".

rudder-core/src/test/resources/configuration-repository/expected-share/root-default-install/common/1.0/cf-served.cf
VinceMacBuche: Found why, it's because we changed booleans in string template to string, which change conditionals behavior in stringtemplate

fanf:
PR rebased

fanf:
OK, merging this PR

PR #1900, opened 2018-04-11
Fixes #12419: Make Groups and Rules API accessible for internal use

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12419


VinceMacBuche:
OK, merging this PR

PR #1910, opened 2018-04-23
Fixes #12429: make the compliance graph clickable

Haypierre:
https://www.rudder-project.org/redmine/issues/12429

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/homePage.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/homePage.js
@@ -196,7 +196,7 @@ function doughnutChart (id,data,count,colors) {
 
         // redirects to the corresponding node's information when clicking on a node's diagram
          , onClick: function(event, data){
-             //check if the user click on something relevant
+             //check if the user clicks on something relevant
              if (data[0] !== undefined){
               var query = {query:{select:"nodeAndPolicyServer",composition:"And"}};
               switch (id) {
@@ -234,12 +234,20 @@ function doughnutChart (id,data,count,colors) {
                         , value     : data[0]._model.label
                        }];
                       break;
-
-                  // TODO : in the default statement, make the compliance graph clickable
-
-                  default:
+                  case 'nodeCompliance':
+                     var compliance = {
+                        "Poor"   : {min: 0,   max: 50}
+                       ,"Average": {min: 50,  max: 75}
+                       ,"Good"   : {min: 75,  max: 100}
+                       ,"Perfect": {min: 100}
+                       };
+                     var interval = compliance[data[0]._model.label.split(' ')[0]];
+                     var complianceFilter = {complianceFilter:interval};
+                     window.location = contextPath + "/secure/nodeManager/nodes#" + JSON.stringify(complianceFilter);
                      return;
 
+                   default:
+                     return;
               }
                var url = contextPath + "/secure/nodeManager/searchNodes#" +  JSON.stringify(query);
                window.location = url;
--- a/rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
@@ -50,12 +50,49 @@ $.fn.dataTable.ext.order['dom-checkbox'] = function  ( settings, col )
 }
 
 /*
- * This function is used to resort a table after its sorting datas were changed ( like sorting function below)
+ * This function is used to resort a table after its sorting data were changed ( like sorting function below)
  */
 function resortTable (tableId) {
   var table = $("#"+tableId).DataTable({"retrieve" : true});
   table.draw();
 }
+
+
+$.fn.dataTable.ext.search.push(
+    function(settings, data, dataIndex ) {
+        // param needs to be JSON only so we remove the hash tag # at index 0
+
+        var param = decodeURIComponent(window.location.hash.substring(1));
+        if (param !== "" && window.location.pathname === contextPath + "/secure/nodeManager/nodes") {
+            var obj = JSON.parse(param);
+            var min = obj.complianceFilter.min;
+            var max = obj.complianceFilter.max;
+
+            if (min === undefined)
+                return true;
+
+            if (nodeCompliances !== undefined) {
+
+                // here, we get the id of the row element by looking deep inside settings...
+                // maybe there exists something cleaner.
+
+                var complianceArray = nodeCompliances[settings.aoData[dataIndex]._aData.id];
+                 if (complianceArray !== undefined) {
+                     var compliance = computeCompliancePercent(complianceArray);
+
+                     if (max === undefined)
+                        return compliance >= min;
+                      else
+                        return compliance >= min && compliance < max;
+                 }
+            }
+        }
+      return true;
+    });
+
+
+
+
 
 $.fn.dataTableExt.afnSortData['compliance'] = function ( oSettings, iColumn )
 {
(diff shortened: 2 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -165,6 +165,19 @@ object ComplianceLevel {
...
+    val r = new scala.util.Random()
+
+    ComplianceLevel(r.nextInt(100)
+      , r.nextInt(100)
+      , r.nextInt(100)
+      , r.nextInt(100)
+      , r.nextInt(100)
+      , r.nextInt(100)
+      , r.nextInt(100)
+      , (r.nextInt(100))
```
ncharles: I'm pretty sure you don't want to keep this

rudder-web/src/main/resources/version.properties
```diff
@@ -7,7 +7,7 @@
 # Major.minor without micro
 rudder-major-version=${rudder-major-version}
 # Full version, with SNAPSHOT etc
-rudder-full-version=${version}
+rudder-full-version=4.3
```
ncharles: you need to leave ${version}

rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -234,12 +234,18 @@ function doughnutChart (id,data,count,colors) {
                         , value     : data[0]._model.label
                        }];
                       break;
-
                   // TODO : in the default statement, make the compliance graph clickable
-
                   default:
-                     return;
-
+                     console.log(data[0]._model.label.split(' ')[0]);
```
ncharles: can you remove the log ?

rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
```diff
@@ -1041,8 +1085,19 @@ function createRuleComponentValueTable (contextPath) {
...
  */
 function createNodeTable(gridId, data, contextPath, refresh) {
-  //base element for the clickable cells
+
+
+ // console.log(gridId);
+ // console.log(contextPath);
+ // console.log(refresh);
+
+ //console.log(data);
```
ncharles: lots of unused comment - please remove

Haypierre:
This PR should not be merged because it triggers some "Compliance" incoherence between the home page's graph and the page where it should redirects.

Haypierre:
Commit modified

Haypierre:
Commit modified

Haypierre:
Commit modified

Haypierre:
Commit modified

Haypierre:
Commit modified

rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
```diff
@@ -1129,13 +1170,14 @@ function createNodeTable(gridId, data, contextPath, refresh) {
...
+
         $('[data-toggle="tooltip"]').bsTooltip();
         var rows = this._('tr', {"page":"current"});
-        $.each(rows, function(index,row) {
-          // Display compliance progress bar if it has already been computed
-          var compliance = nodeCompliances[row.id]
-          if (compliance !== undefined) {
-            $("#compliance-bar-"+row.id).html(buildComplianceBar(compliance));
+        $.each(rows, function(index,row){
+           // Display compliance progress bar if it has already been computed
```
VinceMacBuche: indentation was correct here

rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
```diff
@@ -50,13 +50,55 @@ $.fn.dataTable.ext.order['dom-checkbox'] = function  ( settings, col )
...
   table.draw();
 }
 
+
+$.fn.dataTable.ext.search.push(
+    function(settings, data, dataIndex ) {
+
+        var param = window.location.hash;
+
+        if (param !== "" && window.location.pathname === "/rudder/secure/nodeManager/nodes") {
```
VinceMacBuche: You should use contextPath+"/secure/nodeManager/nodes"

rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
```diff
@@ -50,13 +50,55 @@ $.fn.dataTable.ext.order['dom-checkbox'] = function  ( settings, col )
...
+
+            //! small parsing of the url argument to get the interval of compliance.
+            //! we return true if an error is triggered, so that everything goes normally
+            //! since it is not a critical error.
+
+            var from = param.indexOf('[');
```
VinceMacBuche: I don't think you should have [min,max] in your hash but a json that would be like this { complianceFilter { min : <value_min>, max : <value_max> } } and you should use JSON utilities to parse it

rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -234,10 +234,17 @@ function doughnutChart (id,data,count,colors) {
                         , value     : data[0]._model.label
                        }];
                       break;
+                  case 'nodeCompliance':
+                     var compliance = {
+                        "Poor"   : "[0,50]"
```
VinceMacBuche: Here it should be equals to an object { min : 0 , max : 50 }

rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -234,10 +234,17 @@ function doughnutChart (id,data,count,colors) {
...
+                       ,"Average": "[50,75]"
+                       ,"Good"   : "[75,100]"
+                       ,"Perfect": "[100,100]"
+                       };
```
VinceMacBuche: Here you could create an object "var complianceFilter = { complianceFilter : compliance[data[0]._model.label.split(' ')[0]] }

VinceMacBuche: In fact you should also have a compliance[data[0]._model.label.split(' ')[0]] stored in a var (let's call it complianceLevel )

Haypierre:
Commit modified

rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -234,10 +234,20 @@ function doughnutChart (id,data,count,colors) {
                         , value     : data[0]._model.label
                        }];
                       break;
+                  case 'nodeCompliance':
+                     var compliance = {
+                        "Poor"   : {min: 0,   max: 49}
```
VinceMacBuche: max should be 50 !!  (compliance is a float so it need to to 49.99999.... )

rudder-web/src/main/webapp/javascript/rudder/rudder-datatable.js
```diff
@@ -50,13 +50,45 @@ $.fn.dataTable.ext.order['dom-checkbox'] = function  ( settings, col )
...
+
+            if (nodeCompliances !== undefined) {
+
+                //! here, we get the id of the row element by looking deep inside settings...
+                //! maybe there exists something cleaner.
+
+                var complianceArray = nodeCompliances[settings.aoData[dataIndex]._aData.id];
+                 if (complianceArray !== undefined) {
+                     var compliance = computeCompliancePercent(complianceArray);
+                     return compliance >= min && compliance <= max;
```
VinceMacBuche: In fact I think it should be strictly less than max

I think max can be undefined so you only check if compliance >= min 

and Perfect should not define a value for max and should be { min : 100 }

rudder-web/src/main/webapp/javascript/rudder/homePage.js
VinceMacBuche: Same for all other value, Also perfect should only have a min value (see my other comment)

PR #1917, opened 2018-04-24
Fixes #12236: Missing border in Default policy mode setting

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/12236


rudder-web/src/main/webapp/style/rudder/rudder-menu.css
```diff
@@ -2325,53 +2325,6 @@ hr.separation{
   padding:0;
   font-weight:900;
 }
-
```
VinceMacBuche: +1

PR #1921, opened 2018-04-29
Fixes #12450: JS sandbox permission must be defined in a file

fanf:
https://www.rudder-project.org/redmine/issues/12450

This PR is done against 4.3. We will keep a basic, limited support of Java 9/10 for rudder 4.1, and we will have an extendable one in 4.3.


VinceMacBuche:
To me it's ok ! Not that i am an expert in Java security model buuuut it seems logic and can be merged

therefore @fanf  there is a conflict and it needs to be rebased

fanf:
PR rebased

PR #1930, opened 2018-05-14
Fixes #12616: ${const.dollar} in generic method parameter leads to missing report

fanf:
https://www.rudder-project.org/redmine/issues/12616

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -233,8 +233,10 @@ case class ComputeCompliance(
  */
 
 object ExecutionBatch extends Loggable {
-  final val matchCFEngineVars = """.*\$(\{.+\}|\(.+\)).*""".r
-  final private val replaceCFEngineVars = """\$\{.+\}|\$\(.+\)"""
+  //these patterns must be reluctant matches to avoid strange things
+  //when two variables are presents, or something like: ${foo}xxxxxx}.
+  final val matchCFEngineVars = """.*\$(\{.+?\}|\(.+?\)).*""".r
+  final private val replaceCFEngineVars = """\$\{.+?\}|\$\(.+?\)"""
 
   /**
    * containers to store common information about "what are we
@@ -270,10 +272,12 @@ object ExecutionBatch extends Loggable {
    * Takes a string, that should contains a CFEngine var ( $(xxx) or ${xxx} )
    * replace the $(xxx) (or ${xxx}) part by .*
    * and doubles all the \
-   * Returns a string that is suitable for a being used as a regexp
+   * Returns a string that is suitable for a being used as a regexp, with anything not
+   * in ".*" quoted with \Q...\E.
+   * For example, "${foo}(bar)$(baz)foo" => "\Q\E.*\Q(bar)\E.*\Qfoo\E"
    */
-  final def replaceCFEngineVars(x : String) : String = {
-    x.replaceAll(replaceCFEngineVars, ".*").replaceAll("""\\""", """\\\\""")
+  final def replaceCFEngineVars(x : String) : Pattern = {
+    Pattern.compile("""\Q"""+ x.replaceAll(replaceCFEngineVars, """\\E.*\\Q""") + """\E""")
   }
 
   case class ContextForNoAnswer(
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -270,10 +272,15 @@ object ExecutionBatch extends Loggable {
...
+   * in ".*" quoted with \Q...\E.
+   * For example, "${foo}(bar)$(baz)foo" => ".*\Q(bar)\E.*\Qfoo\E"
+   *
+   * We also change "\" into "\\" because cfengine escape them that way.
+   *
    */
-  final def replaceCFEngineVars(x : String) : String = {
-    x.replaceAll(replaceCFEngineVars, ".*").replaceAll("""\\""", """\\\\""")
+  final def replaceCFEngineVars(x : String) : Pattern = {
+    Pattern.compile("""\Q"""+ x.replaceAll("""\\""", """\\\\""").replaceAll(replaceCFEngineVars, """\\E.*\\Q""") + """\E""")
```
fanf: I don't understand the goal of the "\\" => "\\\\" here. Is it because CFEngine change "\" into "\\"? 
If so, why don't we do it *always* (not only when the expression contains a cfengine var), because here, only one of the test on "\\" passe.

fanf:
PR rebased

rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -510,6 +510,36 @@ class ExecutionBatchTest extends Specification {
...
+      , reports  = Success ("ayyyy)"    ) :: Repaired("bxxx}") :: Nil
+    )
+
+    // cfengine escape "\" into "\\"
+    test("""escape '\' into '\\' with cfengine vars"""
+      , patterns = Repaired("${foo}x\\x}") :: Nil
+      , reports  = Repaired("yx\\\\x}") :: Nil
+    )
+
+    test("""escape '\' into '\\' *without* cfengine vars"""
```
fanf: This does not passe (the other one above does pass).

fanf:
PR rebased

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -270,10 +272,12 @@ object ExecutionBatch extends Loggable {
    * Takes a string, that should contains a CFEngine var ( $(xxx) or ${xxx} )
    * replace the $(xxx) (or ${xxx}) part by .*
    * and doubles all the \
-   * Returns a string that is suitable for a being used as a regexp
+   * Returns a string that is suitable for a being used as a regexp, with anything not
+   * in ".*" quoted with \Q...\E.
+   * For example, "${foo}(bar)$(baz)foo" => ".*\Q(bar)\E.*\Qfoo\E"
    */
-  final def replaceCFEngineVars(x : String) : String = {
-    x.replaceAll(replaceCFEngineVars, ".*").replaceAll("""\\""", """\\\\""")
```
fanf: The replace all seems to have been here to avoid having "\x" be interpreted as a regex predifined character classes. Now that we correctly quote the regex, we don't need it anymore.

rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -270,10 +272,12 @@ object ExecutionBatch extends Loggable {
    * Takes a string, that should contains a CFEngine var ( $(xxx) or ${xxx} )
    * replace the $(xxx) (or ${xxx}) part by .*
    * and doubles all the \
-   * Returns a string that is suitable for a being used as a regexp
+   * Returns a string that is suitable for a being used as a regexp, with anything not
+   * in ".*" quoted with \Q...\E.
+   * For example, "${foo}(bar)$(baz)foo" => ".*\Q(bar)\E.*\Qfoo\E"
```
VinceMacBuche: are you missing the first Q/E in your example ? I think it hsould be: "\Q\E ".*\Q(bar)\E.*\Qfoo\E"

fanf: yes! Sorry

fanf:
PR rebased

fanf:
OK, merging this PR

PR #1931, opened 2018-05-15
Fixes #7616: Overreriden directive are not listed in reports information

fanf:
https://www.rudder-project.org/redmine/issues/7616


fanf:
PR rebased

PR #1936, opened 2018-05-19
Fixes #12661: Rudder fails to boot because cleaning TTL property is added after rudder reboot by migration script

fanf:
https://www.rudder-project.org/redmine/issues/12661


fanf:
Self merging to attempt a retag tomorrow. I believe the root cause it not there, thought.

PR #1937, opened 2018-05-19
Fixes #12660: JS error on node compliance details for rule with overiden directives

fanf:
https://www.rudder-project.org/redmine/issues/12660


fanf:
The code logic is still buggy, I think.

PR #1939, opened 2018-05-19
Fixes #12629: While adding network in General settings CIDR notation validation fails in special cases

fanf:
https://www.rudder-project.org/redmine/issues/12629


fanf:
Self merging to attempt a retag tomorrow.

PR #1945, opened 2018-05-23
Fixes #12650: Directive  tree filter is very slow

ncharles:
https://www.rudder-project.org/redmine/issues/12650

### DIFF
--- a/rudder-web/src/main/webapp/javascript/rudder/angular/filters.js
+++ b/rudder-web/src/main/webapp/javascript/rudder/angular/filters.js
@@ -200,8 +200,13 @@ app.controller('filterTagCtrl', function ($scope, $http, $location, $timeout, $r
     return !$scope.only.key && !$scope.only.value;
   }
 
+  // Add a grace period before effectively do the search
+  var to;
   $scope.searchTree = function(treeId) {
-    $(treeId).jstree('searchtag', $scope.searchStr, $scope.tags, $scope.only);
+    if (to !== undefined) { clearTimeout(to); }
+    to = setTimeout(function() {
+      $(treeId).jstree('searchtag', $scope.searchStr, $scope.tags, $scope.only);
+    }, 200);
   }
   
   $scope.clearSearchFieldTree = function(treeId) {


rudder-web/src/main/webapp/javascript/rudder/angular/filters.js
```diff
@@ -200,8 +200,13 @@ app.controller('filterTagCtrl', function ($scope, $http, $location, $timeout, $r
     return !$scope.only.key && !$scope.only.value;
   }
 
+  // Add a grace periot before effectively do the search
```
VinceMacBuche: period :)

rudder-web/src/main/webapp/javascript/rudder/angular/filters.js
```diff
@@ -200,8 +200,13 @@ app.controller('filterTagCtrl', function ($scope, $http, $location, $timeout, $r
...
+  var to = null;
```
VinceMacBuche: you should use undefined 

by declaring only

> var to;

rudder-web/src/main/webapp/javascript/rudder/angular/filters.js
```diff
@@ -200,8 +200,13 @@ app.controller('filterTagCtrl', function ($scope, $http, $location, $timeout, $r
...
   $scope.searchTree = function(treeId) {
-    $(treeId).jstree('searchtag', $scope.searchStr, $scope.tags, $scope.only);
+    if (to) { clearTimeout(to); }
```
VinceMacBuche: and here check that

> if to === undefined

VinceMacBuche: In fact you can remove the check, calling clearTimeout on undefined cause no problem

ncharles:
Commit modified

ncharles:
Commit modified

rudder-web/src/main/webapp/javascript/rudder/angular/filters.js
```diff
@@ -200,8 +200,13 @@ app.controller('filterTagCtrl', function ($scope, $http, $location, $timeout, $r
     return !$scope.only.key && !$scope.only.value;
   }
 
+  // Add a grace period before effectively do the search
+  var to;
   $scope.searchTree = function(treeId) {
-    $(treeId).jstree('searchtag', $scope.searchStr, $scope.tags, $scope.only);
+    if (typeof to !== 'undefined') { clearTimeout(to); }
```
VinceMacBuche: we do not use it and i clearly think you should use 

> to !== undefined

VinceMacBuche:
OK, merging this PR

PR #1947, opened 2018-05-23
Fixes #12690: Rule tree filter is a bit slow

ncharles:
https://www.rudder-project.org/redmine/issues/12690


VinceMacBuche:
OK, merging this PR

PR #1963, opened 2018-06-07
Fixes #12754: No reports on dsc only techniques on a Linux Node

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12754

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -219,12 +219,18 @@ class ClassicTechniqueWriter extends AgentSpecificTechniqueWriter {
   import ResultHelper._
   def writeAgentFile( technique : Technique, methods : Map[BundleName, GenericMethod] )  : Result[Option[String]] = Right(None)
   def agentMetadata ( technique : Technique, methods : Map[BundleName, GenericMethod] )  : Result[NodeSeq] = {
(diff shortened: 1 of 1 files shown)


rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -219,12 +219,18 @@ class ClassicTechniqueWriter extends AgentSpecificTechniqueWriter {
...
+                           ! (gm.agentSupport.contains(AgentType.CfeEnterprise) || (gm.agentSupport.contains(AgentType.CfeCommunity)))
+                         )
+                       )
+    // We need to add a reporting bundle for this method to generate a na report for any method with a condition != any/cfengine (which ~= true
     val needReportingBundle = technique.methodCalls.exists(m => m.condition != "any" && m.condition != "cfengine-community" )
     val xml = <AGENT type="cfengine-community,cfengine-nova">
       <BUNDLES>
         <NAME>{technique.bundleName.value}</NAME>
-        {if (needReportingBundle) <NAME>{technique.bundleName.value}_rudder_reporting</NAME>}
+        {if (noAgentSupportReporting || needReportingBundle) <NAME>{technique.bundleName.value}_rudder_reporting</NAME>}
```
ncharles: this fails if the technique has only generic method from one type:
<pre>
Undefined bundle DSC_only_rudder_reporting with type usebundle
</pre>

PR #1964, opened 2018-06-07
Fixes #12758: Invalid policies generated when using an ncf Technique with generic method non compatible with the agent

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12758


VinceMacBuche:
OK, merging this PR

PR #1965, opened 2018-06-11
Fixes #11889: The red cross in case of generation problem is not visible enough

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/11889


rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
```diff
@@ -92,8 +92,8 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
 
   val deployementErrorMessage = """(.*)!errormessage!(.*)""".r
 
-  private[this] def statusIcon : NodeSeq = {
-
+  private[this] def statusBg : String = {
+/*
```
VinceMacBuche: Don't keep the commented code 

and rename this function to statusBackground

RaphaelGauthier:
Commit modified

PR #1966, opened 2018-06-11
Fixes #12765: System Techniques must not be added by reload technique, only updated, else they are duplicated

fanf:
https://www.rudder-project.org/redmine/issues/12765


VinceMacBuche:
OK, merging this PR

PR #1973, opened 2018-06-20
Fixes #12702: Method copyResourceFile is quite inefficient

ncharles:
https://www.rudder-project.org/redmine/issues/12702

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PromiseWriteDataStructures.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PromiseWriteDataStructures.scala
@@ -120,5 +120,14 @@ case class TechniqueTemplateCopyInfo(
   , destination: String
   , content    : String //template content as a file
 ) extends HashcodeCaching {
-  override def toString() = s"Promise template ${id.name}; destination ${destination}"
+  override def toString() = s"Technique template ${id.name}; destination ${destination}"
+}
+
+
+case class TechniqueResourceCopyInfo(
+    id         : TechniqueResourceId
+  , destination: String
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PromiseWriteDataStructures.scala
```diff
@@ -122,3 +122,12 @@ case class TechniqueTemplateCopyInfo(
...
   override def toString() = s"Promise template ${id.name}; destination ${destination}"
 }
+
+
+case class TechniqueResourceCopyInfo(
+    id         : TechniqueResourceId
+  , destination: String
+  , content    : String //template resource as a file
+) extends HashcodeCaching {
+  override def toString() = s"Promise resource ${id.name}; destination ${destination}"
```
fanf: Can you use something else then "promise" here? Technique resource, or policy resource.

fanf:
Appart for the little comment, this is a very nice pr! 

Oh: could you try to align "<-" too?

ncharles:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PromiseWriteDataStructures.scala
ncharles: ok, renaming previous message also then

ncharles:
Commit modified

ncharles:
Commit modified

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #1983, opened 2018-07-09
Fixes #12471: wrong text encoding in technique editor

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/12471

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -358,9 +358,15 @@ class DSCTechniqueWriter(
                 ) ( e =>
                   IOError(s"Could not find dsc Technique '${technique.name}' in path ${basePath}/${techniquePath}", Some(e))
                 )
+      // Powershell files needs to have a BOM added at the beginning of all files when using UTF8 enoding
+      // See https://docs.microsoft.com/en-us/windows/desktop/intl/using-byte-order-marks
+      contentWithBom : List[Byte] =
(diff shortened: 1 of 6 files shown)


rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -358,9 +358,14 @@ class DSCTechniqueWriter(
                 ) ( e =>
                   IOError(s"Could not find dsc Technique '${technique.name}' in path ${basePath}/${techniquePath}", Some(e))
                 )
+      // Powershell files needs to have a BOM added at the beginning of all files
+      contentWithBom : List[Byte] =
+        // Bom, three bytes: EF BB BF
```
fanf: Can you add a link towards the spec/convention/wikipedia/etc? If one day we want to change to something else?

rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -358,9 +358,14 @@ class DSCTechniqueWriter(
...
+        239.toByte :: 187.toByte :: 191.toByte  ::
+        content.getBytes(StandardCharsets.UTF_8).toList
       files <-  execute {
                   Files.createDirectories(path.getParent)
-                  Files.write(path, content.getBytes(StandardCharsets.UTF_8))
+                  Files.write(path, contentWithBom.toArray)
```
ncharles: Why not using utf_8_sig ?
https://docs.python.org/2.5/lib/module-encodings.utf-8-sig.html

VinceMacBuche:
Commit modified

PR #1986, opened 2018-07-11
Fixes #12940: Can not create token

fanf:
https://www.rudder-project.org/redmine/issues/12940

That variable is just not used anymore.


fanf:
OK, merging this PR

PR #1988, opened 2018-07-17
Fixes #12474: root node disapeared while upgrading from 4.1 to 4.3 on debian 9

fanf:
https://www.rudder-project.org/redmine/issues/12474


fanf:
PR rebased

PR #1990, opened 2018-07-20
Fixes #12987: Style to define behaviour regarding duplicated reports in 4.3 is not consistent with the rest of the option

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/12987


RaphaelGauthier:
Commit modified

PR #2010, opened 2018-08-17
Fixes #13227: Incorrect name of support info file result used

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/13227

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/system/SupportInfoService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/system/SupportInfoService.scala
@@ -50,8 +50,14 @@ import scala.concurrent.Await
 import scala.concurrent.duration.Duration
 import scala.util.{Success, Try, Failure => TryFailure}
 
+final case class SupportInfoScriptResult (
+    serverName : String
+  , result     : Array[Byte]
+)
+
+
 trait SupportInfoService {
-  def launch() : Box[Array[Byte]]
+  def launch() : Box[SupportInfoScriptResult]
 }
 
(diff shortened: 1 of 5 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/system/SupportInfoService.scala
```diff
@@ -66,28 +72,32 @@ class SupportInfoServiceImpl extends SupportInfoService with Loggable {
...
   // We want to get its binary representation into an Array of byte
   // In order for the API to build an InMemoryResponse
 
-  private[this] def getScriptResult() : Box[Array[Byte]] = {
+  private[this] def getScriptResult() : Box[SupportInfoScriptResult] = {
+    import java.net.InetAddress.getLocalHost
 
-    val resultPath = "/tmp/support-info-server.tar.gz"
     Try {
+      val serverHostname = getLocalHost.getHostName
```
ncharles: getLocalHost may fail - you should catch the UnknownHostException and set a default name in this case
also, the security manager need to allow this operation

VinceMacBuche: This is in a Try block to prevent exceptions from this function

VinceMacBuche: The script already define the name of the file, we just try to find back the file in /tmp, if java could not find back localhost I prefer to have an error, than trying to guess the file.

I think the security manager is only set for javascript engine, not for Rudder

PR #2024, opened 2018-09-05
Fixes #13396: We need to be able to decide what workflow to use based on the change request

fanf:
https://www.rudder-project.org/redmine/issues/13396


rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
```diff
@@ -217,10 +217,10 @@ case class RestExtractorService (
   }
 
   private[this] def toWorkflowStatus (value : String) : Box[Seq[WorkflowNodeId]] = {
-    val possiblestates = workflowService.stepsValue
+    val possiblestates = workflowLevelService.getWorkflowService().stepsValue
```
VinceMacBuche: In this fonction, workflowLevelService.getWorkflowService() could be stored in a val at the beginning.

But I don't think it has much impact :)

rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
```diff
@@ -355,11 +350,11 @@ class GroupApiService2 (
     val actor = RestUtils.getActor(req)
     val groupId = NodeGroupId(req.param("id").getOrElse(uuidGen.newUuid))
 
-    def actualGroupCreation(restGroup : RestGroup, baseGroup : NodeGroup, nodeGroupCategoryId: NodeGroupCategoryId) = {
-      val newGroup = restGroup.updateGroup( baseGroup )
+    def actualGroupCreation(change: NodeGroupChangeRequest) = {
+      //val newGroup = restGroup.updateGroup( baseGroup )
```
VinceMacBuche: can be removed

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
```diff
@@ -235,14 +217,17 @@ class ModificationValidationPopup(
   }
 
   private[this] val disabled = item match {
-    case Left(item) => ! item._4.isEnabled
```
VinceMacBuche: omg, tuple with different ordering, people who did this was mad (was a younger naive me, ha my lost youth ....)

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
```diff
@@ -312,32 +297,33 @@ class ModificationValidationPopup(
...
+          else
+            NodeSeq.Empty
+        }
+      </div>
+    )
+    val (buttonName, classForButton, titleWorkflow) = (validationRequired, action) match {
+      case (false, DGModAction.CreateSolo       )   => ("Create"      , "btn-success", NodeSeq.Empty)
+      case (false, DGModAction.CreateAndModRules)   => ("Create"      , "btn-success", NodeSeq.Empty)
+      case (false, DGModAction.Delete           )   => ("Delete"      , "btn-danger", NodeSeq.Empty)
+      case (false, DGModAction.Update             ) => ("Update"      , "btn-success", NodeSeq.Empty)
```
VinceMacBuche: misaligned parenthesis :open_mouth:

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
```diff
@@ -312,32 +297,33 @@ class ModificationValidationPopup(
...
+      case (false, DGModAction.Enable           )   => ("Enable"      , "btn-primary", NodeSeq.Empty)
+      case (false, DGModAction.Disable          )   => ("Disable"     , "btn-primary", NodeSeq.Empty)
+      case (true , DGModAction.CreateSolo       )   => ("Create"      , "btn-success", NodeSeq.Empty)
+      case (true , DGModAction.CreateAndModRules)   => ("Open request", "btn-primary", workflowMessage(true))
+      case (true , DGModAction.Delete           )   => ("Open request", "btn-primary", workflowMessage(false))
+      case (true , DGModAction.Update             ) => ("Open request", "btn-primary", workflowMessage(false))
```
VinceMacBuche: Same :open_mouth:

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/RuleModificationValidationPopup.scala
```diff
@@ -74,61 +76,57 @@ object RuleModificationValidationPopup extends Loggable {
...
-      "enable"  -> "enable"
-    , "disable" -> "disable"
-    , "delete"  -> "delete"
-    , "save"    -> "update"
-    , "create"  -> "create"
-  )
-  private def explanationMessages = Map(
-      "enable"  ->
-    <div class="row">
+  private def titles(action: RuleModAction) = action match {
```
VinceMacBuche: Could be "${action.name.capitalize} a Rule"

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/RuleModificationValidationPopup.scala
```diff
@@ -74,61 +76,57 @@ object RuleModificationValidationPopup extends Loggable {
...
+    case RuleModAction.Enable  => "Enable a Rule"
+    case RuleModAction.Disable => "Disable a Rule"
+    case RuleModAction.Delete  => "Delete a Rule"
+    case RuleModAction.Update  => "Update a Rule"
+    case RuleModAction.Create  => "Create a Rule"
+  }
+
+  private def explanationMessages(action: RuleModAction) = action match {
+    case RuleModAction.Enable =>
```
VinceMacBuche: Whole function could be replaced by <pre>`
      <div class="row">
        <h4 class="col-lg-12 col-sm-12 col-xs-12 text-center">
            Are you sure that you want to ${action.name} this Rule?
        </h4>
    </div>`</pre>

rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/Groups.scala
```diff
@@ -90,12 +90,12 @@ class Groups extends StatefulSnippet with DefaultExtendableSnippet[Groups] with
 
   val mainDispatch = {
     RudderConfig.configService.rudder_workflow_enabled match {
```
VinceMacBuche: We don't need this pattern matching anymore, since we do not use the value

rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
fanf: Hum. In fact, it may be the opposite. We need to ensure all over the place that workflow status (and so, the workflow we user anywhere) is dynamic and change it the user enable the license or the license become unavailable or if the user property "enable workflow" value is switched. So perhaps it should be a `def` here, I will look

rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
fanf: s/Save/Update/g does not correct valign :/

rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/Groups.scala
fanf: It's needed because of StatefulSnipper + used in nodeManagement/groups.html

VinceMacBuche: Do you still need the pattern matching ? just put mainDispatch = Map ( "head" .... "groupHierarchy" )

fanf: AFAIK, it is till needed

PR #2025, opened 2018-09-06
Fixes #13416: Rudder system api token only have RO access

VinceMacBuche:
https://www.rudder-project.org/redmine/issues/13416


VinceMacBuche:
Also we define that the user for the account is a USER and not an API account, Should we change that too ? I don't know yet ... (User could be replaced by API)

https://github.com/Normation/rudder/pull/2025/files#diff-447696956d7831677269b4ee5e86a8dcL230

fanf:
So, the token is correctly a system token. RW seems to be the correct rights to be independant from ACL.

PR #2027, opened 2018-09-07
Fixes #13432: Git merge issue between Rudder 4.3 and Rudder 5.0

fanf:
https://www.rudder-project.org/redmine/issues/13432


fanf:
For now, I let the different commits separated. I'm not sure what is the best option.

PR #2040, opened 2018-09-19
Fixes #13517: Policy generation fails definitively when postgresql is shut down during policy generation

fanf:
https://www.rudder-project.org/redmine/issues/13517


fanf:
OK, merging this PR

PR #2043, opened 2018-10-09
Fixes #13609: Change request must not be saved when no validation is needed

fanf:
https://www.rudder-project.org/redmine/issues/13609

Some explanation: we had a problem because we used to have two repo, one false (in memory) and the real one, for change requests. When we switched the "enable change requests" status, it used to switch the repo. 
That does not work in a world where the status also depends of the actual modification. 

But that was not the whole problem: we used to always save a change request, even when the validation workflow was not used, so that the only step of the "no workflow" workflow was deleting the saved change in the in memory repos...

So, that PR change the flow of event so that: 

- the change is created, 
- if correct workflow is chosen, 
- if the workflow is a persistent one (the one with validation), its first step is to actually save the change in base, 
- if that happens, it returns the correct change ID,
- when the workflow status evolves, it is its responsability to change things in DB. 

In the end, that allows to completly split appart the "two step validation workflow" from Rudder, including all the repository part. I didn't touched the table creation part, because it has no consequences regarding API version or user or dev experience.

That's a big, very very welcome change. It simplifies a lot what a workflow is and what is needed to make one. 

Hereby, change Validation is really independant from Rudder.


rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
```diff
@@ -144,6 +146,14 @@ trait WorkflowLevelService {
   def getForDirective  (actor: EventActor, change: DirectiveChangeRequest  ): Box[WorkflowService]
   def getForNodeGroup  (actor: EventActor, change: NodeGroupChangeRequest  ): Box[WorkflowService]
   def getForGlobalParam(actor: EventActor, change: GlobalParamChangeRequest): Box[WorkflowService]
+
+  /*
+   * These method allow to get change request impacting a rule/directive/etc.
+   * Used to display information on them on corresponding update screens.
+   */
+  def getByDirective(id : DirectiveId, onlyPending:Boolean) : Box[Vector[ChangeRequest]]
```
VinceMacBuche: Just a question here, why a Vector and not an IndexedSeq ?   (or even a Seq/List? )

But don't change it, it's just for my knowledge

fanf: I just moved (or copy/pasted) interfaces that where provided elsewhere. 

Not sure for the rationnal, but the idea should have been along the line: be precise about what you give to let consumer adpat their behaviors.

rudder-rest/src/main/scala/com/normation/rudder/rest/RoleApiMapping.scala
```diff
@@ -44,14 +44,138 @@ import com.normation.rudder.api.ApiAclElement
...
+        case Compliance.Read      => ComplianceApi.GetGlobalCompliance.x :: ComplianceApi.GetRulesCompliance.x ::
+                                     ComplianceApi.GetRulesComplianceId.x :: ComplianceApi.GetNodesCompliance.x ::
+                                     ComplianceApi.GetNodeComplianceId.x :: Nil
+        case Compliance.Write     => Nil
+        case Compliance.Edit      => Nil
+
+        // Configuration doesn't give API rights.
+        // But as nothing manage parameter API, I think it's the correct place
+        // to add it.
+        case Configuration.Read   => ParameterApi.ListParameters.x :: ParameterApi.ParameterDetails.x :: Nil
```
VinceMacBuche: I though Configuration was giving access to all configs (Rules/Directives/Parameters/Techniques), am i wrong ?
So it should provide access to all other element not only parameter

VinceMacBuche: In fact it's already like that in 4.3

PR #2045, opened 2018-10-11
Fixes #13643: In Rudder 5.0, user password encoding is broken for some passwords. 

fanf:
https://www.rudder-project.org/redmine/issues/13643


fanf:
PR rebased

PR #2048, opened 2018-10-13
Fixes #13014: StackOverflowError during policy generation in JavascriptEngine - debian 9.5 with jdk 1.8.0_181

fanf:
https://www.rudder-project.org/redmine/issues/13014

So, more information: the problem is due to security check that are initiated during other security checks. Typically, trying to execute:
```
             ....
681          case x: FilePermission if( x.getActions == "read" && (x.getName.contains(".class") || x.getName.endsWith(".jar")  || x.getName.endsWith(".so") ) ) => // ok
```
needs to load FilePermission class file, and that need to resolve URL (more permission), and then load a file from fs (more perm).
It seems that something changed in later JVM 1.8 JDK releases about what classes are loaded when. And that results in infinite loops due to impossibility to linearize the graphe of dependance check.

I thought about two resolutions, a correct one and a hacky one. Unfortunatly, I'm not able to implement the correct one for now, and I'm not sure I want to amass the quantity of java priviledges & security to be able to do it.
So, the correct solution is to be able to make the check permission implementation logic not subject to the permission check. This would be correct because the code for the check is statically knows and immutable, so we can assess its security. But I don't know of a way to escape the current security contexte (and it seems logical, since it's kind of the goal of the security context...)
S, the hacky one is to preload relevant classes out of the security context so that we don't have to do the resolution check later on. Of course, this is extremelly fragile.

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
@@ -69,6 +69,7 @@ import com.normation.cfclerk.domain.AixPasswordHashAlgo
 import com.normation.cfclerk.domain.AbstactPassword
 import java.security.NoSuchAlgorithmException
 import java.net.URL
+import java.net.URLPermission
 import java.nio.file.Paths
 import java.security.CodeSource
 import java.security.ProtectionDomain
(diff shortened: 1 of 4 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -491,6 +506,12 @@ final object JsEngine {
 
   final class SandboxedJsEngine private (jsEngine: ScriptEngine, sm: SecurityManager, pool: ExecutorService, maxTime: (Int, TimeUnit)) extends JsEngine {
 
+    // we need to preload Permission to avoid loops - see #13014
```
VinceMacBuche: >_< sadness overflow

fanf:
OK, merging this PR

PR #2049, opened 2018-10-23
Fixes #13695: Authentication providers are not loaded before authentication is initialized

fanf:
https://www.rudder-project.org/redmine/issues/13695

As we can't control when and how spring is initializing its beans, I changed the spring authentication provider that is in charge of asking to backends is authentication is OK and did one for Rudder which dynamically get the list of beans depending of plugin status. 

The part that give the list of allowed backends is controled by Rudder, and is independant from the spring init part. So it accpets latter binding of plugins. 

Still, that means that all the code/config files must be present at boot time. To restrict the scope of possible unwanted error, we only look (and let spring load) code for plugins in the `rudder.auth.provider` list. 

(that also mean that allowing runtime change of authentication is still out of reach)

### DIFF
--- a/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -39,13 +39,13 @@ package bootstrap.liftweb
 
 import java.util.Collection
 
-import bootstrap.liftweb.RudderProperties.config
 import com.normation.rudder.Role
 import com.normation.rudder.RoleToRights
 import com.normation.rudder.RudderAccount
 import com.normation.rudder.api._
 import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.web.services.UserSessionLogEvent
+import com.typesafe.config.Config
 import com.typesafe.config.ConfigException
 import javax.servlet.Filter
 import javax.servlet.FilterChain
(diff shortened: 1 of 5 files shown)


fanf:
PR rebased

rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -365,12 +385,79 @@ class DefaultAuthBackendProviders() extends AuthBackendsProvider {
...
         }
       }
     }
+  }
+}
+
+// and default implementation: provides 'file', 'rootAdmin'
+class AuthBackendProvidersManager() extends DynamicRudderProviderManager {
+
+  val defaultAuthBackendsProvider = new AuthBackendsProvider() {
```
VinceMacBuche: I don't find this val used in the code, nor on the repository ...

rudder-web/src/main/scala/bootstrap/liftweb/DynamicRudderProviderManager.java
```diff
@@ -0,0 +1,18 @@
+package bootstrap.liftweb;
```
VinceMacBuche: :fearful:

VinceMacBuche: Maybe we can add a license hear

rudder-web/src/main/scala/bootstrap/liftweb/RudderProviderManager.java
```diff
@@ -0,0 +1,256 @@
+/*
```
VinceMacBuche: :fearful:

fanf:
PR rebased

rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
VinceMacBuche: Ok it's now used :)

PR #2053, opened 2018-10-31
Fixes #12829:  We have no way to know that a Technique is disabled

RaphaelGauthier:
https://www.rudder-project.org/redmine/issues/12829


VinceMacBuche:
OK, merging this PR

PR #2059, opened 2018-11-12
Fixes #13671: Missing node policy mode and node state variable expansion in directive

VinceMacBuche:
https://issues.rudder.io/issues/13671


VinceMacBuche:
Commit modified

rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -303,6 +304,15 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
         case "id" :: Nil => Full(context.nodeInfo.id.value)
         case "hostname" :: Nil => Full(context.nodeInfo.hostname)
         case "admin" :: Nil => Full(context.nodeInfo.localAdministratorAccountName)
+        case "state" :: Nil => Full(context.nodeInfo.state.name)
+        case "policyMode" :: Nil =>
```
fanf: This must be "policymode"

PR #2061, opened 2018-11-12
Fixes #13810: Update ldap node entries from  ipNetworkNumber => networkSubnet

fanf:
https://issues.rudder.io/issues/13810


rudder-web/src/main/resources/ldapObjectAndAttributes.properties
```diff
@@ -175,6 +175,7 @@ ldap.attr.networkInterfaceType = Interface type
 ldap.attr.networkInterfaceTypeMib = Interface type (Mib)
 ldap.attr.ipHostNumber = IP address
 ldap.attr.ipNetworkNumber = Network
+ldap.attr.networkSubnet = Network
```
VinceMacBuche: Should it not be different ? like "Subnet" instead of "Network"

fanf:
Actually, this should be done in a migration script. There is no reason to do it at each start of rudder. Moreover, with a migration script, we are sure that dyn-groups are not impacted, since rudder is always shutdown during the whole migration process.

PR #2070, opened 2018-11-13
Fixes #13654: [API] Can you add rudder server version to the API ?

fanf:
https://issues.rudder.io/issues/13654


fanf:
OK, merging this PR

PR #2077, opened 2018-11-13
Fixes #13674: Compliance error (missing) when a directive is applied by two rules on a node

fanf:
https://issues.rudder.io/issues/13674

In fact, we don't want to put "skipped" for the directive in other rules (meaning: already exec in previous rule) because that's not the same semantic as "skipped" for unique technique (meaning: not exec at all)


fanf:
OK, merging this PR

PR #2082, opened 2018-11-14
Fixes #9149: Inventory keeps its key certified when node is deleted

fanf:
https://issues.rudder.io/issues/9149

### DIFF
--- a/rudder-core/src/main/scala/com/normation/rudder/services/servers/RemoveNodeService.scala
+++ b/rudder-core/src/main/scala/com/normation/rudder/services/servers/RemoveNodeService.scala
@@ -12,13 +12,15 @@ import com.normation.rudder.repository.RoNodeGroupRepository
 import com.normation.eventlog.EventActor
 import com.normation.rudder.domain.nodes.ModifyNodeGroupDiff
 import com.normation.inventory.ldap.core.LDAPFullInventoryRepository
-import com.normation.inventory.domain.{AcceptedInventory,RemovedInventory}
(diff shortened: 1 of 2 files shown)


rudder-core/src/main/scala/com/normation/rudder/services/servers/RemoveNodeService.scala
```diff
@@ -156,6 +161,13 @@ class RemoveNodeServiceImpl(
             timePostHooks =  (System.currentTimeMillis - postHooksTime)
             _             = logger.debug(s"Node-post-deletion scripts hooks ran in ${timePreHooks} ms")
           } yield {
+            removeKeyCertification(nodeId) match {
+              case Full(_) => //ok
+              case eb: EmptyBox =>
+                val e = (eb ?~! "Error when removing the certification status of node key")
```
VinceMacBuche: Maybe a message to tell if the node is added again with a new key it will be refused and a link explaining how to reset a key (/opt/rudder/bin/rudder-keys reset-status ${report.node.main.id.value} ), like we do in https://github.com/Normation/ldap-inventory/pull/140/files#diff-5f54250574733d77e36bb05ce8e5332aL280 ?

 I don't know if the script can manage deleted node though ...

VinceMacBuche: A part that remark it's ok to me :)

VinceMacBuche: no idea :(

fanf:
OK, merging this PR

PR #2087, opened 2018-11-14
Fixes #12198: The password field allows storing a password without any selected hash

VinceMacBuche:
https://issues.rudder.io/issues/12198


VinceMacBuche:
Commit modified

PR #2090, opened 2018-11-16
Fixes #13843:  Inventory keeps its key certified when node is deleted even is keystatus is erased

fanf:
https://issues.rudder.io/issues/13843


fanf:
OK, merging this PR

PR #2098, opened 2018-12-17
Fixes #13987: Massive performance penalty in policy generation due to invalid usage of StringTemplate

ncharles:
https://issues.rudder.io/issues/13987

### DIFF
--- a/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
+++ b/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
@@ -72,7 +72,37 @@ final case class STVariable(
   , isSystem  :Boolean
 )
 
+
 class FillTemplatesService extends Loggable {
+
+  /*
+   * The cache is managed template by template
+   * If a content is not in the cache, it will create the StringTemplate instance, and return it
+   *
+   */
+  private[this] var cache = Map[String,StringTemplate]()
+
+
+  def clearCache(): Unit = this.synchronized {
+    cache = Map()
+  }
+
+  def getTemplateFromContent(templateName: String,content: String): Box[StringTemplate] = this.synchronized {
+    cache.get(content) match {
+      case Some(template) => Full(template.getInstanceOf())
+      case _ =>
+        try {
+          val template = new StringTemplate(content, classOf[NormationAmpersandTemplateLexer])
+          cache = cache + ((content, template))
+          Full(template.getInstanceOf())
+        } catch {
+          case ex: Exception =>
+            val m = s"Error when trying to parse template '${templateName}': ${ex.getMessage}"
+            logger.error(m, ex)
+            Failure(m)
+        }
+    }
+  }
 
   /**
    * Replace all occurences of parameters in the 'content' string with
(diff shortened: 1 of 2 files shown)


rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
```diff
@@ -72,8 +72,32 @@ final case class STVariable(
...
+
+  def clearCache(): Unit = this.synchronized {
+    cache = Map()
+  }
+
+  def getTemplateFromContent(content: String): Box[StringTemplate] = this.synchronized {
+    cache.get(content) match {
+      case Some(template) => Full(template.getInstanceOf())
+      case _ =>
+        val template = new StringTemplate(content, classOf[NormationAmpersandTemplateLexer])
```
fanf: I believe that can throw exception, no ? If so, can you wrapp it in `try/catch ex => Failure...`. And the same for `getInstanceOf` if needed.

rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
```diff
@@ -84,9 +108,7 @@ class FillTemplatesService extends Loggable {
   def fill(templateName: String, content: String, variables: Seq[STVariable]): Box[String] = {
     for {
       template <- try {
-                    //string template does not allows "." in path name, so we are force to use a templateGroup by policy template (versions have . in them)
-                    val template = new StringTemplate(content, classOf[NormationAmpersandTemplateLexer])
-                    Full(template)
+                    getTemplateFromContent(content)
```
fanf: no more need for try/catch here once `getTemplateFromContent` catch exception.

rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
ncharles: you are totally right

ncharles:
Commit modified

rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
```diff
@@ -72,8 +72,38 @@ final case class STVariable(
...
+   *
+   */
+  private[this] var cache = Map[String,StringTemplate]()
+
+
+  def clearCache(): Unit = this.synchronized {
+    cache = Map()
+  }
+
+  def getTemplateFromContent(templateName: String,content: String): Box[StringTemplate] = this.synchronized {
```
ncharles: i doubled checked - the templateName is indeed the destination of the file, note its source. Which is pretty confusing

ncharles: i dugged into the detail, to see how much ram it took to store the parsed stringtemplate
I have less than 1MB of antlr related class in the heapdump, which is peanuts compared to anything else. Granted, it contains a lot of java instance, but including ALL java instances, I barely reach 200MB (100 MB being Object, 65MB being String)
so i'm pretty convinced the memory usage is not problematic there (at least, we have easier battle to fight to save memory)

PR #2099, opened 2018-12-19
Fixes #14004: Wasted disk space due to inefficient columns ordering 

ncharles:
https://issues.rudder.io/issues/14004

### DIFF
--- a/rudder-core/src/main/resources/Migration/dbMigration-4.1.x-4.1.12-add-compliancelevel-table.sql
+++ b/rudder-core/src/main/resources/Migration/dbMigration-4.1.x-4.1.12-add-compliancelevel-table.sql
@@ -36,10 +36,7 @@
 */
 
 CREATE TABLE nodecompliancelevels (
-  nodeId             text NOT NULL CHECK (nodeId <> '')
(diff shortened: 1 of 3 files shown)


fanf:
There is several things in that PR and I'm not sure we want all of them, or all in master: 

- the idx name should be changed in 4.3, for new install only (ie no migration needed: old bases keep the old name, new base use the ones ok with convention), 
- the unused table should be removed in the first that does not use it, and at least 5.0 (new install will just not have it),
- for the column order... Now that I see the extended modification, I'm not so much at ease with it anymore. It feels really unatural to have nodeid of other identification parts, sometime part of table key, to be in latter columns. This is very unusual and, even if it's likelly not a problem for us (rudder dev), it will be *very* strange for users or contributors. 

If the gain are substantials, I believe we can argue for the change (with a lot of comments for impacted table about the rationnal). If they are not, I'm not sure the cost of surprise worth it.

ncharles:
ok, closing this PR. For the index, its because they are invalid (we are indexing 3 time the same column) ( see https://issues.rudder.io/issues/14007 )

PR #2103, opened 2018-12-24
Fixes #14041: Replace all RudderUniqueID in paths for inputs list and rudder.json, and correctly generate paths based on agent type

ncharles:
https://issues.rudder.io/issues/14041

### DIFF
--- a/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
+++ b/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
@@ -41,9 +41,12 @@ import CfclerkXmlConstants._
 import com.normation.cfclerk.domain._
 import com.normation.cfclerk.services.SystemVariableSpecService
 import com.normation.cfclerk.exceptions.ParsingException
+
 import scala.xml._
 import net.liftweb.common._
 import com.normation.inventory.domain.AgentType
+
+
 
 /**
  * Parse a technique (metadata.xml file)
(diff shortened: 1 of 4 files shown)


ncharles:
must be merged after https://github.com/Normation/ldap-inventory/pull/143

rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
```diff
@@ -246,9 +251,9 @@ class TechniqueParser(
    *
    * if name content start with RUDDER_CONFIGURATION_REPOSITORY, the path must be considered relative
    * to root of configuration repository in place of relative to the technique.
-   *
+   * TODO: pass the AgentType here
    */
-  private[this] def parseResource(techniqueId: TechniqueId, xml: Node, isTemplate:Boolean): (TechniqueResourceId, String) = {
+  private[this] def parseResource(techniqueId: TechniqueId, xml: Node, isTemplate:Boolean, agentType:Option[AgentType]): (TechniqueResourceId, String) = {
```
fanf: Here, it seems that the problem is that `isTemplate` is not sufficiently precise, but having both that parameter and `agentType` is redondant. Can you change `(isTemplate, agentType)` by `templateForAgent:Option[AgentType]`

So that we encode the whole information in just that parameter?

rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -378,6 +378,13 @@ final object Policy {
     path.replaceFirst(subPath, s"${p.technique.id.name.value}/${p.id.getRudderUniqueId}")
   }
 
+  /*
+   * Replace the Rudder Unique Id also in the path for input list
+   */
+  def replaceRudderUniqueId(path: String, p: Policy): String = {
+    path.replace(TAG_OF_RUDDER_MULTI_POLICY, p.id.getRudderUniqueId)
```
fanf: it should be a `replaceAll` to avoid problems in a case where someone put several tag in the path for whatever reason

rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -378,6 +378,13 @@ final object Policy {
     path.replaceFirst(subPath, s"${p.technique.id.name.value}/${p.id.getRudderUniqueId}")
```
fanf: Is their a case where we want to make a unique dest *without* replacing the tag? I think not (and it would avoid to walk the path several times)

fanf:
OK, I'm merging like that, but please open an other ticket for me to try to normalize things

PR #2106, opened 2019-01-02
Fixes #14056: Change request cannot be created -  "Couldn't find newly created entry when saving Change Request"

fanf:
https://issues.rudder.io/issues/14056


fanf:
Merging this one by hand to avoid auto-up-merge.

PR #2107, opened 2019-01-04
Fixes #14091: Configure report value of technique from technique editor

VinceMacBuche:
https://issues.rudder.io/issues/14091


VinceMacBuche:
Commit modified

PR #2118, opened 2019-01-23
Fixes #14212: In technique editor, ncf index is cached which defeats #14194 correction on migration

VinceMacBuche:
https://issues.rudder.io/issues/14212


VinceMacBuche:
OK, merging this PR

PR #2120, opened 2019-01-23
Fixes #14219: A var, leading to compilation error, was introduced

fanf:
https://issues.rudder.io/issues/14219


fanf:
OK, merging this PR

PR #2126, opened 2019-01-28
Fixes #14243: Rudder node id is confused with machine id and not selectable by triple-click

fanf:
https://issues.rudder.io/issues/14243


rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -377,7 +377,9 @@ object DisplayNode extends Loggable {
       <h4 class="tablemargin">General</h4>
 
         <div class="tablepadding">
-          <b>Hostname:</b> {sm.node.main.hostname}<br/>
+          {/* style "inline-block" is for selection by triple-click of the whole content */}
```
fanf: Only works on firefox :/

PR #2130, opened 2019-02-01
Fixes #10773: Rounding error in compliance bar

VinceMacBuche:
https://issues.rudder.io/issues/10773


VinceMacBuche:
At least one thing that could be changed: renmame number attribute to reports

fanf:
OK, merging this PR

PR #2148, opened 2019-02-25
Fixes #14359: Change Scala project structure so that parent-pom is a real parent project

fanf:
https://issues.rudder.io/issues/14359


fanf:
PR rebased

PR #2156, opened 2019-03-25
Fixes #7922: [API] Extend node information by "Last seen"

fanf:
https://issues.rudder.io/issues/7922

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/NodeDetailLevel.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/NodeDetailLevel.scala
@@ -43,6 +43,7 @@ import com.normation.inventory.domain._
 import com.normation.rudder.web.components.DateFormaterService
 import com.normation.rudder.domain.nodes.JsonSerialisation._
 import com.normation.rudder.domain.nodes.NodeInfo
+import org.joda.time.DateTime
 
 sealed trait NodeDetailLevel {
   def fields : Set[String]
@@ -51,12 +52,12 @@ sealed trait NodeDetailLevel {
    * Build the JSON object, for the given list of fields.
    * The object is always built with the fields on the same order.
    */
-  final def toJson(nodeInfo: NodeInfo, status: InventoryStatus, inventory: Option[FullInventory], software: Seq[Software]): JObject = {
+  final def toJson(nodeInfo: NodeInfo, status: InventoryStatus, optRunDate: Option[DateTime], inventory: Option[FullInventory], software: Seq[Software]): JObject = {
 
     val jsonFields: List[JField] = NodeDetailLevel.allFields.filter(fields.contains).map { field =>
       //map the field to its value with the correct context
       val json =                      NodeDetailLevel.statusInfoFields.   get(field).map( _(status)).
-        orElse(                       NodeDetailLevel.nodeInfoFields.     get(field).map( _(nodeInfo))).
+        orElse(                       NodeDetailLevel.nodeInfoFields.     get(field).map( _((nodeInfo, optRunDate)))).
         orElse(inventory.flatMap(i => NodeDetailLevel.fullInventoryFields.get(field).map( _(i)))).
         orElse(NodeDetailLevel.softwareFields.get(field).map( _(software))).
         getOrElse(JNothing)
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
```diff
@@ -562,12 +567,12 @@ class NodeApiService6 (
                        Full(Map[NodeId, Seq[Software]]())
                      }
     } yield {
-      val nodeIds = nodeFilter.getOrElse(nodeInfos.keySet)
       for {
         nodeId    <- nodeIds
         nodeInfo  <- nodeInfos.get(nodeId)
       } yield {
-        serializeInventory(nodeInfo, state, inventories.get(nodeId), software.getOrElse(nodeId, Seq()), detailLevel, version)
+        val runDate = runs.get(nodeId).map( _.map(_.agentRunId.date)).flatten
```
VinceMacBuche: flatmap instead of first map + flatten ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/NodeDetailLevel.scala
```diff
@@ -159,61 +161,63 @@ object NodeDetailLevel {
...
+  private val nodeInfoFields: Map[String, INFO => JValue] = {
+
+    val id           : INFO => JValue = (info:INFO) => info._1.id.value
+    val hostname     : INFO => JValue = (info:INFO) => info._1.hostname
+    val state        : INFO => JValue = (info:INFO) => info._1.state.name
+    val description  : INFO => JValue = (info:INFO) => info._1.description
+    val policyServer : INFO => JValue = (info:INFO) => info._1.policyServerId.value
+    val ram          : INFO => JValue = (info:INFO) => info._1.ram.map(MemorySize.sizeMb)
+    val arch         : INFO => JValue = (info:INFO) => info._1.archDescription
+    val runDate      : INFO => JValue = (info:INFO) => info._2.map(d => JString(DateFormaterService.getFormatedDate(d))).getOrElse(JNothing)
```
VinceMacBuche: So here you won't have "lastRunDate" field if you ask for it and there was no run present, maybe we should always have a value, saying 'never' ? what do you think of this ?

fanf: I wasn't sure. It seems easier/more idiomatic for user to test for "field is null" than for an arbitrary value. But really, I'm not sure of idiomatism for these thing

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
fanf: done

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/NodeDetailLevel.scala
VinceMacBuche: "idiotmatism"  would be fine for JSON ;)

What do our team members thinks ? what do people thinks ? what does the internet say ? let's look !

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/NodeDetailLevel.scala
```diff
@@ -159,61 +161,63 @@ object NodeDetailLevel {
...
+        ( "version"       -> info._1.osDetails.version.value ) ~
+        ( "fullName"      -> info._1.osDetails.fullName ) ~
+        ( "servicePack"   -> info._1.osDetails.servicePack ) ~
+        ( "kernelVersion" -> info._1.osDetails.kernelVersion.value )
     }
 
     val machine = {
-      ( inv : NodeInfo ) =>
-        val (machineType,provider) = inv.machine.map(_.machineType match {
+      ( info : INFO ) =>
```
VinceMacBuche: Here we could have unapplied the info to a tuple (inv, optLastRunDate ) but no pb il like the _1 too and would be quite PITA for you to rewrite it !!

PR #2166, opened 2019-03-26
Fixes #14312: Missing eventlogs for technique editor action and technique update

fanf:
https://issues.rudder.io/issues/14312

the provided commitid is the revision tree one, not a real commit id. The problem is that we totally can have several git commit without a reload, and to we can only log the state of the git when the reload is done. It is not possible to change that without changing all the logic on technique library, which is extremelly dangerous.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogFactory.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogFactory.scala
@@ -151,6 +151,16 @@ trait EventLogFactory {
     , reason      : Option[String]
   ) : ModifyNodeGroup
 
+  def getAddTechniqueFromDiff(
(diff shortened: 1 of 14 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogFactory.scala
```diff
@@ -592,7 +627,12 @@ class EventLogFactoryImpl(
     , severity    : Int = 100
     , reason      : Option[String]
   ) : DeleteTechnique = {
-    val details = EventLog.withContent(techniqueXmlSerializer.serialise(deleteDiff.technique) % ("changeType" -> "delete"))
+val details = EventLog.withContent{
```
ncharles: wrong identation

fanf:
Commit modified

fanf:
OK, squash merging this PR

PR #2169, opened 2019-03-29
Fixes #14550:  UI "settings" for management of hooks need Unix right

fanf:
https://issues.rudder.io/issues/14550


fanf:
That strange, before that commit, it seems to work on centos 7 but not on debian 9. Perhaps the problem was with the debian 9 installation, it may not be necessary. 
On the other hand, it would allow to have the same right management for both hooks, which is certainly better for the maintenability.

PR #2173, opened 2019-04-02
Fixes #14573: JS error in rule pages, and Directives don't show in the list of directives

RaphaelGauthier:
https://issues.rudder.io/issues/14573


ncharles:
OK, merging this PR

PR #2202, opened 2019-04-24
Fixes #14758: when accepting a change request on dynamic group, the group is empty, policy generated are so invalid, and it fixes itself only at next dynamic group refresh 

ncharles:
https://issues.rudder.io/issues/14758


ncharles:
this fixes the issue
BUT it causes double policy generation when accepting a change request that changes group, as dynGroupUpdate updates the group, so need a deployment

ncharles:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -111,6 +115,7 @@ class CommitAndDeployChangeRequestServiceImpl(
   , xmlSerializer         : XmlSerializer
   , xmlUnserializer       : XmlUnserializer
   , sectionSpecParser     : SectionSpecParser
+  , queryProcessor        : QueryProcessor
```
VinceMacBuche: I think you should use `DynGroupUpdaterService` which already has the logic and not the Query processor directly

And maybe we should do it at repository level in update function, to treat all call to update correctly, but maybe not since a repository should not modify data

ncharles: the DynGroupUpdaterService takes the group within ldap - so it would need to save first the group, with invalid node list (hence invalid nodelist) and then call update on it

VinceMacBuche: But you could do something like this: https://github.com/VinceMacBuche/rudder/commit/40671095ad721e93363eb8308b0f614b8e7f6109#diff-49b7910e1c5346540a45122d00fed7e2R94

ncharles:
> Last night i made a change to prevent empty groups when resotring archive [VinceMacBuche@4067109](https://github.com/VinceMacBuche/rudder/commit/40671095ad721e93363eb8308b0f614b8e7f6109)
> 
> Do we still need to trigger an automatic update during policy generation ?

There are pleinty of edge cases where we need it - if we accept a CR while accept a node, or if two people accept in a row, etc
so yes, it's safer to trigger an automatic update during policy generation, even if it doesn't do anything

Your change on archive looks good, could you make a PR out of that ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
ncharles: ha, indeed, it was not existent when I was coding it

ncharles:
Commit modified

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #2206, opened 2019-04-29
Use ZIO to manage Rudder result type

fanf:
Ready for review!

The intersting part is `com.normation.errors` which defines the main error types for rudder, along with two main result types: 
- `PureResult[A]` is a no effect result that can fail with a `RudderError` (for example because of business logic) or succeed with a `A` value. Under the hood it's a `Either[RudderError, A]`.
- `IOResult[A]` is a value describing a result with effects that can fail with a `RudderError` (for ex because an exception happened. Under the hood, it's a `ZIO[Any, RudderError, A]`.
Plus theres some generic mapping from/to `Option` (`notOptional`), `Box`, and exceptions. 

### `ZIO` parameters

`ZIO[R,E,A]` => 
- `R` == context, it is always `Any` for now but can be use for method-level dependency injection (to do. A big one). 
- `E` == error type. In our case, it's always a subtype of `RudderError`.
- `A` == the return value. 

There is 2 alias of `ZIO`: 
- `IO[A]` == `ZIO[Any,Throwable,A]` when your errors are exception (ie: java)
- `UIO[A]` == `ZIO[Any,Nothin,A]` for effectful programs that can't fail (for ex, `Ssytem.currentTimeMillis` neither fail).

And so `IOResult[A]` if one more alias. 

### Importing effectful code

`ZIO` has the very clean notion of importing an effectful code into a `ZIO` value. It is the clear limit between the two world, and the step you need to do each time you deal with effectuf (java) code. 

I added a `IORestult` variant, so that we (almost) always use it to be consistant in our types. 
Ex: `val hello =  IOResult.effect(println("hello!"))`
You can also import an other effectful prog, for ex is some logic is needed: 
```
val hello = IOResult.effectM {
  if(isAlice) IOResult.effect(println("hello Alice!")) 
  else Unexpected("it's not Alice!").fail
}
```
There is variant of `IOResult.effect/effectM` that takes an error message in case the imported code throws an exception. The message is used to build a `RudderError.System(message, exception)` error. 

### Combinators

There is a lot of wonderful combinators in ZIO. You can work as easely on the success value (`map`, etc) than on the error one (`mapError`, `catchAll`) or on both (`fold`, foldM`). 
The `xxxM` variant of these combinators tells that the arguments is also an effect (a `ZIO[R, E, A]`)

The most interesting ones for use are: 

- `ZIO.foreach`: it's traverse, renamad. `ZIO.foreach(iterable){ elt => doStuff(elt) }` with `doStuff:ZIO`. The first errors stop the traversal. 
- `ZIO.bracket(get resources)(release resource)(use resource)` which ensure that the resource release code is always executed whatever happen in `use resource` code. No more leak!

(... to be contnued...)


webapp/sources/rudder/rudder-web/doc/dev/desPaquetsDePolitiquesAuxPromisesCFE.mkdwn
```diff
@@ -105,17 +105,17 @@ Le passage des rules aux "promises + assets" n'est pas direct:
 Remarques
 ---------
 
-- on se fiche des assets entre le policy package et la fin: c'est un problème purement de 
+- on se fiche des assets entre le policy package et la fin: c'est un problème successment de
```
ncharles: sounds like an eager replace :)

VinceMacBuche:
I can't find com.normation.errors package in code change through github UI ... I'll look in the branch directly

VinceMacBuche:
OK ! found it here! webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala

webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
```diff
@@ -0,0 +1,377 @@
...
+    val prog =
+      log.error("wouhou") *>
+      ZIO.bracket(Task.effect{
+      val checkRelativePath = "file:///tmp/plop.txt"
+      val url = new URL(checkRelativePath)
+      url.openStream()
+    })(is =>
+      Task.effect(is.close).run // here, if I put `UIO.unit`, I can have the content
+    )(is =>
+      Task.effect(println(new String(is.readAllBytes(), "utf-8") ))
```
VinceMacBuche: is.readAllBytes does not compile with Java8

fanf:
Superseeded by #2218

PR #2209, opened 2019-04-30
Fixes #14797: A technique can have two agent section for the same kind of agent 

fanf:
https://issues.rudder.io/issues/14797

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
@@ -106,7 +106,16 @@ class TechniqueParser(
 
               parseAgentConfig(id, forCompatibilityAgent)
             }
-          )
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/TechniqueParser.scala
```diff
@@ -106,7 +106,16 @@ class TechniqueParser(
 
               parseAgentConfig(id, forCompatibilityAgent)
             }
-          )
+          // we need to filter back for totally empty agent (most likely default one added for nothing)
+          ).filter(a => List(a.templates, a.files, a.bundlesequence, a.runHooks).exists(_.nonEmpty))
```
VinceMacBuche: We could use .forall(_.empty) instead of .exists(_.nonEmpty) which seems easier to understand, or maybe just changing the comment to match the code because  i exepected something else

fanf:
Commit modified

fanf:
OK, squash merging this PR

PR #2214, opened 2019-05-07
Fixes #14834: Default archive retention is too large, which cause very large disk usage 

ncharles:
https://issues.rudder.io/issues/14834

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
@@ -141,11 +141,11 @@ rudder.jdbc.maxPoolSize=25
 # If the archive TTL is greater than the delete TTL then archiving will be disabled
 # as there will be no reports to archive (they would be deleted first).
 #
-# Defaults: archive after 4 days, delete after 90 days.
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -141,11 +141,11 @@ rudder.jdbc.maxPoolSize=25
 # If the archive TTL is greater than the delete TTL then archiving will be disabled
 # as there will be no reports to archive (they would be deleted first).
 #
-# Defaults: archive after 4 days, delete after 90 days.
+# Defaults: archive after 4 days, delete after 15 days.
 #
 
 rudder.batch.reportscleaner.archive.TTL=4
```
amousset: why keeping archives in the default config as they are not directly usable anyway? I could be an option for advanced users who need to keep data even if not accessible by the web interface.

ncharles: great idea, but i'm willing to do this change only in 5.1 (that's a massive change, and a big change in documentation)

PR #2217, opened 2019-05-13
Fixes #14241: Missing rule name in compliance API

ncharles:
https://issues.rudder.io/issues/14241


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -199,6 +199,7 @@ object JsonCompliance {
 
     def toJson(level: Int) = (
         ("id" -> rule.id.value)
+      ~ ("name" -> rule.name)
```
amousset: :boom:

fanf:
OK, merging this PR

PR #2241, opened 2019-06-06
Fixes #15035: Inefficient fetch of nodeconfigurations when runs are received 

ncharles:
https://issues.rudder.io/issues/15035


ncharles:
Commit modified

PR #2243, opened 2019-06-06
Ust 10551 rudder 5 1/make policy generation be node by node

fanf:
Version for branche 5.1 of node by node, depending where it should go.


fanf:
I'm closing that one as it won't be done for 5.1. We need to build a better story on:
1/ performance. We need to have *better* performance for that, not worse. 
2/ user feedback. We need to let the user know what is happening, what is actually configured on a node (ie warn on node details if the node didn't got the last config for ex), change generation progression, etc.

PR #2247, opened 2019-06-10
Fixes #15042: Updating changes/repairs (probably) uses too much memory 

ncharles:
https://issues.rudder.io/issues/15042

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
@@ -40,7 +40,6 @@ package com.normation.rudder.reports.execution
 
 import scala.concurrent.Future
 import scala.concurrent.ExecutionContext.Implicits.global
-
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.batch.FindNewReportsExecution
 import com.normation.rudder.domain.logger.ReportLogger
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
```diff
@@ -195,18 +193,19 @@ class ReportsExecutionService (
     Future {
       //update changes by rules
       (for {
-        changes <- reportsRepository.getChangeReportsOnInterval(lowestId, highestId)
-        updated <- cachedChanges.update(changes)
+        changes         <- reportsRepository.getChangeReportsOnInterval(lowestId, highestId)
```
ncharles: as an afterthought, it should probably be the getChangeReportsOnInterval which ought to have a different query

fanf: Afterthought of #15041?

fanf: OK, after some more hours on the code, I understand that you want to change the query to only return the relevant part (ruleId, executionTimestamp). That seems to be a much better solution to minimize memory, I agree!

fanf:
Superseeded by https://github.com/Normation/rudder/pull/2281

PR #2251, opened 2019-06-11
Fixes #15057: Repair broken nodeslist tests after parent ticket

ArmandSy:
https://issues.rudder.io/issues/15057

### DIFF
--- a/relay/sources/relayd/src/data/node.rs
+++ b/relay/sources/relayd/src/data/node.rs
@@ -216,7 +216,7 @@ mod tests {
 
         let file_nodelist = NodesList::new("tests/files/nodeslist.json", None).unwrap();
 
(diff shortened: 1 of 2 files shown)


relay/sources/relayd/src/data/node.rs
```diff
@@ -216,7 +216,7 @@ mod tests {
 
         let file_nodelist = NodesList::new("tests/files/nodeslist.json", None).unwrap();
 
-        assert_eq!(file_nodelist.get_neighbours().sort(), my_string_vec.sort());
+        assert_eq!(file_nodelist.get_neighbours(&"root".to_string()).sort(), my_string_vec.sort());
```
amousset: `"root"` should be enough, it is a `&'static str`, so also an `&str`

PR #2256, opened 2019-06-12
Fixes #15062: store run agent batch may never catch up on very loaded system because of inefficient index and the way it handles reports

ncharles:
https://issues.rudder.io/issues/15062

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
@@ -280,6 +280,26 @@ class ReportsJdbcRepository(doobie: Doobie) extends ReportsRepository with Logga
     }
   }
 
+  def getReportsWithLowestIdFromDate(from:DateTime) : Box[Option[(Long, Reports)]] = {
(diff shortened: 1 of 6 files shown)


ncharles:
Commit modified

fanf:
OK, merging this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ReportsJdbcRepository.scala
```diff
@@ -280,6 +280,26 @@ class ReportsJdbcRepository(doobie: Doobie) extends ReportsRepository with Logga
...
+    }
+  }
+
+  /**
+    *  utilitary methods
+    */
+
+  // Get max ID before a datetime
+  def getMaxIdBeforeDateTime(fromId: Long, before: DateTime): Box[Long] = {
+    (query[Long](s"select max(id) as id from RudderSysEvents where where id > ${fromId} and executionTimeStamp <  '${new Timestamp(before.getMillis)}''").
```
ncharles: extra ' here

ncharles: extra where also

PR #2260, opened 2019-06-12
Fixes #14465: Exception: Task java.util.concurrent.FutureTask rejected from java.util.concurrent.ThreadPoolExecutor

fanf:
https://issues.rudder.io/issues/14465

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -37,12 +37,11 @@
 
 package com.normation.rudder.services.policies
 
-import scala.Option.option2Iterable
+import java.util.concurrent.TimeUnit
+
 import org.joda.time.DateTime
-import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.Constants
 import com.normation.rudder.domain.nodes.NodeInfo
-import com.normation.rudder.domain.parameters.GlobalParameter
 import com.normation.rudder.domain.parameters.ParameterName
 import com.normation.rudder.domain.policies._
 import com.normation.rudder.repository._
@@ -51,9 +50,7 @@ import com.normation.rudder.services.nodes.NodeInfoService
 import com.normation.rudder.services.policies.nodeconfig.NodeConfigurationHash
 import com.normation.utils.Control._
 import net.liftweb.common._
-import com.normation.rudder.domain.parameters.GlobalParameter
 import com.normation.inventory.services.core.ReadOnlyFullInventoryRepository
-import com.normation.inventory.domain.NodeInventory
 import com.normation.inventory.domain.AcceptedInventory
 import com.normation.inventory.domain.NodeInventory
 import com.normation.rudder.domain.parameters.GlobalParameter
@@ -88,6 +85,14 @@ import com.normation.rudder.domain.logger.TimingDebugLogger
 import com.normation.rudder.domain.logger.PolicyLogger
 import cats.data.NonEmptyList
 import com.normation.rudder.domain.reports.OverridenPolicy
+import monix.eval.Task
+import monix.eval.TaskSemaphore
+import monix.execution.ExecutionModel
+import monix.execution.Scheduler
+import org.joda.time.Period
+
+import scala.concurrent.Await
+import scala.concurrent.duration.FiniteDuration
 
 
 /**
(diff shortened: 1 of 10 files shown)


fanf:
Commit modified

fanf:
Commit modified

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -762,11 +959,15 @@ object BuildNodeConfiguration extends Loggable {
       byNodeDrafts.toMap
     }
 
+
+    implicit val p = parallelism
+
     // 1.3: build node config, binding ${rudder./node.properties} parameters
     // open a scope for the JsEngine, because its init is long.
-    JsEngineProvider.withNewEngine(scriptEngineEnabled) { jsEngine =>
+    // Hardcoded 4 thread for test case.
```
ncharles: this comment should not be there

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -204,53 +365,58 @@ trait PromiseGenerationService extends Loggable {
...
-      _                     =  logger.debug(s"Node contexts built in ${timeNodeContexts} ms, start to build new node configurations.")
+      _                     =  PolicyLogger.debug(s"Node contexts built in ${timeNodeContexts} ms, start to build new node configurations.")
 
       buildConfigTime       =  System.currentTimeMillis
       /// here, we still have directive by directive info
-      configs               <- buildNodeConfigurations(activeNodeIds, ruleVals, nodeContexts, allNodeModes, scriptEngineEnabled, globalPolicyMode) ?~! "Cannot build target configuration node"
-      nodeConfigs           =  configs.map(c => (c.nodeInfo.id, c)).toMap
+      configsAndErrors      <- buildNodeConfigurations(activeNodeIds, ruleVals, nodeContexts, allNodeModes, scriptEngineEnabled, globalPolicyMode, parallelism, jsTimeout) ?~! "Cannot build target configuration node"
+      /// only keep successfull node config. We will keep the failed one to fail the whole process in the end if needed
+      nodeConfigs           =  configsAndErrors.ok.map(c => (c.nodeInfo.id, c)).toMap
```
ncharles: this is a huge behavioural change - it means some nodes will have policy not like the one in the interface, and some other will. This is likely to cause some massive headache, and should be optionnal

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -104,11 +109,129 @@ trait PromiseGenerationHooks {
...
+    //exec
+    Await.result(tasks.runAsync(parallelism.scheduler), timeout)
+  }
+
+  def parallelSequenceGen[U, T, E](seq: Seq[U])(f:U => Either[E, T], recover:(U, Throwable) => Either[E, T])(implicit parallelism: Parallelism): Either[NonEmptyList[E], Seq[T]] = {
+
+    val updated = parallelApplicativeGen(seq)(f, recover)
+
+    // gather results
+    val gatherErrors = updated.foldLeft[Either[NonEmptyList[E], Seq[T]]](Right(Nil)) {
```
VinceMacBuche: Funny, errors and success will be in different ordrer (error are added as head of list, and result on tail) Don't knwo if it's a problem though ... since it's unordered and parallel i think it's ok

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -132,56 +255,94 @@ trait PromiseGenerationService extends Loggable {
...
     import HooksImplicits._
 
+    val maxParallelism = {
+      // We want to limit the number of parallel execution and threads to the number of core/2 (minimum 1) by default.
+      // This is taken from the system environment variable because we really want to be able to change it at runtime.
+      def threadForProc(mult: Double): Int = {
+        Math.max(1, (Runtime.getRuntime.availableProcessors * mult).ceil.toInt)
+      }
+      val t = try {
+        getMaxParallelism().getOrElse("x0.5") match {
```
VinceMacBuche: Value of maxParallelism should be directly a double or a float (and maybe should be name parallelismFactor) with value between 0 and 1 (maybe not limited by 1 but i do fell it would cause problem to have more than 1)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -132,56 +255,94 @@ trait PromiseGenerationService extends Loggable {
...
+      }
+    }
+    val jsTimeout = {
+      // by default 5s but can be overrided
+      val t = getJsTimeout().getOrElse(5)
+      FiniteDuration(try {
+        Math.max(1, t.toLong) // must be a positive number
+      } catch {
+        case ex: NumberFormatException =>
+          PolicyLogger.error(s"Impossible to user property '${t}' for js engine timeout: not a number")
```
VinceMacBuche: error should state that default value of "5 seconds will be use instead"

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
```diff
@@ -612,35 +615,42 @@ final object JsEngine {
...
+          task.get(maxTime.toMillis, TimeUnit.MILLISECONDS)
       } catch {
         case ex: ExecutionException => //this can happen when rhino get security exception... Yeah...
           throw RudderFatalScriptException(s"Evaluating script '${name}' was forced interrupted due to ${ex.getMessage}, aborting.", ex)
 
         case ex: TimeoutException =>
           //try to interrupt the thread
           try {
-            // try to gently terminate the thread
-            pool.shutdownNow()
```
VinceMacBuche: What this throwing the 'threadpool' exception we used to get sometimes ? would be logic though, threadpool was shutdown but other thread could still be created/using the threadpool

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
VinceMacBuche: Ok we can use it to set a fix value ...

VinceMacBuche:
Apart small question, and to add a flag to activate the feature (would allow to fail fast like current 5.0, preventing partial updates, but keeping current behavior), we shoudl merge this

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/JavascriptEngine.scala
ncharles: exactly

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
VinceMacBuche: I do think it's a better behavior but I totally agree with Nicolas that we can't change a behavior that is such important in Rudder without adding a flag to enable/disable it

VinceMacBuche:
OK, merging this PR

PR #2273, opened 2019-06-17
Fixes #15056: Implementing agents effectively in remote-run API

ArmandSy:
https://issues.rudder.io/issues/15056


relay/sources/relayd/hashtest.txt
```diff
@@ -0,0 +1 @@
+hashtesting
```
amousset: Tests should go in a dedicated folder in `tests/`

relay/sources/relayd/src/api.rs
```diff
@@ -28,26 +28,37 @@
 // You should have received a copy of the GNU General Public License
 // along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
 
+extern crate data_encoding;
```
amousset: these `extern crate` lines should not be necessary (as we are using Rust 2018)

relay/sources/relayd/src/api.rs
```diff
@@ -28,26 +28,37 @@
...
 };
-use warp::{Filter};
+use warp::Filter;
 
-use crate::remote_run::{nodes_handle};
+use crate::remote_run::{nodes_handle, nodes_handle2, AgentParameters, RemoteRun, RemoteRunTarget};
+use crate::shared_files::{
+    compute_hash, metadata_writer, parse_hash_from_raw, parse_path_from_peek,metadata_hash_checker
+};
+use ring::digest::{Context, Digest, SHA1, SHA256, SHA512};
```
amousset: I don't think we should use ring at all (it relies on C code that would make it harder to build and package)

relay/sources/relayd/Cargo.toml
```diff
@@ -48,6 +48,16 @@ reqwest = "0.9"
...
 regex = "1.1"
 itertools = "0.8"
+tokio-process = "0.2.3"
+tokio-io = "0.1.12"
+hyper = "0.12.31"
+sha1 = "0.6.0"
+sha2 = "0.8.0"
+io = "0.0.2"
+file = "1.1.2"
+blake2 = "0.8.0"
```
amousset: I don't think we support blake2 for file hashes in the API

relay/sources/relayd/src/shared_files.rs
```diff
@@ -0,0 +1,136 @@
...
+// along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+extern crate data_encoding;
+extern crate ring;
+use crate::error::Error;
+use std::collections::HashMap;
+
+use std::fs;
+
+use ring::digest::{Context, Digest, SHA1, SHA256, SHA512};
```
amousset: we should use crates from https://github.com/RustCrypto/hashes instead.

relay/sources/relayd/Cargo.toml
```diff
@@ -48,6 +48,16 @@ reqwest = "0.9"
 chrono = { version = "0.4", features = ["serde"] }
 regex = "1.1"
 itertools = "0.8"
+tokio-process = "0.2.3"
+tokio-io = "0.1.12"
+hyper = "0.12.31"
+sha1 = "0.6.0"
+sha2 = "0.8.0"
+io = "0.0.2"
```
amousset: This crate seems to only contain https://docs.rs/io/0.0.2/src/io/lib.rs.html#1-3 so I guess it is not what you expected :)

PR #2282, opened 2019-06-21
Fixes #15117: Use a queue for "changes" cache update

fanf:
https://issues.rudder.io/issues/15117

This need to be tested in the load test plateform to check that it didn't make things worse.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
@@ -37,12 +37,10 @@
 
 package com.normation.rudder.services.reports
 
-import scala.concurrent._
-import scala.concurrent.ExecutionContext.Implicits.global
+import com.github.ghik.silencer.silent
 import org.joda.time.DateTime
 import org.joda.time.Interval
 import com.normation.rudder.domain.policies.RuleId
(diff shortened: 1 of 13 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -263,49 +302,75 @@ class CachedNodeChangesServiceImpl(
...
+   * Get changes and update cache accordingly
+   */
+  protected def updateCache(lowestId: Long, highestId: Long): IOResult[Unit] = {
+    (for {
+      time1   <- currentTimeMillis
+      //
+      // actually, in syncUpdate we check if the cache was inited or not.
+      // when the cache is not, we do it - but what was the point to get
+      // changes here in that case ?
+      //
```
fanf: I think the cache should keep the highestId it knows (either after init or after runs), and that we should get changes in `[lowestId, highestId]` interval, but in `[max(lowestId, cacheHighestId), highestId]`

fanf: it is now done

fanf:
We will need to test all of 5.1 for load. But it must be after initial release of 5.1 given the planning.

PR #2286, opened 2019-06-27
Fixes #15127: Create system variables for the new protocol

ncharles:
https://issues.rudder.io/issues/15127


ncharles:
I still need to edit unit test for generated policies

ncharles:
Commit modified

PR #2294, opened 2019-07-09
Fixes #14389: Duplicate dependency "jsr305" in rudder-core/pom.xml

amousset:
https://issues.rudder.io/issues/14389


amousset:
OK, merging this PR

PR #2297, opened 2019-07-10
Fixes #15199: Update relay component - handle broken reports

amousset:
https://issues.rudder.io/issues/15199


amousset:
Commit modified

PR #2298, opened 2019-07-11
Fixes #15196: make the  remote run agent parameters configurable

ArmandSy:
https://issues.rudder.io/issues/15196

### DIFF
--- a/relay/sources/relayd/src/configuration.rs
+++ b/relay/sources/relayd/src/configuration.rs
@@ -82,6 +82,7 @@ pub struct Configuration {
     pub general: GeneralConfig,
     pub processing: ProcessingConfig,
     pub output: OutputConfig,
+    pub remote_run: RemoteRun,
 }
 
 impl Configuration {
@@ -165,6 +166,13 @@ impl OutputSelect for InventoryOutputSelect {
     fn is_enabled(&self) -> bool {
         *self != InventoryOutputSelect::Disabled
     }
+}
+
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
+
+pub struct RemoteRun
+{
+    pub command : PathBuf,
 }
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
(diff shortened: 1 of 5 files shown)


relay/sources/relayd/src/configuration.rs
```diff
@@ -167,6 +168,13 @@ impl OutputSelect for InventoryOutputSelect {
     }
 }
 
+#[derive(Deserialize, Debug, PartialEq, Eq, Clone)]
+
+pub struct RemoteRun
+{
+    pub command : PathBuf,
```
amousset: No space before :, but `cargo fmt` should take care of that

relay/sources/relayd/src/remote_run.rs
```diff
@@ -175,14 +176,13 @@ impl RunParameters {
...
         let program = if test_mode {
-            "echo"
+            PathBuf::from("echo")
         } else {
-            // FIXME make it configurable
-            "/opt/rudder/bin/rudder"
-        }
-        .to_string();
+            let jobconfig2 = &job_config.clone();
+            PathBuf::from(jobconfig2.clone().cfg.remote_run.command.to_str().unwrap())
```
amousset: `job_config.clone().cfg.remote_run.command.clone()` should be enough

relay/sources/relayd/src/remote_run.rs
```diff
@@ -38,6 +38,7 @@ use std::{
     process::{Command, Stdio},
     str::FromStr,
     sync::Arc,
+    path::{Path, PathBuf},
```
amousset: `Path` is not used and produces a warning, we should not have warnings in merged code.

relay/sources/relayd/src/remote_run.rs
amousset: We have to clone because a reference is not enough as we have to return an owned `PathBuf`

ArmandSy:
Commit modified

PR #2299, opened 2019-07-11
Fixes #15201: support "classes" and "conditions" as parameters in the remote-run API

ArmandSy:
https://issues.rudder.io/issues/15201

### DIFF
--- a/relay/sources/relayd/src/remote_run.rs
+++ b/relay/sources/relayd/src/remote_run.rs
@@ -56,7 +56,7 @@ impl RemoteRun {
             run_parameters: RunParameters::new(
                 options.get("asynchronous"),
                 options.get("keep_output"),
-                options.get("classes"),
+                if options.contains_key("conditions") {options.get("conditions")} else {options.get("classes")},
             )?,
         })
     }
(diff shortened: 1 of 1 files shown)


relay/sources/relayd/src/remote_run.rs
```diff
@@ -56,7 +56,11 @@ impl RemoteRun {
             run_parameters: RunParameters::new(
                 options.get("asynchronous"),
                 options.get("keep_output"),
-                options.get("classes"),
+                if options.contains_key("conditions") {
```
peckpeck: Doesn't this allow the user to pass any parameter to the remote run command ?

amousset: `RunParameters` constructor parses the conditions field and validates them (https://github.com/Normation/rudder/blob/master/relay/sources/relayd/src/remote_run.rs#L161) as valid conditions separated by commas, so it should not be possible to inject arbitrary commands (but arbitrary conditions indeed, which could be a problem in some cases, but we would need to distribute a list of valid conditions for remote run)

PR #2300, opened 2019-07-11
Fixes #15204: Update relay component - Make tokio runtime configurable

amousset:
https://issues.rudder.io/issues/15204

### DIFF
--- a/relay/sources/relayd/src/configuration.rs
+++ b/relay/sources/relayd/src/configuration.rs
@@ -109,6 +109,9 @@ pub struct GeneralConfig {
     pub nodes_certs_file: NodesCertsFile,
     pub node_id: NodeId,
     pub listen: SocketAddr,
+    /// None means using the number of available CPUs
+    pub core_threads: Option<usize>,
+    pub blocking_threads: usize,
 }
 
 #[derive(Deserialize, Debug, PartialEq, Eq, Copy, Clone)]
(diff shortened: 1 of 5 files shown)


relay/sources/relayd/src/configuration.rs
```diff
@@ -425,6 +427,8 @@ mod tests {
                 nodes_certs_file: PathBuf::from("tests/keys/nodescerts.pem"),
                 node_id: "root".to_string(),
                 listen: "127.0.0.1:3030".parse().unwrap(),
+                core_threads: None,
+                blocking_threads: 100,
```
peckpeck: how was 100 chosen ? isn't it a lot ?

relay/sources/relayd/tests/files/config/main.conf
```diff
@@ -4,6 +4,10 @@ nodes_certs_file = "tests/keys/nodescerts.pem"
 node_id = "root"
 listen = "127.0.0.1:3030"
 
+# By default, the number of CPUs
+#core_threads = "4"
```
peckpeck: This does not seem to match the None above

relay/sources/relayd/src/configuration.rs
amousset: It's the default, and this pool is only used for io blocking operations. We can move to something like 50, it would probably be enough for most usecases.

relay/sources/relayd/tests/files/config/main.conf
amousset: It gives `None` when the key is not is the config. Once you use the configuration line, you have to give a precise number of thread.
We could add another option like a boolean "autoscale_core_threads" (or a magic value like 0, but I don't like it)

PR #2303, opened 2019-07-11
Fixes #15207: Error when modifying a Directive when having the Change Request plugin

ncharles:
https://issues.rudder.io/issues/15207


ncharles:
caution: i didn't test it

fanf:
The proposed change only normalise the log message, not the comparison, so it won't work. I will do the correction again.

fanf:
Superseeded by https://github.com/Normation/rudder/pull/2304

PR #2304, opened 2019-07-11
Fixes #15207: Error when modifying a Directive when having the Change Request plugin

fanf:
https://issues.rudder.io/issues/15207


fanf:
OK, merging this PR

PR #2311, opened 2019-07-17
[TEST] do not merge

amousset:


VinceMacBuche:
Great change, merging ... (i don't see the scala checks :( i would have wanted two checks) maybe something not ok in the way we call github api

amousset:
I think I saw it but it is too slow and does not keep up with my commits

amousset:
Here it comes

PR #2314, opened 2019-07-17
Fixes #15238: Move /opt/rudder/etc/ssl/nodeslist.cert to /var/rudder/lib/ssl

amousset:
https://issues.rudder.io/issues/15238


amousset:
Commit modified

PR #2319, opened 2019-07-18
Fixes #15245: Limiting conditions length to 1024 characters

ArmandSy:
https://issues.rudder.io/issues/15245


relay/sources/relayd/src/remote_run.rs
```diff
@@ -148,11 +148,16 @@ impl FromStr for Condition {
...
         if !re.is_match(s) {
-            Err(Error::InvalidCondition(format!(
+            return Err(Error::InvalidCondition(format!(
                 "Wrong condition: '{}', it should match {}",
                 s.to_string(),
                 condition_regex
-            )))
+            )));
+        }
+        if s.len() > 1024 {
```
amousset: Could you move it before the regex check? It could help prevent a DoS based on the expensive regex check (though unlikely to happen :) )

ArmandSy:
Commit modified

PR #2321, opened 2019-07-18
Fixes #15239: Generate a list of directives with their id and mode

VinceMacBuche:
https://issues.rudder.io/issues/15239


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -613,7 +614,23 @@ class PolicyWriterServiceImpl(
...
 
-  // just write an empty file for now
+
+  private[this] def writeDirectiveCsv(paths: NodePromisesPaths, policies: Seq[Policy], policyMode : GlobalPolicyMode) =  {
+    val path = new File(paths.newFolder, "rudder-directives.csv")
+
+    val csvContent = for {
+      policy <- policies
+    } yield {
+      s"""${policy.id.directiveId.value};${policy.policyMode.getOrElse(policyMode.mode).name};${policy.technique.generationMode.name};${policy.technique.agentConfig.runHooks.nonEmpty};${policy.technique.id.name};${policy.technique.id.version}"""
```
fanf: What is going to be displayed to user for them to choose? It can't be the uuid, user won't understand. I believe it must be directive name set in UI, or an ID for that.  
Also, it need to be sorted by some user interesting thing. I don't know if it's technique/directive, or just directive, or rule/directive... But someone must find it back quickly (even in that file, be it just to minimize changes between run and so that we can compare file by diff). 
Also, we need to know if the directive is system (and perhaps just ignore them totally, I'm not sure we want to let the user only run one of them). 
Finally, I think we need to group directive that we merged - I'm not sure it's the case at that point, but perhaps it is, I don't remember

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -613,7 +614,23 @@ class PolicyWriterServiceImpl(
...
 
-  // just write an empty file for now
+
+  private[this] def writeDirectiveCsv(paths: NodePromisesPaths, policies: Seq[Policy], policyMode : GlobalPolicyMode) =  {
+    val path = new File(paths.newFolder, "rudder-directives.csv")
+
+    val csvContent = for {
+      policy <- policies
+    } yield {
+      s"""${policy.id.directiveId.value};${policy.policyMode.getOrElse(policyMode.mode).name};${policy.technique.generationMode.name};${policy.technique.agentConfig.runHooks.nonEmpty};${policy.technique.id.name};${policy.technique.id.version};"${policy.directiveOrder.value}""""
```
fanf: So, we converged with a normalized solution and still easy to parse in shell!
- fully quoted csv, ie: `"field 1","field 2","some \" in field 3","some , in last field"
- the first fields MUST be comma-free, only the last can contain comma. 
In shell, we will use `cut` with the numbered fields for first columns (uuid,etc) and a `cut` "everything remaining" for last part.

VinceMacBuche:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: The first fields are missing the quotes

VinceMacBuche:
Commit modified

VinceMacBuche:
last push was a push dto rebase, because it was not compiling due to change in ZioRuntime;) here is the change

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -613,7 +614,31 @@ class PolicyWriterServiceImpl(
...
+    } yield {
+      ( policy.id.directiveId.value ::
+      policy.policyMode.getOrElse(policyMode.mode).name ::
+      policy.technique.generationMode.name ::
+      policy.technique.agentConfig.runHooks.nonEmpty ::
+      policy.technique.id.name ::
+      policy.technique.id.version ::
+      policy.directiveOrder.value ::
+      Nil ).mkString("\"","\",\"","\"")
+
```
fanf: I believe we need the list to be sorted by `directiveOrder` and that we need to know if the technique is system or not.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #2322, opened 2019-07-19
Fixes #15247: Validating signatures in shared-files API

ArmandSy:
https://issues.rudder.io/issues/15247

### DIFF
--- a/relay/sources/relayd/src/shared_files.rs
+++ b/relay/sources/relayd/src/shared_files.rs
@@ -33,9 +33,11 @@ use chrono::prelude::DateTime;
 use chrono::{Duration, Utc};
 use hyper::StatusCode;
 use openssl::error::ErrorStack;
+use openssl::hash::MessageDigest;
 use openssl::pkey::PKey;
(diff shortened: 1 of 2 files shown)


relay/sources/relayd/src/shared_files.rs
```diff
@@ -76,6 +78,13 @@ impl HashType {
             }
         }
     }
+
+    fn hashtype_to_openssl_hash(&self) -> MessageDigest {
```
amousset: `to_openssl_hash` is enough as it is already a method on `HashType`

relay/sources/relayd/src/shared_files.rs
```diff
@@ -177,6 +186,19 @@ pub fn get_pubkey(pubkey: String) -> Result<PKey<Public>, ErrorStack> {
     )?)
 }
 
+pub fn validate_signature(file: String, pubkey: PKey<Public>, hash_type: HashType, digest: &[u8]) -> Result<bool, ErrorStack>
```
amousset: Looks good!

PR #2328, opened 2019-07-22
Fixes #15261: Update relay component - allow arbitrary log filter

amousset:
https://issues.rudder.io/issues/15261


amousset:
Commit modified

PR #2358, opened 2019-08-01
Fixes #15374: rudder-server-relay fails during postinstall

peckpeck:
https://issues.rudder.io/issues/15374


relay/sources/rudder-server-relay-postinst
```diff
@@ -76,13 +83,6 @@ if [ ! -f /var/rudder/lib/ssl/nodescerts.pem ]; then
   touch /var/rudder/lib/ssl/nodescerts.pem
```
amousset: shoudl be allnodescerts.pem

amousset: (only needed and generated on root server, but does no harm if present elsewhere)

peckpeck:
PR updated with a new commit

relay/sources/rudder-server-relay-postinst
peckpeck: Done: it will be empty anyway

peckpeck:
OK, squash merging this PR

PR #2379, opened 2019-08-08
Fixes #15434: Cannot start relayd

ncharles:
https://issues.rudder.io/issues/15434


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/WriteNodeCertificatesPem.scala
```diff
@@ -96,14 +100,18 @@ class WriteNodeCertificatesPemImpl(reloadScriptPath: Option[String]) extends Wri
                 })}
       writen <- writeCertificatesToNew(allCertsNew, certs)
       moved  <- IOResult.effect(allCertsNew.moveTo(file, overwrite = true))
+      //perm   <- IOResult.effect(file.setGroup("rudder")) -> fails with SystemError: An error occured; cause was: java.io.IOException: 'owner' parameter can't be a group
```
ncharles: pointing this sadeness to @fanf  so that he'll have an enlighted advice

fanf:
For reference, it's trivial to reproduce and it's now reported here: https://github.com/pathikrit/better-files/issues/332

PR #2381, opened 2019-08-09
Fixes #15441: change call to rudder relay reload to check permission of generated file

ncharles:
https://issues.rudder.io/issues/15441


webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -535,7 +535,7 @@ rudder.hooks.ignore-suffixes= .swp, ~, .bak, \
 # After change, we need to notify relayd to reload that file. This is the relad
```
amousset: reload

ncharles:
Commit modified

PR #2384, opened 2019-08-13
Fixes #15456: Handle ttl parameter correctly in shared-files API

ArmandSy:
https://issues.rudder.io/issues/15456

### DIFF
--- a/relay/sources/relayd/src/shared_files.rs
+++ b/relay/sources/relayd/src/shared_files.rs
@@ -28,7 +28,7 @@
 // You should have received a copy of the GNU General Public License
 // along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
 
-use crate::error::Error;
+use crate::{error::Error, JobConfig};
 use chrono::prelude::DateTime;
 use chrono::{Duration, Utc};
 use hyper::StatusCode;
@@ -46,6 +46,7 @@ use std::io::BufRead;
 use std::io::Read;
 use std::path::Path;
 use std::str::FromStr;
+use std::sync::Arc;
 use warp::Buf;
 
 pub enum HashType {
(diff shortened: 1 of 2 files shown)


ArmandSy:
Commit modified

ArmandSy:
Commit modified

ArmandSy:
Commit modified

relay/sources/relayd/src/shared_files.rs
```diff
@@ -235,13 +242,40 @@ pub fn file_writer(buf: &mut warp::body::FullBody, path: &str) {
...
+        return StatusCode::from_u16(404).unwrap();
+    }
+
     let _ = fs::create_dir_all(format!("shared-files/{}/{}/", target_uuid, source_uuid)); // on cree les folders s'ils existent pas
                                                                                           //fs::create_dir_all(format!("/var/rudder/configuration-repository/shared-files/{}/{}/", target_uuid, source_uuid)); // real path
-    fs::write(format!("shared-files/{}.metadata", peek), metadata_string)
-        .expect("Unable to write file");
+    fs::write(
+        format!("shared-files/{}.metadata", peek),
+        format!("{}expires={}\n", metadata_string, timestamp.unwrap()),
```
amousset: we shouldn't unwrap here, we need to properly manage the error

amousset: I just saw it is actually correct as the error cases returns early. Could you change the definition of timestamp to something like ```rust
let timestamp = match parse_ttl(..) {
    Ok(ttl) => ttls,
    Err(_) => return StatusCode::from_u16(500).unwrap(),
}```

to make it clearer?
}

ArmandSy:
Commit modified

relay/sources/relayd/src/data/node.rs
```diff
@@ -97,7 +97,7 @@ impl RawNodesList {
 }
 
 pub struct NodesList {
-    list: RawNodesList,
+    pub list: RawNodesList,
```
amousset: It should not be public, but you can add a get_... method in `impl`

ArmandSy:
Commit modified

relay/sources/relayd/src/data/node.rs
```diff
@@ -123,6 +123,10 @@ impl NodesList {
         Ok(NodesList { list: nodes, my_id })
     }
 
+    pub fn get_nodeslist(&self) -> &NodesList {
```
amousset: The idea of the fields not being public is to prevent code from outside of this module to rely on the internal representation of the nodes list, which may evolve. Instead you should create a dedicated method for what you are trying to to. In this case, it is even easier is it seems to already be there :)

ArmandSy:
Commit modified

PR #2386, opened 2019-08-13
Fixes #15461: reports are not forwarded to the server du to errer

ncharles:
https://issues.rudder.io/issues/15461


amousset:
OK, merging this PR

PR #2388, opened 2019-08-14
Fixes #15472: upgrade API settings to v12 to include default report protocol and syslog protocol disabled 

ncharles:
https://issues.rudder.io/issues/15472

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
@@ -95,8 +95,6 @@ class SettingsApi(
       RestRelaySynchronizePolicies ::
       RestRelaySynchronizeSharedFiles ::
       RestReportingProtocol ::
-      RestReportProtocolDefault ::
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
```diff
@@ -124,8 +122,18 @@ class SettingsApi(
 
   val allSettings_v8 = RestUseReverseDNS :: allSettings_v10
 
+  val allSettings_v12 =   RestReportProtocolDefault :: RestSyslogProtocolDisabled :: allSettings_v10
```
amousset: why is it based on 10 and not 11?

amousset: Because version 11 did not change anything in settings?

ncharles: exactly, latest API settings is v10 (v11 is simply forwarding to v10)

PR #2408, opened 2019-08-20
Arch 15521/add hardening config in relayd systemd unit

amousset:


amousset:
OK, squash merging this PR

PR #2415, opened 2019-08-21
Fixes #15536: Updating Apache's reverse proxy configuration to handle HTTPS requests on remote-run API

ArmandSy:
https://issues.rudder.io/issues/15536

### DIFF
--- a/relay/sources/apache/rudder-apache-relay-common.conf
+++ b/relay/sources/apache/rudder-apache-relay-common.conf
@@ -77,6 +77,8 @@ Alias /reports /var/rudder/reports/incoming
 # Load relay-api
 WSGIScriptAlias /rudder/relay-api /opt/rudder/share/relay-api/relay-api.wsgi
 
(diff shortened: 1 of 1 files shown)


relay/sources/apache/rudder-apache-relay-common.conf
```diff
@@ -75,7 +75,9 @@ Alias /reports /var/rudder/reports/incoming
 </Directory>
 
 # Load relay-api
-WSGIScriptAlias /rudder/relay-api /opt/rudder/share/relay-api/relay-api.wsgi
+#WSGIScriptAlias /rudder/relay-api /opt/rudder/share/relay-api/relay-api.wsgi
```
amousset: In fact we need to keep this for old API not migrated to relayd yet.

amousset: We may need to make it more precise or put the ProxyPass before to avoid conflict

peckpeck: i think the proxypass below is already precise enough

ArmandSy:
Commit modified

PR #2417, opened 2019-08-22
Fixes #15547: Empty keyhashes in nodeslist.json on Linux nodes

ncharles:
https://issues.rudder.io/issues/15547

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeInfo.scala
@@ -165,8 +165,13 @@ final case class NodeInfo(
           ""
         }
       case Some(cert : Certificate) =>
-        logger.info(s"Node '${hostname}' (${id.value}) is a Dsc node and a we do not know how to generate a hash yet")
-        ""
+        CFEngineKey.getSha256Digest(cert) match {
+          case Full(hash) => hash
+          case eb:EmptyBox =>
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -323,15 +323,18 @@ final class AsyncDeploymentActor(
 
         case Full(nodeIds) =>
           logger.info(s"Successful policy update '${id}' [started ${startTime.toString(timeFormat)} - ended ${endTime.toString(timeFormat)}]")
+          println("prout")
```
amousset: :thinking:

VinceMacBuche:
There is to commits, can you rebase it ? or edit commit messageof second on to be a fixup!

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
VinceMacBuche: :boom:

VinceMacBuche: I think you did not want to commit that file and was automatically commited by rudder-dev

ncharles:
rudder-dev fails, and pushed extra files the second time. Now github doesn't refresh the code :)

VinceMacBuche:
It's not changing expected result of policy generation tests ? I thought it would change in rudder.json but maybe not ...

IF tests are ok it's good to me!

ncharles:
Tests are all great for me

PR #2423, opened 2019-08-23
Fixes #15567: rudder-pkg command in /opt/rudder/bin should not be created in postinst but at build time

Fdall:
https://issues.rudder.io/issues/15567


Fdall:
Commit modified

PR #2425, opened 2019-08-23
Fixes #15568: Display warning when plugin license is near expiration date 

fanf:
https://issues.rudder.io/issues/15568


fanf:
The style/displayed info can be improved but at least backend info and the main notification logic is present.

PR #2431, opened 2019-08-28
Fixes #15423: Filtering tags with more than one tag with same "key" did not return correct results

RaphaelGauthier:
https://issues.rudder.io/issues/15423


VinceMacBuche:
OK, merging this PR

PR #2438, opened 2019-09-04
Fixes #15629: Rudder-pkg create a symlink pointing to a non existing path

Fdall:
https://issues.rudder.io/issues/15629


Fdall:
Commit modified

PR #2458, opened 2019-09-11
Bug 15675/leak in cache of node compliance and nodeinfo and perfs improvement

ncharles:
i'm creating a PR to be able to comment on it easily

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -401,11 +401,14 @@ trait PromiseGenerationService {
       configsAndErrors      <- buildNodeConfigurations(activeNodeIds, ruleVals, nodeContexts, allNodeModes, scriptEngineEnabled, globalPolicyMode, parallelism, jsTimeout, generationContinueOnError) ?~! "Cannot build target configuration node"
       /// only keep successfull node config. We will keep the failed one to fail the whole process in the end if needed
       nodeConfigs           =  configsAndErrors.ok.map(c => (c.nodeInfo.id, c)).toMap
+      configsError          =  configsAndErrors.errors
       timeBuildConfig       =  (System.currentTimeMillis - buildConfigTime)
       _                     =  PolicyLogger.debug(s"Node's target configuration built in ${timeBuildConfig} ms, start to update rule values.")
 
       updatedNodeConfigIds  =  getNodesConfigVersion(nodeConfigs, nodeConfigCaches, generationTime)
-      updatedNodeConfigs    =  nodeConfigs.filterKeys(id => updatedNodeConfigIds.keySet.contains(id))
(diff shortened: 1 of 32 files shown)


ncharles:
there's something suboptimal here, as the memory usage is enormous

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -415,10 +418,11 @@ trait PromiseGenerationService {
       ///// so now we have everything for each updated nodes, we can start writing node policies and then expected reports
 
       // WHY DO WE NEED TO FORGET OTHER NODES CACHE INFO HERE ?
-      _                     <- forgetOtherNodeConfigurationState(nodeConfigs.keySet) ?~! "Cannot clean the configuration cache"
+      _                     <- forgetOtherNodeConfigurationState(nodeConfigs.keysIterator.toSet) ?~! "Cannot clean the configuration cache"
```
ncharles: this could be allNodeConfigsId

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/HomePage.scala
```diff
@@ -421,7 +421,7 @@ class HomePage extends Loggable {
       }.toMap.mapValues{_.maxBy(_.value)} // Get the max version available
 
       // take back the initial set of nodes to be sure to have one agent for each
-      val allAgents = nodeInfos.keySet.toSeq.map(nodeId => agentVersionByNodeEntries.get(nodeId) match {
+      val allAgents = nodeInfos.keysIterator.toSeq.map(nodeId => agentVersionByNodeEntries.get(nodeId) match {
```
ncharles: maybe overkill, as it was a keySet.toSeq.map

fanf: `keySet.toSet.map` is less effecient than `keysIterator.map` (the seq is not created)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -213,20 +214,23 @@ class PolicyWriterServiceImpl(
   override def writeTemplate(
       rootNodeId      : NodeId
     , nodesToWrite    : Set[NodeId]
-    , allNodeConfigs  : Map[NodeId, NodeConfiguration]
+    , nodeConfigsToWrite  : Map[NodeId, NodeConfiguration]
+    , allNodeInfos    : Map[NodeId, NodeInfo]
```
ncharles: ok, this part may be useful, as there's some memory constraint right before Promises prepared in xxx ms

ncharles:
Rejecting, as there are no leaks. Some part will be cherry picked later

PR #2464, opened 2019-09-16
Fixes #14209: Impossible to reset the status of a Node using rudder-keys cli

fanf:
https://issues.rudder.io/issues/14209

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/services/provisioning/CheckInventoryDigest.scala
+++ b/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/services/provisioning/CheckInventoryDigest.scala
@@ -199,33 +199,20 @@ class InventoryDigestServiceV1(
 
     repo.get(receivedInventory.node.main.id) match {
       case Full(storedInventory) =>
-        val status = storedInventory.node.main.keyStatus
-        val inventory  : NodeInventory = status match {
-          case UndefinedKey =>
-            storedInventory.node.agents.map(_.securityToken).headOption match {
-              case None =>
-                // There is no key and status is undefined, use received key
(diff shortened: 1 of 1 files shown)


webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/services/provisioning/CheckInventoryDigest.scala
```diff
@@ -199,33 +199,20 @@ class InventoryDigestServiceV1(
 
     repo.get(receivedInventory.node.main.id) match {
       case Full(storedInventory) =>
-        val status = storedInventory.node.main.keyStatus
-        val inventory  : NodeInventory = status match {
-          case UndefinedKey =>
-            storedInventory.node.agents.map(_.securityToken).headOption match {
-              case None =>
-                // There is no key and status is undefined, use received key
+        val keyStatus = storedInventory.node.main.keyStatus
```
VinceMacBuche: keyStatus is the status of the stored inventory, even if we use the received inventory ? what if the removed inventory was certified and the new received inventory is not Certified ?

fanf: Yes, we want to ensure that a certified inventory is not becoming "uncertified" just because a new inventory without a signature arrive. 
When we delete a node, we reset status key to "undefined".

So for a new inventory for a node which was deleted, we will have "undefined" here, and then, depending of the presence of the signature, we will switch the status to "certified" or let it to "undefined".

VinceMacBuche: :+1:  thanks !

fanf:
OK, merging this PR

PR #2468, opened 2019-09-17
Fixes #15741: Update scala to 2.12.10

fanf:
https://issues.rudder.io/issues/15741

### DIFF
--- a/webapp/sources/rudder-parent-pom/pom.xml
+++ b/webapp/sources/rudder-parent-pom/pom.xml
@@ -361,8 +361,10 @@ limitations under the License.
     <rudder-version>5.0.14-SNAPSHOT</rudder-version>
     <spring-run-dep-version>5.0.14-SNAPSHOT</spring-run-dep-version>
 
-    <scala-version>2.12.7</scala-version>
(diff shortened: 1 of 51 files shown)


webapp/sources/rudder-parent-pom/pom.xml
```diff
@@ -433,6 +435,16 @@ limitations under the License.
         <artifactId>scala-reflect</artifactId>
         <version>${scala-version}</version>
       </dependency>
+      <dependency>
+        <groupId>org.scala-lang.modules</groupId>
+        <artifactId>scala-parser-combinators_${scala-binary-version}</artifactId>
```
ncharles: why do we need it now ?

fanf: I'm *really* not sure, and actually even if I had a good explanation for that one, I don't understand why it needs to be declared in `rudder-web` since it is a compile dep in `rudder-core`. But it seems that they changed the scope of that module in `scala-compiler` from compile to provided, and so we don't get it through `liftweb-json` (I think).

fanf:
OK, merging this PR

PR #2472, opened 2019-09-19
Fixes #15749: A lot of memory (and ressources) seems to be used in PreUnmarshallCheckConsistency

ncharles:
https://issues.rudder.io/issues/15749


fanf:
OK, merging this PR

PR #2473, opened 2019-09-20
Fixes #15767: cf-promise check (development policy check) must be *uncheked* by default

fanf:
https://issues.rudder.io/issues/15767


fanf:
PR updated with a new commit

ncharles:
This should stay activated as dfefault (see ticket)

PR #2474, opened 2019-09-20
Fixes #15768: Make inventory processing less memory consuming

ncharles:
https://issues.rudder.io/issues/15768


fanf:
OK, merging this PR

PR #2477, opened 2019-09-23
Fixes #15609: 'Nodes Configurations' ldap entry is too big

fanf:
https://issues.rudder.io/issues/15609


ncharles:
With this change, the step  Node configuration cached in LDAP went from 22s to 9s for 4500 nodes

fanf:
PR updated with a new commit

fanf:
So, the time is lineare on the size of that entry. We should try to find a better way to store it. UUID are huge :/

fanf:
OK, squash merging this PR

PR #2482, opened 2019-09-24
Fixes #15792: When writing policies, we replace the RudderUniqueId with replaceAll which is twice as slow as StringUtils.replace

ncharles:
https://issues.rudder.io/issues/15792


fanf:
OK, merging this PR

ncharles:
OK, merging this PR

fanf:
OK, merging this PR

PR #2489, opened 2019-09-30
Fixes #15802: Clean up report with level log info more than once a day

fanf:
https://issues.rudder.io/issues/15802

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
@@ -37,6 +37,8 @@
 
 package com.normation.rudder.batch
 
+import com.normation.ldap.sdk.LDAPConnectionProvider
+import com.normation.ldap.sdk.RoLDAPConnection
 import net.liftweb.actor.LAPinger
 import com.normation.rudder.services.system.DatabaseManager
 import net.liftweb.common._
@@ -46,6 +48,7 @@ import com.normation.rudder.domain.reports._
 import com.normation.rudder.domain.logger.ScheduledJobLogger
 import net.liftweb.actor.SpecializedLiftActor
 import com.normation.rudder.services.system.DeleteCommand
+import com.normation.utils.Control
 
 /**
  *  An helper object designed to help building automatic reports cleaning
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
```diff
@@ -64,8 +67,8 @@ object AutomaticReportsCleaning {
   val defaultHour   = 0
   val defaultDay    = "sunday"
 
-  val defaultArchiveTTL = 30
-  val defaultDeleteTTL  = 90
+  val defaultArchiveTTL = 4
```
ncharles: good catch

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
```diff
@@ -304,6 +335,50 @@ class AutomaticReportsCleaning(
...
+        case Full(x) =>
+          logger.debug(s"Found maximun run interval: ${x} minutes")
+          x
+        case eb: EmptyBox =>
+          logger.error((eb ?~! s"Error when trying to get maximun run interval, defaulting to  5 minutes. Error was: ").messageChain)
+          5
+      }
+      interval * r
+    } else {
+      toInt(deleteLogttlString, deleteLogttlString)
```
ncharles: in this case, it default to 2 minutes if number is malformed

fanf: right :/

fanf:
> Also, if we change the agent run frequency, we need to restart the web interface to change the duration of log deleted.

Yes, this is documented like that in the property file, but I don't see any non-very-hard way to do it otherwise, and the consequences are *at worse* that info log are deleted before the next run (and it's solved by a restart), so it seems OK.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -152,6 +152,14 @@ rudder.jdbc.maxPoolSize=25
 rudder.batch.reportscleaner.archive.TTL=4
 rudder.batch.reportscleaner.delete.TTL=15
 
+# Report with level "log" (log_info, log_warn, etc) are not kept more than a couple of runs.
+# They are used for debugging purpose, and so have a limited interest span (but can consume a lot
+# of DB space).
+# The TTL is given as a number of run to keep (the max run period of all nodes will be used),
+# with the format: Nx (with N a positive integer); or by a number of minutes: N (with N a positive integer).
+# The value is only update date when rudder starts.
```
ncharles: The value is only updated at rudder starts

fanf:
PR updated with a new commit

VinceMacBuche:
OK, squash merging this PR

PR #2491, opened 2019-09-30
Fixes #15809: Cannot update editor techniques at webapp startup if the request to get techniques timeout (5 seconds)

VinceMacBuche:
https://issues.rudder.io/issues/15809

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -193,15 +193,19 @@ class TechniqueWriter (
   }
 
   // Write and commit all techniques files
-  def writeAll(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, committer : EventActor) : Result[Seq[String]] = {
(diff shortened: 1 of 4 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -193,15 +193,19 @@ class TechniqueWriter (
   }
 
   // Write and commit all techniques files
-  def writeAll(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, committer : EventActor) : Result[Seq[String]] = {
+  def writeAll(technique : Technique, methods: Map[BundleName, GenericMethod], modId : ModificationId, committer : EventActor, launchLibUpdate : Boolean) : Result[Seq[String]] = {
```
fanf: boolean option are hard to understand (we should never have one I believe). I think we should have two methods here: `writeAll` and `writeAllAndUpdateTechLib` and just call the correct one in either place.

VinceMacBuche:
Commit modified

fanf:
OK, merging this PR

PR #2492, opened 2019-10-01
Fixes #15810: Incorrect permission on /var/rudder/reports

amousset:
https://issues.rudder.io/issues/15810

### DIFF
--- a/relay/sources/rudder-server-relay-postinst
+++ b/relay/sources/rudder-server-relay-postinst
@@ -54,10 +54,10 @@ chown rudder: /var/rudder/shared-files
 chmod 770 /var/rudder/shared-files
 chmod 755 /var/rudder/inventories
 
-for dir in /var/rudder/inventories/incoming /var/rudder/inventories/accepted-nodes-updates /var/rudder/reports/incoming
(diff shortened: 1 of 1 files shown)


relay/sources/rudder-server-relay-postinst
```diff
@@ -54,10 +54,10 @@ chown rudder: /var/rudder/shared-files
 chmod 770 /var/rudder/shared-files
 chmod 755 /var/rudder/inventories
 
-for dir in /var/rudder/inventories/incoming /var/rudder/inventories/accepted-nodes-updates /var/rudder/reports/incoming
+for dir in /var/rudder/inventories/incoming /var/rudder/inventories/accepted-nodes-updates /var/rudder/reports/incoming /var/rudder/reports/failed
 do
   chmod 770 ${dir}
-  chown rudder:${APACHE_GROUP} ${dir}
+  chown ${APACHE_USER}:rudder ${dir}
```
peckpeck: why this change ?

amousset: We need these folder to be writeable by rudder-relayd, which runs as `rudder-relayd:rudder`. Using the apache user and the Rudder group should make this work for both.

PR #2504, opened 2019-10-05
Fixes #15858: Computing dynamic groups is very memory intensive, and can lead to FGC or OOM

ncharles:
https://issues.rudder.io/issues/15858

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -217,6 +217,19 @@ class AcceptedNodesLDAPQueryProcessor(
     } }
   }
 
+  override def processOnlyId(query:Query) : Box[Seq[NodeId]] = {
+    //only keep the one of the form Full(...)
+    queryAndChekNodeId(query, NodeInfoService.nodeInfoAttributes, None).map { seq => seq.flatMap {
+      case QueryResult(nodeEntry, _ , _) =>
+        processor.ldapMapper.entryToNode(nodeEntry) match {
+          case Full(node) => Seq(node.id)
+          case e:EmptyBox =>
+            logger.error((e ?~! "Ignoring entry in result set").messageChain)
+            Seq()
+        }
+    } }
+  }
+
 }
 
 /**
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -219,13 +219,13 @@ class AcceptedNodesLDAPQueryProcessor(
 
   override def processOnlyId(query:Query) : Box[Seq[NodeId]] = {
     //only keep the one of the form Full(...)
-    queryAndChekNodeId(query, NodeInfoService.nodeInfoAttributes, None).map { seq => seq.flatMap {
+    queryAndChekNodeId(query, Seq(A_NODE_UUID), None).map { seq => seq.flatMap {
       case QueryResult(nodeEntry, _ , _) =>
-        processor.ldapMapper.entryToNode(nodeEntry) match {
-          case Full(node) => Seq(node.id)
-          case e:EmptyBox =>
-            logger.error((e ?~! "Ignoring entry in result set").messageChain)
-            Seq()
+        nodeDit.NODES.NODE.idFromDn(nodeEntry.dn) match {
+          case Some(nodeId) => Some(nodeId)
+          case None =>
+            logger.error(s"Error when processing query ${query.toJSONString}: fetched node entry ${nodeEntry.toString()} is not a correct nodeId")
+            None
         }
     } }
   }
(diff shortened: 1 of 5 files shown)


webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/AgentTypes.scala
```diff
@@ -176,44 +180,36 @@ object AgentInfoSerialisation {
       )
   }
 
-  def parseSecurityToken(agentType : AgentType, tokenJson: JValue, tokenDefault : Option[String]) : Box[SecurityToken]= {
-    import net.liftweb.json.compactRender
+  def parseSecurityToken(agentType : AgentType, tokenJson: io.circe.JsonObject, tokenDefault : Option[String]) : Box[SecurityToken]= {
```
ncharles: i'm pretty sure we could do a decoder here, but with the corner cases, i'm not sure it's really possible

webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/AgentTypes.scala
```diff
@@ -227,20 +223,39 @@ object AgentInfoSerialisation {
...
-     token <- json \ "securityToken" match {
-                case JObject(json) => parseSecurityToken(agentType, json, optToken)
-                case _             => parseSecurityToken(agentType, JNothing, optToken)
-              }
+      json: Json <- io.circe.parser.parse(s) match {
+                   case Left(failure) => Failure(s"Can not parse agent info: ${failure.getMessage()}")
+                   case Right(json)   => Full(json)
+                }
+      // Ensure json is valid json
+      validJson <- json.asObject.map(Full(_)).getOrElse(Failure(s"Error when trying to parse string as JSON Agent: json is '$s')"))
```
ncharles: most of this code will probably make sad any real user of circe :/

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
First commit is there only for posterity, as it doesn't improve notably (less than 10% with incomplete measurement), and is really complex and fragile (because of my implementation)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -217,6 +217,19 @@ class AcceptedNodesLDAPQueryProcessor(
     } }
   }
 
+  override def processOnlyId(query:Query) : Box[Seq[NodeId]] = {
+    //only keep the one of the form Full(...)
+    queryAndChekNodeId(query, NodeInfoService.nodeInfoAttributes, None).map { seq => seq.flatMap {
+      case QueryResult(nodeEntry, _ , _) =>
```
fanf: I'm not sure you need to get all these attributes here, they don't seem to be use. 
I believe you can juste request `NODE_UUID`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -217,6 +217,19 @@ class AcceptedNodesLDAPQueryProcessor(
...
+        processor.ldapMapper.entryToNode(nodeEntry) match {
+          case Full(node) => Seq(node.id)
```
fanf: Here, I'm not sure why you need to mapping ? You only want the `NodeId`, and the mapping is not really the goad of that method. Bad nodes would appear elsewhere (generation for ex, or node details, etc). 
I believe you can earn quite a lot more memory here by just usgin `nodeDit.NODES.NODE.idFromDn(nodeEntry.dn)`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/QueryProcessor.scala
```diff
@@ -53,6 +53,12 @@ trait QueryProcessor {
    */
   def process(query:Query) : Box[Seq[NodeInfo]]
 
+  /**
+   * Only get the nodeId, to update dynamic groups
+   * @param query
+   * @return
```
fanf: Can you remove the two param/return comment, they are useless.
Also for the comment, perhaps something like: 
```
Only get node ids corresponding to that request, with minimal consistency check. This method is usefull to maximize performance (low memory, high throughout) for ex for dynamic groups. 
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/QueryProcessor.scala
```diff
@@ -53,6 +53,12 @@ trait QueryProcessor {
...
+   */
+  def processOnlyId(query:Query) : Box[Seq[NodeId]]
```
fanf: Perhaps the method should return a `Set[NodeId]` ?

fanf:
This seems to be a very good PR. Can you remove the json/circe commit, since it means a new dep and ++ complexity for almost no gain compared to the other parts of the PR ?

ncharles:
Commit modified

fanf:
OK, squash merging this PR

PR #2510, opened 2019-10-08
Fixes #15890: Batch which deletes log reports displays an error when default run interval is not defined

fanf:
https://issues.rudder.io/issues/15890

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
@@ -141,8 +141,11 @@ object AutomaticReportsCleaning {
     val runIntervalAttr = "serializedAgentRunInterval"
     for {
       ldap    <- ldapCon
-      e       <- ldap.get("propertyName=agent_run_interval,ou=Application Properties,cn=rudder-configuration", "propertyValue")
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AutomaticReportsCleaner.scala
```diff
@@ -364,7 +367,7 @@ class AutomaticReportsCleaning(
           logger.debug(s"Found maximun run interval: ${x} minutes")
           x
         case eb: EmptyBox =>
-          logger.error((eb ?~! s"Error when trying to get maximun run interval, defaulting to  5 minutes. Error was: ").messageChain)
+          logger.error((eb ?~! s"Error when trying to get maximun run interval, defaulting to 5 minutes. Error was: ").messageChain)
```
ncharles: shouldn't it be
Error was: ${messageChain}
?

fanf: no, `·eb` is perhaps an `Empty` in the type system even if I know it's a failure, so I have to first change it to failure `eb ?~! "..."` and then call `.messageChain` on it. But as I'm sure it's a failure given implementation, I now that it will be displayed as "... error was: <- more info here"

fanf:
OK, merging this PR

PR #2518, opened 2019-10-09
Fixes #15878: We still have an LDAP entry "ou=Nodes Configuration,..." which is not used anymore

fanf:
https://issues.rudder.io/issues/15878


fanf:
Rhaaaa zut, I deleted the branch in place of updating it :/
We want to keep `cn=Nodes Configuration`, which is the entry with all hashes for cache, but delete `ou=Nodes Configuration` (which was a branch with one entry by node and horrible perfomance). 
I'm opening and other PR, sorry :/

PR #2528, opened 2019-10-14
Fixes #15948: Event log page must only load one day back of logs by default

VinceMacBuche:
https://issues.rudder.io/issues/15948


VinceMacBuche:
Commit modified

PR #2536, opened 2019-10-17
Fixes #15990: Remote-run returns empty output

amousset:
https://issues.rudder.io/issues/15990


amousset:
PR updated with a new commit

PR #2540, opened 2019-10-21
Fixes #16004: Unexpected crash of webapp on 6.0

ncharles:
https://issues.rudder.io/issues/16004


ncharles:
i don't know why this part was synchronized
the underlying code is correctly handled, so we are adding an extra level of synchronize

PR #2564, opened 2019-10-29
Fixes #16076: relayd should stop when a thread panics

amousset:
https://issues.rudder.io/issues/16076


amousset:
OK, merging this PR

PR #2570, opened 2019-11-05
Fixes #16104: On centos7 fresh install, generation fails because relayd is not started

amousset:
https://issues.rudder.io/issues/16104


webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/40-nodelist-relay-configuration
```diff
@@ -12,4 +12,4 @@
 # Copy nodelist.json to its final destination and reload relayd
 
 cp /var/rudder/cfengine-community/inputs/distributePolicy/1.0/nodeslist.json /var/rudder/lib/relay/ 
-/opt/rudder/bin/rudder relay reload -p
+systemctl reload-or-restart rudder-relayd
```
ncharles: the -p was made necessary for the file permission, there we break the system. we need to fix that

amousset: indeed.

PR #2572, opened 2019-11-05
Fixes #16122: Stop listenning for inventories on port 80

amousset:
https://issues.rudder.io/issues/16122

### DIFF
--- a/relay/sources/apache/rudder-apache-relay-common.conf
+++ b/relay/sources/apache/rudder-apache-relay-common.conf
@@ -1,12 +1,5 @@
 DocumentRoot /var/www
 
-<FilesMatch "\.(cgi|shtml|phtml|php)$">
-  SSLOptions +StdEnvVars
-</FilesMatch>
-<Directory /usr/lib/cgi-bin>
-  SSLOptions +StdEnvVars
-</Directory>
-
 # We do not require mod_deflate as it is not strictly necessary
 <IfModule mod_deflate.c>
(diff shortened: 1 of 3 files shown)


relay/sources/apache/rudder-apache-relay-common.conf
```diff
@@ -1,12 +1,5 @@
 DocumentRoot /var/www
 
-<FilesMatch "\.(cgi|shtml|phtml|php)$">
```
amousset: I see no reason behind this.

peckpeck: indeed

relay/sources/apache/rudder-apache-relay-common.conf
```diff
@@ -17,79 +10,3 @@ DocumentRoot /var/www
 </IfModule>
 </IfModule>
 
-# Expose the server UUID through http
-Alias /uuid /opt/rudder/etc/uuid.hive
-<Directory /opt/rudder/etc>
```
peckpeck: isn't this used by the inventory code ?

amousset: It seems to always use https in 5.0.

amousset: I also checked usage of relay API, we are using https or ${inventory_protocol} (which is https now) everywhere.

amousset:
OK, merging this PR

PR #2583, opened 2019-11-12
Fixes #10067: If LDAP attribute is missing, related LDAP error are not reported in UI (silent fail)

fanf:
https://issues.rudder.io/issues/10067


fanf:
OK, merging this PR

PR #2584, opened 2019-11-12
Fixes #16158: Error message when serching for RAM size says we can use Mo

ncharles:
https://issues.rudder.io/issues/16158


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #2594, opened 2019-11-13
Fixes #11917: No compliance from a Rule with only one Directive, based on a Technique made with Technique Editor, when the Directive is also applied in another Rule

fanf:
https://issues.rudder.io/issues/11917


fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #2597, opened 2019-11-14
Fixes #14790: Deleted directive are not always remove from rule and not actionnable

fanf:
https://issues.rudder.io/issues/14790


fanf:
PR updated with a new commit

PR #2599, opened 2019-11-14
Fixes #15124: Rule tag is "Enforce" whereas it has both enforce and audit nodes

fanf:
https://issues.rudder.io/issues/15124

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
@@ -566,10 +566,11 @@ class RuleGrid(
     val t0 = System.currentTimeMillis
 
     val nodes = groupsLib.getNodeIds(line.rule.targets, nodesInfo)
+    val nodeModes = nodes.flatMap(id => nodesInfo.get(id).map(_.policyMode))
     // when building policy mode explanation we look into all directives every directive applied by the rule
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -566,10 +566,11 @@ class RuleGrid(
     val t0 = System.currentTimeMillis
 
     val nodes = groupsLib.getNodeIds(line.rule.targets, nodesInfo)
+    val nodeMods = nodesInfo.values.map(node => if(nodes.contains(node.id)) node.policyMode else None).toSet
```
ncharles: i think this is the opposite of what should be done: nodesInfos is a map, so getting the NodeInfo from the nodeid is basically O(1), getting all nodeinfo is linear in the size of nodes target
here, you take all the nodeinfos, to take their nodeis, and traverse once per node to get the policy mode, so complexity is nodes target * total node /2

ncharles: nodesInfo.collect { case (nodeId, nodeinfo) if nodes.contains(nodeId) =>  nodeinfo.policyMode }

ncharles: or something like that

fanf:
PR updated with a new commit

PR #2602, opened 2019-11-14
Fixes #14332: Document hook that trigger an agent run on update

fanf:
https://issues.rudder.io/issues/14332


webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/60-trigger-node-update
```diff
@@ -18,6 +18,10 @@
 # This feature can generate a more run than expected and some load on the network. In big
 # Rudder installation (thousands of nodes), we advice to measure these metrics and disable
 # the hook if necessary.
+# Other specifities:
+# - obviously, node agent will run more frequently than its configured period since a run
+#   will be triggered at each generation;
+# - splay time is not honored (all agent run are triggired at the same time).
```
ncharles:
```suggestion
# - splay time is not honored (all agent runs are triggered simultaneously).
```

PR #2619, opened 2019-11-19
Fixes #16211: Keep "Mo" in group search for RAM

fanf:
https://issues.rudder.io/issues/16211

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/MemorySize.scala
+++ b/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/MemorySize.scala
@@ -76,7 +76,7 @@ object MemorySize {
    *
    */
   private val num_r = """(\d[\d_ ]*)""".r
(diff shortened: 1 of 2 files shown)


webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/MemorySize.scala
```diff
@@ -76,7 +76,7 @@ object MemorySize {
    *
    */
   private val num_r = """(\d[\d_ ]*)""".r
-  private val numunit_r = """(\d[\d_ ]*) *([\p{Alpha}]?[B]?)""".r
+  private val numunit_r = """(\d[\d_ ]*) *([\p{Alpha}]{0,2})""".r
```
ncharles: we used to have ?[BOo], i think we ought to stick to that

fanf: It was a bug. It does not work with just "k", and now it's advertised in the error message.

PR #2626, opened 2019-11-22
Fixes #16223: Regroup tabs from "Hardware" to "Virtual Machines" into one

ElaadF:
https://issues.rudder.io/issues/16223

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
@@ -137,7 +137,7 @@ object DisplayNode extends Loggable {
     val jsId = JsNodeId(nodeId,salt)
     val detailsId = htmlId(jsId,"details_")
     val softGridDataId = htmlId(jsId,"soft_grid_data_")
-    val softPanelId = htmlId(jsId,"sd_soft_")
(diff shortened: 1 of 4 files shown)


ElaadF:
do not merge yet

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -137,7 +137,7 @@ object DisplayNode extends Loggable {
     val jsId = JsNodeId(nodeId,salt)
     val detailsId = htmlId(jsId,"details_")
     val softGridDataId = htmlId(jsId,"soft_grid_data_")
-    val softPanelId = htmlId(jsId,"sd_soft_")
+    val softPanelId = "node_inventory"
```
fanf: if you change that, you will break software async load on line 215:
```
      // for the software tab, we check for the panel id, and the firstChild id
      // if the firstChild.id == softGridId, then it hasn't been loaded, otherwise it is softGridId_wrapper
      JsRaw(s"""
        $$("#${detailsId}").on( "tabsactivate", function(event, ui) {
          if(ui.newPanel.attr('id')== '${softPanelId}' ){
            ${  SHtml.ajaxCall(JsRaw("'"+nodeId.value+"'"), loadSoftware(jsId, softIds) )._2.toJsCmd}
          }
        });
        """)
```

PR #2631, opened 2019-11-25
Fixes #16269: Could not get Group tree details cause is: Could not find Group tree.

ncharles:
https://issues.rudder.io/issues/16269


ncharles:
Commit modified

PR #2637, opened 2019-11-27
Fixes #16283: we are always computing compliance and byNodes on RuleStatusReport, even when it's not necessay

ncharles:
https://issues.rudder.io/issues/16283


fanf:
this one will not go in 5.0.15

ncharles:
Commit modified

ncharles:
once 6.0.4 is released, we should be able to merge that

ncharles:
rejected in favor for https://github.com/Normation/rudder/pull/2754/files

PR #2643, opened 2019-11-30
Fixes #16291: Deadlock when application starts

fanf:
https://issues.rudder.io/issues/16291
Several main parts in that PR - the main one is reverting to ZIO-1.0.0-RC16, because we seems to hit a bug in RC17 (discussed on zio chan, ticket here: https://github.com/zio/zio/issues/2373), but it also does other things that removed other (possible) deadlocks:

- rewrite GitTechniqueReader for it to be mostly immutable and pure. It makes it be created/exec a lot faster (~10x). It does not seems to be directly related with the bug but was a revelator. The new version is relatively safe to use, since we have tests on that. 

- change doobie transactor to directly work with `ZIO` `Task` in place of `cats` `IO`. Again not directly cause of the bug but a revelator, it also allowed us to see that we still had some `cats.unsafeRun`, and that we were using a yet another threadpool for that (which seem unecessary). That does change the postgres connection pool, only the threadpool on the receiving end of the connection. 

- removed last `synchronized` blocked. We should not have any synchronized block, and again I thought they were the cause (but still just a revelator). It's the part where I'm the less sure, because `semaphore` are not reintrant where `synchronize` block are, so if at some point we have a twisted call graphe that need reintrance, it may lead to a (real) deadlock. I didn't got it with my tests, but it means almost nothing. 
UPDATE: confirmed that AT LEAST for `CachedFindRuleNodeStatusReports` in `ReportingServiceImpl`, we have reentrant parts and we need to keep `synchronized`.

### DIFF
--- a/webapp/sources/pom.xml
+++ b/webapp/sources/pom.xml
@@ -387,8 +387,8 @@ limitations under the License.
     <http4s-version>0.19.0</http4s-version>
     <shapeless-version>2.3.3</shapeless-version>
     <cats-effect-version>2.0.0</cats-effect-version>
-    <dev-zio-version>1.0.0-RC17</dev-zio-version>
-    <zio-cats-version>2.0.0.0-RC8</zio-cats-version>
+    <dev-zio-version>1.0.0-RC16</dev-zio-version>
+    <zio-cats-version>2.0.0.0-RC7</zio-cats-version>
 
     <!--
       Hack to make scalac jvm parameters like RAM configurable.
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
@@ -69,6 +69,8 @@ import zio.syntax._
 import GitTechniqueReader._
 import com.normation.rudder.domain.logger.TechniqueReaderLoggerPure
 import org.eclipse.jgit.lib.Repository
+import com.github.ghik.silencer.silent
+
 
 /**
  *
@@ -199,7 +201,7 @@ class GitTechniqueReader(
                             )
                         }
     } yield res
-  ).flatMap(Ref.make(_)).runNow
+  ).flatMap(Ref.make(_)).runNow: @silent // inferred Any - yes, that's what we want
 
   private[this] val nextTechniquesInfoCache: Ref[(ObjectId,TechniquesInfo)] = {
     (for {
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
@@ -63,7 +63,7 @@ import org.eclipse.jgit.diff.DiffEntry.ChangeType
 import java.io.IOException
 
 import com.normation.errors._
-import com.normation.zio.ZioRuntime
+import com.normation.zio._
 import zio._
 import zio.syntax._
 import GitTechniqueReader._
@@ -136,19 +136,10 @@ class GitTechniqueReader(
   val relativePathToGitRepos : Option[String],
   val directiveDefaultName   : String //full (with extension) name of the file containing default name for directive (default-directive-names.conf)
 ) extends TechniqueReader with Loggable {
-  reader =>
 
-  /*
-   * That class is not yet ported to ZIO.
-   * So we accept that any error in a IOResult[A] will lead to an exception, which will
-   * need to be taken care for.
-   */
-  implicit class RunOrThrow[A](effect: IOResult[A]) {
-    def runNow: A = ZioRuntime.runNow(effect.either) match {
-      case Right(x)  => x
-      case Left(err) => throw new RuntimeException(err.fullMsg)
-    }
-  }
+  // semaphore to have consistent read
+  val semaphore = Semaphore.make(1).runNow
+
 
   //the path of the PT lib relative to the git repos
   //withtout leading and trailing /.
@@ -181,7 +172,7 @@ class GitTechniqueReader(
   //can not set private because of "outer ref cannot be checked" scalac bug
   private case class NoRootCategory(msg: String) extends Exception(msg)
 
-  private[this] var currentTechniquesInfoCache : TechniquesInfo = {
+  private[this] var currentTechniquesInfoCache : TechniquesInfo = semaphore.withPermit(IOResult.effect {
     val currentRevTree = revisionProvider.currentRevTreeId.runNow
     try {
       processRevTreeId(currentRevTree)
@@ -203,7 +194,7 @@ class GitTechniqueReader(
         revisionProvider.setCurrentRevTreeId(newRevTreeId).runNow
         processRevTreeId(newRevTreeId)
     }
-  }
+  }).runNow
 
   private[this] var nextTechniquesInfoCache : (ObjectId,TechniquesInfo) = (revisionProvider.currentRevTreeId.runNow, currentTechniquesInfoCache)
   //a non empty list IS the indicator of differences between current and next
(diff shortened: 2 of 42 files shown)


webapp/sources/pom.xml
```diff
@@ -387,8 +387,8 @@ limitations under the License.
     <http4s-version>0.19.0</http4s-version>
     <shapeless-version>2.3.3</shapeless-version>
     <cats-effect-version>2.0.0</cats-effect-version>
-    <dev-zio-version>1.0.0-RC17</dev-zio-version>
-    <zio-cats-version>2.0.0.0-RC8</zio-cats-version>
+    <dev-zio-version>1.0.0-RC16</dev-zio-version>
+    <zio-cats-version>2.0.0.0-RC7</zio-cats-version>
```
ncharles: is it a bug in the PR ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -181,51 +173,55 @@ class GitTechniqueReader(
...
+                              revisionProvider.setCurrentRevTreeId(newRevTreeId) *>
+                              processRevTreeId(db, newRevTreeId)
+                            )
+                        }
+    } yield res
+  ).flatMap(Ref.make(_)).runNow
+
+  private[this] val nextTechniquesInfoCache: Ref[(ObjectId,TechniquesInfo)] = {
+    (for {
+      a <- revisionProvider.currentRevTreeId
```
ncharles: what happens at install, when currentRevTreeId doesn't exist yet ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -507,10 +507,9 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
   override final def getDeletedNodeInfo(nodeId: NodeId): Box[Option[NodeInfo]] = getNotAcceptedNodeInfo(nodeId, RemovedInventory).toBox
 
   /**
-   * Clear cache. Try a reload asynchronously, disregarding
-   * the result
+   * Clear cache.
    */
-  override def clearCache(): Unit = this.synchronized {
```
ncharles: why no semaphore here ??

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -294,20 +295,18 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
    * Clear cache. Try a reload asynchronously, disregarding
    * the result
    */
-  override def clearCache(): Unit = this.synchronized {
+  override def clearCache(): Unit = {
```
ncharles: why not using the semaphore here ?

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchiverUtils.scala
```diff
@@ -112,7 +112,7 @@ trait GitArchiverUtils {
       case Right(dir) => dir
       case Left(err) =>
         val msg = s"Error when checking required directories '${file.getPath}' to archive in git: ${err.fullMsg}"
-        GitArchiveLoggerPure.logEffect.error(msg)
+        GitArchiveLogger.logEffect.error(msg)
```
ncharles: this change doesn't compile

webapp/sources/pom.xml
fanf: nope, it's the core of the resolution. The problem is likely a bug in zio last RC, see: https://github.com/zio/zio/issues/2373

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
fanf: That part didn't changed from before the PR. `currentRevTreeId` looks for the last available git commit id. If no git exists, it fails. I think we just init the git repos before starting rudder at install.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: It was the only `synchronized` in the whole class, and it does not have anything to synchronize. I believe it's a scare from past battles.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: I think it was useless to synch on that. The clear cache is just the first op (`cache = Map()`). 
The other part is just "reload the cache for the new time". But the part with the complex logic is `checkAndUpdateCache`, and that part is guarded by a semaphore. As far as I can see, the only thing that could happen is a secon call to `checkAndUpdateCache`, which is idempotent.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchiverUtils.scala
fanf: My sed capacity drop around 23:00 :/

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchiverUtils.scala
fanf: (the logEffect part should have been removed)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -181,51 +173,55 @@ class GitTechniqueReader(
...
+      b <- currentTechniquesInfoCache.get
+      r <- Ref.make((a, b))
+    } yield r).runNow
   }
 
-  private[this] var nextTechniquesInfoCache : (ObjectId,TechniquesInfo) = (revisionProvider.currentRevTreeId.runNow, currentTechniquesInfoCache)
   //a non empty list IS the indicator of differences between current and next
-  private[this] var modifiedTechniquesCache : Map[TechniqueName, TechniquesLibraryUpdateType] = Map()
+  private[this] val modifiedTechniquesCache: Ref[Map[TechniqueName, TechniquesLibraryUpdateType]] = Ref.make(Map[TechniqueName, TechniquesLibraryUpdateType]()).runNow
```
fanf: In place of `var` we now use `Ref`, which are atomically accessed memory zone (with `get`, `set` or `update`)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -280,21 +276,49 @@ class GitTechniqueReader(
...
     }
+
+    (for {
+      nextId   <- revisionProvider.getAvailableRevTreeId
+      cached   <- nextTechniquesInfoCache.get
+      db       <- repo.db
+      mods     <- if(nextId == cached._1) modifiedTechniquesCache.get
+                  else for {
+                    nextTechniquesInfo <- processRevTreeId(db, nextId)
+                    managedDiffFmt     =  ZManaged.make(IOResult.effect{
```
fanf: `ZManaged.make(open resources)(close resources)` allows to create a safe manager for a resources. You give it an `open` and `close` logic, and then you safely use the resource with `managedThing.use(resources => ...)`. It's typically the good option from `io streams`, and git related things like `DiffFormatter` and `TreeWalk` (below).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -488,80 +535,115 @@ class GitTechniqueReader(
...
-
-      //now, for each potential path, look if the cat or policy
-      //is valid
-      while(tw.next) {
-        val path = toTechniquePath(tw.getPathString) //we will need it to build the category id
-        processTechnique(repo.db.runNow.open(tw.getObjectId(0)).openStream, path.path, techniqueInfos, parseDescriptor, revTreeId).either.runNow match {
-          case Left(err)  => logger.error(s"Error with technique at path: '${path.path}', it will be ignored. Error: ${err.fullMsg}")
-          case Right(()) => // ok
-        }
+      tw
```
fanf: In place of `while` loop, we use a recursive function guarded by the while condition. We just need to be carefull to use `IOResult.effect` everywhere because most things throw exception, and to not use anything `tw` related before the first call to `next` (because in java-logic, it's that call which init things)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
ncharles: ok, make sense

ncharles:
this still doesn't compile

[ERROR] /root/dev/rudder/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala:202: warning: a type was inferred to be `Any`; this may indicate a programming error.
[ERROR]   ).flatMap(Ref.make(_)).runNow
[ERROR]     ^

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
The [ERROR] inferred any is not a real error. It's a warning with "error" text, but it does not stop the compilation. I `silent`ed it nonetheless.

ncharles:
i don't have any deadlock so far with this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
VinceMacBuche: git init is done by packaging

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryProcessor.scala
```diff
@@ -173,23 +171,6 @@ class InventoryProcessor(
       }
     }
 
-    /*
```
VinceMacBuche: So you changed you mind and now you  delete this function (not used anymore) ? o, to me!

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -975,9 +975,9 @@ object RudderConfig extends Loggable {
       , ".sign"
     )
     if(WATCHER_ENABLE) {
-      watcher.startWatcher()
+      //watcher.startWatcher()
```
ncharles: i'm pretty sure this is an error

webapp/sources/utils/src/test/scala/com/normation/ZioCommonsTest.scala
```diff
@@ -504,11 +504,13 @@ import _root_.zio.internal._
           val blockingExecutor: UIO[Executor] = Executor.fromExecutionContext(Int.MaxValue)(scala.concurrent.ExecutionContext.global).succeed
         }
       }
-      override val platform: Platform = PlatformLive.fromExecutionContext(scala.concurrent.ExecutionContext.global) //.withTracingConfig(_root_.zio.internal.tracing.TracingConfig.disabled)
-      override val environment = new Clock.Live with Console.Live with ZSystem.Live with Random.Live with Blocking.Live
+      override val Platform: Platform = PlatformLive.fromExecutionContext(scala.concurrent.ExecutionContext.global) //.withTracingConfig(_root_.zio.internal.tracing.TracingConfig.disabled)
+      override val Environment = new Clock.Live with Console.Live with ZSystem.Live with Random.Live with Blocking.Live
+      def platform = Platform
+      def environmet = Environment
```
ncharles: isn't there a typo ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryProcessor.scala
fanf: it was making a warning at each compile which drives me mad

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
fanf: it isn't :/

webapp/sources/utils/src/test/scala/com/normation/ZioCommonsTest.scala
fanf: it should not be there at all actually :)

fanf:
PR updated with a new commit

PR #2651, opened 2019-12-04
Bug 16308/fix warnings displayed by cargo clippy linter

gpoblon:
https://issues.rudder.io/issues/16308

### DIFF
--- a/rudder-lang/src/ast/enums.rs
+++ b/rudder-lang/src/ast/enums.rs
@@ -557,6 +557,7 @@ impl<'src> EnumList<'src> {
     }
 
     /// There can be only one mapping per item that defines an identical descendant
+    #[allow(dead_code)]
     pub fn mapping_check(&self) -> Vec<Error> {
         let mut errors = Vec::new();
         for (from, tos) in self.direct_mapping_path.iter() {
@@ -637,7 +638,7 @@ impl<'b, 'src> ContextIterator<'b, 'src> {
             };
         }
         ContextIterator {
-            enum_list: enum_list,
+            enum_list,
             var_list,
             iterators,
             current,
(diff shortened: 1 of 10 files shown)


rudder-lang/src/ast/resource.rs
```diff
@@ -33,7 +33,8 @@ use super::enums::*;
 use super::value::Value;
 use crate::error::*;
 use crate::parser::*;
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, hash_map::Entry};
+// use std::collections::hash_map::Entry;
```
amousset: no need to keep this line (as it is now part of the other `use` line)

rudder-lang/src/ast/enums.rs
```diff
@@ -556,7 +556,8 @@ impl<'src> EnumList<'src> {
         errors
     }
 
-    /// There can be only one mapping per item that defines an identical descendant
+    // There can be only one mapping per item that defines an identical descendant
```
amousset: It there a specific reason for this not being a doc comment? (with `///` instead of `//`)?

gpoblon: Not anymore, i'm putting it back. But the following method was unused and before silencing the warning with the `allow` attribute I commented the whole function out... resulting in a compilation error (doc comment right before end of an impl block)

amousset: :+1:

rudder-lang/src/ast/resource.rs
gpoblon: yes, my bad. Wasn't sure about the cleanest way to include it

PR #2658, opened 2019-12-05
Fixes #16334: More generic lib linding of the rudder-lang tester script

gpoblon:
Should work on Rudder servers as I only installed a server/agent on a clean debian distro. The script works on it.
Did not moove paths to external config files as I am unsure it would have been cleaner

### DIFF
--- a/rudder-lang/README.adoc
+++ b/rudder-lang/README.adoc
@@ -10,7 +10,7 @@ This language is not:
 
 It has no:
 
-- recussion
+- recursion
 - generator / generic iterator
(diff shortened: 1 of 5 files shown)


gpoblon:
besoin d'installer ce module perl pour faire fonctionner generate-lib: 
`sudo cpan Config::IniFiles`

amousset:
> The following perl module needs to be installed in order to make generate-lib work:
> `cpan Config::IniFiles`

Could you add this information to the README?

PR #2664, opened 2019-12-09
Fixes #16374: Race condition in technique variable filling

fanf:
https://issues.rudder.io/issues/16374


webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
```diff
@@ -148,34 +148,33 @@ class FillTemplatesService {
...
-        cache = Map()
-      }
+      cache.set(Map())
     }
   }
 
   def getTemplateFromContent(templateName: String,content: String): IOResult[SynchronizedFileTemplate] = {
-    for {
-      c <- semaphore.withPermit(cache.get(content) match {
+    semaphore.withPermit(for {
```
ncharles: here, you are locking with the same semaphore than fill ; you should have a distrinct semaphore (if possible)

fanf: they have the same name but there is two `ZioRuntime.unsafeRun(Semaphore.make(1))` so there are not the same (actually, there is one semaphore for each `SynchronizedFileTemplate`, so for each fill)

PR #2666, opened 2019-12-09
Fixes #16382: Improve performance of policy generation writer

fanf:
https://issues.rudder.io/issues/16382

This PR try to make the abysmal performances of policy generation in 6.0 less abysmal. 

Most of that PR was split in smaller amount, but there is a big part remaining: I change (most) of the the write logic, and that part can't really be done in smaller bit. 

The big changes are:
- port a bit more GitTemplateReader & friends to ZIO to avoid switching from in/out ZIO;
- change the way policies are done so that now, it' not by node but by template, which avoid to lock on them 
- add A LOT of timing information, which are now pretty cool
- use atomic move when possible, but it doesn't yield a lot of perf, 
- parallelise move of policies to their final place, 
- use ArraySeq for string template values, which avoid to switch between list and array which cost a lot in ZIO. 

It can be merded. 

--- for records, what was done before --

Here, we explore three paths to earn some performances:

- 1/ try to use OS atomic move operation in place of copy+delete for moving policies from `rules.new` to `rules`
- 2/ adapt ZIO blocking thread executor, giving it a bit more chance to reuse existing threads
- 3/ refactoring from `Box` to `PureResult` and `IOResult`. There is something extremely costly with `toIO` and `toBox` (and even more for layers of them). 
- 4/ add a parrallel call in `prepareTechniqueTemplate`

Tests and measure were done on `WriteSystemTechniquesTest`, so the test case is rather specific and small (even for the 500 nodes cases). So the experiment should be replicated in real load testing environment to draw any conclusions. 

1/ That change does not seems to change anything, but my test case didn't exibited the big problems seen elsewhere. It could be an easy win. 

2/ Changing the threadpool ergonomics leads to ~7% better perf, but it may be due to the very specific patterns in the unit test. 

3/ This change is the biggest in number of lines. The most dramatic improvement is due to the changes in `STVariable` for validation, which drives on itself a ~7% improvement. Validation is called a lot of time. Other changes leads to 5-7% improvment. 

4/ This last change lead to ~30% improvment (but uniquely on `prepareTechniqueTemplate` phase)

All in all, we get a consistant 20-22% improvment compared to 6.0.0. 

--- oups, I did a rebase, I wanted to just do an added commit :/ ----

Some more change, and we are now near what we had in 5.0 (still 25% worse for writting techniques) but on the other hand, we are 50 times faster on `move promise to final position`. 

New changes: 

- use `Promise` for parsing template (to put it in cache) which allows to block less in semaphore,
- change the way we fill template and use `IOResult.effectNonBlocking` to minimize time in semaphore. This (especially avoiding thread creation with `effectNonBlocking`) leads to a 4x improvement on the most costly step. 
- remove parallelisation below the first level (ie all `traversePar`). It leads to more stable result overall. 

It's also intersting to see that almost nothing change when we add nodes to the test (until we reach memory/gc limit, and then things go to a stop).

Results in images: 

2019-12-17_21 23 33-Untitled_spreadsheet_-_Google_Sheets

2019-12-17_20 25 58-Untitled_spreadsheet_-_Google_Sheets

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
@@ -53,21 +53,12 @@ import com.normation.rudder.hooks.HookEnvPairs
 import com.normation.rudder.hooks.RunHooks
 import com.normation.templates.FillTemplatesService
 import com.normation.templates.STVariable
-import com.normation.utils.Control._
-import java.io.File
-import java.io.IOException
+import java.nio.charset.Charset
 
 import net.liftweb.common._
 import net.liftweb.json.JsonAST
 import net.liftweb.json.JsonAST.JValue
-import net.liftweb.util.Helpers.tryo
-import org.apache.commons.io.FileUtils
-import org.apache.commons.io.IOUtils
 import org.joda.time.DateTime
-
-import scala.util.{Failure => FailTry}
-import scala.util.Success
-import scala.util.Try
 import com.normation.cfclerk.domain.SystemVariable
 import com.normation.cfclerk.domain.Variable
 import com.normation.rudder.services.policies.NodeConfiguration
@@ -75,6 +66,9 @@ import com.normation.rudder.services.policies.ParameterEntry
 import com.normation.rudder.services.policies.PolicyId
 import com.normation.rudder.services.policies.Policy
 import java.nio.charset.StandardCharsets
+import java.nio.file.AtomicMoveNotSupportedException
+import java.nio.file.StandardCopyOption
+import java.nio.file.StandardOpenOption
 import java.util.concurrent.TimeUnit
 
 import com.normation.rudder.hooks.HookReturnCode
@@ -84,13 +78,15 @@ import zio.duration._
 import com.normation.errors._
 import com.normation.box._
 import com.normation.zio._
-import zio.blocking.Blocking
 import zio.duration.Duration
 import cats.data._
 import cats.implicits._
 import com.normation.rudder.domain.logger.NodeConfigurationLogger
-import org.apache.commons.lang.StringUtils
+import better.files._
 import com.normation.rudder.domain.logger.PolicyGenerationLogger
+import com.normation.rudder.domain.logger.PolicyGenerationLoggerPure
+import com.normation.templates.FillTemplateThreadUnsafe
+import com.normation.templates.FillTemplateTimer
 
 /**
  * Write promises for the set of nodes, with the given configs.
@@ -114,6 +110,40 @@ trait PolicyWriterService {
   ) : Box[Seq[NodeId]]
 }
 
+object PolicyWriterServiceImpl {
+  //an utility that write text in a file and create file parents if needed
+  implicit class CreateParentAndWrite(val file: File) extends AnyVal {
+    // open file mode for create or overwrite mode
+    def createParentsAndWrite(text: String) = IOResult.effect {
+      file.parent.createDirectoryIfNotExists(true)
+      file.writeText(text)(Seq(StandardOpenOption.WRITE, StandardOpenOption.CREATE), StandardCharsets.UTF_8)
+    }
+  }
+}
+
+/*
+ * Timer accumulator for write part, in nanoseconds
+ */
+case class WriteTimer(
+    writeTemplate: Ref[Long]
+  , fillTemplate : Ref[Long]
+  , copyResources: Ref[Long]
+  , agentSpecific: Ref[Long]
+  , writeCSV     : Ref[Long]
+  , writeJSON    : Ref[Long]
+)
+
+object WriteTimer {
+  def make() = for {
+    a <- Ref.make(0L)
+    b <- Ref.make(0L)
+    c <- Ref.make(0L)
+    d <- Ref.make(0L)
+    e <- Ref.make(0L)
+    f <- Ref.make(0L)
+  } yield WriteTimer(a, b, c, d, e, f)
+}
+
 class PolicyWriterServiceImpl(
     techniqueRepository       : TechniqueRepository
   , pathComputer              : PathComputer
@@ -124,14 +154,18 @@ class PolicyWriterServiceImpl(
   , HOOKS_D                   : String
   , HOOKS_IGNORE_SUFFIXES     : List[String]
 ) extends PolicyWriterService {
+  import PolicyWriterServiceImpl._
 
-
-  val timingLogger = PolicyGenerationLogger.timing
+  val clock = ZioRuntime.environment
+  val timingLogger = PolicyGenerationLoggerPure.timing
 
   val hookWarnDurationMillis = (60*1000).millis
 
   val newPostfix = ".new"
   val backupPostfix = ".bkp"
+
+  // we use UTF-8 everywhere
+  implicit val charset: Charset = StandardCharsets.UTF_8
 
   private[this] def writeNodePropertiesFile (agentNodeConfig: AgentNodeConfiguration) = {
 
@@ -141,44 +175,31 @@ class PolicyWriterServiceImpl(
       ( "properties" -> properties.toDataJson())
     }
 
-    val fileName = Constants.GENERATED_PROPERTY_FILE
     val path = Constants.GENERATED_PROPERTY_DIR
+    val file = File(agentNodeConfig.paths.newFolder, path, Constants.GENERATED_PROPERTY_FILE)
     val jsonProperties = generateNodePropertiesJson(agentNodeConfig.config.nodeInfo.properties)
     val propertyContent = JsonAST.prettyRender(jsonProperties)
-    PolicyGenerationLogger.trace(s"Create node properties file '${agentNodeConfig.paths.newFolder}/${path}/${fileName}'")
-    Try {
-      val propertyFile = new File ( new File (agentNodeConfig.paths.newFolder, path), fileName)
-      FileUtils.writeStringToFile(propertyFile, propertyContent, StandardCharsets.UTF_8)
-    } match {
-      case FailTry(e) =>
-        val message = s"could not write ${fileName} file, cause is: ${e.getMessage}"
-        Failure(message)
-      case Success(_) =>
-        Full(agentNodeConfig.config.nodeInfo.id)
-    }
+
+    for {
+      _ <- PolicyGenerationLoggerPure.trace(s"Create node properties file '${agentNodeConfig.paths.newFolder}/${path}/${file.name}'")
+      _ <- file.createParentsAndWrite(propertyContent).chainError(s"could not write file: '${file.name} '")
+    } yield ()
   }
 
-  private[this] def writeRudderParameterFile(agentNodeConfig: AgentNodeConfiguration) = {
+  private[this] def writeRudderParameterFile(agentNodeConfig: AgentNodeConfiguration): IOResult[Unit] = {
     def generateParametersJson(parameters : Set[ParameterEntry]): JValue = {
       import com.normation.rudder.domain.nodes.JsonSerialisation._
       import net.liftweb.json.JsonDSL._
       ( "parameters" -> parameters.toDataJson())
     }
 
-    val fileName = Constants.GENERATED_PARAMETER_FILE
+    val file = File(agentNodeConfig.paths.newFolder, Constants.GENERATED_PARAMETER_FILE)
     val jsonParameters = generateParametersJson(agentNodeConfig.config.parameters.map(x => ParameterEntry(x.name.value, x.value, agentNodeConfig.agentType)))
     val parameterContent = JsonAST.prettyRender(jsonParameters)
-    PolicyGenerationLogger.trace(s"Create parameter file '${agentNodeConfig.paths.newFolder}/${fileName}'")
-    Try {
-      val parameterFile = new File ( new File (agentNodeConfig.paths.newFolder), fileName)
-      FileUtils.writeStringToFile(parameterFile, parameterContent, StandardCharsets.UTF_8)
-    } match {
-      case FailTry(e) =>
-        val message = s"could not write ${fileName} file, cause is: ${e.getMessage}"
-        Failure(message)
-      case Success(_) =>
-        Full(agentNodeConfig.config.nodeInfo.id)
-    }
+    for {
+      _ <- PolicyGenerationLoggerPure.trace(s"Create parameter file '${agentNodeConfig.paths.newFolder}/${file.name}'")
+      _ <- file.createParentsAndWrite(parameterContent).chainError(s"could not write file: ' ${file.name}'")
+    } yield ()
   }
 
   /*
@@ -201,10 +222,8 @@ class PolicyWriterServiceImpl(
    * here, f must not throws execption.
    *
    */
-  def parrallelSequence[U,T](seq: Seq[U])(f:U => Box[T])(implicit timeout: Duration, maxParallelism: Int): Box[Seq[T]] = {
-    ZIO.accessM[Blocking]{ b =>
-      seq.accumulateParN(maxParallelism)(a => b.blocking.blocking(f(a).toIO))
-    }.timeout(timeout).foldM(
+  def parrallelSequence[U,T](seq: Seq[U])(f:U => IOResult[T])(implicit timeout: Duration, maxParallelism: Int): IOResult[Seq[T]] = {
+    seq.accumulateParN(maxParallelism)(a => f(a)).timeout(timeout).foldM(
       err => err.fail
     , suc => suc match {
         case None      => //timeout
@@ -212,12 +231,13 @@ class PolicyWriterServiceImpl(
         case Some(seq) =>
           seq.succeed
       }
-    ).provide(ZioRuntime.environment).toBox
+    ).provide(clock)
   }
 
   // a version for Hook with a nicer message accumulation
-  def parallelSequenceNodeHook(seq: Seq[AgentNodeConfiguration])(f: AgentNodeConfiguration => HookReturnCode)(implicit timeout: Duration, maxParallelism: Int): Box[Unit] = {
+  def parallelSequenceNodeHook(seq: Seq[AgentNodeConfiguration])(f: AgentNodeConfiguration => IOResult[HookReturnCode])(implicit timeout: Duration, maxParallelism: Int): IOResult[Unit] = {
 
+    // a dedicated error for hooks
     final case class HookError(nodeId: NodeId, errorCode: HookReturnCode.Error) extends RudderError {
       def limitOut(s: String) = {
         val max = 300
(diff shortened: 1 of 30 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -264,20 +262,29 @@ class PolicyWriterServiceImpl(
...
     //debug - but don't fails for debugging !
-    logNodeConfig.log(nodeConfigsToWrite.values.toSeq) match {
-      case eb:EmptyBox =>
-        val e = eb ?~! "Error when trying to write node configurations for debugging"
-        logger.error(e)
-        e.rootExceptionCause.foreach { ex =>
-          logger.error("Root exception cause was:", ex)
-        }
-      case _ => //nothing to do
+    val logNodeConfigurations = ZIO.when(logNodeConfig.isDebugEnabled) {
```
ncharles: this one must also help a lot - we were always duplicating all the nodeconfig written

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -311,143 +317,138 @@ class PolicyWriterServiceImpl(
...
-                            } yield {
-                              "OK"
-                            }) ?~! s"Error when writing configuration for node '${prepared.paths.nodeId.value}'"
-                          }
-      promiseWrittenTime = System.currentTimeMillis
-      promiseWrittenDur  = promiseWrittenTime - preparedPromisesTime
-      _                  = policyLogger.debug(s"Policies written in ${promiseWrittenDur} ms")
+      preparedPromises     <- parrallelSequence(configAndPaths) { case agentNodeConfig =>
+                                val nodeConfigId = versions(agentNodeConfig.config.nodeInfo.id)
+                                prepareTemplate.prepareTemplateForAgentNodeConfiguration(agentNodeConfig, nodeConfigId, rootNodeId, templates, allNodeConfigs, Policy.TAG_OF_RUDDER_ID, globalPolicyMode, generationTime, maxParallelism).chainError(
```
ncharles: i'm a bit confused by the expected gain of a paralelised computation within a parrallelSequence

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -512,58 +568,76 @@ class PolicyWriterServiceImpl(
...
+               for {
+                 copyInfo <- techniqueRepository.getTemplateContent(templateId) { optInputStream =>
+                               optInputStream match {
+                                 case None =>
+                                   Unexpected(s"Error when trying to open template '${templateId.toString}'. Check that the file exists with a ${TechniqueTemplate.templateExtension} extension and is correctly commited in Git, or that the metadata for the technique are corrects.").fail
+                                 case Some(inputStream) =>
+                                   for {
+                                     _       <- PolicyLoggerPure.trace(s"Loading template: ${templateId}")
+                                               //string template does not allows "." in path name, so we are force to use a templateGroup by polity template (versions have . in them)
+                                     content <- IOResult.effect(s"Error when copying technique template '${templateId.toString}'")(inputStream.asString(false))
```
ncharles: is switching from IOUtils to better.files as a noticable impact ? if so, we could do that in 5.0 also

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -512,58 +568,76 @@ class PolicyWriterServiceImpl(
...
+      _   <- timingLogger.debug(s"${templatesToRead.size} promises templates read in ${t1-t0} ms")
+    } yield res
   }
 
 
+  /*
+   * This method is an hot spot in Rudder. It is where most of the generation is taking time, and that for two reasons:
+   * - the fisrt one is that we write a lot of template. Not much can be done here.
+   * - the second is around template filling with StringTemplate. Parsing a template is very slow, so we cache them.
+   *   But even once parsed, we can't just use the parsed value, we need to
```
ncharles: the comment is truncated here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -581,33 +655,36 @@ class PolicyWriterServiceImpl(
...
-      } yield {
-        ((copyInfo.id, agentType), copyInfo)
-      }
-    }).map( _.toMap)
-
-    logger.debug(s"${staticResourceToRead.size} techniques resources read in ${System.currentTimeMillis-now} ms")
-    res
+    for {
+      t0  <- currentTimeMillis
+      res <- (parrallelSequence(staticResourceToRead) { case (templateId, agentType, templateOutPath) =>
```
ncharles: maybe this one also should not be parralel?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -789,75 +862,91 @@ class PolicyWriterServiceImpl(
...
+      val src = File(nodeFolder)
+      if (src.isDirectory()) {
+        val dest = File(backupFolder)
+        if (dest.isDirectory) {
+          // force deletion of previous backuup
+          dest.delete(false, File.LinkOptions.noFollow)
+        }
+        PolicyLogger.trace(s"Backup old '${nodeFolder}' into ${backupFolder}")
+        dest.parent.createDirectoryIfNotExists(true)
+        src.moveTo(dest)(File.CopyOptions.atomically)
```
ncharles: ok, if we do that, we need to enforce that users are using the same FS for backup and for policies, otherswise it raises a AtomicNotSupportedException

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -789,75 +862,91 @@ class PolicyWriterServiceImpl(
...
-      val dest = new File(backupFolder)
-      if (dest.isDirectory) {
-        // force deletion of previous backuup
-        FileUtils.deleteDirectory(dest)
+  private[this] def backupNodeFolder(nodeFolder: String, backupFolder: String): IOResult[Unit] = {
+    IOResult.effect {
+      val src = File(nodeFolder)
+      if (src.isDirectory()) {
+        val dest = File(backupFolder)
+        if (dest.isDirectory) {
```
ncharles: just a stupid question: if the folder is there, and we move, do we really nned to delete first ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
ncharles: and do you know how much of the improvement comes from the atomically, compared to switching to betterFiles ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -789,75 +862,91 @@ class PolicyWriterServiceImpl(
...
+             val dest = File(destinationFolder)
+             for {
+               _ <- PolicyLoggerPure.trace(s"Moving folders: \n  from ${src.pathAsString}\n  to   ${dest.pathAsString}")
+               _ <- ZIO.whenM(IOResult.effect(dest.isDirectory)) {
+                      // force deletion of previous promises
+                      IOResult.effect(dest.delete(false, File.LinkOptions.noFollow))
+                    }
+               _ <- IOResult.effect {
+                      dest.parent.createDirectoryIfNotExists(true)
+                      src.moveTo(dest)(File.CopyOptions.default)
```
ncharles: this ought to be atomically, there is no reason for the backup to be and not this one

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: No, this one is ok, it doesn't have any parallel on top of it.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: I've added a test that choose between `atomic` or not, and write a warning if it's not supported. But the more I test, the less I believe `atomic` is the reason of the performance boost (if my test always return `default`, I get the same result - but perhaps linux is smart and it does an atomic move by default if no option is given ?)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: I think it was for the case when something bad happened in a previous generation. But now that we can be much more fine grained on our error management, perhaps we could delete on error. I'm not sur it changes much in perf (to avoid the unused delete).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: it's not all consistant

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
ncharles: delete is fairly costly when there are many files and already a lors of I/O
I don't remember exactly, but it was between 2 and 6 minutes to delete with rm the backup folders with 10 000 nodes

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
ncharles: :+1:

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
ncharles: Linux is smart. The change is in the cache FS first, then is commited when it's full

ncharles:
This fails:
```[2019-12-18 04:57:13] ERROR policy.generation - Error when updating policy, reason was: Cannot write nodes configuration <- Error when testing for policy 'mv' mode
[2019-12-18 04:57:13] ERROR policy.generation - Root exception was: /var/rudder/share/de0727bb-b594-4929-bebd-2ca258c47204/rules.new/test-rudder-policy-mv-options8612784980732448 -> /var/rudder/share/de0727bb-b594-4929-bebd-2ca258c47204/rules
[2019-12-18 04:57:13] INFO  policy.generation - Flag file '/opt/rudder/etc/policy-update-running' successfully removed
```
i don't have more explaination than that

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -795,6 +801,49 @@ class PolicyWriterServiceImpl(
...
+      _  <- IO.effect(f3.delete(true))
+    } yield {
+      atomically
+    }).catchAll {
+      case ex: AtomicMoveNotSupportedException =>
+        PolicyLoggerPure.warn(s"Node policy directories (${examplePath.baseFolder}, ${examplePath.newFolder}, ${examplePath.backupFolder}) " +
+                              s"are not on the same file system. Rudder won't be able to use atomic move for policies, which may have dire " +
+                              s"effect on policy generation performance. You should use the file system.") *>
+        default.succeed
+      case ex: Throwable => SystemError("Error when testing for policy 'mv' mode", ex).fail
```
ncharles: root exception is missing in my error
[2019-12-18 04:57:13] ERROR policy.generation - Error when updating policy, reason was: Cannot write nodes configuration <- Error when testing for policy 'mv' mode
[2019-12-18 04:57:13] ERROR policy.generation - Root exception was: /var/rudder/share/de0727bb-b594-4929-bebd-2ca258c47204/rules.new/test-rudder-policy-mv-options8612784980732448 -> /var/rudder/share/de0727bb-b594-4929-bebd-2ca258c47204/rules

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -300,8 +300,10 @@ object RunHooks {
    * Hooks must be executable and not ends with one of the
    * non-executable extensions.
    */
-   def getHooks(basePath: String, ignoreSuffixes: List[String]): Box[Hooks] = {
-     tryo {
+   def getHooks(basePath: String, ignoreSuffixes: List[String]): Box[Hooks] = getHooksPure(basePath, ignoreSuffixes).toBox
+
+   def getHooksPure(basePath: String, ignoreSuffixes: List[String]): IOResult[Hooks] = {
+     IOResult.effect {
```
ncharles: it seems this causes a huge drop in performance

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: Ok :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: it's the box displaying that is not nice. It say that it can't move to dest, but we miss the reason. Most likely, the problem is because parent doesn't exists and I didn't created the correct level.

fanf: I will change the exception reporting too

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
fanf: OK, I will need to test specifically that part. I think I spawn thread for non blocking part (the blocking part already spawn a thread via NuProcess. I need to check carefully

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: I don't see what the problem is, I added logs - I fear we will need to test it again to understand.

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: I added logs and it was because I forgot the filename in the move, so it was trying to overwrite a directory with file :/. Now corrected. Perf are not better

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
fanf: so it seems that performance are no better

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -305,168 +335,293 @@ class PolicyWriterServiceImpl(
...
+     */
+
+    // template are created based on "content"
+    val byTemplate = (for {
+      p  <- preparedTemplates
+      pt <- p.preparedTechniques
+      t  <- pt.templatesToProcess
+    } yield {
+      (t.content, TemplateFillInfo(t.id, t.destination, p.paths.newFolder, pt.environmentVariables, pt.reportIdToReplace))
+    }).groupMap(_._1)(_._2)
```
ncharles: This is really clever

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -305,168 +335,293 @@ class PolicyWriterServiceImpl(
...
+                                           filled
+                                         }
+                       (replaced, dest) = replacedDest
+                       t1               <- currentTimeNanos
+                       _                <- File(info.newFolder, dest).createParentsAndWrite(replaced).chainError(
+                                               s"Bad format in Technique ${info.id.toString} (file: ${info.destination})"
+                                           )
+                       t2               <- currentTimeNanos
+                       _                <- writeTimer.writeTemplate.update(_ + t2 - t1)
+                     } yield ()
```
ncharles: I am a bit scared by the potential memory usage here, as we keep a lot of data in memory
I'm going to accept it, because it seems ok on our test platform, but let's keep an eye on this

webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
```diff
@@ -139,51 +177,126 @@ class SynchronizedFileTemplate(templateName: String, localTemplate: Either[Rudde
...
+/*
+ * A performance optimized version of fill template which assumes that thread safety is managed for it.
+ * Main changes:
+ * - IT IS NOT THREADSAFE.
+ * - set variable's values directly as list and not one by one
+ * - unwrappe STVariable array seq to avoid the huge conversion cost.
+ */
+object FillTemplateThreadUnsafe {
+  ////////// Hottest method on whole Rudder //////////
+  def fill(templateName: String, sourceTemplate: StringTemplate, variables: Seq[STVariable], timer: FillTemplateTimer, replaceId: Option[(String, String)]): IOResult[(String, String)] = {
```
ncharles: why is there 2 fill in this file ? how can we know which one to use ?

fanf:
OK, squash merging this PR

PR #2673, opened 2019-12-11
Fixes #16391: Rudder package list command should not list plugins that are not available for the user

Fdall:
https://issues.rudder.io/issues/16391

### DIFF
--- a/relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
+++ b/relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
@@ -184,6 +184,13 @@ def download(completeUrl, dst=""):
     return fileDst
 
 """
+    Make a HEAD request on the given url, return true if result is 200, false instead
+"""
+def check_download(completeUrl):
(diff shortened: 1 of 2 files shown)


relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
```diff
@@ -183,6 +183,13 @@ def download(completeUrl, dst=""):
            fail("Received a HTTP %s error when trying to get %s"%(r.status_code, completeUrl))
     return fileDst
 
+"""
+    Make a HEAD request on the given url, return true if result is 200, false instead
+"""
+def check_download(completeUrl):
+    r = requests.head(completeUrl, auth=(USERNAME, PASSWORD))
+    return (r.status_code < 400)
```
peckpeck: why not < 300 ?

Fdall: We are doing redirection, so 301 is ok.
Could be <= 301

Fdall:
PR updated with a new commit

PR #2687, opened 2019-12-20
Fixes #16449: fuzzy matching to help with error messages

gpoblon:
https://issues.rudder.io/issues/16449

Compiling and implemented fuzzy string matching feature which is a useful tool to have when dealing with compilation errors such as type or variable not found. Finds a similar variable/type name available in the actual context (global + current scope) and indicates similar names.

### DIFF
--- a/rudder-lang/src/ast.rs
+++ b/rudder-lang/src/ast.rs
@@ -157,9 +157,10 @@ impl<'src> AST<'src> {
                 for em in new_mappings {
                     self.errors.push(err!(
                         em.to,
-                        "Enum {} not found when trying to define mapping {}",
+                        "Enum {} not found when trying to define mapping {}{}",
                         em.from,
-                        em.to
+                        em.to,
+                        get_suggestion_message(em.from.fragment(), enum_list.iter().map(|(k, _)| k)),
                     ));
                 }
                 break;
@@ -254,9 +255,10 @@ impl<'src> AST<'src> {
             if !self.resource_list.contains(&parent) {
                 self.errors.push(err!(
                     &child,
-                    "Resource {} declares {} as a parent, but it doesn't exist",
+                    "Resource {} declares {} as a parent, but it doesn't exist.{}",
                     child,
-                    parent
+                    parent,
+                    get_suggestion_message(parent.fragment(), self.resource_list.iter()),
                 ));
             } else {
                 children.entry(parent).or_insert_with(HashSet::new);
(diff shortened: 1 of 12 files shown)


rudder-lang/src/error.rs
```diff
@@ -205,3 +209,50 @@ impl fmt::Display for Error {
         }
     }
 }
+
+
+// Must have a trait to add the following method as ngrammatic::corpus only has an `add_text()` method
+trait FuzzyChecker<'src, I> {
+	fn add_vec(&mut self, list: I);
```
peckpeck: you call it add_vec but you don't add a vector ?

peckpeck: You have a generic parameter here but you only have a single implementation for an iterator

rudder-lang/src/error.rs
```diff
@@ -205,3 +209,50 @@ impl fmt::Display for Error {
...
+}
+
+impl<'src, I: Iterator<Item = &'src Token<'src>>> FuzzyChecker<'src, I> for Corpus {
```
peckpeck: It works here, but usually having a life time "&'a Object<'a>" often ends up with the constraint beging 'a == 'static 
which can become problematic, so avoid it when possible

rudder-lang/src/error.rs
```diff
@@ -205,3 +209,50 @@ impl fmt::Display for Error {
...
+}
+
+/// Searches for a matching string in an Iterator of Token
+fn fuzzy_search<'src, I>(token_fragment: &str, list: I) -> Option<String>
+where 
+	I: Iterator<Item = &'src Token<'src>>,
+{
+	let mut corpus = CorpusBuilder::new().finish();
+	corpus.add_vec(list);
+    let results = corpus.search(token_fragment, FUZZY_THRESHOLD);
```
peckpeck: indent

rudder-lang/src/ast/enums.rs
```diff
@@ -195,10 +195,11 @@ impl<'src> EnumList<'src> {
                     if !(values.contains(&f) || *f == "*") {
                         fail!(
                             f,
-                            "Value {} used in mapping {} does not exist in {}",
+                            "Value {} used in mapping {} does not exist in {}{}",
```
peckpeck: Missing separator

rudder-lang/src/ast/enums.rs
```diff
@@ -379,11 +386,17 @@ impl<'src> EnumList<'src> {
                         } else {
                             match self.enums.get(&e1) {
                                 // or get it from a global enum
-                                None => fail!(e1, "No such enum {}", e1),
+                                None => fail!(
+                                    e1,
+                                    "No such enum {}{}",
```
peckpeck: Missing separator

rudder-lang/src/ast.rs
```diff
@@ -311,29 +314,68 @@ impl<'src> AST<'src> {
         }
     }
 
+    // binding_check fn dependance: compares library and user function's parameters. if diff, output an error
```
amousset: Should probably be a doc comment `///`

PR #2688, opened 2019-12-23
Fixes #16458: Set HTTPS mode by default for new install, keep syslog for upgrade

ncharles:
https://issues.rudder.io/issues/16458


ncharles:
Commit modified

ncharles:
Commit modified

PR #2689, opened 2019-12-27
Fixes #16461: prettify output of rudder-lang compiler

gpoblon:
Until now rudder-lang was not verbose about what was going on behind the scenes. Plus and foremost every compilation error led to a `panic!`, which was wrapping errors, making them ugly and hard to read.

So the goal of this PR is to enhance output of the compiler, displaying what is happening (compilation stages and files) and clearly displaying errors so they are easier to be dealt with.

### DIFF
--- a/rudder-lang/src/error.rs
+++ b/rudder-lang/src/error.rs
@@ -28,8 +28,6 @@
 // You should have received a copy of the GNU General Public License
 // along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
 
-use ngrammatic::CorpusBuilder;
-use std::collections::HashMap;
 ///
 /// We write our own error type to have a consistent error type through all our code.
 /// We translate other types to this one when necessary.
@@ -41,10 +39,12 @@ use std::collections::HashMap;
 /// - Compilation error: usually we can skip what we are doing and go to next iteration
 /// - List: aggregate compilation errors so that user can fix them all ant once
 ///
+use crate::parser::Token;
+use colored::Colorize;
+use ngrammatic::CorpusBuilder;
+use std::collections::HashMap;
 use std::fmt;
 use std::hash::Hash;
-use colored::Colorize;
-use crate::parser::Token;
 
 const FUZZY_THRESHOLD: f32 = 0.5; // pub + `crate::` before calling since it may end up in the main.rs file.
 
--- a/rudder-lang/src/error.rs
+++ b/rudder-lang/src/error.rs
@@ -42,7 +42,7 @@ use std::collections::HashMap;
 ///
 use std::fmt;
 use std::hash::Hash;
-use colored::*;
+use colored::Colorize;
 
 #[derive(Debug, PartialEq, Clone)]
 pub enum Error {
@@ -104,7 +104,7 @@ impl Error {
 macro_rules! err {
     ($origin:expr, $ ( $ arg : tt ) *) => ({
         use crate::error::Error;
-        use colored::*;
+        use colored::Colorize;
 
         Error::User(format!(
                 "{}:\n{} {}",
--- a/rudder-lang/src/error.rs
+++ b/rudder-lang/src/error.rs
@@ -42,6 +42,7 @@ use std::collections::HashMap;
 ///
 use std::fmt;
 use std::hash::Hash;
+use colored::*;
 
 #[derive(Debug, PartialEq, Clone)]
 pub enum Error {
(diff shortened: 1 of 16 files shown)


rudder-lang/src/error.rs
```diff
@@ -42,6 +42,7 @@ use std::collections::HashMap;
 ///
 use std::fmt;
 use std::hash::Hash;
+use colored::*;
```
amousset: wildcard use should generally be avoided as it may break in the future in case new items added to the library may collide with existing names. You should be able to just import the required trait for color formatting.

amousset:
This needs a rebase now the other PR has been merged.

rudder-lang/src/main.rs
```diff
@@ -103,10 +104,13 @@ fn add_file<'a>(
...
-    let content = fs::read_to_string(path)
-        .unwrap_or_else(|_| panic!("Something went wrong reading the file {}", filename));
-    let content_str = source_list.append(content);
-    past.add_file(filename, &content_str)
+    println!("|- {} {}", "Parsing".bright_green(), filename);
+    if let Ok(content) = fs::read_to_string(path) {
+        let content_str = source_list.append(content);
+        past.add_file(filename, &content_str)
+    } else {
+        Err(Error::User(format!("could not read {}", &filename.bright_yellow())))
```
amousset: could be good to also display the underlying io error

rudder-lang/src/main.rs
```diff
@@ -103,10 +104,18 @@ fn add_file<'a>(
...
+    println!("|- {} {}", "Parsing".bright_green(), filename);
+    match fs::read_to_string(path) {
+        Ok(content) => {
+            let content_str = source_list.append(content);
+            past.add_file(filename, &content_str)
+        },
+        Err(e) => {
+            let tk = Token::new(filename, "");
+            println!("{:#?}", tk);
+            Err(err!(Token::new(filename, ""), "{}", e))
```
amousset: You should be able to use `tk` in the `Err` (you will need to pass `&tk` to the `println` though)

rudder-lang/src/main.rs
```diff
@@ -103,10 +104,18 @@ fn add_file<'a>(
...
-    past.add_file(filename, &content_str)
+    println!("|- {} {}", "Parsing".bright_green(), filename);
+    match fs::read_to_string(path) {
+        Ok(content) => {
+            let content_str = source_list.append(content);
+            past.add_file(filename, &content_str)
+        },
+        Err(e) => {
+            let tk = Token::new(filename, "");
+            println!("{:#?}", tk);
```
amousset: The next step will be to make this kind of output only appear in a verbose mode as it will be mostly useless for users.

rudder-lang/src/main.rs
gpoblon: It was actually just some debug I forgot, correcting right now

rudder-lang/src/main.rs
amousset: This will likely be one of the next output formatting changes, but should go into another pull request. Use the `log` crate with a simple cli flag would be a good start.

rudder-lang/src/technique.rs
```diff
@@ -64,33 +66,35 @@ struct MethodCall {
     component: String,
 }
 
-pub fn translate_file(json_file: &Path, rl_file: &Path) -> Result<()> {
-    let config_data = fs::read_to_string("data/config.toml").expect("Cannot read config.toml file");
-    let config: toml::Value = toml::from_str(&config_data).expect("Invalig config.toml file");
+pub fn translate_file(json_file: &Path, rl_file: &Path) -> Result<()>
+{
+    let input_filename = &json_file.to_string_lossy();
```
amousset: I don't really like this - converting Path to string to use them as paths just after - but I guess it makes formatting easier, so why not.

rudder-lang/src/technique.rs
```diff
@@ -64,33 +66,35 @@ struct MethodCall {
...
+    let output_filename = &rl_file.to_string_lossy();
+    let config_filename = "data/config.toml";
+    let file_error = |filename: &str, err| err!(Token::new(&filename.to_owned(), ""), "{}", err);
```
amousset: I'm not sure a file error should be represented using a Token, but this will do the job for now.

rudder-lang/src/technique.rs
gpoblon: Yes I agree but here both the Path and a string are used (and somewhat needed)

rudder-lang/src/technique.rs
gpoblon: Indeed, Clunky... but it allows to have only 1 format function to deal with 100% of errors. 
Tokens store both a filename and it's location

PR #2693, opened 2020-01-02
Fixes #16326: Display compliance bar in Node details

ElaadF:
https://issues.rudder.io/issues/16326

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/AsyncComplianceService.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/AsyncComplianceService.scala
@@ -200,4 +200,17 @@ class AsyncComplianceService (
     compliance(kind,tableId)
 
   }
+
+  def nodeCompliance(nodeId: NodeId): Box[ComplianceLevel] = {
+    val node = new NodeCompliance(Set(nodeId), Set())
+    node.computeCompliance match {
+      case Full(contentMap) =>
+        contentMap.get(nodeId) match {
+          case Some(Some(compliance)) => Full(compliance)
+          case _                      => Failure(s"No compliance found for ${nodeId.value}")
+        }
+      case eb: EmptyBox     =>
+        eb ?~! s"Compute compliance of ${nodeId.value} failed"
+    }
+  }
 }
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -85,10 +86,21 @@ object DisplayNode extends Loggable {
...
   private[this] val errorPopupHtmlId = "errorPopupHtmlId"
   private[this] val successPopupHtmlId = "successPopupHtmlId"
 
+
+  private def loadComplianceBar(nodeInfo: Option[NodeInfo]): Option[JsArray] = {
+    for {
+      node       <- nodeInfo
+      t          <- asyncComplianceService.nodeCompliance(node.id)
+    } yield {
+      ComplianceLevelSerialisation.ComplianceLevelToJs(t(node.id).get).toJsArray()
```
VinceMacBuche: never use get !! this thows exception ! in fact your service should return a Box[ComplianceLevel], not the Map[NodeId,Option[ComplianceLevel]]

Also don't do t(node.id), it's also a get and it throws exception !!

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -468,6 +485,7 @@ object DisplayNode extends Loggable {
 
         <div class="status-info col-lg-6 col-sm-5 col-xs-12">
           <h3>Status information</h3>
+          <div id={s"compliance-bar-${nodeInfo.get.id.value}"}></div>
```
VinceMacBuche: Here you should have a class "node-compliance-bar" and define all the css done below in a css file targeting that class

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/AsyncComplianceService.scala
```diff
@@ -200,4 +200,17 @@ class AsyncComplianceService (
     compliance(kind,tableId)
 
   }
+
+  def nodeCompliance(nodeId: NodeId): Box[ComplianceLevel] = {
+    val node = new NodeCompliance(Set(nodeId), Set())
+    node.computeCompliance match {
+      case Full(contentMap) =>
+        contentMap.values.head match {
```
VinceMacBuche: head throws exception too, especially if this is empty

You shoud should do 

<pre>contentMap.get(nodeId) match 
  case Some(Some(compiance)) => Full(Compliance)
  case _ =>
</pre>

PR #2695, opened 2020-01-06
Fixes #16480: Add default values for relayd configs whenever possible

amousset:
https://issues.rudder.io/issues/16480


amousset:
PR updated with a new commit

PR #2698, opened 2020-01-07
Fixes #16483: nonblocking errors parser, more precise context capture

gpoblon:
Integration of a non-blocking error parser
Integration of another way to implement and catch error contexts allowing to catch more accurate informations about errors

https://issues.rudder.io/issues/16483

### DIFF
--- a/rudder-lang/src/parser.rs
+++ b/rudder-lang/src/parser.rs
@@ -31,6 +31,7 @@
 mod error;
 mod token;
 
+use nom::error::*;
 use nom::branch::*;
 use nom::bytes::complete::*;
 use nom::character::complete::*;
@@ -190,23 +191,42 @@ macro_rules! sequence {
     ( { $($f:ident : $parser:expr;)* } => $output:expr ) => {
         move |i| {
             $(
-                let (i,$f) = $parser (i)?;
+                // intercept error to update its context if it should lead to a handled compilation error
+                let (j, $f) = match $parser (i) {
+                    Ok(res) => res,
+                    Err(e) => return Err(update_error_context(e, get_accurate_context(i)))
+                };
+                let i = j;
             )*
             Ok((i, $output))
         }
     };
 }
+
 /// wsequence is the same a sequence, but we automatically insert space parsing between each call
 macro_rules! wsequence {
     ( { $($f:ident : $parser:expr;)* } => $output:expr ) => {
         move |i| {
             $(
-                let (i,$f) = $parser (i)?;
-                let (i,_) = strip_spaces_and_comment(i)?;
+                // intercept error to update its context if it should lead to a handled compilation error
+                let (j, $f) = match $parser (i) {
+                    Ok(res) => res,
+                    Err(e) => return Err(update_error_context(e, get_accurate_context(i)))
+                };
+                let (i,_) = strip_spaces_and_comment(j)?;
             )*
             Ok((i, $output))
         }
     };
+}
+
+/// Tool function allowing to save last state of parsing offset line before any error
+fn get_accurate_context(i: PInput) -> PInput {
+    // Might be useful to expand via a second parameter the limit pattern or even a function to combine with
+    let single_line_result: nom::IResult<PInput, PInput> = take_until("\n")(i);
+    // return the received input by default
+    let (_next, updated_context) = single_line_result.unwrap_or((i, i));
+    updated_context
 }
 
 /// A source file header consists of a single line '@format=<version>'.
@@ -223,7 +243,7 @@ fn pheader(i: PInput) -> PResult<PHeader> {
             version: or_fail(
                 map_res(take_until("\n"), |s: PInput| s.fragment.parse::<u32>()),
                 || PErrorKind::InvalidFormat
-                );
+            );
             _x: tag("\n");
         } => PHeader { version }
     )(i)
@@ -260,10 +280,10 @@ fn penum(i: PInput) -> PResult<PEnum> {
         {
             metadata: pmetadata_list; // metadata unsupported here, check done after 'enum' tag
             global: opt(tag("global")); // TODO at least one space here
-            e:      tag("enum"); // TODO at least one space here
-            _fail: or_fail(verify(peek(anychar), |_| metadata.is_empty()), || PErrorKind::UnsupportedMetadata(metadata[0].key.into()));
+            e:      or_err(tag("enum"), || PErrorKind::ExpectedReservedWord("enum")); // TODO at least one space here
+            _fail:  or_fail(verify(peek(anychar), |_| metadata.is_empty()), || PErrorKind::UnsupportedMetadata(metadata[0].key.into()));
             name:   or_fail(pidentifier, || PErrorKind::InvalidName(e));
-            b:      tag("{"); // do not fail here, it could still be a mapping
+            b:      or_err(tag("{"), || PErrorKind::ExpectedToken("{")); // do not fail here, it could still be a mapping
             items:  separated_nonempty_list(sp(tag(",")), pidentifier);
             _x:     opt(tag(","));
             _x:     or_fail(tag("}"), || PErrorKind::UnterminatedDelimiter(b));
@@ -290,12 +310,12 @@ fn penum_mapping(i: PInput) -> PResult<PEnumMapping> {
     wsequence!(
         {
             metadata: pmetadata_list; // metadata unsupported here, check done after 'enum' tag
-            e:    tag("enum");// TODO at least one space here
+            e:     tag("enum");// TODO at least one space here
             _fail: or_fail(verify(peek(anychar), |_| metadata.is_empty()), || PErrorKind::UnsupportedMetadata(metadata[0].key.into()));
-            from: or_fail(pidentifier,|| PErrorKind::InvalidName(e));
-            _x:   or_fail(tag("~>"),|| PErrorKind::UnexpectedToken("~>"));
-            to:   or_fail(pidentifier,|| PErrorKind::InvalidName(e));
-            b:    or_fail(tag("{"),|| PErrorKind::UnexpectedToken("{"));
+            from:  or_fail(pidentifier,|| PErrorKind::InvalidName(e));
+            _x:    or_fail(tag("~>"),|| PErrorKind::ExpectedToken("~>"));
+            to:    or_fail(pidentifier,|| PErrorKind::InvalidName(e));
+            b:     or_fail(tag("{"),|| PErrorKind::ExpectedToken("{"));
             mapping:
                 separated_nonempty_list(
                     sp(tag(",")),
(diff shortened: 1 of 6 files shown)


rudder-lang/src/parser.rs
```diff
@@ -190,25 +191,47 @@ macro_rules! sequence {
...
+                };
                 let (i,_) = strip_spaces_and_comment(i)?;
             )*
             Ok((i, $output))
         }
     };
 }
 
+/// Tool function allowing to save last state of parsing offset line before any error
+fn get_focalized_input(i: PInput) -> PInput {
```
peckpeck: Is focalized the right term, isn't there a more explicit word ?

rudder-lang/src/parser.rs
```diff
@@ -190,25 +191,47 @@ macro_rules! sequence {
...
+    // Might be useful to expand via a second parameter the limit pattern or even a function to combine with
+    let err_result: nom::IResult<PInput, PInput> = take_until("\n")(i);
+    // return the received input by default
+    let (_next, err_holder) = err_result.unwrap_or((i, i));
```
peckpeck: why name it err_holder ?

rudder-lang/src/parser.rs
```diff
@@ -190,25 +191,47 @@ macro_rules! sequence {
...
             )*
             Ok((i, $output))
         }
     };
 }
 
+/// Tool function allowing to save last state of parsing offset line before any error
+fn get_focalized_input(i: PInput) -> PInput {
+    // Might be useful to expand via a second parameter the limit pattern or even a function to combine with
+    let err_result: nom::IResult<PInput, PInput> = take_until("\n")(i);
```
peckpeck: why name it err_result ?

rudder-lang/src/parser.rs
```diff
@@ -190,25 +191,47 @@ macro_rules! sequence {
...
         move |i| {
             $(
-                let (i,$f) = $parser (i)?;
+                // save last line (cursor is updated at parser calls). context of an potential error
+                let focalized_error = get_focalized_input(i);
+                // intercept error to update its context if it should lead to a handled compilation error
+                let (i, $f) = match $parser (i) {
+                    Ok(res) => res,
+                    Err(e) => return Err(update_error_context(e, focalized_error))
+                };
```
peckpeck: you could also write it: 
```
  let (j,$f) ) ... {
       ... => return Err(update_error_context(e, get_focalized_input(i)))
  };
  let i = j;
```

Then you still have access to the original i and may some day find more information to extract from it

rudder-lang/src/parser.rs
```diff
@@ -260,10 +283,10 @@ fn penum(i: PInput) -> PResult<PEnum> {
         {
             metadata: pmetadata_list; // metadata unsupported here, check done after 'enum' tag
             global: opt(tag("global")); // TODO at least one space here
-            e:      tag("enum"); // TODO at least one space here
-            _fail: or_fail(verify(peek(anychar), |_| metadata.is_empty()), || PErrorKind::UnsupportedMetadata(metadata[0].key.into()));
-            name:   or_fail(pidentifier, || PErrorKind::InvalidName(e));
-            b:      tag("{"); // do not fail here, it could still be a mapping
+            e:      or_err(tag("enum"), || PErrorKind::ExpectedReservedWord("enum")); // TODO at least one space here
+            _fail:  or_err(verify(peek(anychar), |_| metadata.is_empty()), || PErrorKind::UnsupportedMetadata(metadata[0].key.into()));
```
peckpeck: why did you remove or_fail ? 
it seems to me that if the user has already written "enum blah" an invalid blah is a fail.
and if it is not, please rename the variable too

rudder-lang/src/parser.rs
```diff
@@ -260,10 +283,10 @@ fn penum(i: PInput) -> PResult<PEnum> {
...
+            name:   or_err(pidentifier, || PErrorKind::InvalidName(e));
```
peckpeck: This seems to be a fail too

rudder-lang/src/parser.rs
```diff
@@ -967,6 +990,7 @@ pub enum PDeclaration<'src> {
     Alias(PAliasDef<'src>),
 }
 fn pdeclaration(i: PInput) -> PResult<PDeclaration> {
+    end_of_pfile(i)?;
```
peckpeck: please add a comment on why this is here

rudder-lang/src/parser.rs
```diff
@@ -977,6 +1001,17 @@ fn pdeclaration(i: PInput) -> PResult<PDeclaration> {
     ))(i)
 }
 
+fn end_of_pfile(i: PInput) -> PResult<()> {
```
peckpeck: A little comment on what this parser does

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
-            Ok(y) => Ok(y),
+            Err(Err::Incomplete(_)) => panic!("Incomplete should never happen"),
+            Ok(y) => Ok(y)
         }
     }
 }
 
+/// This function turns our own `Error`s (not nom ones) into `Failure`s so they are properly handled
+/// by the `nom::multi::many0()` function which abstracts Errors, only breaking on failures which was an issue
+pub fn or_fail_perr<'src, O, F>(f: F) -> impl Fn(PInput<'src>) -> PResult<'src, O>
```
peckpeck: i think there is a better name, "_perr" here is emaningless

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
     move |input| {
-        let x = f(input);
-        match x {
+        match f(input) {
             // a non nom error cannot be superseded
             // keep original context when possible
             Err(Err::Failure(err)) => match err.kind {
                 PErrorKind::Nom(_) => Err(Err::Failure(PError {
-                    context: err.context,
+                    context: None,
```
peckpeck: why are you removing the original context ? 
it may contain useful data, especially if you forget or don't need to override it

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
             })),
-            Err(Err::Incomplete(_)) => Err(Err::Failure(PError {
-                context: input,
+            Err(Err::Incomplete(_)) => panic!("Incomplete should never happen"),
+            Ok(y) => Ok(y),
+        }
+    }
+}
+
+/// Similar code than `or_fail()` yet usage and behavior differ:
```
peckpeck: similar .. to

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
+/// primarly it is non-blockant as it does not turn Errors into Failures
```
peckpeck: non blocking

peckpeck: the proper should be non terminating

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
+where
+    F: Fn(PInput<'src>) -> PResult<'src, O>,
+    E: Fn() -> PErrorKind<PInput<'src>>,
+{
+    move |input| {
+        match f(input) {
+            Err(Err::Failure(err)) => match err.kind {
+                PErrorKind::Nom(_) => Err(Err::Failure(PError {
+                    context: None,
+                    kind: e(),
```
peckpeck: i'm not sure we should override the error kind here, could you give an example ?

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
+                })),
+                _ => Err(Err::Failure(e)),
+            },
+            Err(Err::Incomplete(_)) => panic!("Incomplete should never happen"),
+            Ok(y) => Ok(y)
+        }
+    }
+}
+
+/// Updates content of on error to fit and capture a better context
```
peckpeck: "on error" ?

rudder-lang/src/parser/error.rs
```diff
@@ -166,30 +184,103 @@ where
...
+/// Solely exists for (w)sequence macro
+pub fn update_error_context<'src>(e: Err<PError<PInput<'src>>>, new_ctx: PInput<'src>) -> Err<PError<PInput<'src>>> {
+    match e {
+        Err::Failure(err) => {
+            let context = match err.context {
+                None => Some(new_ctx),
+                Some(context_to_keep) => Some(context_to_keep)
+            };
+            Err::Failure(PError {
+                context: context,
```
peckpeck: this can be written " context, "

rudder-lang/src/parser/error.rs
```diff
@@ -123,27 +124,44 @@ impl<I: Clone> ParseError<I> for PError<I> {
...
-            self.context.fragment,
-            "-->".bright_blue(),
-            message.bold(),
-            
-        ))
+
+        // simply removes superfluous line return (prettyfication)
+        match self.context {
+            Some(ctx) => {
+                let mut context = ctx.fragment.to_owned();
```
amousset: Using https://doc.rust-lang.org/std/primitive.str.html#method.trim_end_matches would avoid creating a new mutable string, but it's not mandatory and current solution is acceptable.

rudder-lang/src/parser.rs
gpoblon: Just checked, this was a mistake: I needed to have every enum error non-terminated for testing purpose and forgot to revert it. Fixed

rudder-lang/src/parser.rs
gpoblon: I wanted to make it explicit that it is only call it whenever we hit an error.
But yeah, I've updated the entire function naming so it's not ambiguous

rudder-lang/src/parser.rs
gpoblon: renamed

rudder-lang/src/parser/error.rs
gpoblon: nice tip, thanks, updated

rudder-lang/src/parser/error.rs
gpoblon: Do you mean that it is just superfuous code since anyway `or_fail()` will be called in case of a failure ?

gpoblon: fixed, it was superfluous indeed

gpoblon:
All fixes related to previous comments pushed

rudder-lang/src/parser.rs
```diff
@@ -209,27 +208,26 @@ macro_rules! wsequence {
...
-                let focalized_error = get_focalized_input(i);
                 // intercept error to update its context if it should lead to a handled compilation error
-                let (i, $f) = match $parser (i) {
+                let (j, $f) = match $parser (i) {
                     Ok(res) => res,
-                    Err(e) => return Err(update_error_context(e, focalized_error))
+                    Err(e) => return Err(update_error_context(e, get_accurate_context(i)))
                 };
-                let (i,_) = strip_spaces_and_comment(i)?;
+                let (j,_) = strip_spaces_and_comment(j)?;
```
peckpeck: let (i,_)

gpoblon: Corrigé

PR #2699, opened 2020-01-07
Fixes #16491: Port Rudder to Scala 2.13

fanf:
https://issues.rudder.io/issues/16491

There is a lot of little things, only one tricky, so start by that one:
In scala.xml, `Node ++ Node` leads to `Seq[Node]` in place of `NodeSeq`. There is nothing special for method awaiting `NodeSeq` because there's an implicit conversion, but when a method accepts `Seq[A]`, it keeps the former. This is a problem for lift `#>` which accept both `Seq[A]` et `NodeSeq` with different semantic. On `Seq[A]`, it applies the mapping on all `A`, else it applies one whole `NodeSeq`. So we need to force everything to `NodeSeq` to be sure.

Other things: 

- some lib needed to be updated to have a scala 2.13, but all to version with similar semantic (doobie just need an explicit import for `Timestamp` now)
- changes in scalac flags lead to various change in imports 
- the biggest change: there is no more automatic conversion between mutable and immutable collection, so that leads to a lot of `.toSeq`
- `mapValues` and `filterKey` are now method on `view`, which has a different type to make explicit when there is a view. So we must force its evaluation with `toMap` afterward. 
- compiler is more strict on several things, so we needed to add type annotation, missing escape, 1L in place of 1l, etc
- we need to call `toString` before `+` (thing not coerced to string automatically by default, which is a big win against silly errors)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

PR #2702, opened 2020-01-08
Fixes #16496: StatusReportTest leads to inconsistant results

fanf:
https://issues.rudder.io/issues/16496


amousset:
Trying against 6.0 to see what it gives.

PR #2703, opened 2020-01-08
Fixes #16503: Normalize loggers for timing performance on write promises

fanf:
https://issues.rudder.io/issues/16503


fanf:
PR updated with a new commit

PR #2705, opened 2020-01-08
Fixes #16505: Change ZIO runtime ergonomics

fanf:
https://issues.rudder.io/issues/16505

### DIFF
--- a/webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
+++ b/webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
@@ -311,8 +311,12 @@ object zio {
      * a hierarchy of calls.
      */
     val internal = new DefaultRuntime() {
-
-
+      import _root_.zio.internal._
+      import _root_.zio.clock._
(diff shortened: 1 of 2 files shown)


fanf:
John proposed: ```You might create a Runtime that is designed for blocking actions, which limits threads to 100 (say), and has a nice queue, and keeps threads alive for a while. That should perform pretty well for "mixed loads" in your scenario. And then you don't have to use Blocking at all anywhere.```. 
I would like to test that before merging that PR.

fanf:
So I tried to use the same thead pool for async and blocking fibers, and the results are not good. I tried:
- JDK 8 ForkJoinPool
- ThreadPoolExecutor with: LinkedBlockingQueue, ArrayBlockingQueue (with 32k size), both fair or not, with a core thread pool of nb CPU, and a max of 256. 

The result are similar or worse than with that code, and I got some deadlock (with fairness - not sure why). 

So I propose to just use that PR modification which are at least safe, and to come back to that subject latter, when other perf gain will be explored.

PR #2706, opened 2020-01-08
Fixes #16506: Improve Hook exec  performance

fanf:
https://issues.rudder.io/issues/16506


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -300,8 +303,10 @@ object RunHooks {
    * Hooks must be executable and not ends with one of the
    * non-executable extensions.
    */
-   def getHooks(basePath: String, ignoreSuffixes: List[String]): Box[Hooks] = {
-     tryo {
+   def getHooks(basePath: String, ignoreSuffixes: List[String]): Box[Hooks] = getHooksPure(basePath, ignoreSuffixes).toBox
```
ncharles: no using try makes it easier on the heap, see https://medium.com/@muuki88/follow-the-stacktraces-jvm-performance-profiling-3c371d323e5f

PR #2709, opened 2020-01-10
Fixes #16513: JVM GC cannot clean objects in scope in a for { } yield {} even if they are not referenced anymore

ncharles:
https://issues.rudder.io/issues/16513

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -335,113 +335,132 @@ trait PromiseGenerationService {
       timeCodePreGenHooks  =  (System.currentTimeMillis - codePreGenHooksTime)
       _                    =  PolicyLogger.debug(s"Pre-policy-generation modules hooks in ${timeCodePreGenHooks} ms, start getting all generation related data.")
 
+      // Objects need to be created in separate for {} yield to be garbage collectable,
+      // otherwise they remain in the scope and are not freed.
+      // We need to do that as many time as necessary to free memory during policy generation
+
       fetch0Time           =  System.currentTimeMillis
-      allRules             <- findDependantRules() ?~! "Could not find dependant rules"
-      fetch1Time           =  System.currentTimeMillis
-      _                    =  PolicyLogger.trace(s"Fetched rules in ${fetch1Time-fetch0Time} ms")
-      allNodeInfos         <- getAllNodeInfos.map( _.filter { case(_,n) =>
-                                if(n.state == NodeState.Ignored) {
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -335,113 +335,132 @@ trait PromiseGenerationService {
...
-
-      writeTime             =  System.currentTimeMillis
-      writtenNodeConfigs    <- writeNodeConfigurations(rootNodeId, updatedNodeConfigIds, nodeConfigs, allLicenses, globalPolicyMode, generationTime, parallelism) ?~!"Cannot write nodes configuration"
-      timeWriteNodeConfig   =  (System.currentTimeMillis - writeTime)
-      _                     =  PolicyLogger.debug(s"Node configuration written in ${timeWriteNodeConfig} ms, start to update expected reports.")
+      (updatedNodesId, updatedNodeInfo, expectedReports, allErrors, errorNodes, timeFetchAll, timeHistorize, timeRuleVal, timeBuildConfig, timeWriteNodeConfig, timeSetExpectedReport) <- for {
+        (updatedNodeConfigIds, allNodeConfigurations, allNodeConfigsInfos, updatedNodesId, updatedNodeInfo, allLicenses, globalPolicyMode, allNodeModes, allErrors, errorNodes, timeFetchAll, timeHistorize, timeRuleVal, timeBuildConfig) <- for {
+          (activeNodeIds, ruleVals, nodeContexts, allNodeModes, scriptEngineEnabled, globalPolicyMode, nodeConfigCaches, allLicenses, timeFetchAll, timeHistorize, timeRuleVal) <- for {
```
fanf: The scoping is quickly becoming hard to follow I believe. Wouldn't it be clearer to factor out the inner-most `for` in dedicated support function ? (I'm not sure, because we already have tons of them, and we will likely loose the general understanding of the policy generation workflow). What's your opinion on that?

fanf: Or perhaps build dedicated `case` objects that only handle the relevant part and are passed around ? 
(well, I fear that it will make refactoring harder, since we will have a hard time to know when we can remove an object from such a structure, so perhaps a bad idea actually)

ncharles: we already have a lot of back and forth with def, making it hard to follow. Plus, there are timestamp for loging within this, so adding them in a def is nightmarish

ncharles: I wish there were a way to make it better (an annotation that the compiler could use to know to release quickly some variable)

fanf:
OK, merging this PR

PR #2710, opened 2020-01-10
Fixes #16524: we should not use .size to check if a collection is empty

ncharles:
https://issues.rudder.io/issues/16524


ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

ncharles:
OK, merging this PR

PR #2711, opened 2020-01-12
Fixes #16525: Improve hook performance caused test errors

fanf:
https://issues.rudder.io/issues/16525


fanf:
I'm self merging that one: it's a revert of a bit of the parent issue to previous code, which is known to work.

PR #2714, opened 2020-01-14
Fixes #16484: rl integration tests v1

gpoblon:
Iterative and early work on integration tests for rudder-lang + some structural changes to sources for better naming clarity and to allow for library function calls from integration tests_. 

WIP, current notes:
For now, only compile tests are integrated.
3 tests: 
- every file from the *compile/* folder,
- a static array of tuples from inside the *compile.rs* file,
- a list of files from the *virtual_files.toml*
For the last two tests, a file is created in the *tmp/* folder, and called from the base crate compile function.

TODO:
- Use a macro to make it so 1 (virtual or not) file = 1 test. If it is even possible
- Clean the compile.rs file, improve features
- Add translation tests
- Post-PR note: add this comment into a doc file directly in the *test* folder

Test Rules: 
Tests supposed to fail (ie the compilation handles a failure properly) are meant to be prefixed with a `f_`. If test is supposed to succeed, prefix it with a `s_`. Any file that does not match this criteria shall be ignored.
To add a test create a file or add an entry in either the static array or the toml file.

https://issues.rudder.io/issues/16484


rudder-lang/tests/compile.rs
```diff
@@ -0,0 +1,115 @@
...
+use colored::Colorize;
+use test_case::test_case;
+
+// ======= Tests every file listed below, from the */compile* folder =======
+#[test_case("f_enum")]
+#[test_case("f_fuzzy")] // should_fail
+#[test_case("s_basic")]
+#[test_case("s_does_not_exist")] // should fail
+fn real_files(filename: &str) {
+    test_real_file(filename);
```
peckpeck: why not put the content of the function directly here ?

PR #2717, opened 2020-01-15
Fixes #16557: CachedFindRuleNodeStatusReports is a huge source of contention

fanf:
https://issues.rudder.io/issues/16557

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
@@ -57,7 +57,10 @@ import com.normation.utils.Control.sequence
 import com.normation.box._
 import com.normation.errors._
 import com.normation.rudder.domain.logger.ReportLogger
+import com.normation.rudder.domain.logger.ReportLoggerPure
+import zio._
 import com.normation.zio._
+import com.github.ghik.silencer.silent
 
 object ReportingServiceUtils {
 
@@ -94,6 +97,7 @@ class ReportingServiceImpl(
 class CachedReportingServiceImpl(
     val defaultFindRuleNodeStatusReports: ReportingServiceImpl
   , val nodeInfoService                 : NodeInfoService
+  , val batchSize                       : Int
 ) extends ReportingService with RuleOrNodeReportingServiceImpl with CachedFindRuleNodeStatusReports {
   val confExpectedRepo = defaultFindRuleNodeStatusReports.confExpectedRepo
   val directivesRepo   = defaultFindRuleNodeStatusReports.directivesRepo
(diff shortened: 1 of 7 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -251,38 +305,41 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+        _            <- invalidate(toUpdateIds).unit.toBox
       } yield {
-        //here, newStatus.keySet == expired.keySet, so we have processed all nodeIds that should be modified.
-        ReportLogger.debug(s"Compliance cache miss (updated):[${newStatus.keySet.map(_.value).mkString(" , ")}], " +
-                           s" hit:[${upToDate.map(_.value).mkString(" , ")}]")
-        cache = cache ++ newStatus
-        val toReturn = cache.filterKeys { id => nodeIds.contains(id) }
-        ReportLogger.trace("Compliance cache content: " + cacheToLog(toReturn))
+        ReportLogger.Cache.debug(s"Compliance cache to load (expired, missing):[${toUpdateIds.map(_.value).mkString(" , ")}]")
+        ReportLogger.Cache.trace("Compliance cache hit: " + cacheToLog(toReturn))
```
ncharles: We should at least advertise loudly that the returned map may not contain all the nodes from the method parameters, but only those in cache (expired or not).

fanf:
PR updated with a new commit

ncharles:
this breaks, as I never have the compliance for all nodes available

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -204,85 +250,100 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+                             case UnexpectedNoVersion(_, _, lastRunExpiration, _, _)
+                                                     => lastRunExpiration.isBefore(now)
+                             case _                  => false
+                           }
+                           !expired
+                         }
+        // starting with nodeIds, is all accepted node passed in parameter,
+        // we don't miss node ids not yet in cache
+        requireUpdate =  nodeIds -- upToDate.keySet
+        _             <- invalidate(requireUpdate).unit.toBox
```
ncharles: this will never return compliance for node with unexpected reports (like nodes with broken policies)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -204,85 +250,100 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
-        //here, newStatus.keySet == expired.keySet, so we have processed all nodeIds that should be modified.
-        ReportLogger.debug(s"Compliance cache miss (updated):[${newStatus.keySet.map(_.value).mkString(" , ")}], " +
-                           s" hit:[${upToDate.map(_.value).mkString(" , ")}]")
-        cache = cache ++ newStatus
-        val toReturn = cache.filterKeys { id => nodeIds.contains(id) }
-        ReportLogger.trace("Compliance cache content: " + cacheToLog(toReturn))
-        toReturn
+        ReportLogger.Cache.debug(s"Compliance cache to reload (expired, missing):[${requireUpdate.map(_.value).mkString(" , ")}]")
+        ReportLogger.Cache.trace("Compliance cache hit: " + cacheToLog(upToDate))
+        upToDate
```
ncharles: maybe we should return inCache

PR #2731, opened 2020-01-22
Fixes #16592: Error when creating a directive with a given id from the API

VinceMacBuche:
https://issues.rudder.io/issues/16592


VinceMacBuche:
Commit modified

PR #2743, opened 2020-01-27
Fixes #16629: Use only one ZIO threadpool to enhance performances

fanf:
https://issues.rudder.io/issues/16629

This change leads to a 20% improvement in Rudder policy generation time, from around 25 minutes to below 20 minutes. 

This change allowed to workaround three things described here: https://github.com/zio/zio/issues/1275#issuecomment-578734553

- 1/ switching from a thread to an other cost a bit, 
- 2/ switching on the same threadpool is idempotent but all the fixe costs, like fiber pause etc, are paid (while it should be `identity`)
- 3/ `effectBlocking` manage the discarding of the underlying thread, which we generally don't want. 

The last two are already corrected in ZIO-RC18! And the first one is not that bad anymore.
So, I'm not sure I we should commit that complicated PR or just wait for RC18 (which is stabilising and will likely be ZIO 1.0.0)

### DIFF
--- a/webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
+++ b/webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
@@ -39,6 +39,7 @@ import _root_.zio.syntax._
 import com.normation.zio.ZioRuntime
 import org.slf4j.Logger
 
+
 /**
  * This is our based error for Rudder. Any method that can
  * error should return that RudderError type to allow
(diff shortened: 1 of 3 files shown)


fanf:
We need to test ZIO-RC18 snapshot before merging that one.

fanf:
Following perf feedback from `RC18-2` (https://github.com/Normation/rudder/pull/2817) we don't need that one anymore.

PR #2745, opened 2020-01-29
Fixes #16643: Lots and lots of unecessary compliance computation when compliance cache is invalidate because of serialization when no serialization is asked

ncharles:
https://issues.rudder.io/issues/16643

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -924,11 +924,11 @@ object ExecutionBatch extends Loggable {
 
     TimingDebugLogger.trace(s"Compliance: mergeCompareByRule - compute buildRuleNodeStatusReport: ${t11-t10}ms")
 
-    val (computed, newStatus) = currentRunReports.foldLeft((nil, nil)) { case ( (c,n), currentStatusReports) =>
+    val (computed, newStatus) = currentRunReports.foldLeft((List[RuleNodeStatusReport](), List[RuleNodeStatusReport]())) { case ( (c,n), currentStatusReports) =>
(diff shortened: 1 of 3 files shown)


ncharles:
Memory impact seems crazy. But it looks like going to the rule page for the first time is painfully slow as well

ncharles:
actually, perfs looks ok, it's just the first instanciation

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -924,6 +924,27 @@ object ExecutionBatch extends Loggable {
...
+          val currentDirectives = currentStatusReports.directives
+
+          //don't keep directive that were removed between the two configs
+          val toKeep = runDirectives.filterKeys(k => currentDirectives.keySet.contains(k))
+
+          //now override currentDirective with the one to keep in currentReport
+          val updatedDirectives = currentDirectives ++ toKeep
+          val newCompliance = runStatusReport.copy(directives = updatedDirectives)
+
+          (newCompliance :: c, n)
```
ncharles: i'm a bit unsure about this change: I prepend in a list, so it should be better.

ncharles:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: you can't be better than prepending a list, so it seems safe

ncharles:
OK, squash merging this PR

ncharles:
OK, merging this PR

PR #2748, opened 2020-01-30
Fixes #16649: Prevent logger string interpolation when logger is not configured in the used level

ncharles:
https://issues.rudder.io/issues/16649

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
@@ -130,13 +130,16 @@ class CachedReportsExecutionRepository(
   }
 
   override def updateExecutions(executions : Seq[AgentRun]) : Seq[Box[AgentRun]] = this.synchronized {
-    logger.trace(s"Update runs for nodes [${executions.map( _.agentRunId.nodeId.value ).mkString(", ")}]")
+    if (logger.isTraceEnabled) {
+      logger.trace(s"Update runs for nodes [${executions.map(_.agentRunId.nodeId.value).mkString(", ")}]")
+    }
     val n1 = System.currentTimeMillis
     val runs = writeBackend.updateExecutions(executions)
     //update complete runs
     val completed = runs.collect { case Full(x) if(x.isCompleted) => x }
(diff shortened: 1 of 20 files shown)


ncharles:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
```diff
@@ -130,13 +130,16 @@ class CachedReportsExecutionRepository(
   }
 
   override def updateExecutions(executions : Seq[AgentRun]) : Seq[Box[AgentRun]] = this.synchronized {
-    logger.trace(s"Update runs for nodes [${executions.map( _.agentRunId.nodeId.value ).mkString(", ")}]")
+    if (logger.isTraceEnabled) {
+      logger.trace(s"Update runs for nodes [${executions.map(_.agentRunId.nodeId.value).mkString(", ")}]")
```
fanf: Are you sure it does anything? The parameter is marked by name, they should not be evaluated if they are not used?

ncharles: i'm not 100% sure; but isn't the map and mkstring evaluated anyway?

fanf: good question. Can you make a quick test with a println in the param?

fanf: No, nothing is evaluated is the log level is not reached: 
```
object TestLogger {

  import org.slf4j.LoggerFactory
  import net.liftweb.common.Logger
  object MyLogger extends Logger {
    override protected def _logger = LoggerFactory.getLogger("test")
  }
  org.slf4j.LoggerFactory.getLogger("test").asInstanceOf[ch.qos.logback.classic.Logger].setLevel(ch.qos.logback.classic.Level.DEBUG)

  def main(args: Array[String]): Unit = {
    println("start")
    MyLogger.debug(s"debug: ${println("I'm in debug"); "ok"}")
    MyLogger.trace(s"trace: ${println("I'm in trace"); "ok"}")
    println("done")
  }
}
```
=>
```
start
I'm in debug
00:39:36.382 [main]            DEBUG test - debug: ok
done
```

So I believe that PR is useless

ncharles:
ok, closing this PR then

PR #2761, opened 2020-02-06
Fixes #16692: Authorized networks are actually called allowed networks

fanf:
https://issues.rudder.io/issues/16692

- change "authorized networks" to "allowed networks" so that it matches UI/documentation, 
- if a policy server is missing allowed network policy, we want to return an empty list of networks in place of failing,
- if the user didn't specify "allowed_networks" param, don't erase all allowed networks but fail.


fanf:
wait, we have back the erasure of everything when we supply incorrect name

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #2764, opened 2020-02-11
Fixes #16712: We must allows to limit the upload API by configuration

fanf:
https://issues.rudder.io/issues/16712


fanf:
Actually, we decided that api token were sufficient

PR #2765, opened 2020-02-11
Fixes #16674: We should use BCrypt by default for storing user's local password

ElaadF:
https://issues.rudder.io/issues/16674

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
@@ -368,6 +368,16 @@ object RudderConfig extends Loggable {
     }
   }
 
+  val RUDDER_BCRYPT_COST = {
+    try {
+      config.getInt("rudder.bcrypt.cost")
+    } catch {
+      case ex: ConfigException =>
+        ApplicationLogger.debug("Property 'rudder.bcrypt.cost' is missing or empty in rudder.configFile. Default cost to 12.")
+        12
+    }
+  }
+
   val RUDDER_BATCH_PURGE_DELETED_INVENTORIES_INTERVAL = {
     try {
       config.getInt("rudder.batch.purge.inventories.delete.interval")
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
@@ -42,13 +42,9 @@ import java.io.FileInputStream
 import java.io.InputStream
 import java.nio.charset.StandardCharsets
 import java.security.MessageDigest
+import java.security.SecureRandom
 import java.util.Collection
-
-import com.normation.rudder.AuthorizationType
-import com.normation.rudder.Rights
-import com.normation.rudder.Role
-import com.normation.rudder.RoleToRights
-import com.normation.rudder.RudderAccount
+import com.normation.rudder._
 import com.normation.rudder.api._
 import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.domain.logger.PluginLogger
@@ -80,7 +76,10 @@ final case class UserFile(
 object PasswordEncoder {
   type Rudder = org.springframework.security.crypto.password.PasswordEncoder
 
+  import org.bouncycastle.crypto.generators.OpenBSDBCrypt
   import org.springframework.security.crypto.password.PasswordEncoder
+
+  val random = new SecureRandom()
 
   class DigestEncoder(digestName: String) extends PasswordEncoder {
     override def encode(rawPassword: CharSequence): String = {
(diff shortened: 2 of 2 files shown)


ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -368,6 +368,16 @@ object RudderConfig extends Loggable {
     }
   }
 
+  val RUDDER_BCRYPT_COST = {
+    try {
+      config.getInt("rudder.bcrypt.cost")
+    } catch {
+      case ex: ConfigException =>
+        ApplicationLogger.info("Property 'rudder.bcrypt.cost' is missing or empty in rudder.configFile. Default cost to 12.")
```
fanf: use debug here, because that information doesn't have any meaning safe if you use rudder-user plugin.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
```diff
@@ -41,9 +41,10 @@ import java.io.File
 import java.io.FileInputStream
 import java.io.InputStream
 import java.nio.charset.StandardCharsets
-import java.security.MessageDigest
+import java.security.{MessageDigest, SecureRandom}
```
fanf: prefer to use two lines, it's better for diff/linters/autoformating/etc

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
```diff
@@ -81,6 +84,8 @@ object PasswordEncoder {
   type Rudder = org.springframework.security.crypto.password.PasswordEncoder
 
   import org.springframework.security.crypto.password.PasswordEncoder
+  import org.springframework.security.crypto.bcrypt.BCrypt
```
fanf: I don't think you should have that import.

ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
```diff
@@ -100,10 +105,23 @@ object PasswordEncoder {
...
   }
+  // Unsalted hash functions :
   val MD5    = new DigestEncoder("MD5"    )
   val SHA1   = new DigestEncoder("SHA-1"  )
   val SHA256 = new DigestEncoder("SHA-256")
   val SHA512 = new DigestEncoder("SHA-512")
+  // Salted hash functions :
+  val BCRYPT = new PasswordEncoder() {
+    override def encode(rawPassword: CharSequence): String = {
+      val random = new SecureRandom()
```
fanf: you should have only one secure random, not one new for each call to `encode`.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
```diff
@@ -100,10 +105,23 @@ object PasswordEncoder {
...
+      val salt: Array[Byte] = new Array(16)
+      random.nextBytes(salt)
+      OpenBSDBCrypt.generate("2b", rawPassword.toString.toCharArray, salt, RudderConfig.RUDDER_BCRYPT_COST)
```
fanf: Please document the "2b" here, with a reference to `bcrypt` documentation and why we use that one.

ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetails.scala
```diff
@@ -37,21 +37,15 @@
 
 package bootstrap.liftweb
 
-import java.io.File
-import java.io.FileInputStream
-import java.io.InputStream
+import java.io.{File, FileInputStream, InputStream}
```
fanf: Soory, I mis-explained: it's the opposite that is needed: one import by line, ie we want: 

```
import java.io.File
import java.io.FileInputStream
import java.io.InputStream
```
Not: 
```
import java.io.{File, FileInputStream, InputStream}
```

fanf:
> In fact, I think we took a bad turn at some point. All the write part has nothing to do here and should only be in the user-management plugin. They are never used in Rudder and create a hard dependency in API between rudder base and the plugin. If we want to evolve the encoding part latter (for example to add new algo or whatever), we will be forced to modify rudder while it's the plugin responsibility to do that. Same for Rudder cost property actually.
> 
> So: I think you need to trim down that PR and only include the `BCRYT` password encode type withe the `matches` method.
> And in rudder user, add the logic for cost property, create new hashes, etc.

Looking a bit better to the code, we can't do that because spring implemented interface need the `encode` part, even if it's not used. 
So the work would be to get ride of `PasswordEncoder` enterily, but that's an other task with a very different cost. So we need to keep it that way (even if the coupling is bad).

ElaadF:
Commit modified

PR #2777, opened 2020-02-19
Fixes #16448: Trigger remote run in node details says it timeeout

fanf:
https://issues.rudder.io/issues/16448


amousset:
The scala test fails with:

amousset:
Looks like another timing issue when testing PRs

PR #2798, opened 2020-02-28
Fixes #16839: Restoring directive archive timeout and break LDAP

fanf:
https://issues.rudder.io/issues/16839


amousset:
@ncharles approves.

PR #2800, opened 2020-03-02
Fixes #16844: Changing one node policy mode change the policy mode for all nodes

fanf:
https://issues.rudder.io/issues/16844


fanf:
OK, merging this PR

PR #2808, opened 2020-03-04
Fixes #16866: Add relay API and webapp v12 to api doc

amousset:
https://issues.rudder.io/issues/16866


amousset:
OK, merging this PR

PR #2815, opened 2020-03-06
Fixes #16886: rudder-lang packaging update

gpoblon:
https://issues.rudder.io/issues/16886

no module dependencies left so rudder-lang should be easy to integrate now


amousset:
This needs a rebase

PR #2819, opened 2020-03-12
Fixes #16906: Be able to save in LDAP a Policy Server Target

ElaadF:
https://issues.rudder.io/issues/16906

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
@@ -151,6 +151,49 @@ class RudderDit(val BASE_DN:DN) extends AbstractDit {
   }
 
 
+//  def activeTechniqueModel(
+//                            uuid:String
+//                            , parentDN:DN
+//                            , techniqueName:TechniqueName
+//                            , acceptationDateTimes:JObject
+//                            , isEnabled:Boolean
+//                            , isSystem : Boolean
+//                          ) : LDAPEntry = {
+//    val mod = LDAPEntry(new DN(buildRDN(uuid),parentDN))
+//    mod +=! (A_OC, OC.objectClassNames(OC_ACTIVE_TECHNIQUE).toSeq:_*)
+//    mod +=! (A_TECHNIQUE_UUID, techniqueName.value)
+//    mod +=! (A_IS_ENABLED, isEnabled.toLDAPString)
+//    mod +=! (A_IS_SYSTEM, isSystem.toLDAPString)
+//    mod +=! (A_ACCEPTATION_DATETIME, compactRender(acceptationDateTimes))
+//    mod
+//  }
+//  def policyServerTarget(
+//      uuid : String
+//    , parentDN : DN
+//    , name :  String = ""
+//    , description : String = ""
+//  ) : LDAPEntry = {
+//  val mod = LDAPEntry(new DN(buildRDN(uuid),parentDN))
+//  mod +=! (A_OC, OC.objectClassNames(OC_SPECIAL_TARGET).toSeq:_*)
+//  mod +=! (A_NAME, name)
+//  mod +=! (A_DESCRIPTION, description)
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
```diff
@@ -150,6 +150,22 @@ class RudderDit(val BASE_DN:DN) extends AbstractDit {
     )
   }
 
+  def policyServerTarget(
+      uuid : String
+    , parentDN : DN
+    , name :  String = ""
+    , description : String = ""
+  ) : CATEGORY = {
```
fanf: I don't think it should be here. Category are for the branches in DIT, like "ou=nodes", etc. Here, IIUC, you define a new entry type for leaf

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
```diff
@@ -309,6 +325,40 @@ class RudderDit(val BASE_DN:DN) extends AbstractDit {
...
+      uuid = "rootRuleCategory"
+    , parentDN = BASE_DN
+    , name = "Rules"
+    , description = "This is the main category of Rules"
+    , isSystem = true
+    , objectClass = OC_SPECIAL_TARGET
+    , objectClassUuid = A_RULE_CATEGORY_UUID
+  ) {
+    ruleTargetServer =>
+
```
fanf: I don't think it's a category, just a new entry type. 
I think it should go under `SYSTEM` and use `SYSTEM` dn in its `ruleTargetDN`

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
```diff
@@ -748,6 +749,17 @@ class WoLDAPNodeGroupRepository(
     }
   }
 
+  override def createPolicyServerTarget(target : PolicyServerTarget, modId: ModificationId, actor:EventActor, reason:Option[String]): IOResult[LDIFChangeRecord] = {
+    for {
+      con    <- ldap
+      categoryEntry    <-  getCategoryEntry(con, ???).notOptional(s"Entry with ID '${id}' was not found")
```
ElaadF: Here I don't know how to get the parentDN needed in `ruleTargetModel`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
```diff
@@ -309,6 +352,32 @@ class RudderDit(val BASE_DN:DN) extends AbstractDit {
     }
   }
 
+  object RULETARGET {
+
+      def ruleTargetDN(targetId:String) : RDN = new RDN(A_ACTIVE_TECHNIQUE_UUID, targetId)
```
ElaadF: What should I put instead of `A_ACTIVE_TECHNIQUE_UUID`

fanf: `RDN` is the attribute used as identifier for entry. So for node, it's `nodeid` (`A_NODE_UUID`). 
Ldap entries are located by their full name, the path from root of tree to them, that we call a distinguished name (dn). 
rdn is the last part of the path, so that `entry DN = entry RDN, parent DN` (recursively until root)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
fanf: on which branch of LDAP directory does it go?

PR #2826, opened 2020-03-18
Fixes #16929: Git configuration-repository object is created at each evaluation, impacting performance

fanf:
https://issues.rudder.io/issues/16929

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/GitUtils.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/GitUtils.scala
@@ -51,9 +51,9 @@ trait GitRepositoryProvider {
   /**
    * Obtain access to JGit porcelain API.
    */
-  def git: IOResult[Git]
+  def git: Git
 
-  def db: IOResult[Repository]
+  def db: Repository
 }
 
 object GitRepositoryLogger extends NamedZioLogger() { def loggerName = "git-repository" }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitUtilsImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitUtilsImpl.scala
@@ -61,43 +61,48 @@ import com.normation.zio._
  * If a .git repository is found in it, it is considered as the repository
  * to use, else if none is found, one is created.
  */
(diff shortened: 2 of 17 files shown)


fanf:
PR rebased

webapp/sources/rudder/rudder-rest/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,85 @@
+<configuration>
```
fanf: This file is needed to avoid millions of logs lines during tests of rudder rest about `jgit`

webapp/sources/utils/src/test/scala/com/normation/ZioCommonsTest.scala
```diff
@@ -162,6 +162,18 @@ object TestImplicits {
   }
 }
 
+object SimpleEvalTest {
+
+  val hello = IOResult.effect(println("plop"))
+
```
fanf: A reminder that effect are executed at each call of `runNow`. Really really. Even if it's a val.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/GitUtils.scala
```diff
@@ -51,9 +51,9 @@ trait GitRepositoryProvider {
   /**
    * Obtain access to JGit porcelain API.
    */
-  def git: IOResult[Git]
+  def git: Git
```
fanf: This is the important part of the change. 
We don't want to create a Git object each time, just have it ready. The creation should only happen one time at boot.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitUtilsImpl.scala
```diff
@@ -61,43 +61,48 @@ import com.normation.zio._
...
-    IOResult.effect {
-      val db = (new FileRepositoryBuilder().setWorkTree(root).build).asInstanceOf[FileRepository]
-      if(!db.getConfig.getFile.exists) {
-        GitRepositoryLogger.logEffect.info(s"Git directory was not initialised: create a new git repository into folder '${root.getAbsolutePath}' and add all its content as initial release")
-        db.create()
-        val git = new Git(db)
-        git.add.addFilepattern(".").call
-        git.commit.setMessage("initial commit").call
+  def make(gitRootPath: String): IOResult[GitRepositoryProviderImpl] = {
+    /**
```
fanf: This is the other interesting change, where the creation of git repo is effectfull, but once we have it, we just give it to `new GitRepositoryProviderImpl()`

PR #2828, opened 2020-03-19
Fixes #16931: By default we should not archive reports

ncharles:
https://issues.rudder.io/issues/16931

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
@@ -137,7 +137,7 @@ rudder.jdbc.batch.max.size=500
 # Defaults: archive disabled, delete after 4 days.
 #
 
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -137,7 +137,7 @@ rudder.jdbc.batch.max.size=500
 # Defaults: archive disabled, delete after 4 days.
 #
 
-rudder.batch.reportscleaner.archive.TTL=4
+rudder.batch.reportscleaner.archive.TTL=0
```
fanf: I think we should keep that one, so that if someone set "archive reports", it doesn't blow up too much

fanf: OK, so no: it's the time for archiving, so we really want it to be 0 to disable archiving.

PR #2852, opened 2020-04-01
Fixes #16898: Hide remote run button for Windows agent

fanf:
https://issues.rudder.io/issues/16898


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
```diff
@@ -783,10 +791,16 @@ class NodeApiService8 (
...
+      val okNodes = nodes.toList.flatMap {
+        case (_, node) if(node.agentsName.exists(a => a.agentType == AgentType.CfeEnterprise || a.agentType == AgentType.CfeCommunity)) =>
+          Some(node)
+        case _ => None
+      }
 
       val res =
       for {
-        node <- nodes.values.toList
+        node <- okNodes
```
VinceMacBuche, line 803: I would have kept the whole list here but instead  in the result for non compatible nodes i would have changed the result ro "Skipped, agent is not compatible with remote-run" or just "skipped"

fanf: good idea

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #2853, opened 2020-04-01
Fixes #16902: technique editor test loop integration

gpoblon:
https://issues.rudder.io/issues/16902


amousset:
Tests are not passing.

peckpeck:
ok, for me, but I need a second thought from @VinceMacBuche

amousset:
You have some `.pyc` files, they should be ignore by gitignore (and removed from the PR)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -248,6 +249,7 @@ class TechniqueWriter (
       libUpdate  <- techLibUpdate.update(modId, committer, Some(s"Update Technique library after creating files for ncf Technique ${technique.name}")).
                       toIO.chainError(s"An error occured during technique update after files were created for ncf Technique ${technique.name}")
      } yield {
+      Seq("/opt/rudder/share/rudder-lang/tools/tester.sh", technique.name).run
```
VinceMacBuche, line 252: You should not run a command with sys.process. 

It can throw exception and it will break the beahvior of Rudder here

You want to have something like this 

```
RunNuCommand.run(Cmd("/opt/rudder/share/rudder-lang/tools/tester.sh", technique.name :: Nil, Map())).either match {
  case Left(error: Error) => ApplicationLogger.error(error.getMessage)
  case Right(_)  => ApplicationLogger.debug(s"rudder lang tester called for technique ${technique.bundleName.value}")
}

Comment (VinceMacBuche) [reply webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala:252]: On a side note, don't you want the technique id/bundle_name instead ? here you are using the technique human name which is not the technique identifier!!

Comment (gpoblon) [reply webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala:252]: I'll look into it, thank you
More fixes are coming along

Well, I need the `technique.name` because it allows me to get the technique path of the created file. 
But maybe I could use the technique id for log purposes

**Comment by peckpeck** [comment]
rebase needed

File: rudder-lang/tools/cfjson_tester [hunk 2]
```diff
@@ -133,54 +161,35 @@ def remove_wilful_entries(diff):
...
     exit(1)
 
   cfg = configparser.ConfigParser()
-  cfg.read(cfg_full_path)
+  cfg.read(cfgpath)
   return cfg
 
 if __name__ == "__main__":
-  arguments = docopt.docopt(__doc__)
+  import docopt
```
peckpeck, line 176: if it is not installed this will fail, i think you meant to put this in the try catch

peckpeck: you can make this work everywhere by just adding rudder python path all the time
 sys.path.append("/opt/rudder/share/python/")

gpoblon: because we always have docopt installed ?

gpoblon:
Still have scala changes to do, do not merge

rudder-lang/tools/cfjson_tester
gpoblon: Ok fixed. Still kept the try execpt as dev env might not have `/opt/rudder/share/python/docopt`

peckpeck: you don't need it, you may want to put a try around the import itself to provide an error message to the developper, but the developper should already know how to fix this.
The rest of your try/except is just duplicate code

gpoblon: fixed

rudder-lang/Makefile
```diff
@@ -14,6 +14,10 @@ install:
 	mkdir -p $(DESTDIR)/opt/rudder/share/python
 	mkdir -p $(DESTDIR)/opt/rudder/share/rudder-lang/lib
 	mkdir -p $(DESTDIR)/opt/rudder/share/rudder-lang/tools
+	mkdir -p $(DESTDIR)/var/log/rudder/rudder-lang/
+	
+	# allow for non sudo users to write rudder-lang logs
+	chmod 755 $(DESTDIR)/var/log/rudder/rudder-lang
```
amousset, line 20: these logs should only be readable by the right/group. They may contain configuration and it is sensitive data.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/reports/StatusReportTest.scala
```diff
@@ -290,7 +290,7 @@ class StatusReportTest extends Specification {
         }
         val t1 = System.nanoTime
         println(s"Time to run test is ${(t1-t0)/1000} µs")
-        (t1-t0) must be lessThan( 200000*1000 ) // tests show 60030µs
+        (t1-t0) must be lessThan( 900000*1000 ) // tests show 60030µs
```
VinceMacBuche, line 293: You should not change this value !!

gpoblon: fixed

peckpeck:
OK, squash merging this PR

peckpeck:
fixup were broken, the rebase ddn't happen on this branch so it is not detceted as merged

PR #2855, opened 2020-04-01
Fixes #17019: Policy generation must fail when report is HTTPS only and old agent are present

fanf:
https://issues.rudder.io/issues/17019

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -228,7 +228,7 @@ trait PromiseGenerationService {
       fetch0Time           =  System.currentTimeMillis
       (updatedNodesId, updatedNodeInfo, expectedReports, allErrors, errorNodes, timeFetchAll, timeHistorize, timeRuleVal, timeBuildConfig, timeWriteNodeConfig, timeSetExpectedReport) <- for {
         (updatedNodeConfigIds, allNodeConfigurations, allNodeConfigsInfos, updatedNodesId, updatedNodeInfo, globalPolicyMode, allNodeModes, allErrors, errorNodes, timeFetchAll, timeHistorize, timeRuleVal, timeBuildConfig) <- for {
-          (activeNodeIds, ruleVals, nodeContexts, allNodeModes, scriptEngineEnabled, globalPolicyMode, nodeConfigCaches, timeFetchAll, timeHistorize, timeRuleVal) <- for {
+          (activeNodeIds, ruleVals, nodeContexts, allNodeModes, scriptEngineEnabled, globalPolicyMode, nodeConfigCaches, timeFetchAll, timeHistorize, timeRuleVal, errors) <- for {
              allRules             <- findDependantRules() ?~! "Could not find dependant rules"
              fetch1Time           =  System.currentTimeMillis
              _                    =  PolicyGenerationLogger.timing.trace(s"Fetched rules in ${fetch1Time-fetch0Time} ms")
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -789,7 +796,7 @@ trait PromiseGeneration_BuildNodeContext {
     , globalAgentRun        : AgentRunInterval
     , globalComplianceMode  : ComplianceMode
     , globalPolicyMode      : GlobalPolicyMode
-  ): Box[Map[NodeId, InterpolationContext]] = {
+  ): Box[NodesContextResult] = {
```
fanf, line 799: Here, the semantic was to not stop on error, and just print a log. I'm not sure why we did that without showing the error also in generation status, because with that choice, someone can't know that a node is node updated without looking at logs, which is almost certainly never done. 
In our case, the goal is to show that there is a problem to user, so I choose to now display these error in generation status, ie: the same semantic is kept (these errors don't stop generation, as it was before that change) but error are shown in status at the end in place of success.

VinceMacBuche:
Why do you change the permissions of hooks ?

fanf:
because I didn't run the test before pushing

fanf:
PR updated with a new commit

VinceMacBuche:
tests are failling because agent version in tests is 4.0.0 and we are in HTTPS only :)

fanf:
So, what should do? Update agent? Change https only to mixed mode? @amousset @Fdall at least is seems to work :rofl:

VinceMacBuche:
Dunno, would say update agent version from tests ;)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -530,18 +532,47 @@ class SystemVariableServiceImpl(
...
+          case _              => getReportProtocolDefault()
+        }
+      }.flatMap{ reportingProtocol => // If HTTPS is used on a node that does support it, we fails.
+        // as we don't have capalities and version is not reliable, we only fail when we are sure:
+        // - version provided and starts by 2.x, 3.x, 4.x, 5.x
+        // - agent is DSC
+        // Also, special case root, because not having root cause strange things.
+        val badVersion = nodeInfo.agentsName.find { agent => !versionHasHttps(agent.version) }
+        val dsc = nodeInfo.agentsName.find { agent => agent.agentType == AgentType.Dsc }
+        if(nodeInfo.id != Constants.ROOT_POLICY_SERVER_ID && (badVersion.nonEmpty || dsc.nonEmpty)) {
```
ncharles, line 563: shouldn't it fail only if HTTPS ?

fanf: Hm. There was a check `reportingProtocol == AgentReportingHTTPS` but I beleive it was erased when I put appart version/dsc for error reporting.

fanf: Never, ever change something after testing, even something that seems trivial :)

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -530,17 +533,50 @@ class SystemVariableServiceImpl(
...
+            true // we don't know, so say HTTPS is supported
+          case Some(v) =>
+            val forbiden = List("2.", "3.", "4.", "5.")
+            !forbiden.exists(x => v.value.startsWith(x))
+        }
+      }
+
+      def failure(nodeInfo: NodeInfo, badVersion: Option[AgentInfo], dsc: Option[AgentInfo]) = {
+        Failure(s"Node ${nodeInfo.hostname} (${nodeInfo.id.value}) has an agent which doesn't support sending compliance reports in HTTPS: ${
+                  dsc match {
```
ncharles, line 553: we should not blindly refuse dsc - the code that decides if it can send https reports or not is in the plugin - so we should at least limit to version known not to support HTTPS

fanf: no version in 6.0 will suport it, and it's expected that all version in 6.1 support it. So I propose to let it like that for 6.0, and change it in 6.1 (but once we know it works) or think how to limit it at that time, knowing that the only correct solution is out os reach in 6.0 (because capabilities are not parsed).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -530,17 +533,50 @@ class SystemVariableServiceImpl(
...
-      case eb:EmptyBox =>
-        val e = eb ?~! s"Failed to get information on syslog protocol global status, fallbacking to default value."
-        logger.error( e.messageChain )
-        getProp("REPORTING_PROTOCOL", () => getReportProtocolDefault().map(_.value))
+    // If Syslog is disabled, we force HTTPS.
+    val varNodeReportingProtocol = {
+      // By default, we are tolerant: it's far worse to fail a generation while support is ok than to
+      // succeed and have a node that doesn't rerport.
+      // => version must exists and and starts by 2.x, 3.x, 4.x, 5.x
+      def versionHasHttps(maybeVersion: Option[AgentVersion]): Boolean = {
```
VinceMacBuche, line 541: You could reverse the logic since you are doing looking for a False in badVersion
You could instead do a versionHasSyslogOnly
You could Put the None case to be false
and remove the not in the Some case

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -530,17 +533,50 @@ class SystemVariableServiceImpl(
...
+                    case None => s"agent version ${badVersion.flatMap(_.version.map(_.value)).getOrElse("")} is too old"
+                    case _    => s"Windows agent has only syslog reporting"
+                  }
+                }. You need to either disable that node (in node details > settings > Node State) or use HTTPS + syslog (in " +
+                s"Settings > General > Reporting protocol)")
+      }
+      // as we don't have capalities and version is not reliable, we only fail when we are sure:
+      // - version provided and starts by 2.x, 3.x, 4.x, 5.x
+      // - agent is DSC
+      val badVersion = nodeInfo.agentsName.find { agent => !versionHasHttps(agent.version) }
```
VinceMacBuche, line 563: If you switch to versionHasSyslogOnly, you can remove the not here

VinceMacBuche:
apart this it's ok to me

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #2859, opened 2020-04-04
Fixes #17046: Replace SourceList with a typed arena

peckpeck:
https://issues.rudder.io/issues/17046

### DIFF
--- a/rudder-lang/src/compile.rs
+++ b/rudder-lang/src/compile.rs
@@ -8,51 +8,40 @@ use crate::{
     parser::{Token, PAST},
 };
 
+use typed_arena::Arena;
 use colored::Colorize;
-use std::{cell::UnsafeCell, fs, path::Path, pin::Pin};
+use std::{fs, path::Path};
 
-/// Read file, parse it and store it
-fn add_file<'a>(
-    past: &mut PAST<'a>,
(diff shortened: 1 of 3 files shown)


rudder-lang/src/compile.rs
```diff
@@ -8,82 +8,64 @@ use crate::{
...
-    path: &'a Path,
-    filename: &'a str,
-) -> Result<()> {
+/// Parses the whole stdlib
+pub fn parse_stdlib<'src>(past: &mut PAST<'src>, sources: &'src Arena<String>, libs_dir: &'src Path) -> Result<()> {
+    let core_lib = libs_dir.join("corelib.rl");
+    parse_file(past, sources, &core_lib, "corelib.rl")?;
+
+    let oses = libs_dir.join("oslib.rl");
+    parse_file(past, sources, &oses, "oslib.rl")?;
```
amousset, line 21: Does the compiler really have to know about stdlib's content? Could it load all `.st` file?

peckpeck: it could, its just that we did not do that until now

amousset: Ok, I'll dot it then :)

PR #2864, opened 2020-04-05
Fixes #16974: Full review of rudder-lang suported systems

peckpeck:
https://issues.rudder.io/issues/16974

### DIFF
--- a/rudder-lang/libs/oslib.rl
+++ b/rudder-lang/libs/oslib.rl
@@ -1,18 +1,35 @@
 @format=0
 
-# do not extend system (or so ?)
+# @cfengine_name in list format only are useful for translate
+# remove them once we don't need it anymore
+
 global enum system {
-  unix,
-  linux,
   windows
+  linux,
+  aix,
+  bsd,
+  hp_ux,
+  solaris,
+  *
 }
 
+# == Windows
 items in windows {
-  windows_XP,
-  windows_Vista,
+  windows_desktop,
+  windows_server,
+  *
+}
+
+items in windows_desktop {
+  windows_xp,
+  windows_vista,
   windows_7,
   windows_8,
   windows_10,
+  *
+}
+
+items in windows_server {
   windows_2000,
   windows_2003,
   windows_2008,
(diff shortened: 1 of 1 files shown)


peckpeck:
Do not merge before we discuss and solve all the TODO in this file

peckpeck:
Commit modified

rudder-lang/libs/oslib.rl
```diff
@@ -39,16 +65,17 @@ items in bsd {
 }
 
 items in linux {
-  alpine_linux,
-  arch_linux,
+# TODO do we keep android ?
```
amousset, line 68: I don't think it really matters as it is not a list of supported OS. I'm fine with both.

amousset:
We probably need a SUSE fmaily with SLES and OpenSUSE

rudder-lang/libs/oslib.rl
```diff
@@ -88,13 +111,275 @@ items in solaris {
...
+items in redhat_8 {
+  redhat_8_0,
+  redhat_8_1,
+  *
+}
+
+# == redhat: cfengine + wikipedia checked
+
+# == ubuntu
+# TODO Should we have major only for ubuntu ?
```
amousset, line 258: 18_04 is not major+minor but a major version by itself. There is no particular link between 18_04 and 18_10, so I don't think so.

rudder-lang/libs/oslib.rl
```diff
@@ -30,6 +30,32 @@ items in unix {
...
 
+# == AIX
+items in aix {
+  aix_5,
+  aix_6,
+  aix_7,
+  *
+}
+
+# TODO should we make it extendable ?
```
peckpeck, line 41: The general question, is for all OS, should we add a * (extensible list) in os subversion, given that there won't be anymore such version. 
I think we should not have it

rudder-lang/libs/oslib.rl
peckpeck: let's go with a no, since this list is extensible

rudder-lang/libs/oslib.rl
peckpeck: I agree

rudder-lang/libs/oslib.rl
```diff
@@ -88,13 +111,275 @@ items in solaris {
...
 }
+
+enum alias lenny    = debian_5
+enum alias squeeze  = debian_6
+enum alias wheezy   = debian_7
+enum alias jessie   = debian_8
+enum alias stretch  = debian_9
+enum alias buster   = debian_10
+enum alias bullseye = debian_11
+# TODO cfengine doesn't have proper minor version for debian 5 and 6
```
peckpeck, line 328: The question is, if we add them here, how can we implement them ?

peckpeck:
PR updated with a new commit

peckpeck:
PR updated with a new commit

peckpeck:
PR updated with a new commit
I think it's mergreable now

peckpeck:
PR updated with a new commit

PR #2865, opened 2020-04-06
Fixes #16552: Webdav password is ignored and access is granted for all nodes in allowed networks

amousset:
https://issues.rudder.io/issues/16552

### DIFF
--- a/relay/sources/apache/rudder-apache-relay-ssl.conf
+++ b/relay/sources/apache/rudder-apache-relay-ssl.conf
@@ -44,10 +44,14 @@ Alias /inventories /var/rudder/inventories/incoming
   AuthName "WebDAV Storage"
   AuthType Basic
   AuthUserFile /opt/rudder/etc/htpasswd-webdav-initial
-  Require valid-user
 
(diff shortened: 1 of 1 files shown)


relay/sources/apache/rudder-apache-relay-ssl.conf
```diff
@@ -44,10 +44,14 @@ Alias /inventories /var/rudder/inventories/incoming
   AuthName "WebDAV Storage"
   AuthType Basic
   AuthUserFile /opt/rudder/etc/htpasswd-webdav-initial
-  Require valid-user
 
-  # rudder-networks-24.conf is automatically generated according to the hosts allowed by rudder.
-  Include /opt/rudder/etc/rudder-networks-24.conf
+  <RequireAll>
+    Require valid-user
+    <RequireAny>
```
peckpeck, line 50: why any with only one entry ?

amousset: because the included file contains a list of require

peckpeck: oh yes

PR #2868, opened 2020-04-06
Fixes #17062: technique Resources are not provided to Windows agent

VinceMacBuche:
https://issues.rudder.io/issues/17062

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -603,6 +603,7 @@ class DSCTechniqueWriter(
             |  )
             |
             |  $$local_classes = New-ClassContext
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -636,6 +637,15 @@ class DSCTechniqueWriter(
         <FILE name={s"RUDDER_CONFIGURATION_REPOSITORY/${computeTechniqueFilePath(technique)}"}>
           <INCLUDED>true</INCLUDED>
         </FILE>
+        { for {
```
fanf, line 640: cn you put the open `{` on previoux line? Sometimes, when we open it on a new line, scalac parser just ignore the code produced. It can depend on scala version, which makes it hard to always notice. Having it on the same line prevent that. 
And in miror, put the last `}` just before `<FILES>.

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
OK, squash merging this PR

PR #2869, opened 2020-04-06
Fixes #16258: Folded categories in the Directive tree should be kept between sessions

RaphaelGauthier:
https://issues.rudder.io/issues/16258


RaphaelGauthier:
Commit modified

PR #2872, opened 2020-04-06
Fixes #17070: Display a warning in place of login form if no user are defined

fanf:
https://issues.rudder.io/issues/17070


fanf:
This need some css love, perhaps @RaphaelGauthier could you take over that one?

PR #2876, opened 2020-04-07
Fixes #17075: translation into rudder-lang doesn't handle any class

peckpeck:
https://issues.rudder.io/issues/17075

### DIFF
--- a/rudder-lang/src/translate.rs
+++ b/rudder-lang/src/translate.rs
@@ -60,12 +60,12 @@ pub fn translate_file(json_file: &Path, rl_file: &Path, config_filename: &Path) 
 
     info!(
         "|- {} {}",
-        "Serializating".bright_green(),
+        "Reading".bright_green(),
         config_filename.bright_yellow()
     );
     let config_data =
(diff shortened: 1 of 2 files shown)


peckpeck:
It also contains a litte rewrite to always have a translation context and some bugfix in oslib

rudder-lang/src/translate.rs
```diff
@@ -148,115 +124,231 @@ resource {bundle_name}({parameters})
...
-    let method = match mconf.get(&call.method_name) {
-        None => {
-            return Err(Error::User(format!(
-                "Unknown generic method call: {}",
-                &call.method_name
-            )))
+    fn translate_condition(&self, expr: &str) -> Result<String> {
+        lazy_static! {
+            static ref CLASS_RE: Regex = Regex::new(r"([\w${}.]+)").unwrap();
+            static ref ANY_RE: Regex = Regex::new(r"(any\.)").unwrap();
```
peckpeck, line 280: The interesting part is this, and the replace below

PR #2877, opened 2020-04-08
Fixes #17079: Test files are not correctly clean

fanf:
https://issues.rudder.io/issues/17079

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/reports/StatusReportTest.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/reports/StatusReportTest.scala
@@ -37,13 +37,13 @@
 
 package com.normation.rudder.domain.reports
 
+import com.github.ghik.silencer.silent
+
 import scala.collection.TraversableOnce.flattenTraversableOnce
 import scala.io.Source
-
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.policies.DirectiveId
 import com.normation.rudder.domain.policies.RuleId
-
 import org.junit.runner.RunWith
 import org.specs2.mutable.Specification
 import org.specs2.runner.JUnitRunner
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/reports/StatusReportTest.scala
```diff
@@ -297,13 +295,10 @@ class StatusReportTest extends Specification {
 
       "run fast enough to sum" in {
         ComplianceLevel.sum(initData.map( x => ComplianceLevel.compute(x._2)))
-
-        val source = initData.map(x => (x._1, ComplianceLevel.compute(x._2))).map( _._2)
         val t0 = System.nanoTime
 
         for (i <- 1 to 100) {
           val t0_0 = System.nanoTime
-          val result = ComplianceLevel.sum(source)
```
ncharles, line 306: these should not be removed. They are used to warm the JVM, plus are not at all related to this ticket

VinceMacBuche: Those variable are unused, and really look like dead code

fanf: @ncharles if they are useful, we need a way to mark them as that. I regularly remove things that lead to warning or looks like dead code. At minima, they need to have: 
- a comment to explain why they are here,
- a silent annotation to remove warning. 
Rudder should build with 0 warning

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/hooks/HooksTest.scala
```diff
@@ -74,7 +75,9 @@ class HooksTest() extends Specification with AfterAll {
   }
 
   override def afterAll(): Unit = {
-   // tmp.delete()
+    if(System.getProperty("tests.clean.tmp") != "false") {
+      File("/tmp/rudder-test-hook/").delete()
```
VinceMacBuche, line 79: In think @amousset will be very happy (test will now be cleaned up ....)

fanf:
OK, squash merging this PR

PR #2883, opened 2020-04-10
Fixes #17100: add a method to find all activeTechnique based on the Technique name

ncharles:
https://issues.rudder.io/issues/17100


ncharles:
PR updated with a new commit

fanf:
I think we don't need at all these modificationsn and that PR should be closed.

fanf:
Closing this PR (cf above comments)

PR #2890, opened 2020-04-13
Fixes #16347: Group properties

fanf:
https://issues.rudder.io/issues/16347
First working version of group properties. 
Works:
- inheritance of subgroups is undertood, 
- if a property is defined on two unrelated groups, it's an error (appart if they are exactly the same prop)
- group properties can be added by API
- saved in Ldap
- event log and save in XML
- no UI 
- when a value is a json object overriden by an other json object, json is merged.

Does not work:
- global parameters follows same rules

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
@@ -37,28 +37,28 @@
 
 package com.normation.rudder.domain.nodes
 
+import com.normation.errors.Inconsistency
+import com.normation.errors.PureResult
+import com.normation.inventory.domain.FullInventory
+import com.normation.inventory.domain.KeyStatus
 import com.normation.inventory.domain.NodeId
+import com.normation.inventory.domain.PublicKey
+import com.normation.inventory.domain.SecurityToken
 import com.normation.rudder.domain.policies.PolicyMode
+import com.normation.rudder.domain.policies.SimpleDiff
 import com.normation.rudder.reports.AgentRunInterval
 import com.normation.rudder.reports.HeartbeatConfiguration
 import com.normation.rudder.reports.ReportingConfiguration
-import com.normation.utils.Control.sequence
-import com.normation.rudder.services.policies.ParameterEntry
-import org.joda.time.DateTime
-import com.normation.rudder.domain.policies.SimpleDiff
-import com.normation.inventory.domain.FullInventory
-import com.normation.inventory.domain.KeyStatus
-import com.normation.inventory.domain.PublicKey
-import com.normation.inventory.domain.SecurityToken
-import net.liftweb.json.JsonAST.JValue
-import net.liftweb.json.JsonAST.JString
-import net.liftweb.json.JsonParser.ParseException
-import net.liftweb.common.Box
-import net.liftweb.common.Full
-import net.liftweb.common.Failure
-import com.normation.rudder.repository.json.DataExtractor.OptionnalJson
 import com.normation.rudder.repository.json.DataExtractor.CompleteJson
+import com.normation.rudder.repository.json.DataExtractor.OptionnalJson
+import com.normation.rudder.services.policies.ParameterEntry
+import net.liftweb.common.Box
+import net.liftweb.common.Failure
+import net.liftweb.common.Full
 import net.liftweb.http.S
+import net.liftweb.json.JsonAST._
+import net.liftweb.json.JsonParser.ParseException
+import org.joda.time.DateTime
 
 /**
  * The entry point for a REGISTERED node in Rudder.
@@ -75,11 +75,11 @@ final case class Node(
   , isPolicyServer            : Boolean
   , creationDate              : DateTime
   , nodeReportingConfiguration: ReportingConfiguration
-  , properties                : Seq[NodeProperty]
+  , properties                : List[NodeProperty]
   , policyMode                : Option[PolicyMode]
 )
 
-final case object Node {
+case object Node {
   def apply (inventory : FullInventory) : Node = {
     Node(
         inventory.node.main.id
@@ -90,14 +90,14 @@ final case object Node {
       , false
       , inventory.node.inventoryDate.getOrElse(new DateTime(0))
       , ReportingConfiguration(None,None, None)
-      , Seq()
+      , Nil
       , None
     )
   }
 }
 
 sealed trait NodeState { def name: String }
-final object NodeState {
+object NodeState {
 
   final case object Enabled          extends NodeState { val name = "enabled"       }
   final case object Ignored          extends NodeState { val name = "ignored"       }
@@ -133,14 +133,10 @@ final case class NodePropertyProvider(value: String) extends AnyVal
 
 /**
  * A node property is a key/value pair + metadata.
- * For now, only metadata availables are:
+ * For now, only metadata availables is:
  * - the provider of the property. By default Rudder.
- * - the mode: read-write (default) or read-only.
- *   (TODO: add mode="none" ?)
  *
- * If mode is read-only, only the provider can modify
- * the properties. This is typically what happens
- * when the property is set by an external source.
+ * Only the provider of a property can modify it.
  */
 final case class NodeProperty(
     name    : String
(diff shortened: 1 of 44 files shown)


VinceMacBuche:
Hooks permissions were modified ;)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Node.scala
```diff
@@ -224,30 +230,30 @@ object CompareProperties {
...
+                           Right(updateOrRemoveProp(newProp))
                          case Some(oldProp@NodeProperty(name, value, provider)) =>
                              if(canBeUpdated(old = provider, newer = newProp.provider)) {
-                               Full(updateOrRemoveProp(newProp))
+                               Right(updateOrRemoveProp(newProp))
                              } else {
                                val old = provider.getOrElse(NodeProperty.rudderNodePropertyProvider).value
                                val current = newProp.provider.getOrElse(NodeProperty.rudderNodePropertyProvider).value
-                               Failure(s"You can not update property '${name}' which is owned by provider '${old}' thanks to provider '${current}'")
+                               Left(Inconsistency(s"You can not update property '${name}' which is owned by provider '${old}' thanks to provider '${current}'"))
```
VinceMacBuche, line 249: Inconsistency here is an inventory error 

Don't you want to use the Inconstancy from com.normation.erros ? 

Maybe We should fi the typo in the class in master/6.1 ...

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeGroup.scala
```diff
@@ -37,15 +37,165 @@
...
  * UUId type for Node Groups, so that they
  * can be uniquely identified in our world.
  */
 final case class NodeGroupId(value:String) extends AnyVal
 
+final case class GroupProperty(
+    name    : String
+  , value   : JValue
+) {
+  def renderValue: String = value match {
```
VinceMacBuche, line 68: Maybe this could be in GenericPropertyUtils

Or having a trait for Properties having this type and then a function in this trait or an implicit class providing this function

fanf: I thought a bit to that, but I went to `GroupProperties` because:
- `NodeProperties` are reasonably different, both in code (because of provider) and in semantic (because the overriding is totally different, they are the only one that matters in the end: they are the only one to go to node
- I'm not sure that if we add other properties, they will be the same.

But in current way of doing things, it's rather bad. We have double serialisation for each of them, which mean a change to define incompatible format and make impossible sharing things like the property editor. 
So I think it should be done, not sure how (perhaps a `property` which is just a `name` + `value`, and a `case class NodeProperty(property: Property, provider: ...)` and smart constructor/extractors...)

VinceMacBuche: I think having an inner Property with common behavior for all properties would be interesting, but maybe we can keep it for another PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/MergeNodeProperties.scala
```diff
@@ -0,0 +1,389 @@
...
+ * - on a group by an other group, if the second group is a subgroup of the first,
+ *   i.e if he has a "AND" query composition, same nature (static/dynamic), and
+ *   a `SubGroupComparator` criterion with value the parent groupid.
+ * - on a node, if provider if the same.
+ *
+ * When overriding happens, we use a merge-override strategy, ie:
+ * - if the overriding value is a simple value or an array, it replaces the previous one
+ * - if the overriding value is an object and the overriden value is a simple type or an array,
+ *   the latter one is replaced by the former
+ * - if boeth overriding/overriden are an array, overriding values are appended to overriden array
```
VinceMacBuche, line 72: not sure here if we should only keep the overrding values

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/MergeNodeProperties.scala
```diff
@@ -0,0 +1,389 @@
...
+ *
+ * When overriding happens, we use a merge-override strategy, ie:
+ * - if the overriding value is a simple value or an array, it replaces the previous one
+ * - if the overriding value is an object and the overriden value is a simple type or an array,
+ *   the latter one is replaced by the former
+ * - if boeth overriding/overriden are arrays, overriding values are appended to overriden array
+ * - if both overriding/overriden are objects, then each key is overriden recursively as explained,
+ *   and new keys are added.
+ */
+
```
VinceMacBuche, line 76: To sum up,

if overriding is:
* a value => overrides everything
* an array => overrides value and object, append value to overriden array
* an object => overrides value and array but if object :
** if key in both object : merge recursive on the values of theses keys
** keep keys that are only in one of the two object (either overriding or overriden)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/NodeGroup.scala
fanf: Yes, we can. And I'm not sure we can makes much more than what is in `GenericPropertyUtils`, because for three cases (global param, node prop, group prop), we have 3 differents way of serializing (totally different for param, and node has more info than group). Perhaps there's something that can be done regarding event log, and even there, we have similar problems

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/MergeNodeProperties.scala
fanf: now, we have both: `[overriden, overriding]`. I think it may not be what people want, but I also thing that whatever the choice, half people will want the other one

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/MergeNodeProperties.scala
fanf: something like that

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
```diff
@@ -148,6 +140,7 @@ class GroupsApi(
       if(req.json_?) {
         req.json match {
           case Full(arg) =>
+            println("****** json:" + compactRender(arg))
```
VinceMacBuche, line 143: You can remove this print

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/MergeNodeProperties.scala
```diff
@@ -0,0 +1,389 @@
...
+      }
+    }
+
+    /*
+     * Here, we look up for parent groups. We know that "all" contains all parents.
+     */
+    def buildUp(root: GroupProp, all: Map[NodeGroupId, GroupProp]): PureResult[InvertedTree[GroupProp]] = {
+      // recursively takes parent, keeping criterion line order
+      for {
+        parents <- root.parentGroups.accumulatePure { id =>
```
VinceMacBuche, line 373: What happens if there is a cycle between two groups ?

It should not happen, because in this case there is lot of chance that that the groups are empty (if combined with an AND),

or the composition is OR and then the groups are filtered from parents and it will trigger an error in case of common property...

So it should be ok

VinceMacBuche:
I think I'm ok with the PR and did not find any problems within the code.

It should be tested live to see how it behaves within a real Rudder and how we can use this brand new properties and its limits !

I can't merge because I think a commit is not named fixup!

VinceMacBuche:
And no issue number in commits too ;)

VinceMacBuche:
And the small feedback about a println too

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/MergeNodeProperties.scala
fanf: cycle are forbidden by construction and check when a group is modified

fanf:
PR updated with a new commit

VinceMacBuche:
You still need to handle the two first commits to have issue number in fisrt one and a fixup! in second commit

PR #2892, opened 2020-04-14
Fixes #17112: Missing LDAP write operations to promote a node to relay

ElaadF:
https://issues.rudder.io/issues/17112

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
@@ -601,7 +601,6 @@ class WoLDAPDirectiveRepository(
 
   override def loggerName: String = this.getClass.getName
 
-
   /**
    * Save the given directive into given active technique
    * If the directive is already present in the system but not
@@ -703,6 +702,17 @@ class WoLDAPDirectiveRepository(
     })
   }
 
+  /**
+   * Delete a directive's system.
+   * No dependency check are done, and so you will have to
+   * delete dependent rule (or other items) by
+   * hand if you want.
+   *
+   * If no directive has such id, return a success.
+   */
+  override def deleteSystemDirective(id: DirectiveId, modId: ModificationId, actor: EventActor, reason: Option[String]): IOResult[Option[DeleteDirectiveDiff]] = {
+    internalDeleteDirective(id, modId, actor, reason:Option[String], callSystem = true)
+  }
 
   /**
    * Delete a directive.
@@ -710,7 +720,11 @@ class WoLDAPDirectiveRepository(
    * delete dependent rule (or other items) by
    * hand if you want.
    */
-  override def delete(id:DirectiveId, modId: ModificationId, actor:EventActor, reason:Option[String]) : IOResult[Option[DeleteDirectiveDiff]] = {
+  override def delete(id: DirectiveId, modId: ModificationId, actor: EventActor, reason: Option[String]): IOResult[Option[DeleteDirectiveDiff]] = {
+    internalDeleteDirective(id, modId, actor, reason:Option[String], callSystem = false)
+  }
+
+  private[this] def internalDeleteDirective(id:DirectiveId, modId: ModificationId, actor:EventActor, reason:Option[String], callSystem: Boolean = false) : IOResult[Option[DeleteDirectiveDiff]] = {
     getActiveTechniqueAndDirectiveEntries(id).flatMap {
       case None => // directive already deleted, do nothing
         None.succeed
@@ -720,9 +734,11 @@ class WoLDAPDirectiveRepository(
           //for logging, before deletion
           activeTechnique <- mapper.entry2ActiveTechnique(uptEntry).chainError(s"Error when mapping active technique entry to its entity. Entry: ${uptEntry}").toIO
           directive       <- mapper.entry2Directive(entry).chainError(s"Error when transforming LDAP entry into a directive for id '${id.value}'. Entry: ${entry}").toIO
-          okNotSystem     <- ZIO.when(directive.isSystem) {
-                               s"Error: system directive (like '${directive.name} [id: ${directive.id.value}])' can't be deleted".fail
-                             }
+          checkSystem     <- (directive.isSystem, callSystem) match {
+                              case (true, false) => Unexpected(s"System directive '${id.value}' can't be deleted").fail
+                              case (false, true) => Inconsistancy(s"Non-system directive '${id.value}' can not be deleted with that method").fail
+                              case _             => directive.succeed
+                            }
           technique       <- techniqueRepository.get(TechniqueId(activeTechnique.techniqueName,directive.techniqueVersion)).succeed
           //delete
           deleted         <- userLibMutex.writeLock { con.delete(entry.dn) }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
@@ -760,6 +760,25 @@ class WoLDAPNodeGroupRepository(
     }
   }
 
+  /**
+   * Delete the given policyServerTarget.
+   * If no policyServerTarget has such id in the directory, return a success.
+   */
+  override def deletePolicyServerTarget(policyServer : PolicyServerTarget): IOResult[PolicyServerTarget] = {
+    val rdn = rudderDit.RULETARGET.ruleTargetDN(policyServer.target)
+    for {
+      con           <- ldap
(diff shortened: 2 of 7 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -169,6 +169,27 @@ class WoLDAPRuleRepository(
     }
   }
 
+  def deleteSystemRule(id:RuleId, modId: ModificationId, actor:EventActor, reason:Option[String]) : IOResult[DeleteRuleDiff] = {
```
VinceMacBuche, line 172: You should not duplicate code.

you can either:
* Add a boolean parameter to deleteRule to allow deletion of systemRules
* Do like groups add a private function to internalDelete with a boolean parameter to allow systemRules deletion and two functions deleteRule and deleteSystemRules

I would go for the first one, either to understand to me, but maybe @fanf has another preference on this, but don( duplicate code ;)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -703,6 +702,46 @@ class WoLDAPDirectiveRepository(
...
 
+  /**
+   * Delete a directive's system.
+   * No dependency check are done, and so you will have to
+   * delete dependent rule (or other items) by
+   * hand if you want.
+   *
+   * If no directive has such id, return a success.
+   */
+  def deleteSystemDirective(id:DirectiveId, modId: ModificationId, actor:EventActor, reason:Option[String]) : IOResult[Option[DeleteDirectiveDiff]] = {
```
VinceMacBuche, line 713: Same as rules, don't duplicate code

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -703,6 +702,46 @@ class WoLDAPDirectiveRepository(
...
+    getActiveTechniqueAndDirectiveEntries(id).flatMap {
```
fanf, line 714: can you create a private methods with the common code, and a "onlySystem" boolean parameter that only call the common method with the corresponding error check ? We should never duplicate that much code, especially it it's so much similar.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
fanf: I prefer to expose different APIs. 

But actually, I'm thinking these methods shoud just go in the plugin (and so duplicate code :) ). 
If we don't have them so far in rudder, I'm almost sure their use case it extremely narrow, and perhaps even limited to that plugin. So I would just enterily factor them out in the plugin, where they are used. It would make evolution (for both rudder and the plugin) easier.

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -169,6 +169,44 @@ class WoLDAPRuleRepository(
     }
   }
 
+  private[this] def internalRuleSystemDirective(id:RuleId, modId: ModificationId, actor:EventActor, reason:Option[String]) : IOResult[DeleteRuleDiff] = {
```
VinceMacBuche, line 172: only this should be private and should have a Boolean to accept deleting system Rules

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -147,7 +147,7 @@ class WoLDAPRuleRepository(
     })
   }
 
-  def delete(id:RuleId, modId: ModificationId, actor:EventActor, reason:Option[String]) : IOResult[DeleteRuleDiff] = {
+  private[this] def deleteRule(id:RuleId, modId: ModificationId, actor:EventActor, reason:Option[String]) : IOResult[DeleteRuleDiff] = {
```
VinceMacBuche, line 150: Should not be private and call internalDelete with boolean at false

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -169,6 +169,44 @@ class WoLDAPRuleRepository(
...
+        } yield {
+          archive
+        }
+      }
+    } yield {
+      diff
+    }
+  }
+
+  private[this] def deleteRuleCommon(ruleId: RuleId, modId: ModificationId, actor:EventActor, reason:Option[String], isSystem: Boolean) ={
```
VinceMacBuche, line 193: Should only call InternalDelete with boolean value to True to force system Rule Deletion

VinceMacBuche: and not be private

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
VinceMacBuche: ok i missread

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -169,6 +169,44 @@ class WoLDAPRuleRepository(
...
+    if(isSystem){
+      internalRuleSystemDirective(ruleId, modId, actor, reason)
+    } else {
+      deleteRule(ruleId, modId, actor, reason)
+    }
+  }
+
+  override def deleteSystemRule(id: RuleId, modId: ModificationId, actor: EventActor, reason: Option[String]): IOResult[DeleteRuleDiff] = {
+    deleteRuleCommon(id, modId, actor, reason:Option[String], isSystem = true)
```
VinceMacBuche, line 202: This should only call InternalDelete with the boolean to true

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -169,6 +169,44 @@ class WoLDAPRuleRepository(
...
+  }
+
+  override def delete(id: RuleId, modId: ModificationId, actor: EventActor, reason: Option[String]): IOResult[DeleteRuleDiff] = {
```
VinceMacBuche, line 205: This should only call InternalDelete with the boolean to false

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
VinceMacBuche: Should be renamed InternalDelete + a boolean named 'allowSystemDeletion'

VinceMacBuche: below line 

ZIO.when(oldCr.isSystem)

ZIO.when(!allowSystemDeletion && oldCr.isSystem)

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
```diff
@@ -760,6 +760,25 @@ class WoLDAPNodeGroupRepository(
...
+   */
+  override def deletePolicyServerTarget(policyServer : PolicyServerTarget): IOResult[PolicyServerTarget] = {
+    val rdn = rudderDit.RULETARGET.ruleTargetDN(policyServer.target)
+    for {
+      con           <- ldap
+      _             <- getCategoryEntry(con, NodeGroupCategoryId("SystemGroups")).flatMap {
+                         case Some(categoryEntry) =>
+                           val entry = LDAPEntry(new DN(rdn, categoryEntry.dn))
+                           con.delete(entry.dn)
+                         case None => policyServer.target.succeed
```
fanf, line 775: it could be `UIO.unit` if you don't care about the result

PR #2896, opened 2020-04-14
Fixes #17093: Add "technical preview" status for plugins

ElaadF:
https://issues.rudder.io/issues/17093

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
@@ -94,6 +94,14 @@ trait DefaultPluginDef extends RudderPluginDef {
       PluginVersion(0,0,1, s"ERROR-PARSING-VERSION: ${versionString}")
     )
   }
+  override lazy val versionInfo = {
+    try {
+      Some(buildConf.getString("version-info"))
+    } catch {
+      case e:ConfigException.Missing => None
+    }
+  }
+
 
   override def description : NodeSeq  = (
      <div>
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/secure/plugins/pluginInformation.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/secure/plugins/pluginInformation.html
@@ -23,7 +23,8 @@
 
       <div class="lift:administration.PluginManagement.display">
         <div class="inner-portlet">
-          <h4 class="page-subtitle" data-plugin="name">[Here comes the plugins name]</h4>
+          <div data-plugin="name"></div>
+
 
           <div class="col-xs-12 callout-fade callout-info">
             <div class="marker">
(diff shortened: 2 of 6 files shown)


ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
```diff
@@ -94,6 +94,28 @@ trait DefaultPluginDef extends RudderPluginDef {
       PluginVersion(0,0,1, s"ERROR-PARSING-VERSION: ${versionString}")
     )
   }
+  override lazy val versionInfo = {
+    val info = buildConf.getString("version-info")
+    if (info.isEmpty) None else Some(info)
+
+//    val content = for {
+//      info <- IO.effect {buildConf.getString("version-info")}
+//    } yield {
```
fanf, line 103: you can remove that

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
```diff
@@ -94,6 +94,28 @@ trait DefaultPluginDef extends RudderPluginDef {
...
+//      case missingErr: ConfigException.Missing =>
+//        ApplicationLogger.info("Missing version-info in build.conf")
+//        None.succeed
+//      case typeErr: ConfigException.WrongType =>
+//        ApplicationLogger.error("Wrong type version-info in build.conf")
+//        None.succeed
+//      case e =>
+//        ApplicationLogger.error(s"Error occurs when trying to get version-info in build.conf, cause by : ${e}")
+//        None.succeed
+//    }
```
ElaadF, line 116: My Intention was to try to get the `version-info` parameter in build.conf, as this parameter is not mandatory, `getString` can fail and throw exception. In this scenario I think we don't want the program fail because of the missing parameter, so I try to catch the exceptions and log the error and return `None`
If the `getString` succeed (no Throwable) I would like to get an `Some(String)`, Nose otherwise bu,t I don't think it's the right thing to do in ZIO

Does `versionInfo` have to be an :
 + `Option[String]` -> match/case at the end (when we want to display in the webapp)
+ `IO[Throwable, Option[String]]`

fanf: just a ```try { val x = buildConf.getString("version-info"); if(x.isEmpty) None else Some(x) } catch { ex:ConfigException => None }```

(I'm not sure about exception name, but you have several example for rudder properties in `RudderConfig` file)

VinceMacBuche: You can catch any exception also, or use net.liftweb.util.tryo that returns a box and then transform into an option

ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/webapp/secure/plugins/pluginInformation.html
```diff
@@ -39,6 +40,7 @@ <h4 class="page-subtitle" data-plugin="name">[Here comes the plugins name]</h4>
               <ul class="pluginpadd">
                 <li><b>Plugin ID:</b> <span data-plugin="fullid">[Here comes the plugin full id]</span></li>
                 <li><b>Plugin version:</b> <span data-plugin="version">[Here comes the plugin version]</span></li>
+<!--                <li><b>Version information:</b> <div data-plugin="versioninfo">[Here comes the plugin version]</div></li>-->
```
fanf, line 43: can you remove that comment please, if unused ?

ElaadF:
Commit modified

ElaadF:
OK, merging this PR

PR #2900, opened 2020-04-16
Fixes #17133: Rudder should be able to read techniques by itself and not rely on legacy ncf rest api

VinceMacBuche:
https://issues.rudder.io/issues/17133

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
@@ -44,8 +44,14 @@ import com.normation.errors.Unexpected
 import cats.data.NonEmptyList
 import com.normation.inventory.domain.Version
 import com.normation.inventory.domain.AgentType
+import com.normation.rudder.ncf
 import com.normation.rudder.ncf.Constraint.Constraint
 import com.normation.rudder.ncf.Constraint.CheckResult
+import com.normation.rudder.ncf.ParameterType.ParameterTypeService
+import net.liftweb.json.JField
+import net.liftweb.json.JObject
+import net.liftweb.json.JString
+import net.liftweb.json.JValue
 
 sealed trait NcfId {
   def value : String
@@ -132,6 +138,7 @@ object ParameterType {
 
   trait ParameterTypeService {
     def create(value : String) : PureResult[ParameterType]
+    def value(parameterType: ParameterType) : PureResult[String]
     def translate(value : String, paramType : ParameterType, agentType: AgentType) : PureResult[String]
   }
 
@@ -145,6 +152,14 @@ object ParameterType {
       }
     }
 
+    def value(parameterType: ParameterType) = {
+      parameterType match {
+        case StringParameter => Right("string")
+        case Raw             => Right("raw")
+        case HereString      => Right("HereString")
+        case _               => Left(Unexpected(s"parameter type '${parameterType}' has no value defined"))
+      }
+    }
     def translate(value : String, paramType : ParameterType, agentType: AgentType) : PureResult[String] = {
       (paramType,agentType) match {
         case (Raw,_) => Right(value)
@@ -170,6 +185,12 @@ object ParameterType {
         case(service, _) => service.create(value)
       }
     }
+    def value(parameterType: ParameterType) = {
+      (innerServices foldRight (Left(Unexpected(s"parameter type '${parameterType}' has no value defined"))   : PureResult[String])) {
+        case(_, res @ Right(_)) => res
+        case(service, _) => service.value(parameterType)
+      }
+    }
 
     def translate(value : String, paramType : ParameterType, agentType: AgentType) : PureResult[String] = {
       (innerServices foldRight (Left(Unexpected(s"'${value}' is not a valid method parameter type")) : PureResult[String])) {
@@ -191,9 +212,9 @@ object Constraint {
   }
   import NonEmptyList.one
 
-  case object NonEmpty extends  Constraint {
+  case class AllowEmpty(allow : Boolean) extends  Constraint {
     def check(value: String): CheckResult = {
-      if (value.nonEmpty) {
+      if (allow || value.nonEmpty) {
         OK
       } else {
         NOK(one("Must not be empty"))
@@ -201,9 +222,9 @@ object Constraint {
     }
   }
 
-  case object NoWhiteSpace extends  Constraint {
+  case class AllowWhiteSpace(allow: Boolean) extends  Constraint {
     def check(value: String): CheckResult = {
-      if (Pattern.compile("""^(?!\s).*(?<!\s)$""", Pattern.DOTALL).asPredicate().test(value)) {
+      if (allow || Pattern.compile("""^(?!\s).*(?<!\s)$""", Pattern.DOTALL).asPredicate().test(value)) {
         OK
       } else {
         NOK(one("Must not have leading or trailing whitespaces"))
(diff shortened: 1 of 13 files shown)


VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala
```diff
@@ -191,19 +212,19 @@ object Constraint {
   }
   import NonEmptyList.one
 
-  case object NonEmpty extends  Constraint {
+  case class AllowEmpty(allow : Boolean) extends  Constraint {
```
fanf, line 215: I'm not sure I understand why we need that? If the field can be empty, we used to not have the constraint, what changed that need that evolution? (Same for `AllowWhiteSpace` vs `NoWhiteSpace`) ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -0,0 +1,85 @@
...
+  restExtractor  : RestExtractorService
+) {
+
+
+  import zio.syntax._
+  import better.files.File._
+  val configuration_repository = root / "var" / "rudder" / "configuration-repository"
+  val methodsFile = configuration_repository / "ncf" / "generic_methods.json"
+  def getAllTechniqueFiles(currentPath : File): List[File]= {
+    val (subdirs, files) = currentPath.children.partition(_.isDirectory)
```
fanf, line 26: this should be in an `IO`, because it's io plus is can block

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -0,0 +1,85 @@
...
+    val techniqueFilePath = currentPath / "technique.json"
+    val checkSubdirs = subdirs.flatMap(getAllTechniqueFiles).toList
+    if (techniqueFilePath.exists) {
```
fanf, line 29: alos that in `IO`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -0,0 +1,85 @@
...
+    } else {
+      checkSubdirs
+    }
+  }
+  def readTechniquesMetadataFile: IOResult[List[Technique]] = {
+    for {
+      methods <- readMethodsMetadataFile
+      techniques <- ZIO.foreach(getAllTechniqueFiles(configuration_repository / "techniques"))(
+        techniqueFile =>
+            restExtractor.extractNcfTechnique(parse(techniqueFile.contentAsString), methods, false) match {
```
fanf, line 40: You need to use `toIO` to correctly suspend the effect in box, else evaluation won't happen when it should. 
Then, you can just use `.chainError` for contextualisation:

```
restExtractor.extractNcfTechnique(parse(techniqueFile.contentAsString), methods, false).toIO
  .chainError("An Error occured while extracting data from techniques ncf API")
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -0,0 +1,85 @@
...
+      techniques
+    }
+  }
+
+  def readMethodsMetadataFile : IOResult[Map[BundleName, GenericMethod]] = {
+    for {
+      genericMethodContent <- IOResult.effect(s"error while reading ${methodsFile.pathAsString}")(methodsFile.contentAsString)
+      methods <- parse(genericMethodContent) match {
+        case JObject(fields) =>
+          restExtractor.extractGenericMethod(JArray(fields.map(_.value))).map(_.map(m => (m.id, m)).toMap) match {
```
fanf, line 56: Same as above

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -0,0 +1,85 @@
...
+              Inconsistency(fail.messageChain).fail
+          }
+        case a => Inconsistency(s"Could not extract methods from ncf api, expecting an object got: ${a}").fail
+      }
+    } yield {
+      methods
+    }
+  }
+
+  def updateMethodsMetadataFile = {
```
fanf, line 69: Can you add return type for public methods? It makes it easire to follow review and check for API changes

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -194,13 +194,29 @@ object NcfApi extends ApiModuleProvider[NcfApi] {
...
+    val description = "Get all Techniques metadata"
+    val (action, path)  = GET / "ncf" / "techniques"
+  }
+  final case object GetMethods extends NcfApi with ZeroParam with StartsAtVersion15 with SortIndex { val z = implicitly[Line].value
+    val description = "Get all methods metadata"
+    val (action, path)  = GET / "ncf" / "methods"
+  }
+  final case object UpdateMethods extends NcfApi with ZeroParam with StartsAtVersion15 with SortIndex { val z = implicitly[Line].value
+    val description = "Update methods metadata"
+    val (action, path)  = POST / "ncf" / "methods" / "update"
```
fanf, line 214: Why no just `POST` on `ncf/methods` ? I think the update is not necessary and it's a break compared with what is used everywhere else in rudder

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckNcfTechniqueUpdate.scala
```diff
@@ -90,104 +80,48 @@ class CheckNcfTechniqueUpdate(
...
-                              )) match {
-                                case Full(m) => Right(m)
-                                case eb : EmptyBox =>
-                                  val fail = eb ?~! "An error occured while writing ncf Techniques"
-                                  Left(WriteTechniqueError(fail.messageChain, None))
-                              }
-        // Update technique library once all technique are updated
-        libUpdate          <- techLibUpdate.update(ModificationId(uuidGen.newUuid), EventActor(systemApiToken.name.value), Some(s"Update Technique library after updating all techniques at start up")) match {
-                                case Full(techniques) => Right(techniques)
+                                  match {
```
fanf, line 102: you need to `toIO` the box to correctly suspend `IO` for when the will be evaluated. 
You can also just `chainError` appart if you want to keep `WriteTechniqueError` type, in which case you can `mapError(err => WriteTechniqueError(err.fullMsg, None))`

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckNcfTechniqueUpdate.scala
```diff
@@ -90,104 +80,48 @@ class CheckNcfTechniqueUpdate(
...
+                                  case Full(m) => m.succeed
+                                  case eb : EmptyBox =>
+                                    val fail = eb ?~! "An error occured while writing ncf Techniques"
+                                    WriteTechniqueError(fail.messageChain, None).fail
+                                }
+                              )
+                                                        // Update technique library once all technique are updated
+        libUpdate   <- techLibUpdate.update(ModificationId(uuidGen.newUuid), EventActor(systemApiToken.name.value), Some(s"Update Technique library after updating all techniques at start up")) match {
```
fanf, line 110: same here

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckNcfTechniqueUpdate.scala
```diff
@@ -90,104 +80,48 @@ class CheckNcfTechniqueUpdate(
...
-          BootstrapLogger.logEffect.warn(s"All ncf techniques were updated, but we could not delete flag file ${ncfTechniqueUpdateFlag}, please delete it manually")
-        case Left(e : NcfTechniqueUpgradeError) =>
-          BootstrapLogger.logEffect.error(s"An error occured while updating ncf techniques: ${e.msg}")
       }
     }
 
     try {
       if (ncfTechniqueUpdateFlag.exists) {
-        ZioRuntime.internal.unsafeRunAsync(Task.effect(updateNcfTechniques).delay(10.seconds))( _ => ())
+        ZioRuntime.runNow(updateNcfTechniques)
```
fanf, line 124: You need to handle errors here to avoid stack traces:
```
val prog = updateNcfTechniques.catchAll(err => theCorrectLogger.error(err.chainError("Error when updating ncf techniques").fullMsg)

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala (hunk 1 above)
Comment (VinceMacBuche) [reply webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/Technique.scala:215]: Since we now need to generate that constraint in the technique editor (mandatory field for technique editor), and the logic to serialize it is much simpler if we have it exlicitly https://github.com/Normation/rudder/pull/2900/files#diff-7bc55a791aa1d54ac3f6abaf79e4142cR351 (or else to set an allow_empty constraint if have to check the non presence of NonEmpty, which is kind of weird), it also simplify the parsing of the json https://github.com/Normation/rudder/pull/2900/files#diff-c54639a4587468db45968f0b40c358b5L1171 ) 

But maybe also we could adapt technique editor have a default value if not defined

Maybe I should also try with the old structure, maybe it would be ok

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala (hunk 2 above)
Comment (VinceMacBuche) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala:26]: Everything should be a big IO

**Comment by VinceMacBuche** [comment]
PR updated with a new commit

**Comment by VinceMacBuche** [comment]
PR rebased

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala (hunk 7 above)
Comment (fanf) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala:214]: up?

**Comment by VinceMacBuche** [comment]
PR updated with a new commit

**Comment by VinceMacBuche** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala (hunk 7 above)
Comment (VinceMacBuche) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala]: I do think we need to have different url for this methods, it's not a POST where you give data directly, but here we start a reload of methods/techniques from the file system.

So i guess a better URL maybe POST /ncf/reload/(techniques|methods) would be better

Comment (VinceMacBuche) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala]: I also don't like that url prefix here ncf/

other api from this api (write a technique/ parameter check don't have this prefix)

Maybe we could have ncf prefix everywhere for this api ...

**Comment by VinceMacBuche** [comment]
PR updated with a new commit

**Comment by VinceMacBuche** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala [hunk 11]
```diff
@@ -179,27 +179,43 @@ object NcfApi extends ApiModuleProvider[NcfApi] {
...
+  }
+  final case object GetTechniques extends NcfApi with ZeroParam with StartsAtVersion15 with SortIndex { val z = implicitly[Line].value
+    val description = "Get all Techniques metadata"
+    val (action, path)  = GET / "internal" / "techniques"
+  }
+  final case object GetMethods extends NcfApi with ZeroParam with StartsAtVersion15 with SortIndex { val z = implicitly[Line].value
+    val description = "Get all methods metadata"
+    val (action, path)  = GET / "internal" / "methods"
+  }
+  final case object UpdateMethods extends NcfApi with ZeroParam with StartsAtVersion15 with SortIndex { val z = implicitly[Line].value
```
fanf, line 212: can you change version to 12? I think that since all is internal, version number should not change and use the last one.

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
OK, squash merging this PR

PR #2902, opened 2020-04-16
Fixes #17141: when we add a multiline description in Technique, it breaks policy generation and technique loading

VinceMacBuche:
https://issues.rudder.io/issues/17141

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -354,7 +354,7 @@ class ClassicTechniqueWriter(basePath : String, parameterTypeService: ParameterT
 
     val content =
       s"""# @name ${technique.name}
-         |# @description ${technique.description}
+         |# @description ${technique.description.replaceAll("\\R", "\n# ")}
          |# @version ${technique.version.value}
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -354,7 +354,7 @@ class ClassicTechniqueWriter(basePath : String, parameterTypeService: ParameterT
 
     val content =
       s"""# @name ${technique.name}
-         |# @description ${technique.description}
+         |# @description ${technique.description.split("\n").mkString("\n# ")}
```
fanf, line 357: why not `replaceAll` ?

fanf: and perhaps `[\n\r]{1,2}` ? Not sure what browser on windows do?

ncharles: according to https://stackoverflow.com/questions/454908/split-java-string-by-new-line , which point to https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html, \R should do the trick

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
OK, squash merging this PR

PR #2903, opened 2020-04-16
Fixes #16301: Add JSON support for global parameters

fanf:
https://issues.rudder.io/issues/16301

This commit is based on `groups properties` one, PR https://github.com/Normation/rudder/pull/2890 need to be review/merged first. Also, you should only look at last commit for that PR until that time. 

This PR main changes are: 

- GlobalParameter value is now a `JValue` to be homogeneous with `GroupProperties` and `NodeProperties`
- also, removed `overridable` since that field wasn't ever used, and overriding logic don't match it. 

Then, the other big change is about inheritance logic. To be able to use `parameters` value when we build `NodeProperties` in `getNodeContexts > buildParams`, we need to evaluate `Param` interpollation there. It's ok, because we have all needed info for that at that point, and we could have done it all along here, but it means that we need to split `InterpollationContext` in two: one for params (with param not yet evaluated, to use when parsing params), and one for nodes (with all params value evaluated, ie they are just plain string, to use everywhere else). 
That part is a bit verbose because it need to add a type parameter and change "what it means to get a parameter value" in the evaluation, and it needed to split the evaluation part of the compiler from the compilation part. But the code is rather automatic and guided by types.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/parameters/GlobalParameter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/parameters/GlobalParameter.scala
@@ -37,6 +37,9 @@
 package com.normation.rudder.domain.parameters
 import java.util.regex.Pattern
 
+import com.normation.rudder.domain.nodes.GenericPropertyUtils
+import net.liftweb.json._
+
 final case class ParameterName(value:String) extends AnyVal
 
 object ParameterName {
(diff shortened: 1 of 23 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/parameters/GlobalParameter.scala
```diff
@@ -37,32 +37,36 @@
...
 
+  /**
+   * A builder with the logic to handle the value part.
+   *
+   * For compatibity reason, we want to be able to process
+   * empty (JNothing) and primitive types, especially string, specificaly as
+   * a JString *but* a string representing an actual JSON should be
+   * used as json.
+   */
+  def apply(name: String, value: String, description: String =""): GlobalParameter = {
```
VinceMacBuche, line 68: Don(t you think we would want to rename GlobalParameter to GlobalProperty so that we can relate them more easily ?

fanf: It may make a lot of sense, but it seems to be a diruptive change for a 6.1: breaking API, doc (perhaps internal things). 
I would prefer to keep that kind of things for 7.0 cc @amousset for feedback

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
```diff
@@ -899,9 +899,12 @@ final case class RestExtractorService (
     for {
       description <- extractJsonString(json, "description")
       overridable <- extractJsonBoolean(json, "overridable")
```
VinceMacBuche, line 901: I think you can remove overridable here and in RestParameter as wel

VinceMacBuche: I'll make another pr for this, and prefere to merge it for more test on real Rudder!

PR #2904, opened 2020-04-19
Fixes #17154: Allow HTTP reporting in changes-only mode

fanf:
https://issues.rudder.io/issues/17154


amousset:
I think there's also something in the settings form to prevent switching to HTTP.

amousset:
Or maybe it's this and I'm just not able to do this review :)

fanf:
Actually, it seems we were just "hidding" in the form the possibility, no really forbidding it (:scream:). 
So it's just letting the button be seen again (and making heartbeat title a bit smaller, else it's very confusing)

PR #2909, opened 2020-04-21
Fixes #17161: Historize node count information

fanf:
https://issues.rudder.io/issues/17161

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/metrics/Scheduler.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/metrics/Scheduler.scala
@@ -1,0 +1,130 @@
+/*
+*************************************************************************************
+* Copyright 2020 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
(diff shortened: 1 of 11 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/metrics/Scheduler.scala
```diff
@@ -0,0 +1,130 @@
...
+  , queue  : Queue[A]
+  , action : A => UIO[Unit]
+  , default: UIO[A] // the default value to used for consumer started by cron, not event
+  , zclock : Clock
+) {
+
+  /**
+   * External API to use when an event happens.
+   */
+  def triggerSchedule(a: A): UIO[Unit] = {
```
VinceMacBuche, line 96: There is nothing to trigger a Schedule ? is it expected ? 

Nonetheless, it's not needed for this feature and we can add it after.

fanf: I wasn't sure about what exactly we wanted. Likely for added node, and deleted?
But I wanted to get the core feature in before validating what events we would like to add.

PR #2910, opened 2020-04-21
Fixes #17172: Concurrent access to node info cache cause exception to be thrown when accessing cache content

VinceMacBuche:
https://issues.rudder.io/issues/17172

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
@@ -413,7 +413,10 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
                      }
                   } else {
                     logPure.debug(s"NodeInfo cache is up to date, ${nodeCache.map(c => s"last modification time: '${c.lastModTime}' for: '${c.lastModEntryCSN.mkString("','")}'").getOrElse("")}") *>
-                    nodeCache.get.nodeInfos.succeed //get is ok because in a synchronized block with a test on isEmpty
+                    (nodeCache match {
+                      case Some(cache) => cache.nodeInfos.succeed
+                      case None => Inconsistancy("When trying to access Node information from cache was empty, but it should not, this is a developper issue, please open an issue on https://issues.rudder.io").fail
+                    })
                   }
         res    <- useCache(info)
         t1     <- currentTimeMillis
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -413,7 +413,10 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
                      }
                   } else {
                     logPure.debug(s"NodeInfo cache is up to date, ${nodeCache.map(c => s"last modification time: '${c.lastModTime}' for: '${c.lastModEntryCSN.mkString("','")}'").getOrElse("")}") *>
-                    nodeCache.get.nodeInfos.succeed //get is ok because in a synchronized block with a test on isEmpty
+                      (nodeCache match {
```
ncharles, line 416: could you correct the indentation here ?

VinceMacBuche: indentation was not correct, this is a parameter of the first line 

logPure.debug (xxxx) *> (nodeCache match ....) 

if really find it easier to understand that the second line depends on the first one with one more indent level.

When aligned on the same line it took me some time to understand why it did not compile without parenthesis, because first line is too long to see *> on screen

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
fanf: `*>` does not means it's a parameter, it's a sequence (like with `;`).
You can put `*>` at the begining to make it more visible. 
You may be force to use parenthesis for the `else` to make scalac happy. 
The parenthesis around  `nodeCache match {}` should not be necessary though.

VinceMacBuche: no without parenthesis it considers that 

logPure.debug(s"NodeInfo cache is up to date, ${nodeCache.map(c => s"last modification time: '${c.lastModTime}' for: '${c.lastModEntryCSN.mkString("','")}'").getOrElse("")}") *> nodeCache

is the variable to match and not just nodeCache

it does not work to put it before the second line it does not compile (*> cannot be applied )

anyway i'll go back to old indentation

VinceMacBuche:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #2928, opened 2020-04-24
Fixes #17224: Update some libraries for 6.1

fanf:
https://issues.rudder.io/issues/17224

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPSwapGroupLibrary.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPSwapGroupLibrary.scala
@@ -259,7 +259,7 @@ class ImportGroupLibraryImpl(
         }
       }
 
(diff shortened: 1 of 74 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPSwapGroupLibrary.scala
```diff
@@ -259,7 +259,7 @@ class ImportGroupLibraryImpl(
         }
       }
 
-      def recSanitizeCategory(content:NodeGroupCategoryContent, parent: NodeGroupCategory, isRoot:Boolean = false) : Option[NodeGroupCategoryContent] = {
+      def recSanitizeCategory(content: NodeGroupCategoryContent, parent: NodeGroupCategory, isRoot: Boolean) : Option[NodeGroupCategoryContent] = {
```
ncharles, line 262: can't we have default value anymore?

fanf: default in recursive methods lead to a warning saying "be careful, it's using a default value!", I'm not sure about the intent but I wasn't in a mood to fight scalac

PR #2931, opened 2020-04-24
Fixes #17237: Do not compute dynamic groups if nothing changed in LDAP

fanf:
https://issues.rudder.io/issues/17237

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
@@ -41,7 +41,6 @@ import cats.implicits._
 import com.normation.box._
 import com.normation.inventory.domain.NodeId
 import com.normation.inventory.ldap.core.LDAPConstants
-import com.normation.ldap.sdk.BuildFilter._
 import com.normation.ldap.sdk._
 import com.normation.rudder.domain.RudderLDAPConstants._
 import com.normation.rudder.domain.logger.NodeLogger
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
```diff
@@ -106,6 +123,61 @@ class DynGroupServiceImpl(
       dyngroupIds.sortBy(numberOfQuery)
     }
   }.toBox
+
+  override def changesSince(lastTime: DateTime): Box[Boolean] = {
+    val n0 = System.currentTimeMillis
+    if(n0 < lastTime.getMillis || n0 - lastTime.getMillis < 100) {
```
ncharles, line 129: does that mean a change in the future?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
```diff
@@ -106,6 +123,61 @@ class DynGroupServiceImpl(
...
+                  // ou=Removed Inventories,ou=Inventories,cn=rudder-configuration
+                  Filter.create(s"entryDN:dnSubtreeMatch:=ou=Removed Inventories,ou=Inventories,cn=rudder-configuration")
+                  // ou=Accepted Inventories,ou=Inventories,cn=rudder-configuration
+                , Filter.create(s"entryDN:dnSubtreeMatch:=ou=Accepted Inventories,ou=Inventories,cn=rudder-configuration")
+                , AND(IS(OC_SOFTWARE), Filter.create(s"entryDN:dnSubtreeMatch:=ou=Software,ou=Inventories,cn=rudder-configuration"))
+                  // ou=Nodes,cn=rudder-configuration - the objectClass is used only here
+                , AND(IS(OC_RUDDER_NODE), Filter.create(s"entryDN:dnOneLevelMatch:=ou=Nodes,cn=rudder-configuration"))
+                , AND(IS(OC_RUDDER_NODE_GROUP), Filter.create(s"entryDN:dnSubtreeMatch:=groupCategoryId=GroupRoot,ou=Rudder,cn=rudder-configuration"))
+              )
+            , GTEQ("modifyTimestamp", GeneralizedTime(lastTime).toString)
```
ncharles, line 153: we don't have index for that. It will be really costly on ldap

ncharles: also, is there a way to stop the query if one result is found ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
fanf: ah yes, it's not usefull in fact (because if the first is true, the second is also). I will remove it.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
fanf: it's actually stop on the first (or second?) hit, so it is (a bit, but still much less that group search) costly only if there's no modifications. We do something very similar for cache, and it's quite fast. 
(openldap seem to be very efficient in pruning entries that he don't care on that. The fact that keys are sorted in mdb may be the reason, I don't know).

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #2935, opened 2020-04-27
Fixes #17251: Initialize `rudder` global parameter to default value at boot

fanf:
https://issues.rudder.io/issues/17251

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
@@ -1,0 +1,10 @@
+[{
+  "name": "rudder"
+  "description": "This parameter defines important properties for Rudder and should not be modified. If you modify them, they will be reset at next restart. This value can be overridden by setting a node/group property named \"rudder\" containing the keys you want to override."
+  "value": {
+    "packages": {
+      "installed_cache_expire": 60,
+      "updates_cache_expire": 240
+    }
+  }
+}]
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckRudderGlobalParameter.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckRudderGlobalParameter.scala
@@ -1,0 +1,123 @@
+/*
+*************************************************************************************
+* Copyright 2020 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
+* (at your option) any later version.
+*
+* In accordance with the terms of section 7 (7. Additional Terms.) of
+* the GNU General Public License version 3, the copyright holders add
+* the following Additional permissions:
+* Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+* Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+* Public License version 3, when you create a Related Module, this
+* Related Module is not considered as a part of the work and may be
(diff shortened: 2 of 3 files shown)


fanf:
PR rebased

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
```diff
@@ -0,0 +1,10 @@
+[{
+  "name": "rudder"
+  "description": "This parameter defines important properties for Rudder and shoudl not be modifier. If you modifiy them, they will be resetted at next restart."
```
ncharles, line 3: typo on shoudl

ncharles: and on modifiy

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckRudderGlobalParameter.scala
```diff
@@ -0,0 +1,124 @@
...
+ * So, if a full export wasn't done until know, just do one.
+ *
+ */
+class CheckRudderGlobalParameter(
+    roParamRepo: RoParameterRepository
+  , woParamRepo: WoParameterRepository
+  , uuidGen    : StringUuidGenerator
+) extends BootstrapChecks {
+
+  val resource = "rudder-system-global-parameter.conf"
```
ncharles, line 75: we could make it overridable, as a poor man solultion to replace it

webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
```diff
@@ -0,0 +1,10 @@
+[{
```
amousset, line 1: shouldn't it be a .json file instead?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/CheckRudderGlobalParameter.scala
ncharles: ha wait, it's simply to restore defaut, so no need to have such cleverness, you can ignore that

webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
fanf: I prefer to have a `.conf` or whatever to not be bound to json if we evolve format to hocon. And since it's internal, I think it doesn't matter.

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
```diff
@@ -0,0 +1,10 @@
...
+  "name": "rudder"
+  "description": "This parameter defines important properties for Rudder and should not be modified. If you modify them, they will be resetted at next restart."
```
VinceMacBuche, line 3: I think the descprition should explain how to override it. 

This property can be overriden by defining property rudder on groups / nodes

VinceMacBuche, line 3:
```suggestion
  "description": "This parameter defines important properties for Rudder and should not be modified. If you modify them, they will be reset at next restart. This value can be overridden by setting a node/group property named \"rudder\" containing the keys you want to override."
```

fanf: But it will be true for all properties. It should be in the global documentation, not here. OK since it's the first.

fanf:
PR rebased

PR #2937, opened 2020-04-29
Fixes #17252: translate should not use hardcoded path for libraries

gpoblon:
https://issues.rudder.io/issues/17252


peckpeck:
OK, merging this PR

PR #2942, opened 2020-04-29
Fixes #7006: Parameter validation in Rudder should accept rudder variables ( ${ } )

fanf:
https://issues.rudder.io/issues/7006

"oups". 

It was *really* too complicated to add non-rudder variable to our compiler with scala lib parser combintor. Since we now have `fastparse` in the path, I tried to see if it was easier... And it was thanks to its amazing log possibilities (the parsing by itself is a bit tricky). 
The most interesting is how little code changes from scala regex combinator to fastparse. And the perf are just in other league (it does not change much in our case, since node prop interpolation is in second range, but now we are in the 10s of millisec - it may still be interesting with group properties and many more parsing to do). 
So I also switch the last parser we had (the one for our "search anything" bar queries), what allows to get ride of one dependancy and use the (very cool!) same combinator parser everywhere.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
@@ -39,6 +39,8 @@ package com.normation.cfclerk.domain
 
 import com.normation.errors.Inconsistency
 import com.normation.errors.PureResult
+import com.normation.rudder.services.policies.PropertyParser
+import com.normation.rudder.services.policies.PropertyParserTokens
 import org.joda.time.format.ISODateTimeFormat
 
 class ConstraintException(val msg: String) extends Exception(msg)
@@ -123,6 +125,20 @@ object VTypeConstraint {
   }
 
   val allTypeNames = validTypes(None,Seq()).map( _.name ).mkString(", ")
+
+  def checkIsProperty(value: String, forField: String)(elsePart: String => PureResult[Unit]): PureResult[Unit] = {
+    PropertyParser.parse(value) match {
+      case Left(err) => //likely an error with property name
+        Left(Inconsistency(s"Wrong value ${value} for field '${forField}': check property syntax"))
+      case Right(tokens) => // we need to check that there's at least some var, not just a pure string
+        if(PropertyParserTokens.containsVariable(tokens)) {
+          Right(())
+        } else { //pure string
+          elsePart(value)
+        }
+    }
+
+  }
 }
 
 sealed trait StringVType extends VTypeConstraint with VTypeWithRegex with STString {
@@ -157,12 +173,15 @@ object MailVType extends FixedRegexVType {
 final case class IntegerVType(regex: Option[RegexConstraint] = None) extends VTypeConstraint with VTypeWithRegex  {
   override val name = "integer"
   override def getFormatedValidated(value:String, forField:String, escapeString: String => String) : PureResult[String] = {
-    super.getFormatedValidated(value, forField, escapeString).flatMap( _ =>
-      (try {
-        Right(value.toInt)
-      } catch {
-        case _:Exception => Left(Inconsistency(s"Wrong value ${value} for field '${forField}': expecting an integer."))
-      }).map( _ => value )
+    super.getFormatedValidated(value, forField, escapeString).flatMap( escaped =>
+      VTypeConstraint.checkIsProperty(value, forField) { v =>
+        try {
+          v.toInt
+          Right(())
+        } catch {
+          case _:NumberFormatException => Left(Inconsistency(s"Wrong value ${v} for field '${forField}': expecting an integer."))
+        }
+      }.map(_ => escaped)
     )
   }
 }
@@ -177,12 +196,15 @@ final case class SizetbVType(regex: Option[RegexConstraint] = None) extends Size
 final case class DateTimeVType(regex: Option[RegexConstraint] = None) extends VTypeConstraint with VTypeWithRegex {
   override val name = "datetime"
   override def getFormatedValidated(value:String, forField:String, escapeString: String => String) : PureResult[String] = {
-    super.getFormatedValidated(value, forField, escapeString).flatMap( _ =>
-      (try
-        Right(ISODateTimeFormat.dateTimeParser.parseDateTime(value))
-      catch {
-        case _:Exception => Left(Inconsistency(s"Wrong value ${value} for field '${forField}': expecting a datetime in ISO 8601 standard."))
-      }).map( _.toString(ISODateTimeFormat.dateTime()))
+    super.getFormatedValidated(value, forField, escapeString).flatMap( escaped =>
+      VTypeConstraint.checkIsProperty(value, forField) { v =>
+        try {
+          ISODateTimeFormat.dateTimeParser.parseDateTime(v)
+          Right(())
+        } catch {
+          case _:Exception => Left(Inconsistency(s"Wrong value ${v} for field '${forField}': expecting a datetime in ISO 8601 standard."))
+        }
+      }.map(_ => escaped)
     )
   }
 }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
@@ -40,6 +40,8 @@ import java.util.regex.Pattern
 
 import com.normation.errors.Inconsistency
 import com.normation.errors.PureResult
+import com.normation.rudder.services.policies.PropertyParserTokens
+import com.normation.rudder.services.policies.PropertyParser
 
 /**
  * We require a non empty regex pattern
@@ -52,15 +54,26 @@ case class RegexConstraint(pattern: String, errorMsg: String) {
 
   val compiled = Pattern.compile(pattern)
 
-  val variablePattern = Pattern.compile(".*?\\$\\{.*?\\}.*?")
-
   /* throw a ConstraintException if the value doesn't match the pattern */
-  def check(varValue: String, varName: String) : PureResult[String]=
-    if(variablePattern.matcher(varValue).matches || compiled.matcher(varValue).matches) {
+  def check(varValue: String, varName: String) : PureResult[String] = {
+    // don't fail on interpolated values. That value can be in the middle of a string, and
+    // we can't say much about the respect of the resulting expanded string regarding the regex,
+    // so just accept and believe in our users
+    if(compiled.matcher(varValue).matches) {
       Right(varValue)
     } else {
-      Left(Inconsistency(s"Please modify ${varName} to match the requested format ${if (errorMsg != "") " : " + errorMsg else ""}"))
+      PropertyParser.parse(varValue) match {
+        case Left(err) => // most likely the user wanted to use a property but made a typo
+          Left(Inconsistency(s"Please modify ${varName}: error while parsing property: ${varValue}")) // parser error are not hyper user friendly
+        case Right(tokens) => // check that we don't have a pure string
+          if(PropertyParserTokens.containsVariable(tokens)) {
+            Right(varValue)
+          } else {
+            Left(Inconsistency(s"Please modify ${varName} to match the requested format ${if (errorMsg != "") " : " + errorMsg else ""}"))
+          }
+      }
     }
+  }
 }
 
 object RegexConstraint {
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
@@ -37,14 +37,15 @@
 
 package com.normation.rudder.services.policies
 
-import scala.util.parsing.combinator.RegexParsers
-import net.liftweb.common.{Failure => FailedBox, _}
-import com.normation.rudder.domain.parameters.ParameterName
-import net.liftweb.json.JsonAST.JValue
-import com.normation.inventory.domain.AgentType
-import com.normation.rudder.domain.policies.PolicyModeOverrides
+
+import com.normation.box._
 import com.normation.errors._
-import com.normation.rudder.services.policies.InterpolatedValueCompilerImpl._
+import com.normation.inventory.domain.AgentType
+import com.normation.rudder.domain.parameters.ParameterName
+import com.normation.rudder.domain.policies.PolicyModeOverrides
+import com.normation.rudder.services.policies.PropertyParserTokens._
+import net.liftweb.common._
+import net.liftweb.json.JsonAST.JValue
 
 /**
  * A parser that handle parameterized value of
@@ -141,7 +142,7 @@ trait InterpolatedValueCompiler {
 
 }
 
-object InterpolatedValueCompilerImpl {
+object PropertyParserTokens {
 
   /*
    * Our AST for interpolated variable:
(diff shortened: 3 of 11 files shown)


fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -406,64 +405,88 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
   //also parse full blank string, because no need to look for more case for them.
-  val emptyVar    : Parser[List[CharSeq]] = """(?iums)(\s)*""".r ^^ { x => List(CharSeq(x)) }
-
+  def emptyVar[_: P]: P[List[CharSeq]] = P(CharsWhile(_.isSpaceChar, 0).!).map(x => List(CharSeq(x)) )
+  def s[_:P] = P(CharsWhile(_.isWhitespace, 0))
   // plain string must not match our identifier, ${rudder.* and ${node.properties.*}
-  // here we defined a function to build them, with the possibility to
-  val plainString : Parser[CharSeq] = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties)).)+""").r  ^^ { CharSeq(_) }
+  // here we defined a function to build them
+  def plainStringNoDollar[_: P] : P[CharSeq] = P( (!"${" ~ AnyChar).rep(1).! ).map { CharSeq(_) }
```
ncharles, line 429: i'm worried that everything is a def rather than a val: the number of instanciated objects will explode.

ncharles: see https://github.com/Normation/rudder/pull/2467

fanf: I'm asking on fastparse how to handle that case

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/Constraint.scala
```diff
@@ -157,12 +173,15 @@ object MailVType extends FixedRegexVType {
...
-      (try {
-        Right(value.toInt)
-      } catch {
-        case _:Exception => Left(Inconsistency(s"Wrong value ${value} for field '${forField}': expecting an integer."))
-      }).map( _ => value )
+    super.getFormatedValidated(value, forField, escapeString).flatMap( escaped =>
+      VTypeConstraint.checkIsProperty(value, forField) { v =>
+        try {
+          v.toInt
+          Right(())
```
ncharles, line 180: why is it returning an empty Right ?

fanf: to force the user (me) to `map(_ => escaped)` for the return value. We don't use these value, the parsing/etc is just done to check if the provided input is of the correct type, but we need to use the escaped string char that is already computed before that method is called

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/RegexConstraint.scala
```diff
@@ -52,15 +54,26 @@ case class RegexConstraint(pattern: String, errorMsg: String) {
 
   val compiled = Pattern.compile(pattern)
 
-  val variablePattern = Pattern.compile(".*?\\$\\{.*?\\}.*?")
```
ncharles, line 55: i like that it is removed :)

ncharles:
this looks good to me, but I'm not sure i'll be able to understand the parsing part

fanf:
Fortunately, there is a bunch of test cases. It seems that whoever wrote the first parser had had to deal with a great deal of edge cases. Or he was just a bit paranoid. They helped a lot and caught a lot of problem when adding the `${whatever}` case

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: https://gitter.im/lihaoyi/fastparse?at=5ea9e817f0377f163158e8b4
```
Since fastparse 2.x, parsers no longer have any initialization cost. They are simply methods that you call just like any other methods
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -406,64 +405,88 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+  def emptySQuote[_: P] : P[DefaultValue] = P( "\"\"" ).map { _ => DefaultValue(CharSeq("")::Nil) }
+  def emptyTQuote[_: P] : P[DefaultValue] = P("\"\"\"\"\"\"").map { _ => DefaultValue(CharSeq("")::Nil) }
 
-  //string must be simple or triple quoted string
-  val sqString    : Parser[DefaultValue] = id("\"")  ~> rep1(sqplainStr | interpol )<~ id("\"") ^^ { case x => DefaultValue(x) }
-  val sqplainStr  : Parser[Token]        = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|")).)+""").r  ^^ { str => CharSeq(str) }
 
-  val tqString    : Parser[DefaultValue]   = id("\"\"\"")  ~> rep1(tqplainStr | interpol )<~ id("\"\"\"") ^^ { case x => DefaultValue(x) }
-  val tqplainStr  : Parser[Token]          = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|"""+"\"\"\""+ """)).)+""").r  ^^ { str => CharSeq(str) }
+  def leftTag[_: P] = P( "<" ~ (!">" ~ AnyChar).rep(1).! ~ ">" )
```
VinceMacBuche, line 484: unused, can be removed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -406,64 +405,88 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
-
+  def emptyVar[_: P]: P[List[CharSeq]] = P(CharsWhile(_.isSpaceChar, 0).!).map(x => List(CharSeq(x)) )
+  def s[_:P] = P(CharsWhile(_.isWhitespace, 0))
   // plain string must not match our identifier, ${rudder.* and ${node.properties.*}
-  // here we defined a function to build them, with the possibility to
-  val plainString : Parser[CharSeq] = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties)).)+""").r  ^^ { CharSeq(_) }
+  // here we defined a function to build them
+  def plainStringNoDollar[_: P] : P[CharSeq] = P( (!"${" ~ AnyChar).rep(1).! ).map { CharSeq(_) }
+  def plainStringNoEndCurly[_: P] : P[CharSeq] = P( (!"}" ~ AnyChar).rep(1).! ).map { CharSeq(_) }
+  def plainStringNoStartCurly[_: P] : P[CharSeq] = P( (!"{" ~ AnyChar).rep(1).! ).map { CharSeq(_) }
```
VinceMacBuche, line 431: unused, can be removed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -406,64 +405,88 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+  }
 
-  val space = """\s*""".r
+  def all[_: P] : P[List[Token]] = P( Start ~ ( (plainStringNoDollar | varOrString).rep(1) | emptyVar) ~ End).map(_.toList)
 
   //empty string is a special case that must be look appart from plain string.
   //also parse full blank string, because no need to look for more case for them.
-  val emptyVar    : Parser[List[CharSeq]] = """(?iums)(\s)*""".r ^^ { x => List(CharSeq(x)) }
-
+  def emptyVar[_: P]: P[List[CharSeq]] = P(CharsWhile(_.isSpaceChar, 0).!).map(x => List(CharSeq(x)) )
```
VinceMacBuche, line 425: you take only space character here, 

You could write 

s.!.map(x => List(CharSeq(x)))

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -406,64 +405,88 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
-  val sqString    : Parser[DefaultValue] = id("\"")  ~> rep1(sqplainStr | interpol )<~ id("\"") ^^ { case x => DefaultValue(x) }
-  val sqplainStr  : Parser[Token]        = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|")).)+""").r  ^^ { str => CharSeq(str) }
 
-  val tqString    : Parser[DefaultValue]   = id("\"\"\"")  ~> rep1(tqplainStr | interpol )<~ id("\"\"\"") ^^ { case x => DefaultValue(x) }
-  val tqplainStr  : Parser[Token]          = ("""(?iums)((?!(\Q${\E\s*rudder\s*\.|\Q${\E\s*node\s*\.\s*properties|"""+"\"\"\""+ """)).)+""").r  ^^ { str => CharSeq(str) }
+  def leftTag[_: P] = P( "<" ~ (!">" ~ AnyChar).rep(1).! ~ ">" )
+
+  //string must be simple or triple quoted string
+  def sqString[_: P]    : P[DefaultValue] = P( "\"" ~ (sqplainStr | variable ).rep(1) ~ "\"").map { case x => DefaultValue(x.toList) }
+  def sqplainStr[_: P]  : P[Token]        = P( CharsWhile(c => c != '"').! ).map { str => CharSeq(str) }
```
VinceMacBuche, line 488: This does not managed escaped quote

VinceMacBuche: also, i thin it will ignores variables, it should be something like the begining 

(plainStringNoDollar | varOrString)

But cannot use them, because the condition of underlying strings are not the same (no quote character)

VinceMacBuche:
Also I think you should rename s to space, put long names instead of shortname (ie: sqStr, should be singleQuoteString)

VinceMacBuche:
I started a PR of renaming, a removal, also adding some edge cases tests

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
VinceMacBuche: Also I think it is already treated by plainStringNoDollar

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: no, it's needed, I added a test case for it

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: we never managed escaped quote in node properties default, we have triple quote for that. If you think of a use case node supported, I should go in an other bug report I believe

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
VinceMacBuche: i removed it and added a test case and all tests are ok

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: corrected

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: corrected

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
VinceMacBuche: ok no need to support escaped quote, if it was not like before 

But you are ignoring variables in string 

default=" defaut ${node.properties[datacenterDefault]}" should match Default(CharSeq(" defaut "), Property(datacenterDefault)) but it is a CharSeq" defaut ${node.properties[datacenterDefault]}"

I added a test case in my PR

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
VinceMacBuche: That pr fixes mosr problems with inner variables in quotes, i also need to look some other edge cases (non ending variables, non Rudder variable too https://github.com/fanf/rudder/pull/10/files

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -406,64 +405,83 @@ class InterpolatedValueCompilerImpl extends RegexParsers with InterpolatedValueC
...
+  }
 
-  val space = """\s*""".r
+  def all[_: P] : P[List[Token]] = P( Start ~ ( (noVariableStart | variableStart).rep(1) | emptyVar) ~ End).map(_.toList)
 
   //empty string is a special case that must be look appart from plain string.
   //also parse full blank string, because no need to look for more case for them.
-  val emptyVar    : Parser[List[CharSeq]] = """(?iums)(\s)*""".r ^^ { x => List(CharSeq(x)) }
-
+  def emptyVar[_: P]: P[List[CharSeq]] = P(CharsWhile(_.isSpaceChar, 0).!).map(x => List(CharSeq("")) )
```
VinceMacBuche, line 425: Ha ok it's just empty ?? ok I thought is was whitspace ?

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: the problem is with empty string `""`

VinceMacBuche:
OK, squash merging this PR

PR #2947, opened 2020-05-04
Fixes #17324: Group properties inheritance error when it should not

fanf:
https://issues.rudder.io/issues/17324

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
@@ -37,6 +37,7 @@
 
 package com.normation.rudder.services.nodes
 
+import cats.implicits._
 import com.normation.errors._
 import com.normation.rudder.domain.nodes.FullParentProperty
 import com.normation.rudder.domain.nodes.GenericPropertyUtils
@@ -56,6 +57,13 @@ import com.softwaremill.quicklens._
 import net.liftweb.json.Diff
 import net.liftweb.json.JsonAST.JNothing
 import net.liftweb.json.JsonAST.JValue
+import org.jgrapht.alg.connectivity.ConnectivityInspector
+import org.jgrapht.graph.AsSubgraph
+import org.jgrapht.graph.DefaultDirectedGraph
+import org.jgrapht.graph.DefaultEdge
+import org.jgrapht.traverse.TopologicalOrderIterator
+
+import scala.jdk.CollectionConverters._
 
 /**
  * This file handle how node properties are merged with other (global, groups, etc)
@@ -72,10 +80,10 @@ import net.liftweb.json.JsonAST.JValue
  *
  * When overriding happens, we use a merge-override strategy, ie:
  * - if the overriding value is a simple value or an array, it replaces the previous one
- * - if the overriding value is an object and the overriden value is a simple type or an array,
+ * - if the overriding value is an object and the overridden value is a simple type or an array,
  *   the latter one is replaced by the former
- * - if both overriding/overriden are arrays, overriding values are appended to overriden array
- * - if both overriding/overriden are objects, then each key is overriden recursively as explained,
+ * - if both overriding/overridden are arrays, overriding values are appended to overridden array
+ * - if both overriding/overridden are objects, then each key is overridden recursively as explained,
  *   and new keys are added.
  */
 
@@ -113,7 +121,9 @@ object GroupProp {
 
   /**
    * Utility class to transform RuleTarget (which are the things where we get when
-   * we resolve node belongings, for some reason I don't know about) into GroupProp
+   * we resolve node belongings, for some reason I don't know about) into GroupProp.
+   * Parent order is keep so that if P1 is on line 1, P2 is on line 2, then we get:
+   * P1 :: P2 :: Nil
    */
   implicit class FromTarget(target: FullRuleTargetInfo) {
     def toGroupProp: PureResult[GroupProp] = {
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
```diff
@@ -349,87 +314,94 @@ object MergeNodeProperties {
...
-                                                       s"criterion but the corresponding group was not found: ${id}"))
-                       case Some(g) => buildUp(g, all)
-                     }
-                   }
-      } yield {
-        InvertedTree(root, parents)
-      }
+        _ <- parents.traverse { p =>
+               // P -> S
+               try {
```
VinceMacBuche, line 341: You could use 

```
IOResult.effect(
  s"Error` when trying to build group hierarchy of group '${child.groupName}'  (${child.groupId.value}) for parent '${p.groupName}' (${p.groupId.value})"
) (graph.addEdge(p,child))
```

Don't you think ?

VinceMacBuche: With a .either, i forgot it in my example

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
```diff
@@ -349,87 +314,94 @@ object MergeNodeProperties {
...
+                  }
+                }
+      // add edges
+      _      <- groupList.traverse { group =>
+                  for {
+                    parents <- group.parentGroups.traverse { p =>
+                                 groups.get(p) match {
+                                   case None    =>
+                                     Left(Inconsistency(s"Error when looking for parent group '${p}' of group '${group.groupName}' " +
+                                                        s"[${group.groupId.value}]. Please check criterium for that group."))
```
VinceMacBuche, line 377: criteria or criterion (if you want singular)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
```diff
@@ -349,87 +314,94 @@ object MergeNodeProperties {
...
+                                   case Some(g) =>
+                                     Right(g)
+                                 }
+                               }
+                               // reverse parent here, because we start from the childiest
```
VinceMacBuche, line 382: closest parent 
or direct parent

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
```diff
@@ -349,87 +314,94 @@ object MergeNodeProperties {
...
+                  } yield ()
+                }
+      // get connected components
+      sets   <- try {
+                  Right(new ConnectivityInspector(graph).connectedSets().asScala)
+                } catch {
+                  case ex: Exception => Left(SystemError(s"Error when looking for groups hierarchies", ex))
+                }
+      sorted <- sets.toList.traverse { set =>
+                  try {
```
VinceMacBuche, line 393: Could be replaced by a IOResult.effect (errorMessage) { the block of code).either

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
fanf: the exception are as pure as possible :) (ie: they are just here to denote some information about the code structure, something that would be done by an either in a better language than java). So I would let things in `PureResult` domain, and not bring the whole `ZIO` runtime for that.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
fanf: no, same than earlier, I want to avoir zio runtime here

fanf:
PR updated with a new commit

VinceMacBuche:
You can also replace all overriden by overridden ;)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
VinceMacBuche: Perfect :)

VinceMacBuche: I'm ok with the "not brin the whole zio stack within for this" ;) And pure Exception because not IO... That Should be an execption Type PureException

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/MergeNodeProperties.scala
fanf: just to be clear - an exception is as far from purity as we can go, it breaks referential transparency, lead to horror in concurrent programming, etc.
Nonetheless, this case seems ok since: there's no concurrent context here (all the mutable parts are in the function and don't span longer than it, nothing shard, etc) ; exception only denotes an immutable property on the code and so are purelly deterministic.

fanf:
PR rebased

PR #2951, opened 2020-05-05
Fixes #17334: Property name aren't limited to ascii

fanf:
https://issues.rudder.io/issues/17334


fanf:
OK, squash merging this PR

PR #2952, opened 2020-05-05
Fixes #17326: JSON global parameter are inherited as string

fanf:
https://issues.rudder.io/issues/17326

Implicit conversion are evil.


fanf:
PR rebased

PR #2956, opened 2020-05-06
Fixes #17308: Use cargo-deny to replace cargo-audit

amousset:
https://issues.rudder.io/issues/17308

### DIFF
--- a/relay/sources/relayd/Cargo.toml
+++ b/relay/sources/relayd/Cargo.toml
@@ -7,7 +7,7 @@ description = "Rudder relay component"
 readme = "README.md"
 homepage = "https://www.rudder.io"
 repository = "https://github.com/Normation/rudder"
-license = "GPL3"
+license = "GPL-3.0-or-later"
 
 [[bin]]
 name = "rudder-relayd"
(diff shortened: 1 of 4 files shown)


amousset:
Commit modified

relay/sources/relayd/Cargo.toml
```diff
@@ -7,7 +7,7 @@ description = "Rudder relay component"
 readme = "README.md"
 homepage = "https://www.rudder.io"
 repository = "https://github.com/Normation/rudder"
-license = "GPL3"
+license = "GPL-3.0-or-later"
```
amousset, line 10: Cargo actually expects an SPDX-formatted license name, things are converging somehow.

peckpeck: but or-later is not the sale a gpl-3

amousset: Indeed, and previous value was actually wrong (https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala#L10 or any GPL header in Rudder).

amousset:
OK, merging this PR

PR #2965, opened 2020-05-11
Fixes #17374: Add a cargo-deny config for rudder-lang

amousset:
https://issues.rudder.io/issues/17374

### DIFF
--- a/rudder-lang/deny.toml
+++ b/rudder-lang/deny.toml
@@ -1,0 +1,20 @@
+# Only non-default values are listed here
+
+[licenses]
+# List of explictly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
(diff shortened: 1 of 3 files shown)


rudder-lang/deny.toml
```diff
@@ -0,0 +1,20 @@
...
+# List of explictly allowed licenses
+# See https://spdx.org/licenses/ for list of possible licenses
+# [possible values: any SPDX 3.7 short identifier (+ optional exception)].
+allow = [
+    "MIT",
+    "Apache-2.0",
+    "ISC",
+    "BSD-2-Clause",
+    "BSD-3-Clause",
+    "GPL-3.0",
```
peckpeck, line 13: what about the or-later, is it handled automatically ?

amousset: It seems so, yes

PR #2979, opened 2020-05-15
Fixes #17426: Use hocon as an internal format for node properties

fanf:
https://issues.rudder.io/issues/17426
That PR replace the `JValue` used for node/group properties and global parameter by a `ConfigValue` from `lightbent.config` lib which accepts comments but is still a strict superset of json able to parse/write pur json. 

The behavior after the change should be the same as before safe for two points:
- on override, arrays are replaced in place of being concatenated,
- now, a value is `parsed` and can fail, while before it was always a success. The failure case is if the value starts by `{`, we assume the user want it to be a json-like property, and so we fail if it is not the case. Before, we were accepting things like `{"a":"b"` while there was no chance it's not a typo.

From an implementation point of view, that PR unify the three properties (node, group, global) behind a common `GenericProperty`. They are all backed by a standard `Config` object, with a mandatory `name:String` field, and optionnal `value: ConfigValue`, `provider: String` (mapped to an `Option[PropertyProvider]` and `description: String`. 
The PR mainly replace occurence of the old lib names/types by the new ones. 
A particular attention was put in parsing and serialisation, since it's where there is the most chance that something can change without notice.

### DIFF
--- a/webapp/sources/rudder/rudder-core/pom.xml
+++ b/webapp/sources/rudder/rudder-core/pom.xml
@@ -104,6 +104,11 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
     </dependency>
 
     <dependency>
+      <groupId>com.typesafe</groupId>
+      <artifactId>config</artifactId>
+      <version>${config-version}</version>
+    </dependency>
+    <dependency>
         <groupId>com.lihaoyi</groupId>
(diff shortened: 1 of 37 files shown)


fanf:
PR updated with a new commit

fanf:
PR rebased

PR #3001, opened 2020-05-25
Fixes #17491: Global parameter which are a comment string are ignored

fanf:
https://issues.rudder.io/issues/17491

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
@@ -233,11 +233,10 @@ object GenericProperty {
    */
   def parseValue(value: String): PureResult[ConfigValue] = {
     // find first char that is no in a commented line nor a space
-
     if(value == "") Right(ConfigValueFactory.fromAnyRef(""))
     else firstNonCommentChar(value) match {
-      case None => // here, either we return empty string, or the orginal one. I thing we should return empty string, since user can quote if he wants.case _: scala.None.type =>
-        Right(ConfigValueFactory.fromAnyRef(""))
+      case None => // here, we need to return the original string, user may want to use a comment (in bash for ex) as value
+        Right(ConfigValueFactory.fromAnyRef(value))
       case Some(c) if(c == '{') =>
         PureResult.effect(s"Error: value is not parsable as a property: ${value}") {
           ConfigFactory.parseString(
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/nodes/GenericPropertiesTest.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/nodes/GenericPropertiesTest.scala
@@ -68,7 +68,13 @@ class GenericPropertiesTest extends Specification with Loggable with BoxSpecMatc
       GenericProperty.parseValue("""{}""") must beRight
     }
     "correctly parse non json-like structure as a string" in {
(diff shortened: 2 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -233,11 +233,10 @@ object GenericProperty {
    */
   def parseValue(value: String): PureResult[ConfigValue] = {
```
amousset, line 234: Why is it parsed from a string? We have to chose a type when creating properties, can't we use this information?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -233,11 +233,10 @@ object GenericProperty {
...
     // find first char that is no in a commented line nor a space
-
     if(value == "") Right(ConfigValueFactory.fromAnyRef(""))
     else firstNonCommentChar(value) match {
-      case None => // here, either we return empty string, or the orginal one. I thing we should return empty string, since user can quote if he wants.case _: scala.None.type =>
-        Right(ConfigValueFactory.fromAnyRef(""))
+      case None => // here, we need to return the original string, user may want to use a comment (in bash for ex) as value
+        Right(ConfigValueFactory.fromAnyRef(value))
       case Some(c) if(c == '{') =>
```
amousset, line 240: does that mean that a string variable cannot start by `{`?

amousset: I tested and it says:
> if you intended '}' to be part of a key or string value, try enclosing the key or value in double quotes

which is not true as the quotes are then part of the value

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
fanf: not really - at least, not in 6.1 without changing *a lot* of things in backend. But even if we try to take advantage of that, it means that we need to tag that type everywhere a parameter can come from, including:

- from API 
- from archive of parameters serialized before we were able to specify types 
- from ldap.

And since we have the same format for nodes (and group) which we stored at time when parameters where just string, we can't just assume "string" for already saved values.

Finally, backend parsing will likelly remains more consistent than front-end, since front end doesn't know what hocon is (ie: we will likelly need to take shortcut in UI to change values to json and only check superficially, and then let backend really parsed value and decide)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
fanf: I think you're right. So if you really want  a string that starts with `{` (and is not json escapped, ie not actually `\"{` etc), you will need to build the json:
```
{ "whatever_key":"{ here your string" }
```

fanf: Hum, ok this is a change from `6.0`

fanf: https://issues.rudder.io/issues/17492

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
amousset: Ok!

PR #3003, opened 2020-05-26
Fixes #17496: Inherited properties don't appear anymore

fanf:
https://issues.rudder.io/issues/17496


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -552,9 +555,11 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
     }
   }.toBox
 
-  def getNodeInfo(nodeId: NodeId): Box[Option[NodeInfo]] = withUpToDateCache(s"${nodeId.value} node info") { cache =>
+  def getNodeInfo(nodeId: NodeId): Box[Option[NodeInfo]] = getNodeInfoPure(nodeId).toBox
+
+  def getNodeInfoPure(nodeId: NodeId): IOResult[Option[NodeInfo]] = withUpToDateCache(s"${nodeId.value} node info") { cache =>
```
VinceMacBuche, line 560: Why 'Pure' ?? I know you want to distinguish from the 'Box' function and cannot use the same name  but it's could lead to think that this is a pure function (and it's not)

fanf: It's pure, it's just a value. Effects are not done until `.runNow` is called

VinceMacBuche: I see your point, it wil always give an IO Result, and do nothing.

PR #3006, opened 2020-05-26
Fixes #15557: When adding tags, if anything is wrong in the directive form, all unsaved tags are wiped out

RaphaelGauthier:
https://issues.rudder.io/issues/15557


fanf:
OK, merging this PR

PR #3007, opened 2020-05-26
Fixes #17507: inventory watcher should ignore uuid.hive file

VinceMacBuche:
https://issues.rudder.io/issues/17507

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
@@ -79,10 +79,15 @@ final class Watchers(incoming: FileMonitor, updates: FileMonitor) {
 }
 object Watchers {
 
+  val inventoryExtentions = "gz" :: "xml" :: "ocs" :: "sign" :: Nil
+  def hasValidInventoryExtension(file : File) = {
+    val ext = file.extension(includeDot = false, includeAll = false).getOrElse("")
+    inventoryExtentions.contains(ext)
+  }
   def apply(incoming: File, updates: File, checkProcess: File => Unit): Watchers = {
     def newWatcher(directory: File): FileMonitor = {
       new FileMonitor(directory, recursive = false) {
-        val maxDepth = 0 // see parent class for logic, needed to minimize change in overriden `process` method
+        val maxDepth = 0 // see parent class for logic, needed to minimize change in overridden `process` method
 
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
```diff
@@ -169,14 +169,15 @@ class SchedulerMissedNotify(
...
    */
   def listFiles(d: Duration): UIO[List[File]] = {
     IOResult.effect{
       val ageLimit = DateTime.now().minusMillis(d.toMillis.toInt)
-      val filter = (f:File) => (
-           (f.isRegularFile && ageLimit.isAfter(f.lastModifiedTime.toEpochMilli))
-      )
+      val filter = (f:File) => {
+        val ext = f.extension(includeDot = false, includeAll = false).getOrElse("")
+        ( ext == "gz" || ext == "xml" || ext == "ocs" || ext == "sign") && (f.isRegularFile && ageLimit.isAfter(f.lastModifiedTime.toEpochMilli))
```
fanf, line 179: i think we should have a set of authorized extensions, and just have to check if it contains it; I even believe we already have it somewhere.

fanf: yes, line 102: could you make it a value in the class ?
```val authorizedExtensions = Set("gz", "xml", "ocs", "sign")```

Comment (fanf) [reply webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala:179]: not sure where to defined it exactly, perhaps in object `Watchers`

**Comment by VinceMacBuche** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala [hunk 2]
```diff
@@ -79,10 +79,15 @@ final class Watchers(incoming: FileMonitor, updates: FileMonitor) {
 }
 object Watchers {
 
+  val inventoryExtentions = "gz" :: "xml" :: "ocs" :: "sign" :: Nil
```
fanf, line 82: it should be a set, it's done for that - but actually, for 4 elements, it's still special cased so it shouldn't matter at all.

fanf:
OK, squash merging this PR

PR #3024, opened 2020-05-28
Fixes #17568: Allow to regenerate generic methods when there is a new one

VinceMacBuche:
https://issues.rudder.io/issues/17568

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
@@ -1,5 +1,7 @@
 package com.normation.rudder.ncf
 
+import java.time.Instant
+
 import better.files.File
 import com.normation.errors.IOResult
 import com.normation.errors.Inconsistency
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -53,11 +53,27 @@ class TechniqueReader(
     }
   }
 
+  def doesMethodsMetadataFileNeedsUpdate  ={
+    def getCfFilesInDir(file : File): Iterator[File] = {
+      if (file.isDirectory) {
```
fanf, line 58: why not `file.collectChildren(your extension filter)` ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -53,11 +53,27 @@ class TechniqueReader(
...
+      } else {
+        if (file.extension(false).map(_ == "cf").getOrElse(false)) Iterator(file) else Iterator.empty
+      }
+    }
+    val baseDir = root / "usr" / "share" / "ncf" / "tree" / "30_generic_methods"
+    val userDir = configuration_repository / "ncf" / "30_generic_methods"
+
+    val methodsFileModifiedTime = methodsFile.lastModifiedTime()
+
+    (getCfFilesInDir(baseDir) ++ getCfFilesInDir(userDir)).map{_.lastModifiedTime()}.exists(_.isAfter(methodsFileModifiedTime))
```
fanf, line 69: That seems inefficient. You could either collect directly max time when you iterate on childrens (map is not lazy, so you will go through all children for it), or even pass the comparison date in your collect's filter so that you just have to check for `iterator.nonEmpty` (which should stop on first match)

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -53,11 +53,21 @@ class TechniqueReader(
...
 
+  private[this] def doesMethodsMetadataFileNeedsUpdate : Boolean = {
+    val methodsFileModifiedTime = methodsFile.lastModifiedTime()
+    def isAMethodNewerThanCache(file : File) : Boolean = {
+      file.isRegularFile && file.extension(false).map(_ == "cf").getOrElse(false)  && file.lastModifiedTime.isAfter(methodsFileModifiedTime)
+    }
+    val baseDir = root / "usr" / "share" / "ncf" / "tree" / "30_generic_methods"
+    val userDir = configuration_repository / "ncf" / "30_generic_methods"
+    baseDir.collectChildren(isAMethodNewerThanCache).isEmpty || userDir.collectChildren(isAMethodNewerThanCache).isEmpty
+  }
```
fanf, line 64: I think it's ok but it's hard to test as it is. What would make it easier to test is to have it split in two: 
- one part with the logic and parameters for all states, 
- one methods that set state for user convenience, ie:

```
 // this one contains logic but stateless and so easy to test - actually, we could make it pure
 // with an `IORestul.effect` around last statement
  private[this] def checkNeedsUpdate(lastTime: TimeType, dirs: List[File]) : Boolean = {
    def isAMethodNewerThanCache(file : File) : Boolean = {
      file.isRegularFile && file.extension(false).map(_ == "cf").getOrElse(false)  && lastTime.isAfter(methodsFileModifiedTime)
    }
    dirs.exist(dir => dir.collectChildren(isAMethodNewerThanCache).isEmpty)
  }
  // this one is convenient and meaningful to use in the code (but untestable)
  private[this] def doesMethodsMetadataFileNeedsUpdate : Boolean = {
    val methodsFileModifiedTime = methodsFile.lastModifiedTime()
    val baseDir = root / "usr" / "share" / "ncf" / "tree" / "30_generic_methods"
    val userDir = configuration_repository / "ncf" / "30_generic_methods"
   checkNeedsUpdate(methodsFileModifiedTime, baseDir :: userDir :: Nil)
  }
```

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/ncf/TechniqueReader.scala
```diff
@@ -53,11 +55,32 @@ class TechniqueReader(
...
+    }
+    dirs.exists(_.collectChildren(isAMethodNewerThanCache).isEmpty)
+  }
+  private[this] def doesMethodsMetadataFileNeedsUpdate: IOResult[Boolean]  = {
+    val baseDir = root / "usr" / "share" / "ncf" / "tree" / "30_generic_methods"
+    val userDir = configuration_repository / "ncf" / "30_generic_methods"
+    for {
+      metadataFileExists <- IOResult.effect(methodsFile.exists)
+      needsUpdate <- if (metadataFileExists) {
+                       val methodsFileModifiedTime = methodsFile.lastModifiedTime()
```
fanf, line 70: this is an IO too :)

VinceMacBuche:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3026, opened 2020-05-30
Fixes #17555: error in webapp log after upgrade to 6.1-rc1

ncharles:
https://issues.rudder.io/issues/17555


ncharles:
Commit modified

PR #3028, opened 2020-06-01
Fixes #17549: Property error message for group with parent group

fanf:
https://issues.rudder.io/issues/17549

There was some renaming/moving things around which make the PR seems harder than it is really. 

The renaming/moving allow to create a method similar to the one that merge properties for node but for group, and the only really interesting piece is in that method: before that, we were not getting parents of group, and so when the merge occured, they weren't found (suprisingly). So the new method `getParents` before doing the merge. 

Renaming are: 
- remove useless parameter for `ParentProperty`, since we only have `ConfigValue` and all these parameters were starting to be confusing (and remove corresponding useless trait level)
- extract logic that change a group to a `GroupProp` so that it can be used (also) in new method
- rename `withDefaults` to `forNode` so that we also have `forGroup`


fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3030, opened 2020-06-02
Fixes #17588: Add email path parameter in RUDDER config file

ElaadF:
https://issues.rudder.io/issues/17588


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -258,6 +258,16 @@ object RudderConfig extends Loggable {
 
   //other values
 
+  val RUDDER_EMAIL_CONF = {
+    try {
```
fanf, line 262: I'm not sure why you want to put it in rudder ? Email config is only for one plugin, right?
Is there a constraint I don't see? If so, it need to be documented in the corresponding code, else it will get corrupted/changed at some point.

PR #3035, opened 2020-06-03
Fixes #17596: User with some to access technique editor api

VinceMacBuche:
https://issues.rudder.io/issues/17596


ncharles:
this breaks builds

PR #3040, opened 2020-06-03
Fixes #17594: no policy generation after accepting many nodes in 6.1

ncharles:
https://issues.rudder.io/issues/17594


ncharles:
need testing

ncharles:
this is invalid: last policy generation is doubled

fanf:
This produces a double generation when all nodes are accepted, but the only real solution would be to have a time buffer before triggering generation - not what we can do in a minor.

ncharles:
PR updated with a new commit

ncharles:
i don't have double generation anymore with that, and it's much cleaner to have a way to force dynamic group update when requested

fanf:
OK, squash merging this PR

PR #3041, opened 2020-06-03
Fixes #17602: NodeCacheImpl delay is incorrect and should be configurable

fanf:
https://issues.rudder.io/issues/17602

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
@@ -40,6 +40,7 @@ package bootstrap.liftweb
 import java.io.File
 import java.nio.charset.StandardCharsets
 import java.security.Security
+import java.util.concurrent.TimeUnit
 
 import bootstrap.liftweb.checks._
 import com.normation.appconfig._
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -403,6 +404,16 @@ object RudderConfig extends Loggable {
     }
   }
 
+  val RUDDER_BATCH_CHECK_NODE_CACHE_INTERVAL = {
+    try {
+      Duration.fromScala(scala.concurrent.duration.Duration(config.getString("rudder.batch.check.node.cache.interval")))
+    } catch {
+      case ex: Exception =>
+        ApplicationLogger.info("Property 'rudder.batch.check.node.cache.interval' is missing or empty in rudder.configFile. Default to '15 s'.")
```
ncharles, line 412: shouldn't you add the value in the config file so that we don't have the log message?

fanf: yes, right, even it technical, it will be more discoverable like that

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3053, opened 2020-06-04
Fixes #17625: Group value should only display value for that element in main table

fanf:
https://issues.rudder.io/issues/17625

This PR change a bit the logic for merge properties: 
- in a property hierarchy, we had in head of parents property the original value of the object property in the `overriden` case (when it's purely inherited, nothing change - this is done in method `MergeNodeProperties#mergeDefault`), 
- change name from `parents` to hierarchy since it doesn't hold only parents anymore, 
- for the api case, add the original value in a `origval` attribute, 
- when editing an existing property, if the provider is `overridden`, use `origval`
- for each action among `add`, `delete`, `edit`, reload all properties on save success because inheritance/overriding need to be computed again, and front end doesn't have all information to do it.

When viewing:
image

When editing: 
image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/secure/nodeManager/node.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/secure/nodeManager/node.html
@@ -1,7 +1,7 @@
 <lift:surround with="common-layout" at="content">
 
 <head>
-    <title>Rudder - Search Nodes</title>
+    <title>Rudder - Node details</title>
     <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/style/rudder/rudder-tags.css"  media="screen">
     <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/style/rudder/rudder-node.css" media="screen">
     <script type="text/javascript" data-lift="with-cached-resource" src="/javascript/angular/elastic.min.js"></script>
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/components/ComponentNodeProperties.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/components/ComponentNodeProperties.html
@@ -68,10 +68,10 @@
                      <span ng-if="property.provider !== undefined && property.provider !== 'inherited' && property.provider !== 'overridden'" class='rudder-label label-provider label-sm' data-toggle='tooltip' data-placement='top' data-html='true' title="This property is managed by its provider <b>‘{{ property.provider }}</b>’, and can not be modified manually. Check Rudder’s settings to adjust this provider’s configuration." data-container="body" bs-tooltip>
                        {{ property.provider }}
                      </span>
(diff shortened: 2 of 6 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/webapp/secure/nodeManager/node.html
```diff
@@ -1,7 +1,7 @@
 <lift:surround with="common-layout" at="content">
 
 <head>
-    <title>Rudder - Search Nodes</title>
+    <title>Rudder - Node details</title>
```
VinceMacBuche, line 4: best change

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/components/ComponentNodeProperties.html
```diff
@@ -101,8 +101,13 @@
                      <button class="btn btn-xs btn-default btn-clipboard" data-clipboard-text="{{formatContent(property)}}" data-toggle='tooltip' data-placement='left' data-container="html" title="Copy to clipboard"><i class="ion ion-clipboard"></i></button>
                    </div>
                    <div ng-if="isEdited(property.name)">
-                     <textarea placeholder="Value" msd-elastic rows="1" class="form-control input-sm input-value" ng-model="editedProperties[property.name].new.value"></textarea>
-                     <small class="text-danger" ng-if="!editedProperties[property.name].new.isValid">JSON check is enabled, but the value format is invalid</small>
+                     <div class="col-lg-6">
```
VinceMacBuche, line 104: Why col-lg ? this mean that it wont apply on lower resolution ? in general we use col-sm-x instead

VinceMacBuche: if reducing the size of the window though it's ok

webapp/sources/rudder/rudder-web/src/main/webapp/secure/nodeManager/node.html
fanf: I got bored to have "search node" in my history while it was really node details. We should have added hostname, too, but it wasn't immediate

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/components/ComponentNodeProperties.html
fanf: no idea, I copy pasted from somewhere else in the page.

fanf: (actually, there's a rationnal: in smaller screen, it may be more confortable to have one below the other, since JSON properties can become quite wide)

fanf: even with `lg`, you get existing prop mixed before it goes one below the other: 
image

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3061, opened 2020-06-05
Fixes #17646: Clicking on the warning icon is node list should lead to system status page

ElaadF:
https://issues.rudder.io/issues/17646


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/ShowNodeDetailsFromNode.scala
```diff
@@ -80,7 +80,7 @@ object ShowNodeDetailsFromNode {
   sealed trait DisplayDetailsMode { def tab : Int}
   case object Summary    extends DisplayDetailsMode { override def tab = 0 }
   case object Compliance extends DisplayDetailsMode { override def tab = 1 }
-  case object System     extends DisplayDetailsMode { override def tab = 6 }
```
ElaadF, line 83: I'm not sure if it was a mistake or on purpose

VinceMacBuche: The tab changed between first implem, and was originally in 7 position, but someone changed the tab order, but not here

VinceMacBuche: see #2997

PR #3066, opened 2020-06-09
Fixes #17674: Global parameter "string" doesn't escape json

fanf:
https://issues.rudder.io/issues/17674


fanf:
PR updated with a new commit

PR #3076, opened 2020-06-12
Fixes #17732: Dockerize Rudder Relay Server

ikramBej:
https://issues.rudder.io/issues/17732


docker/relay/cf-execd.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+RUN \
+        ln -sf /bin/true /usr/sbin/service && \
+        ln -sf /bin/true /bin/systemctl && \
+        ln -sf /bin/true /usr/bin/systemctl
+
+RUN \
+        yum -y install diffutils curl && \
+        curl -o rudder-setup https://repository.rudder.io/tools/rudder-setup && \
+        sh rudder-setup add-repository 6.1 && \
+        yum -y install rudder-agent && \
```
amousset, line 12: no need to install the agent + relay, as the agent is already a relay dependency, so relay is enough

docker/relay/http.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM centos:8
+
+RUN yum -y --setopt=tsflags=nodocs update && \
```
amousset, line 3: no need for httpd installatio, it is already a rudder-server-relay dependency

docker/relay/http.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM centos:8
```
amousset, line 1: should probably be `httpd.Dockerfile`

docker/relay/relay.Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM centos:8
```
amousset, line 1: could have `USER rudder-relayd`

docker/relay/http.Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM centos:8
+
```
amousset, line 2: could have `USER apache`

docker/relay/run-httpd.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+# Make sure we're not confused by old, incompletely-shutdown httpd
+# context after restarting the container.  httpd won't start correctly
+# if it thinks it is already running.
+rm -rf /run/httpd/* /tmp/httpd*
```
amousset, line 6: these could maybe be tmpfs, which would make it be reset automatically at each restart?

docker/relay/cf-execd.Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+        ln -sf /data/ssl/agent.cert /opt/rudder/etc/ssl/agent.cert && \
+        ln -sf /data/ppkeys/localhost.pub /var/rudder/cfengine-community/ppkeys/localhost.pub && \
+        ln -sf /data/ppkeys/localhost.priv /var/rudder/cfengine-community/ppkeys/localhost.priv
+COPY \
+        script.sh .
+
+#Make symbolic links to mount volume from cf-execd to httpd
+#To make the files under /apache_conf_file act like origin files, we make a copy of the origin file to /apache_cong_file
+RUN \
+        sh script.sh && \
```
amousset, line 27: why using a script for cp but not ln?

ikramBej: ln didn't work even on the cf-execd container so i thought about making a copy of the files to the directory that we want to share and make then a symbolic link

docker/relay/run-httpd.sh
ikramBej: I had an issue of starting httpd during docker compose execution so i tried this solution to delete temporary files but now it is useless as httpd is installed with relayd

PR #3081, opened 2020-06-16
Fixes #17702: When there is a serialisation error for parameter, their edit screen is unavaible

fanf:
https://issues.rudder.io/issues/17702

Main changes:

- if parsing of serialized global value fails, get value as string. It should be an extremely rare case (likely due to direct modification of LDAP), 
- make delete not fails if the param is not found and use provider, so that we can't delete a parameter with a different provider than our


fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3086, opened 2020-06-18
Fixes #17778: table nodes contains on entry per node per generation, which is too much

ncharles:
https://issues.rudder.io/issues/17778


ncharles:
PR replaced by https://github.com/Normation/rudder/pull/3088

PR #3091, opened 2020-06-19
Fixes #17627: Select node state in settings is confusiing for read_only user 

RaphaelGauthier:
https://issues.rudder.io/issues/17627


fanf:
It doesn't seems to do what it wanted on non read only case: 
image

fanf:
It was just a problem of css/js reloading, pergect!

PR #3094, opened 2020-06-22
Fixes #17725: user with read_only right has button that allows write actions 

RaphaelGauthier:
https://issues.rudder.io/issues/17725


webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/reports_server.html
```diff
@@ -1,6 +1,7 @@
 <batches-list>
-
-  <runAgent></runAgent>
+  <lift:authz role="administration_write">
```
VinceMacBuche, line 2: should be node_write

RaphaelGauthier:
Commit modified

PR #3098, opened 2020-06-23
Fixes #17811: fix nasty random generation failure

gpoblon:
https://issues.rudder.io/issues/17811

### DIFF
--- a/rudder-lang/src/ast/context.rs
+++ b/rudder-lang/src/ast/context.rs
@@ -88,38 +88,44 @@ impl<'src> VarContext<'src> {
     }
 
     /// Inserts a key, value into its related parent content.
-    /// Recursive function, meaning it goes deep to insert an element and check it does not exist yet
-    /// If the path is exactly the same as any existing global variable, throws an error
+    /// Recursive function, meaning it goes deep to insert an element (possibly nested) and check it does not exist yet
+    /// If the full path is exactly the same as any existing global variable, throws an error
     fn push_new_variable(
-        name: Token<'src>,
         stored_ctx: &mut Value<'src>,
         new_var: &Value<'src>,
     ) -> Result<()> {
-        match stored_ctx {
+        let (key_to_push, value_to_push) = match new_var {
+            // always has a single field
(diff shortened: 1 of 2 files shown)


gpoblon:
Generation of a technique in rudder-lang randomly failed for no obvious reason.

After investigation it appeared that this bug were exposed by a duplicate variable declaration located in the rudder-lang library.

The bug was hidden in a recursive function whose purpose was to generate the context tree (which is very important to register variables and their relative namespaces).

Duplicates weren't detected properly (recursion had flow flaws), and when they were (if the duplicate was the first entry of the context tree hashmap) it led to a crash.

Investigating this, I noticed another issue: namespaces weren't push right, ie:
```
sys.arch.linux
sys.arch.windows
```
Would led to
Context { arch { windows } }

instead of:
Context { arch { linux, windows } }

Now program behaves properly:
- warns the user when a duplicate is declared
- properly generates the tree regardless of its elements order
- properly pushes nested content
- added tests unit

rudder-lang/src/ast/context.rs
```diff
@@ -129,10 +135,13 @@ impl<'src> VarContext<'src> {
         name: Token<'src>,
         value: Value<'src>,
     ) -> Result<()> {
+        // println!("new_variable");
+
         if let Some(ref mut existing_kind) = self.variables.get_mut(&name) {
             match existing_kind {
                 VarKind::Generic(ref mut existing_var) => {
-                    Self::push_new_variable(name, existing_var, &value)?
+                    // println!("SOME =>>>>> NAME = {}\n VALUE = {:#?}", name, value);
```
amousset, line 143: this should be a proper debug or trace level log

amousset:
Congratulations for finding this one!

rudder-lang/src/ast/context.rs
gpoblon: fixed

gpoblon:
PR updated with a new commit

gpoblon:
PR rebased

amousset:
OK, squash merging this PR

PR #3109, opened 2020-06-30
Fixes #17882: Shared-files correction in postinst must be recursive

Fdall:
https://issues.rudder.io/issues/17882

### DIFF
--- a/relay/sources/rudder-server-relay-postinst
+++ b/relay/sources/rudder-server-relay-postinst
@@ -49,8 +49,10 @@ chgrp rudder-policy-reader /var/rudder/share
 chmod g+s-w /var/rudder/share
 
 # change some directory to rudder owner
+mkdir -p /var/rudder/configuration-repository/shared-files
 chown -R root:rudder /var/rudder/configuration-repository/shared-files
 chmod 2750 /var/rudder/configuration-repository/shared-files
+chmod -R u+rwx,g+rX /var/rudder/configuration-repository/shared-files/*
 chown -R rudder-relayd:rudder /var/rudder/shared-files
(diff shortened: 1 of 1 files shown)


Fdall:
PR updated with a new commit

relay/sources/rudder-server-relay-postinst
```diff
@@ -49,8 +49,10 @@ chgrp rudder-policy-reader /var/rudder/share
 chmod g+s-w /var/rudder/share
 
 # change some directory to rudder owner
+mkdir -p /var/rudder/configuration-repository/shared-files
```
VinceMacBuche, line 52: This breaks on relay without root server. Should be done in rudder-webapp

VinceMacBuche: ok i was scared because of https://issues.rudder.io/issues/17884 but the mkdir should be ok.

This file is created when the relay synch a shared-file, now it will be created by packaging

Fdall:
OK, squash merging this PR

Fdall:
OK, merging this PR

relay/sources/rudder-server-relay-postinst
```diff
@@ -49,8 +49,10 @@ chgrp rudder-policy-reader /var/rudder/share
...
 chown -R root:rudder /var/rudder/configuration-repository/shared-files
 chmod 2750 /var/rudder/configuration-repository/shared-files
+chmod -R u+rwx,g+rX /var/rudder/configuration-repository/shared-files/*
```
VinceMacBuche, line 55: should be done on /var/rudder/configuration-repository/shared-files/ since the directoy may be empty, especially on install this fails

PR #3114, opened 2020-07-03
Fixes #17821: differenciate classic and dsc generic methods

gpoblon:
https://issues.rudder.io/issues/17821


rudder-lang/tools/generate_lib
```diff
@@ -3,6 +3,41 @@
...
+
+  my $is_default = 0;
+  my $ncf = "/usr/share/ncf/tree/30_generic_methods/";
+  my $dsc_ncf = "/var/rudder/configuration-repository/dsc/ncf/30_generic_methods/";
+  my $resourcelib_path = "/opt/rudder/share/rudder-lang/lib";
+  my $generic_methods = "/opt/rudder/share/rudder-lang/tools/generic_methods.toml";
+  open (FILE, $file);
+  while (<FILE>) {
+    chomp;
+    if ($is_default == 1 && /^\s*ncf\s*=\s*"(?<path>[\/\w\.\-\~]+)"/ ) {
```
peckpeck, line 17: /,~ and . don't need to be backslashed in regex

peckpeck: if you accept ~ explicitly, does it mean you support home directory expansion ?

rudder-lang/tools/rudderc-dev.conf
```diff
@@ -1,7 +1,8 @@
 [shared]
     stdlib="libs/" # only required field for rudderc
     generic_methods="tools/generic_methods.toml"
-    ncf="/usr/share/ncf"
+    ncf="/usr/share/ncf/tree/30_generic_methods/"
```
amousset, line 4: maybe rename "ncf" to "methods" in this case?

peckpeck:
ping

rudder-lang/tools/generate_lib
gpoblon: Ok using tilde would force to import perl libraries + their dependencies. Therefore I remove it, forcing users to write full path

rudder-lang/tests/test_files/source_rl/s_basic.rl.cf
```diff
@@ -1,13 +1,12 @@
-# generated by rudder-lang
+# @generated by rudderc
```
peckpeck, line 1: @ ?

PR #3118, opened 2020-07-07
Fixes #16402: Cannot set Compliance reporting mode on "Non compliant only" in settings tab

RaphaelGauthier:
https://issues.rudder.io/issues/16402


fanf:
OK, merging this PR

PR #3121, opened 2020-07-10
Fixes #17904: make use of conditional and flow statements for dsc

gpoblon:
https://issues.rudder.io/issues/17904


gpoblon:
PR rebased

PR #3132, opened 2020-07-21
Fixes #18015: Remove technique from technique tree when deleting technique in the technique editor

VinceMacBuche:
https://issues.rudder.io/issues/18015

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -79,6 +79,7 @@ import com.normation.rudder.services.workflows.WorkflowLevelService
 import com.normation.utils.Control
 import net.liftweb.common.Box
 import net.liftweb.common.EmptyBox
+import com.normation.rudder.repository.WoDirectiveRepository
 
 trait NcfError extends RudderError {
   def message : String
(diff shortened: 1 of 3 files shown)


VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -149,6 +151,14 @@ class TechniqueWriter (
                      }
 
       _          <- archiver.deleteTechnique(techniqueName,techniqueVersion, category.id.name.value, modId,committer, s"Deleting technique ${techniqueName}/${techniqueVersion}")
```
fanf, line 153: You need to first delete active technique, then on FS, because active technique deletion won't work if there's directives

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
VinceMacBuche: we prevent it if there is directives, or if we force the deletion we delete directives first, but I'm ok with you we must do this before

fanf:
OK, squash merging this PR

PR #3138, opened 2020-07-23
Fixes #17994: add a backtrace option for error handling

gpoblon:
https://issues.rudder.io/issues/17994


gpoblon:
implementation + backtrace for panics but no backtrace for std errors yet. 
The backtrace can be called + created with 1 line, from wherever in the code for debug purposes

rudder-lang/src/logger.rs
```diff
@@ -2,6 +2,8 @@
 // SPDX-FileCopyrightText: 2019-2020 Normation SAS
 
 use crate::Action;
+use backtrace::{ Backtrace, BacktraceSymbol };
```
amousset, line 5: does not look like `rustfmt` was run on this

gpoblon: oopsie

rudder-lang/src/opt.rs
```diff
@@ -57,4 +57,8 @@ pub struct Opt {
     /// Use json logs instead of human readable output
     #[structopt(long, short)]
     pub json_log: bool,
-}
\ No newline at end of file
+
+    /// Generate a backtrace in case an error occurs
```
amousset, line 61: Why is it an option? Do you think people might prefer without backtrace?

gpoblon: Backtrace is useful for rudder-lang contributors, but end users won't really make use of it

gpoblon:
add env_variable to have a backtrace on errors

peckpeck:
rebase needed

amousset:
the pom.xml commits should not be there

gpoblon:
OK, squash merging this PR

PR #3145, opened 2020-07-28
Fixes #17955: Add na reporting to CFEngine generator

amousset:


amousset:
OK, squash merging this PR

PR #3152, opened 2020-07-30
Fixes #18068: Add source information in enum expression

peckpeck:
https://issues.rudder.io/issues/18068

### DIFF
--- a/rudder-lang/src/parser/tests.rs
+++ b/rudder-lang/src/parser/tests.rs
@@ -4,6 +4,7 @@
 use super::*;
 use maplit::hashmap;
 use nom::Err;
+use pretty_assertions::assert_eq;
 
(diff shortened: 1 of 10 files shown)


rudder-lang/src/parser/tests.rs
```diff
@@ -1091,11 +1135,11 @@ fn test_pstatement() {
                 "case".into(),
                 vec![
                     (
-                        map_res(penum_expression, "ubuntu").unwrap().1,
+                        map_res(penum_expression, "ubuntu ").unwrap().1,
```
amousset, line 1138: is this really normal?

peckpeck: yes, parsers are eager on whitespaces, so the "source" field always contains final spaces

-> they are need when comparing the result of 2 parsing

PR #3162, opened 2020-08-06
Fixes #18093: Simplify queries to detects runs in database, as all runs are complete

ncharles:
https://issues.rudder.io/issues/18093


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
```diff
@@ -103,10 +103,9 @@ final case class RoReportsExecutionRepositoryImpl (
                     reportsexecution where (nodeid, insertionid) in (
                       select nodeid, max(insertionid) as insertionid
                         from reportsexecution
-                        where complete = true
                         and nodeid in """ ++
               Fragment.const(s"""(values ${nodes.mkString(",")} )""") ++
-              fr"""
+              fr"""c
```
ElaadF, line 108: Is that a typo?

ncharles: it is

ncharles:
Commit modified

PR #3163, opened 2020-08-07
Fixes #15024: It should be forbidden to create a user technique with the same name as n existing technique

ElaadF:
https://issues.rudder.io/issues/15024


fanf:
Hum, it seems that my review wasn't commited. I will need to review it again.

fanf (approved):

ElaadF:
Commit modified

PR #3164, opened 2020-08-07
Fixes #18094: add inventory parsing perfomance test

ncharles:
https://issues.rudder.io/issues/18094


ncharles:
Commit modified

fanf (changes requested):
We can't commit that many almost identical files, it will be a nightmare to update/maintain (and grep and etc). 
Could it be possible to have one template and generate from it the N hundreds needed for the test?

ncharles:
I'm unsure of a workable process for that. Closing this PR for the moment

PR #3166, opened 2020-08-07
Fixes #18095: update json tests and doc to match new format 

gpoblon:
https://issues.rudder.io/issues/18095


gpoblon:
PR rebased

PR #3173, opened 2020-08-14
Fixes #18107: Context structure members should be private

peckpeck:
https://issues.rudder.io/issues/18107

### DIFF
--- a/rudder-lang/src/ast/resource.rs
+++ b/rudder-lang/src/ast/resource.rs
@@ -345,11 +345,7 @@ impl<'src> Statement<'src> {
         // we must not capture the context in this common getter since it is used
         // as mutable elsewhere in this function
         let common_getter = |ctx: &VarContext<'src>, k: Token<'src>| {
-            ctx
-                .variables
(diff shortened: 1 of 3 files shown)


rudder-lang/src/ast/resource.rs
```diff
@@ -345,11 +345,7 @@ impl<'src> Statement<'src> {
         // we must not capture the context in this common getter since it is used
         // as mutable elsewhere in this function
         let common_getter = |ctx: &VarContext<'src>, k: Token<'src>| {
-            ctx
-                .variables
-                .get(&k)
-                .or_else(|| global_context.variables.get(&k))
-                .map(VarKind::clone)
+            ctx.get(&k).or_else(|| global_context.get(&k))
```
gpoblon, line 348: I guess you could alternatively call directly `or` :  ctx.get(&k).or(global_context.get(&k))

peckpeck: or_else has the advantage of using a function call that is not done when there is no need to

PR #3175, opened 2020-08-15
Fixes #18115: Implement struct type

peckpeck:
https://issues.rudder.io/issues/18115


rudder-lang/libs/cfengine_core.rl
```diff
@@ -118,8 +118,10 @@ let connection.ip
 let connection.hostname
 
 # `match` global variables
-let match
+# this kind of declaration deosn't work at the moment
+#let match
 
+# We should probably remove those
```
amousset, line 124: definitely, especially as we don't provide cf-monitord with rudder

PR #3176, opened 2020-08-16
Fixes #18106: Context should contain only variable->type mapping

peckpeck:
https://issues.rudder.io/issues/18106

### DIFF
--- a/rudder-lang/src/ast.rs
+++ b/rudder-lang/src/ast.rs
@@ -18,6 +18,7 @@ use std::{
     cmp::Ordering,
     collections::{HashMap, HashSet},
 };
+use crate::ast::context::VarType;
 
 // TODO v2: type inference, compatibility metadata
 // TODO aliases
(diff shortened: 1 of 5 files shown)


rudder-lang/src/ast.rs
```diff
@@ -171,16 +172,16 @@ impl<'src> AST<'src> {
...
                 variable,
-                Value::from_static_pvalue(value.clone()).unwrap(),
+                &Value::from_static_pvalue(value.clone()).unwrap(),
             ) {
                 self.errors.push(e);
             }
             let getter = |k| self.context.get(&k);
             match Value::from_pvalue(&self.enum_list, &getter, value) {
-                Err(e) => self.errors.push(e),
+                Err(e) => { self.errors.push(e); },
```
gpoblon, line 184: can't it be inline?

rudder-lang/src/ast.rs
```diff
@@ -171,16 +172,16 @@ impl<'src> AST<'src> {
...
                 Ok(val) => {
                     self.variable_declarations.insert(variable, val);
```
gpoblon, line 186: here too

peckpeck:
PR updated with a new commit

rudder-lang/src/ast.rs
peckpeck: not this one, it needs a ';' to ignore return value

peckpeck:
OK, squash merging this PR

PR #3178, opened 2020-08-18
Fixes #18108: Parsing of variable declaration and definition should be split

peckpeck:
https://issues.rudder.io/issues/18108

### DIFF
--- a/rudder-lang/src/ast.rs
+++ b/rudder-lang/src/ast.rs
@@ -28,9 +28,10 @@ use crate::ast::context::VarType;
 #[derive(Debug)]
 pub struct AST<'src> {
     errors: Vec<Error>,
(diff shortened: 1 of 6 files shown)


rudder-lang/src/ast.rs
```diff
@@ -168,27 +171,42 @@ impl<'src> AST<'src> {
...
             }
         }
     }
 
+    /// Insert the variables declarations into the global context
+    fn add_magic_variables(&mut self, variables: Vec<PVariableDecl<'src>>) {
+        for variable in variables {
+            let PVariableDecl { metadata, name, sub_elts, var_type} = variable;
+            match VarType::from_ptype(var_type, sub_elts) {
+                Ok(_type) =>
```
gpoblon, line 201: type is used, maybe remove `_`

PR #3180, opened 2020-08-18
First go at changing the way compliance is computed

ncharles:
Two differents system coexist:
- the one for stre run agent, that goes over the reportsexecution, compute compliance, and invalidate cache by setting its compliance
- everything else, that work as it used to be

Types are awefull
Class are a mess
Many v2 , many assumption probably wrong, but that's a first change to see if it actually does something


webapp/sources/rudder/rudder-core/src/main/resources/Migration/dbMigration-6.1-7.0-compliance_computation_change.sql
```diff
@@ -0,0 +1,53 @@
...
+
+/*
+  First we need to mark all reports whose compliance was already computed as with a compliance
+  lastid in statusupdate tells us which reports were last considered
+  this is unfortunately quite a slow query on large systems
+*/
+
+ALTER TABLE ReportsExecution
+ADD COLUMN insertiondate timestamp,
+ADD COLUMN  compliancecomputatiodate timestamp;
```
amousset, line 46:
```suggestion
ADD COLUMN  compliancecomputationdate timestamp;
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepositoryImpl.scala
```diff
@@ -158,6 +170,13 @@ final case class WoReportsExecutionRepositoryImpl (
 
   import db._
 
+  def setComplianceComputationDate(runs: List[UncomputedAgentRun]): IOResult[Int] = {
+    val updateKeys = runs.map(x => (x.nodeId, x.date, x.nodeConfigId))
+    val sql = """UPDATE reportsexecution set compliancecomputatiodate = now() where nodeid = ? and date = ? and nodeconfigid = ?"""
```
amousset, line 175:
```suggestion
    val sql = """UPDATE reportsexecution set compliancecomputationdate = now() where nodeid = ? and date = ? and nodeconfigid = ?"""
```

webapp/sources/rudder/rudder-core/src/main/resources/Migration/dbMigration-6.1-7.0-compliance_computation_change.sql
```diff
@@ -42,12 +42,12 @@
...
-ADD COLUMN insertiondate timestamp,
-ADD COLUMN  compliancecomputatiodate timestamp;
+ADD COLUMN insertiondate timestamp default now(),
+ADD COLUMN  compliancecomputationdate timestamp;
 
-UPDATE ReportsExecution set compliancecomputatiodate = now() where insertionId <= (select lastid from statusupdate);
+UPDATE ReportsExecution set compliancecomputationdate = now() where insertionId <= (select lastid from statusupdate);
 
-CREATE INDEX reportsexecution_uncomputedrun_idx on ReportsExecution (compliancecomputatiodate) where compliancecomputatiodate IS NULL;
+CREATE INDEX reportsexecution_uncomputedrun_idx on ReportsExecution (compliancecomputatinodate) where compliancecomputationdate IS NULL;
```
amousset, line 50:
```suggestion
CREATE INDEX reportsexecution_uncomputedrun_idx on ReportsExecution (compliancecomputationdate) where compliancecomputationdate IS NULL;
```

ncharles:
This code is a v0, and it already succeed into computing compliance for received reports -but initial state is broken, and reports get missing after less than 5 minutes)

What is missing:

- cleaning the mess - types are leaking everywhere, methods are in class that should not be there
- correct initial state: cache is not correctly initialized, and a particular attention should be given to this part
- cache invalidation: when we clear cache, when we generate new policies, cache is invalidated. For new policy, we could be more clever and simply pop in the queue that a new NodeConfiguration is available, and give it to the cache to recompute immediately
- the queue shoudl take actions: push new compliance to it, push new config to it, and then poped in order (and don't drop anything from that)
- detecting that a node did not answer in the last x minutes, and store a no report
- change the tests

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,7 +265,7 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
    * It's a List and not a Set, because we want to keep the precedence in
    * invalidation request.
    */
-  private[this] val invalidateComplianceRequest = Queue.dropping[List[NodeId]](1).runNow
+  private[this] val invalidateComplianceRequest = Queue.dropping[List[(NodeId, Option[NodeStatusReport])]](1).runNow
```
ncharles, line 268: Here i'd like an action that can be
NewRunComputed[NodeStatusReports]
InvalidateCompliance[Option[NodeConfiguration]]
NoReports

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -275,12 +278,25 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
     // batch node processing by slice of batchSize.
     // Be careful, sliding default step is 1.
-    ZIO.foreach_(invalidatedIds.sliding(batchSize,batchSize).to(Iterable))(nodeIds =>
-      (for {
-        updated <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(nodeIds.toSet, Set()).toIO
-        _       <- IOResult.effectNonBlocking { cache = cache ++ updated }
-        _       <- ReportLoggerPure.Cache.debug(s"Compliance cache updated for nodes: ${nodeIds.map(_.value).mkString(", ")}")
-      } yield ()).catchAll(err => ReportLoggerPure.Cache.error(s"Error when updating compliance cache for nodes: [${nodeIds.map(_.value).mkString(", ")}]: ${err.fullMsg}"))
+
+    ZIO.foreach_(invalidatedIds.sliding(batchSize,batchSize).to(Iterable))(nodeIdsWithOptionnalCompliance =>
```
ncharles, line 282: here i'd need to do a takewhile

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -275,12 +278,25 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+      // needs to be invalidated
+      // so groupBy NodeId, and take last element for each
+
+      {
+        val (nodeIdsWithoutCompliance, nodeIdsWithCompliance) = nodeIdsWithOptionnalCompliance.groupBy(_._1).map { case (nodeId, list) => (nodeId, list.last._2) }.partition(_._2.isDefined)
+
+        for {
+          updated <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(nodeIdsWithoutCompliance.keys.toSet, Set()).toIO
+          _       <- IOResult.effectNonBlocking {
+            cache = cache ++ updated
```
ncharles, line 294: cache ought to be more complete with
NodeId, LastRunDate, ExpectedConfiguration, NodeConfigurationStatus

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -121,12 +121,14 @@ CREATE TABLE ReportsExecution (
 , complete     boolean NOT NULL
```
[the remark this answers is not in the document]
fanf: we discussed that, and a run could have a bug and be incomplete. But that notion of incomplete is totally different from the old notion of "we didn't get all reports yet". So to avoid confusion (and bug mixing one notion with the other), we will remove that old `complete` notion, and we will treat broken runs with a dedicated state of `RunAndConfigInfo`.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,32 +290,95 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+        } yield ()*/
+      }.catchAll(err => ReportLoggerPure.Cache.error(s"Error when updating compliance cache for nodes: [${queueAction.map(_._1).map(_.value).mkString(", ")}]: ${err.fullMsg}"))
     )
   )
 
   // start updating
   updateCacheFromRequest.forever.forkDaemon.runNow
 
 
+  private[this] def performAction(actions: List[CacheComplianceQueueAction]) = {
```
ncharles, line 348: the type of that is ZIO[Any, RudderError, Unit] and it makes me sad @fanf

ncharles:
at this point, this code complie and "works" - cache invalidation is still not perfect, we should detect automatically run expired.
Sometimes this query is slow
```
select r.nodeid, r.date, r.nodec
onfigid, r.complete, r.insertionid,
                                 c.nodeid, c.nodeconfigid, c.begindate, c.enddate, c.configuration
                            from (
                         select reportsexecution.nodeid, reportsexecution.date, reportsexecution.nodeconfigid, reportsexecution.complete, reportsexecution.insertionid from
                            reportsexecution where (nodeid, insertionid) in (
                              select nodeid, max(insertionid) as insertionid
                                from reportsexecution
                                where nodeid in  (
) 
                                GROUP BY nodeid
                             )
                          ) as r  left outer join nodeconfigurations as c
                                 on r.nodeId = c.nodeid and r.nodeconfigid = c.nodeconfigid

```

which has not been changed yet

Sometimes I have only 152 nodes out of 4559, sometimes 1000, but overall it seems to perform-ish ok-ish

ncharles:
At this point, i have something that looks workable, from a distance.
Pain points:
* initial loading of compliance is very expensive
* i removed the batching in treating nodes cache, hence the previous case. I need to put it back
* in https://github.com/Normation/rudder/pull/3180/commits/30db9c846eed661054c821ff26393ecc6b37df0f I tried to have the number of nodes updated, turns out it prevented it from runing altogether
* cache seems to work as expected
* however, no significant improvement on the load of the system (but the store run agent is always taking around 1 seconds or less, compared to more than 5 more than often

Need to add back batchs in treating node cache update, get the number of updated nodes, and clean the code and tests
This would make a first change

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
```diff
@@ -109,13 +109,15 @@ class ReportsExecutionService (
...
 
-            fetchRunsAndCompliance()
-
-            hookForChanges(lastReportId, maxReportId)
-
-            val executionTime = DateTime.now().getMillis() - startTime.getMillis
-            logger.debug(s"[${FindNewReportsExecution.SERVICE_NAME} #${processId}] (${executionTime} ms) " +
-              s"Added or updated xxxxx agent runs")
+            for {
+              nodes         <- fetchRunsAndCompliance()
```
fanf: nodes         <- fetchRunsAndCompliance().toBox

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,32 +287,87 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+          val impactedNodeIds = actions.map(x => x.nodeId)
+          ZIO.foreach(impactedNodeIds.grouped(batchSize).to(Seq)) { updatedNodes =>
+            for {
+              updated <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(updatedNodes.toSet, Set()).toIO
+              _       <- IOResult.effectNonBlocking {
+                           cache = cache ++ updated
+                         }
+              _ <- ReportLoggerPure.Cache.debug(s"Compliance cache recomputed for nodes: ${updated.keys.map(_.value).mkString(", ")}")
+            } yield ()
+          }.unit
```
ncharles, line 353: ok, this doesn't do anything anymore

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,32 +287,91 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+          } yield ()
+
+
+          /*for {
+            updated <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(impactedNodeIds.toSet, Set()).toIO
+            _       <- IOResult.effectNonBlocking {
+              cache = cache ++ updated
+            }
+            _ <- ReportLoggerPure.Cache.debug(s"Compliance cache recomputed for nodes: ${updated.keys.map(_.value).mkString(", ")}")
+          } yield ()*/
```
fanf, line 364: this can be deleted

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,32 +287,91 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
 
 
+  /**
+   * Do something with the action we received
+   * All actions must have *exactly* the *same* type
+   */
+  private[this] def performAction(actions: List[CacheComplianceQueueAction]): IOResult[Unit] = {
+    ReportLoggerPure.Cache.debug(s"Performing action ${actions.headOption}") *>
+    // get type of action
+      (actions.headOption match {
```
fanf, line 329: one indent too many here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,32 +287,91 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+              x <- ZIO.foreach(impactedNodeIds.grouped(batchSize).to(Seq)) { updatedNodes =>
+                  for {
+                  updated <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(updatedNodes.toSet, Set()).toIO
+                  _       <- IOResult.effectNonBlocking {
+                  cache = cache ++ updated
+                }
+                _ <- ReportLoggerPure.Cache.debug(s"Compliance cache recomputed for nodes: ${updated.keys.map(_.value).mkString(", ")}")
+                } yield ()
+            }
+          } yield ()
```
fanf, line 355: bad indent for `for`, shoud look like: 

```
              x <- ZIO.foreach(impactedNodeIds.grouped(batchSize).to(Seq)) { updatedNodes =>
                       for {
                         updated <- defaultFindRuleNodeStatusReports.findRuleNodeStatusReports(updatedNodes.toSet, Set()).toIO
                         _       <- IOResult.effectNonBlocking {
                                      cache = cache ++ updated
                                    }
etc
```
(can even be `IOResult.effectNonBlocking { cache = cache ++ updated }`)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -262,32 +287,91 @@ trait CachedFindRuleNodeStatusReports extends ReportingService with CachedReposi
...
+              cache = cache ++ updated
+            }
+            _ <- ReportLoggerPure.Cache.debug(s"Compliance cache recomputed for nodes: ${updated.keys.map(_.value).mkString(", ")}")
+          } yield ()*/
+      }
+    })
+  }
+  /**
+   * Group all actions queue by the same type, keeping the global order.
+   */
```
fanf, line 370: I don't understand how order is kept. IIUC, if you have that in queue:
````
List(
  InsertNodeInCache(node1)
, RemoveNodeInCache(node1) 
, InsertNodeInCache(node1)
)
```
It will becomes: 
```
List(
  List(
    InsertNodeInCache(node1)
  , InsertNodeInCache(node1)
  )
, List(
    RemoveNodeInCache(node1) 
  )
)
``` 
Which is semantically different.

fanf (commented):
Lots of small remarks, but I think the core logic of the PR is sound, and it simplifies so much everything!

A question nonetheless with the reordering of compliance cache event for batch, I think I missed something.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -166,6 +174,7 @@ final case class NoUserRulesDefined(
  */
 final case class NoReportInInterval(
     expectedConfig: NodeExpectedReports
+  , expirationDateTime: DateTime
```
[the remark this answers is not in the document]
ncharles: it's because we discover that it has no report in interval at time T, and we need to ensure at T+expire that it's still the case (and if so, serialize it)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
ncharles: no, it does not reorder, it create list of list in same order
`
List(
  InsertNodeInCache(node1)
, InsertNodeInCache(node2)
, RemoveNodeInCache(node1)
, InsertNodeInCache(node2)
, InsertNodeInCache(node1)
)

becomes

List(
  List(
    InsertNodeInCache(node1)
  , InsertNodeInCache(node2)
  )
, List(
    RemoveNodeInCache(node1) 
  )
 , List(
    InsertNodeInCache(node2)
  , InsertNodeInCache(node1)
  )
`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
```diff
@@ -133,67 +135,44 @@ class CachedReportsExecutionRepository(
...
+      runs = agentsRuns.map { case (nodeId, nodeAndConfigId) =>
+        (nodeId,
+          Some(AgentRunWithNodeConfig(
+               AgentRunId(nodeId, lastRunByNode(nodeId).agentRunId.date)
+             , expectedReports.get(nodeAndConfigId).map(optionalExpectedReport => (nodeAndConfigId.version, optionalExpectedReport))
+             , true
+             , lastRunByNode(nodeId).insertionId)))}
+      // and finally mark them read. It's so much easier to do it now than to carry all data all the way long
+      _  <- writeBackend.setComplianceComputationDate(unprocessedRuns.toList)
+    } yield {
```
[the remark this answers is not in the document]
ncharles: I think the cache doesn't have access to the getExpectedReports

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: AH OK :) I could be smarter in a second time, and group same event as long as they belong to different nodes, and simplify action (like insert / delete / insert => insert). Lots of opportunities now that we have a sound ground :)

ncharles: we don't want to be clever on that, because we don't want to loose data when stored in compliancelevels 
(or the necessary cleverness would be too hard to reach a correct result)

fanf: we will, at some point (for perf or whatever). Still, the comment about keeping all intermediate compliance for historization is very interesting and it must be put here!

ncharles:
replaced by https://github.com/Normation/rudder/pull/3227

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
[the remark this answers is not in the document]
ncharles: indeed, the backend is simply jdbc, and doesn't have access to other services. I'd rather leave it like that, as next iteration will change the getExpectedReports part significantly, with a cache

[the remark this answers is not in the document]
fanf: I would prefer to have a totally fine, perfect by itself iteration, because we never know when the following one can be - especially true in current context :)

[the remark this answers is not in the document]
ncharles: ok but it will be messy

PR #3194, opened 2020-08-27
Fixes #10216: "id" setting is silently ignored for new group created through the REST API.

ElaadF:
https://issues.rudder.io/issues/10216


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
```diff
@@ -353,7 +353,10 @@ class GroupApiService2 (
     implicit val prettify = restExtractor.extractPrettify(req.params)
     val modId = ModificationId(uuidGen.newUuid)
     val actor = RestUtils.getActor(req)
-    val groupId = NodeGroupId(req.param("id").getOrElse(uuidGen.newUuid))
```
VinceMacBuche, line 356: It could have been done like we do in Rules / Directives 

      val id = restExtractor.extractId(req)(x => Full(NodeGroupId(x))).map(_.getOrElse(NodeGroupId(uuidGen.newUuid)))

ElaadF:
Commit modified

VinceMacBuche (approved):

ElaadF:
OK, merging this PR

PR #3195, opened 2020-08-28
Fixes #18149: Upgrade to ZIO 1.0.1

fanf:
https://issues.rudder.io/issues/18149
Update went smoothly, three noticable changes: 

- update cats dependencies because of `zio-compatc-cats`. I think I found back a consistent set of lib for `fs2` etc, 
- change in `TestClock.sleep`. Now it actually sleep the fiber, so it needs a `fork`. Also, `TestClock.adjust` doesn't need a sleep to notify the adjust anymore (good!);
- the biggest change is a change in `foreach` signature, which used to take only a `items: Iterable[A]` parameter and was *always* returning a `List[A]` and that now takes 5 differents parameters (`Option[A]`, `Map[A, B]`, `Set[A]`, `NonEmptyChunk[A]`, and `Collection[+Element] <: Iterable[Element]` and return the same type). That need some `.toList` or spelling to scalac that no, really, I want an iteration on map paires, not building a new map all around to make everyone happy, but I think the experience will be more `traverse`-like in the end, so for the better. 

All that being said, I didn't found any deadlock, even with only one thread available, and I did see a noticable perf improvement (around 15% in policy writting test, which is big (on three iteration, we now have `870 ms / 793 ms / 760 ms` versus `1089 ms / 872 ms / 944 ms` - of course, it will need to be tested in real life / real load)

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionReportUnmarshaller.scala
+++ b/webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionReportUnmarshaller.scala
@@ -355,6 +355,7 @@ class FusionReportUnmarshaller(
       }
     }
 
+
     /*
      * Process agents. We want to keep the maximum number of agents,
      * ie if there's two agents, and XML for one is not valid, still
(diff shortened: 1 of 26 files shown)


fanf:
PR rebased

VinceMacBuche (approved):

PR #3207, opened 2020-09-10
Fixes #18168: command line rework

gpoblon:
https://issues.rudder.io/issues/18168

### DIFF
--- a/rudder-lang/src/bin/rudderc.rs
+++ b/rudder-lang/src/bin/rudderc.rs
@@ -56,27 +56,13 @@ use rudderc::{compile::compile_file, io, logger::Logger, opt::Opt, technique::ge
 // TODO use termination
 
 fn main() {
-    // Parse options
-    let opt = Opt::from_args();
-
-    // Compile is the default
-    let action = if opt.translate {
-        Action::Translate
-    } else {
-        Action::Compile
-    };
-
-    let logger = if opt.json_log {
-        Logger::Json
-    } else {
-        Logger::Terminal
-    };
-
+    let command = Opt::from_args();
+    let (logger, log_level, has_backtrace) = command.extract_logging_infos();
+    let action = command.to_action();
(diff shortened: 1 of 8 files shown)


rudder-lang/src/bin/rudderc.rs
```diff
@@ -86,12 +72,19 @@ fn main() {
...
-    // json -> rl (/ cf / dsc ?)
-    // rl -> json / cf / dsc
+    // read = rl -> json
+    // migrate = cf -> rl
+    // compile = rl -> cf / dsc
+    // generate = json -> rl / cf / dsc
+    // TODO make the right calls
     let result = match action {
         Action::Compile => compile_file(&ctx, true),
-        Action::Translate => generate(&ctx),
```
peckpeck, line 94: Where does the generate call go ?

rudder-lang/src/io.rs
```diff
@@ -32,187 +32,168 @@ struct Config {
...
-    pub mode: Action,
+    pub input: PathBuf,
+    pub output: PathBuf,
+    // Unique fields
     pub format: Format,
+    pub action: Action,
+}
+// TODO might try to merge io.rs in here
+impl IOContext {
+    pub fn set(action: Action, opt: &GenericOptions, format: Option<Format>) -> Result<Self> {
```
peckpeck, line 48: set is a too generic name, what does it do ? 
isn't thei a new() ?

rudder-lang/src/bin/rudderc.rs
gpoblon: disappears in next PR, unused since actions are not Translate and Compile anymore

rudder-lang/src/io.rs
gpoblon: yes, will be updated

peckpeck (approved):

PR #3210, opened 2020-09-11
Fixes #18197: pending changes of resources (files deleted / modified / added ) are commited automatically during upgrade

VinceMacBuche:
https://issues.rudder.io/issues/18197


fanf:
Actually, I think that it's what we want. What could be a reason other than "some action was interrupted" to have uncommited files in these directories ?

VinceMacBuche:
we were editing and we did not save, this is why we have uncommitted files

fanf:
ok.
I think we must find way to avoid that case, it's an UI nightmare. Perhaps we should just commit file when they are uploaded. Or delete them if not saved and moving away (with a pop-up "there is uncommited resources, what to do? Delete or save ?").

fanf (approved):

VinceMacBuche:
OK, merging this PR

PR #3217, opened 2020-09-17
Fixes #17250: Webapp does not regerate policies when webdav password is changed, breaking inventories after 5.0 -> 6.0 upgrade

ncharles:
https://issues.rudder.io/issues/17250


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/TriggerPolicyUpdate.scala
```diff
@@ -65,10 +67,11 @@ class TriggerPolicyUpdate(
     try {
       if (file.exists) {
```
fanf, line 68: Why keep that test? I would just remove it and deprecate both flag files. And document it. 
And in 7.0 remove the files enterily.

fanf (changes requested):

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/TriggerPolicyUpdate.scala
ncharles: so that people are not surprised as some are used to use it

fanf: ok, so make the test only to adapt message, and put out the start, wichi is not subject to condition. 
And I think we need to adapt the message to tell "you don't need that file, since 6.1.x start is automatic". 
And I think we must do it only in 6.1: 6.0 is EOL and it complicate the explanation about when the behavior changed.

ncharles: but I do indeed need to remove the other one

ncharles:
PR replaced by https://github.com/Normation/rudder/pull/3218

PR #3218, opened 2020-09-17
Fixes #17250: Webapp does not regerate policies when webdav password is changed, breaking inventories after 5.0 -> 6.0 upgrade

ncharles:
https://issues.rudder.io/issues/17250

Replacing previous PR: https://github.com/Normation/rudder/pull/3217


ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -174,57 +173,16 @@ final class AsyncDeploymentActor(
     )
   }
 
-  // That flag file is used to determine if a policy update was running when the webapp stopped so we can launch a new one when starting
-  val policyUpdateRunningFlagPath = "/opt/rudder/etc/policy-update-running"
```
fanf, line 178: Shouldn't we delete that one during `rudder-upgrade`?

fanf (changes requested):
Appart from two small remark, seems good! Is there only rudder that set file `/opt/rudder/etc/policy-update-running`?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
ncharles: good point

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
ncharles: done

fanf (approved):

fanf:
OK, squash merging this PR

PR #3230, opened 2020-09-30
Fixes #18256: impl new logger

gpoblon:
https://issues.rudder.io/issues/18256

### DIFF
--- a/rudder-lang/src/logger.rs
+++ b/rudder-lang/src/logger.rs
@@ -1,0 +1,174 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// SPDX-FileCopyrightText: 2019-2020 Normation SAS
+
+use crate::error::*;
+use lazy_static::lazy_static;
+use serde::{Deserialize, Serialize};
+use std::sync::RwLock;
+use std::{fmt, str::FromStr};
+
+// to keep LOGS usage as safe as possible, make sure it is only called from this file
+// any outside module call is just a push that happens through macros
+// and a clone to get a safe clone of the global to work on from outside the module
+
+lazy_static! {
+    pub static ref LOGS: RwLock<Logs> = RwLock::new(Logs {
+        logs: Vec::new(),
+        log_level: LogLevel::Warn,
+    });
+}
+
+pub fn clone_logs() -> Logs {
+    LOGS.read().expect("bug: failed to read logs").clone()
+}
+
+// pub static mut LOGS.log_level: LogLevel = LogLevel::Warn;
+pub fn set_global_log_level(log_level: LogLevel) {
+    LOGS.write()
+        .expect("bug: failed to update logs")
+        .update_log_level(log_level)
+}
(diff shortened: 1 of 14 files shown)


gpoblon:
Commit modified

gpoblon:
Commit modified

rudder-lang/src/logger.rs
```diff
@@ -0,0 +1,177 @@
...
+
+// pub static mut LOGS.log_level: LogLevel = LogLevel::Warn;
+pub fn set_global_log_level(log_level: LogLevel) {
+    unsafe { LOGS.update_log_level(log_level) }
+}
+
+#[macro_export]
+macro_rules! error {
+    ($ ( $ arg : tt ) *) => ({
+        unsafe {
```
peckpeck, line 28: why is this unsafe ?

rudder-lang/src/logger.rs
```diff
@@ -0,0 +1,177 @@
...
+            if LOGS.is_level_included(LogLevel::Error) {
```
peckpeck, line 29: i think you need a "use logger::LOGS" in this macro

rudder-lang/src/logger.rs
gpoblon: because `static mut` manipulation can only be done in unsafe blocks

peckpeck: so it should be reblaced with a refcell

gpoblon: static implies Sync so RefCell won't do the trick, I need to use Mutexes/Arcs

gpoblon:
PR updated with a new commit

gpoblon:
PR updated with a new commit

rudder-lang/src/logger.rs
```diff
@@ -2,71 +2,74 @@
...
+        .update_log_level(log_level)
 }
 
 macro_rules! error {
     ($ ( $ arg : tt ) *) => ({
         use crate::logger::LOGS;
-        if LOGS.with(|logs| logs.borrow().is_level_included(LogLevel::Error)) {
-            LOGS.with(|logs| logs.borrow_mut().push(LogLevel::Error, format!( $ ( $ arg ) * )));
+        if LOGS.read().expect("bug: failed to read logs").is_level_included(LogLevel::Error) {
+            LOGS.write().expect("bug: failed to update logs").push(LogLevel::Error, format!( $ ( $ arg ) * ));
```
peckpeck, line 36: Are you sure this work ? 
The read result here still holds (it will be freed after the closing brace) so the write call will wait indefinitely.

peckpeck (approved):

gpoblon:
OK, squash merging this PR

gpoblon:
OK, squash merging this PR

gpoblon:
Commit modified

gpoblon:
OK, squash merging this PR

PR #3232, opened 2020-09-30
Fixes #18266: Uncommited rudder_reporting.cf  when creating a with a condition

VinceMacBuche:
https://issues.rudder.io/issues/18266


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -746,6 +746,7 @@ class TechniqueArchiverImpl (
     val filesToAdd = (
       "metadata.xml" +:
       "technique.cf" +:
+      "rudder_reporting.cf" +:
```
fanf, line 749: can you sort them alpha-num? (or at least let the technique* together at the end?)

fanf: It can be self merged afterward

VinceMacBuche:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3244, opened 2020-10-04
Fixes #18289: Missing system variable "POLICY_SERVER_KEY"

fanf:
https://issues.rudder.io/issues/18289

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
@@ -139,7 +139,11 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                            , constraint = Constraint(mayBeEmpty=true)
                         )
     , SystemVariableSpec(
-                       "MANAGED_NODES_KEY" , "CFEngine KEY (modulus with transformation) of nodes managed by the policy server"
+                       "MANAGED_NODES_KEY" , "Hash of CFEngine KEY of nodes managed by the policy server"
+                                           , constraint = Constraint(mayBeEmpty=true)
+                        )
+    , SystemVariableSpec(
+                       "POLICY_SERVER_KEY" , "Hash of CFEngine KEY of the policy server"
                                            , constraint = Constraint(mayBeEmpty=true)
                         )
       // end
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
```diff
@@ -142,6 +142,10 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                        "MANAGED_NODES_KEY" , "CFEngine KEY (modulus with transformation) of nodes managed by the policy server"
                                            , constraint = Constraint(mayBeEmpty=true)
                         )
+    , SystemVariableSpec(
+                       "POLICY_SERVER_KEY" , "CFEngine KEY (modulus with transformation) of the policy server"
```
ncharles, line 146: it is not the key, but the hash
Could you rename it to "Agent key hash of the policy server"

fanf: it's to be consistant with the `MANAGED_NODES_KEY`, which are the same thing (and should be name otherwise, but are not)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
```diff
@@ -142,6 +142,10 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                        "MANAGED_NODES_KEY" , "CFEngine KEY (modulus with transformation) of nodes managed by the policy server"
```
ncharles, line 142:
```suggestion
                       "MANAGED_NODES_KEY" , "Hash ok CFEngine keys of nodes managed by the policy server"
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
ncharles:
```suggestion
                       "POLICY_SERVER_KEY" , "Hash of CFEngine KEY of the policy server"
```

fanf:
PR updated with a new commit

amousset:
OK, squash merging this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -356,6 +356,7 @@ class SystemVariableServiceImpl(
       val varManagedNodes      = systemVariableSpecService.get("MANAGED_NODES_NAME" ).toVariable(nodesWithCFEKey.map(_.hostname))
       val varManagedNodesId    = systemVariableSpecService.get("MANAGED_NODES_ID"   ).toVariable(nodesWithCFEKey.map(_.id.value))
       val varManagedNodesKey   = systemVariableSpecService.get("MANAGED_NODES_KEY"  ).toVariable(nodesWithCFEKey.map(_.securityTokenHash))
+      val varPolicyServerKey   = systemVariableSpecService.get("POLICY_SERVER_KEY" ).toVariable(Seq(nodeInfo.securityTokenHash))
```
ncharles, line 359: it should have been the server securityTokenHash

fanf: It is on the case where the node is a policy server. Do you mean it should have be the node's policy server?

PR #3245, opened 2020-10-06
Fixes #18300: upgrade testing loop to match new actions and dsc integration

gpoblon:
https://issues.rudder.io/issues/18300


rudder-lang/src/bin/rudderc.rs
```diff
@@ -88,7 +88,7 @@ fn main() {
     };
     let command_status = command_result.is_ok();
```
amousset, line 89: naming it `command_success` or something like that would have prevented the mistake.

gpoblon: True, fixuped. And added an explicit error for cli errors

amousset:
OK for me, I'll let @peckpeck have a look

peckpeck (approved):

gpoblon:
OK, squash merging this PR

gpoblon:
OK, merging this PR

PR #3246, opened 2020-10-06
Fixes #18303: Calling the relay-api to trigger a remote run on all nodes do an infinite loop on root server, killing it

amousset:
https://issues.rudder.io/issues/18303

### DIFF
--- a/relay/sources/relayd/src/data/node.rs
+++ b/relay/sources/relayd/src/data/node.rs
@@ -237,15 +237,27 @@ impl NodesList {
         self.neighbors_from(&self.my_id, nodes)
     }
 
+    /// Get all directly connected downstream policy servers
     pub fn my_sub_relays(&self) -> Vec<Host> {
         let mut relays = HashSet::new();
         for policy_server in self
             .list
             .data
             .values()
-            .filter_map(|v| self.list.data.get(&v.policy_server))
-            .filter(|v| v.policy_server == self.my_id)
-            .map(|v| v.hostname.clone())
+            // Extract policy servers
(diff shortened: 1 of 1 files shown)


relay/sources/relayd/src/data/node.rs
```diff
@@ -445,13 +457,24 @@ mod tests {
...
+            Ok(None)
+        );
+    }
+
     #[test]
     fn it_gets_sub_relays() {
-        let mut reference = vec![
-            "node1.rudder.local",
-            "node2.rudder.local",
-            "server.rudder.local",
```
VinceMacBuche, line 453: Haha

relay/sources/relayd/src/data/node.rs
```diff
@@ -237,15 +237,27 @@ impl NodesList {
...
 
+    /// Get all directly connected downstream policy servers
     pub fn my_sub_relays(&self) -> Vec<Host> {
         let mut relays = HashSet::new();
         for policy_server in self
             .list
             .data
             .values()
-            .filter_map(|v| self.list.data.get(&v.policy_server))
-            .filter(|v| v.policy_server == self.my_id)
```
ncharles, line 247: we had the same code in the scala part in the past, with same consequences :)

ncharles (approved):

relay/sources/relayd/src/data/node.rs
amousset: funny, isn't it?

amousset: unit tests at their best

PR #3247, opened 2020-10-07
Fixes #18302: Make possible to give a revision for directives in rule and get correct technique files

fanf:
https://issues.rudder.io/issues/18302

First version that allows to use past revisions of Techniques and Directives in rule (and directly from directive page by adding `+commid` near directive UUID. 

This is still WIP as we need to have API for directive (and unit tests) allowing to get a specific revision, and set revision for directives in rules (API to list revisions, clone a specific revision, create or clone on a branche different from master will be done in next steps, along with revision for rules).

Since first iteration of draft, main changes are: 
- `revision` is the chosen name to talk about the commit identifier. So there is no more `RevId`, `revisionId`, etc, and everything should be normalized to `revision` (or `rev` when abbreviated in code, but never in serialized form) ;
- now `revision` is a mandatory part of a `DirectiveId` which is now defined as: 
```
final case class DirectiveId(uid: DirectiveUid, rev: Revision = GitVersion.defaultRev) { ... }
```
The old "DirectiveId" type is renamed into `DirectiveUid` (for "unique identifier") and the corresponding `uid` attribute name is used in the new `DirectiveId` to refer to it. 
`uid` was prefed to `uuid` because 1/ that allows to use that name if needed even for system rules/directives for some have an human readable name (system ones), and 2/ perhaps we will want to change from uuid to something else latter on.

Below, the updated list of main changes: 

- add a `revision` for directive and allow to specify it in rule: lots of files change, but it's because we change directive identification from`DirectiveId` to `DirectiveUid + Revision`
- load a directive version given an URL in directive page: deals with web part of querying a past directive and display it in UI
- allows to add/remove directives from fullActiveTechniqueLib: this one is the main change in generation logic: when we look for all directives at start of generation, we also fetch directive with past revision when needed (and we clean-up the resulting structure to have less things in memory). In the long term, this will be replace by just having a pair of `config revision`+`facts revision` and lookup in git repositories all info ; 
- some change to `TechniqueVersion` to add `Revision` to it, and corresponding logic to fetch technique matching directive id (by default or revision specified in directive)

See also internal documentation (development notes and updated architecture in https://www.notion.so/Versionning-des-objets-de-configuration-0d29f58ac58a42d38551a9fe118fe713)


webapp/sources/ldap-inventory/inventory-repository/src/main/resources/ldap/inventory.schema
```diff
@@ -66,6 +66,13 @@ attributetype ( InventoryAttributes:1.5
   DESC 'Unique identifier for the policy server of a node'
   SUP uuid )
 
+# Revision identifier is not used in these objects, but they will be at
```
ncharles, line 69: you are not using it in the inventory, did you put it there in case nodes are versionned also in the future?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
```diff
@@ -37,290 +37,65 @@
...
- * These two steps (comparing and removing initial non-digit strings and initial
- * digit strings) are repeated until a difference is found or both strings are
- * exhausted.
- */
+import com.normation.GitVersion.RevId
+import com.normation.utils.ParseVersion
+import com.normation.utils.Version
+
 
 class TechniqueVersionFormatException(msg: String) extends Exception("The version format of a technique should be : [epoch:]upstream_version\n" + msg)
```
ncharles, line 46: this message about epoch surely doesn't make sense, and maybe the whole exception also

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -379,7 +381,10 @@ class GitTechniqueReader(
 
     val managed = Managed.make(
       for {
-        currentId <- revisionProvider.currentRevTreeId
+        currentId <- revId match {
+                       case None    => revisionProvider.currentRevTreeId
+                       case Some(r) => GitFindUtils.findRevTreeFromRevString(repo.db, r.value)
```
ncharles, line 386: I find this one really scary - i'm unsure of the perf of this one, and maybe we should consider a cache somewhere

fanf: yes, this is a PoC, and I don't want to address potential perf problem until I understand the whole design space and what need to be done. 
I think we will need to completly change the way we use ldap, even remove it and replace it by a pur cache for `(id,revId)`. But it's too early to see the whole scope of change, even more how to iterate on them.

webapp/sources/ldap-inventory/inventory-repository/src/main/resources/ldap/inventory.schema
fanf: they will, it's the only way to have versioned node properties. Same for groups. It leads to a lot of complicated question regarding external providers (for ex, what does it means for a datasource property to be versioned?)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
fanf: I think it's not used anymore and need to be removed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
ncharles: :+1: 
If you need to test that, you can create a bash script that:
creates many non empty files in git
git add these files & git commit with a random message
git remove these files & git commit with a random message
and start again

It makes a serious dent in perf of git after several thousands commit

fanf: it need to make `git gc`, and performances come back to baseline

fanf: actually, perhap not (but yes, we need to test that on big git, and also find what is the best fashion to get what we need from it)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/configuration/ConfigurationRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+ * The repository has a notion of "current version", which is commit id for head in default branch.
+ */
+trait ConfigurationRepository extends RoConfigurationRepository with WoConfigurationRepository
+
+trait RoConfigurationRepository {
+
+  /*
+   * Get a directive and its matching active technique for the given (id, version)
+   */
+  def getDirective(id: DirectiveRId): IOResult[Option[ActiveDirective]]
```
ncharles, line 76: name DirectiveRId is not very human-friendly. when parsed too quickly by the eye, it's hard to know if it's a typo or it's even possible to skip the R
I would prefer DirectiveRevId if you don't see any issue with that

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
```diff
@@ -172,12 +172,12 @@ object Doobie {
   )
 
   implicit val ReportRead: Read[Reports] = {
-    type R = (DateTime, RuleId, DirectiveId, NodeId, Int, String, String, DateTime, String, String)
+    type R = (DateTime, RuleId, DirectiveRId, NodeId, Int, String, String, DateTime, String, String)
```
ncharles, line 175: i didn't realize it would change the format on the reports from the agents

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
```diff
@@ -80,6 +80,29 @@ class CATEGORY(
   }
 }
 
+object RudderDit {
+
+  /*
+   * Create a RDN from an uuid, its attribute name, and a revesition id.
```
ncharles, line 86: typo: i *think* it's revision

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/RudderDit.scala
```diff
@@ -80,6 +80,29 @@ class CATEGORY(
...
+   * If the revesion id is the default one, RDN won't have it, and if it's
```
ncharles, line 87: same here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -1025,8 +1030,8 @@ class WoLDAPDirectiveRepository(
     for {
       con             <- ldap
       activeTechnique <- getUPTEntry(con, uactiveTechniqueId, A_ACCEPTATION_DATETIME).notOptional(s"Active technique with id '${uactiveTechniqueId.value}' was not found")
+      oldAcceptations <- ZIO.fromEither(mapper.unserializeAcceptations(activeTechnique(A_ACCEPTATION_DATETIME).getOrElse("")))
```
ncharles, line 1033: what's the semantic of moving this line?

webapp/sources/pom.xml
```diff
@@ -181,6 +186,8 @@ limitations under the License.
             <arg>-Ycache-plugin-class-loader:last-modified</arg> <!-- Enables caching of classloaders for compiler plugins -->
             <arg>-Ycache-macro-class-loader:last-modified</arg>  <!-- and macro definitions. This can lead to performance improvements. -->
             <arg>-P:silencer:checkUnused</arg>
+<!--            <arg>-P:wartremover:only-warn-traverser:org.wartremover.warts.StringPlusAny</arg>-->
+<!--            <arg>-P:wartremover:only-warn-traverser:org.wartremover.warts.ToString</arg>-->
```
VinceMacBuche, line 190: do you want to keep these commented to use it when needed easily ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
```diff
@@ -37,290 +37,65 @@
...
-    case that:TechniqueVersion => this.epoch == that.epoch && this.upsreamTechniqueVersion == that.upsreamTechniqueVersion
-    case _ => false
-  }
+  def withDefaultRevId = TechniqueVersion(version)
 
-  override lazy val hashCode : Int = 7 + 13 * epoch + 41 * upsreamTechniqueVersion.hashCode
+  // intended for debug
+  def debugString = serialize
+  // for serialisation on path
+  def serialize = revId.fold(
```
VinceMacBuche, line 67: why a fold an not just a pattern matching ?

or even a version.toVersionString + revId.map(_.value).getOrElse("") ?

webapp/sources/pom.xml
fanf: yes, I want to be able to test from time to time, at least until that PR is merged. They lead to a lot of false positive, so we can't keep them always on, but they are the only way to find an implicit `toString` on ID which is no more what we need

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
fanf: not sure, most of the time I find fold clearer (not on `Option`, though, which is a type inference horror), so I default to it. I can change it.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
```diff
@@ -37,12 +37,77 @@
...
+ * Given that, we should (at least in the begining) try to minize distance between API serialisation format and internal
+ * format. So go for the second option, and be careful when evolving methods.
+ *
+ * ==== some more evolution
+ *
+ * - forcing rev everywhere, when we want it to be almost always `defaultRev`, is not efficient. We should use
+ *   Option[RevId]
+ *   BUT it means that special attention need to be used on unserialisation: defaultValue (if serialised)
+ *   must be unserialized to `None`.
+ * - the code is crying for a DirectiveRId(id: DirectiveId, revId: Option[RevId] = None)
```
fanf, line 74: With some more though, I think we should make that change. 
With that PR, the id is really the product of the two parts, and actually the `id` is the two, and we need to find a short, clear name for that tuple.
Also, I'm not thinking of the version par as a revision. Even after all the code and all, it's a "commit id" in my head. I'm not sure it's a good idea to try to use something else to talk about it.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
```diff
@@ -37,290 +37,65 @@
...
-      case (Nil, Nil) => 0
-      case (Nil, l  ) => compareList(defaultList(l.size), l)
-      case (l  , Nil) => compareList(l, defaultList(l.size))
-      case (h1::t1, h2::t2) =>
-        val h = VersionPart.compare(h1, h2)
-        if(h == 0) compareList(t1, t2) else h
+    if(authorizedChar.matcher(v).matches()) {
+      ParseVersion.parse(value).map(x => TechniqueVersion(x, revId))
+    } else {
+      Left(s"Error: technique version can only have [.0-9] chars in main part + revisionId, not: '${value}'")
```
VinceMacBuche, line 98: Don't you want a PureResult[TechniqueVersion] here ?

VinceMacBuche: with an error type instead of a string?

fanf: Having just a string make integration with third parties easier, but for consistancy it should be

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
VinceMacBuche: I think the directiveId should evolve to be a case class that includes the commit 

like 

case class DirectiveId { value : String, commit : Commit ) 

with Commit being a trait with two values Head and CommitId (value) , so that we don't use an Option, but something with a meaning

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/configuration/ConfigurationRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+        parseTechniques.getTechnique(id.name, id.version.version, r)
+    }
+  }
+
+
+  def getDirectiveLibrary(ids: Set[DirectiveRId]): IOResult[FullActiveTechniqueCategory] = {
+    def nonDefaultRev(revId: Option[RevId]): Boolean = revId.isDefined && revId != Some(GitVersion.defaultRev)
+    val versionedDirectives = ids.filter(x => nonDefaultRev(x.revId))
+    for {
+      optDirs   <- ZIO.foreach(versionedDirectives.toList)(getDirective) // TODO: find a way to do that without N git treewalks
```
VinceMacBuche, line 118: I think that would need a new function in GitParseActiveTechniqueLibrary that group directive id by revision and do a git command for each revision to get corresponding  directives

ids.groupMap(_.revision)(_.id).flatMap( (rev, ids) => parseActiveTechniqueLibrary.getDirective(ids,rev) )

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/DB.scala
```diff
@@ -110,7 +111,7 @@ final object DB {
 
   def insertReports(reports: List[com.normation.rudder.domain.reports.Reports]): ConnectionIO[Int] = {
     val dbreports = reports.map { r =>
-      DB.Reports[Unit]((), r.executionDate, r.nodeId.value, r.directiveId.value, r.ruleId.value, r.serial
```
VinceMacBuche, line 113: We want to insert the old directive id or the id with revision ?

VinceMacBuche: And rule is still the id and not the RId (but they should be the same between the Rule and the Directive, i think)

fanf: Yes, for now you can't version rule, it's the next step

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
```diff
@@ -172,12 +172,17 @@ object Doobie {
   )
 
   implicit val ReportRead: Read[Reports] = {
-    type R = (DateTime, RuleId, DirectiveId, NodeId, Int, String, String, DateTime, String, String)
+    type R = (DateTime, RuleId, String, NodeId, Int, String, String, DateTime, String, String)
```
VinceMacBuche, line 175: Why not DirectiveRId here ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
VinceMacBuche: It's now in an IOResult since unserializeAcceptation is now a PureResult see https://github.com/Normation/rudder/pull/3247/files#diff-392a77504d47edf5cb77a4792164ad3b4e876b1229638ef9eb71bbcd31c48656R493

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/HistorizationJdbcRepository.scala
```diff
@@ -244,7 +244,8 @@ class HistorizationJdbcRepository(db: Doobie) extends HistorizationRepository wi
         _  <- Update[DB.SerializedRuleDirectives]("""
                  insert into rulesdirectivesjoin (rulepkeyid, directiveid)
                  values (?, ?)
-               """).updateMany(r.directiveIds.toList.map(d => DB.SerializedRuleDirectives(pk, d.value)))
+               """).updateMany(r.directiveIds.filter(_.revId.isEmpty).toList.map(d => DB.SerializedRuleDirectives(pk, d.id.value)))
```
VinceMacBuche, line 247: I don't understand why we do not serialize with revision id here ? 

It means we don't store the value if there is a revId, and we only keep track of directive with no revision (head)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/DB.scala
```diff
@@ -309,21 +310,23 @@ final case class SerializedRules[T](
...
       , directives : Seq[SerializedRuleDirectives]
     ) : Rule = {
       Rule (
           RuleId(rule.ruleId)
+        , None// TODO: we should perhaps be able to get ride of these table.
         , rule.name
         , RuleCategoryId(rule.categoryId.getOrElse("rootRuleCategory")) // this is not really useful as RuleCategory are not really serialized
         , ruleTargets.flatMap(x => RuleTarget.unser(x.targetSerialisation)).toSet
-        , directives.map(x => DirectiveId(x.directiveId)).toSet
+        , directives.map(x => DirectiveRId(DirectiveId(x.directiveId))).toSet
```
VinceMacBuche, line 329: I think we should store and get back the directive id with revision from de table if we can, or else the directive here won't be the same

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/HistorizationJdbcRepository.scala
fanf: We should be able to totally remove historization repository with that change, since everything will be in git. So I didn't want to loose time on making it correct. 
If we are force to keep it for an unforseen reason, we will see what we historize exactly.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/configuration/ConfigurationRepository.scala
fanf: Before that, I beliieve we will need to be able to "merge" revision of directives. For now, we have nothing to tell that a directive at two point in the git tree is the same (and so, that its revision are the same) if there is no modification on that directive between the two commits.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/DB.scala
fanf: same as above: I would love to delete histororization repo.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/JsonSchema.scala
```diff
@@ -0,0 +1,598 @@
+/*
+*************************************************************************************
```
VinceMacBuche, line 2: Does this file belongs to this PR or the zio-json one ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -279,6 +313,99 @@ class RuleApi(
       ) ("createRuleCategory")
     }
   }
+
```
VinceMacBuche, line 316: ZIO json here too i think it should belong to zio-json pr

VinceMacBuche:
If the toString changes and the zio-json would be in a different PR that would have been easier to read !!

VinceMacBuche:
I reviewed the PR, apart some naming issues / organisation with id and revId this is ok to me

Only thing is why did we have zio-json in this PR and also it would have been easier to read with the toString changes in another PR

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/JsonSchema.scala
fanf: there is different commit and yes, I used the zio-json one. It will need to be rebased, though.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
fanf: it should be transparent (a different provided string to put where it should)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
fanf: I need to sum-up things here, but I agree that the ID is really two parts: an UUID and an optionnal commit id. 
I'm not sure it's a `head|commit`. The reallity for now is that it's optionnal (sometime, we have it, sometime no). But it could be dealt with at system boundaries (read from/to git, unserialize things, api, etc). And we would have a clearer internal model.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
fanf: to be able to parse the string as a directiverid. We could have a separate `Read[DirectiveRId]` but I wanted to limit changes for a first go.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
```diff
@@ -78,27 +77,27 @@ object TechniqueVersion {
    * NOTE: it seems that in existing env, sometimes epoch was written when it should
    * not have. So we must accept also leading number+":".
    */
-  protected def apply(v: Version): Either[String, TechniqueVersion] = {
```
VinceMacBuche, line 93: I'm not all against default parameters, but do you think we may forget some revision some times ? (but i can imagine how many lines it saved you to edit)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
```diff
@@ -83,27 +84,36 @@ import com.normation.cfclerk.domain.SectionSpec
...
-final case class DirectiveId(value : String) extends AnyVal
-
-// there is a lot of place that need that as the real identifier of a directive
-final case class DirectiveRId(id: DirectiveId, revId: Option[RevId] = None) {
+/*
+ * A directive identifier is composed of the directive unique identifier and the directive revision.
+ * For backward compatibility, the UID field is named "id" in most serialized format. We will keep
+ * "uid" in code to avoid code looking like `directive.id.id`.
+ */
+final case class DirectiveId(uid: DirectiveUid, rev: Revision = GitVersion.defaultRev) {
```
VinceMacBuche, line 102: Same here with default value

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Rule.scala
```diff
@@ -55,13 +55,13 @@ final case class RuleRId(id: RuleId, revId: Option[RevId] = None)
  */
 final case class Rule(
     id              : RuleId
-  , revId           : Option[RevId]
```
VinceMacBuche: Revision are still option for Rules ? (maybe that intented has we focus on Directive here)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
```diff
@@ -406,19 +408,19 @@ class GitParseActiveTechniqueLibrary(
 ) extends ParseActiveTechniqueLibrary with GitParseCommon[ActiveTechniqueCategoryContent] {
 
   /**
-   * Get a directive for the specific given revisionId;
+   * Get a directive for the specific given revision;
    */
-  def getDirective(id: DirectiveId, revId: RevId): IOResult[Option[(ActiveTechnique, Directive)]] = {
+  def getDirective(uid: DirectiveUid, rev: Revision): IOResult[Option[(ActiveTechnique, Directive)]] = {
```
VinceMacBuche: Could be a DirectiveId  which regroup both ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlSerialisationImpl.scala
```diff
@@ -197,7 +198,7 @@ class DirectiveSerialisationImpl(xmlVersion:String) extends DirectiveSerialisati
     , directive          : Directive
   ) = {
     createTrimedElem(XML_TAG_DIRECTIVE, xmlVersion) (
-          <id>{directive.id.value}</id>
+          <id>{directive.id.uid.value}</id>
```
VinceMacBuche, line 201: Shouldn't we store the revision here ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitFindUtils.scala
```diff
@@ -127,6 +130,18 @@ object GitFindUtils extends NamedZioLogger {
     }
   }
 
+  /**
+   * Retrieve revisions for the given path
+   */
+  def findRevFromPath(git: Git, path: String): IOResult[Iterable[RevisionInfo]] = {
+    import scala.jdk.CollectionConverters._
+    IOResult.effectM("") {
```
VinceMacBuche: I think you want an error message here :)

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -538,6 +538,10 @@ class TestNodeConfiguration(prefixTestResources: String = "") {
     }))
   }
 
+  println("****** bad bad bad")
```
VinceMacBuche: debug log :)

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -543,9 +543,6 @@ class TestNodeConfiguration(prefixTestResources: String = "") {
     }))
   }
 
-  println("****** bad bad bad")
```
VinceMacBuche: ok they were removed :)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/DirectiveApi.scala
```diff
@@ -592,6 +598,22 @@ class DirectiveApiService14 (
     }
   }
 
+  def getTechniqueWithVerion(optTechniqueName: Option[TechniqueName], opTechniqueVersion: Option[TechniqueVersion], revision: Option[Revision]): IOResult[Technique] = {
```
VinceMacBuche, line 601: typo in function Name (Verion)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/TechniqueVersion.scala
fanf: It was necessary to avoid breaking most of the code. I can imagine removing the default value in first patch release or next minor

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Rule.scala
fanf: yes, it will be in next iteration

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlSerialisationImpl.scala
fanf: no, by definition, it will be given by the git commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
fanf: I wanted to emphasis the fact that you look for a specific revision explictly here. I latter change the named to `getDirectiveRevision` for that

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitFindUtils.scala
fanf: it was added in the lastest commits

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/Directive.scala
fanf: We chose `revision`

VinceMacBuche (approved):

fanf:
PR rebased

PR #3249, opened 2020-10-08
Fixes #18306: Create healthcheck API 

ElaadF:
https://issues.rudder.io/issues/18306

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
@@ -1,0 +1,62 @@
+package com.normation.rudder.services.healthcheck
+
+import java.lang.Runtime.getRuntime
+
+import com.normation.errors.IOResult
+import com.normation.rudder.domain.logger.{HealthcheckLoggerPure => logger}
+import com.normation.rudder.services.healthcheck.HealthcheckResult.Critical
+import com.normation.rudder.services.healthcheck.HealthcheckResult.Warning
+import com.normation.rudder.services.healthcheck.HealthcheckResult.Ok
+import zio.UIO
+import zio.ZIO
+import zio.syntax.ToZio
+
+final case class CheckName(value: String)
+
+trait Check {
+  def name: CheckName
+  def run: IOResult[HealthcheckResult]
+}
+
+sealed trait HealthcheckResult { def msg:String }
+object HealthcheckResult {
+  final case class Ok(val msg: String) extends HealthcheckResult
+  final case class Warning(val msg: String) extends HealthcheckResult
+  final case class Critical(val msg: String) extends HealthcheckResult
+}
+
+class HealthcheckService(checks: List[Check]) {
+
+   def runAll: ZIO[Any, Nothing, List[HealthcheckResult]]  = ZIO.foreach(checks) { c =>
+     for {
+       res <- c.run.catchAll { err =>
+                HealthcheckResult.Critical(
+                  s"A fatal error was encountered when running check '${c.name.value}': ${err.fullMsg}"
+                ).succeed
+              }
+       _   <- logHealthcheck(c.name, res)
+     } yield res
+  }
+
+  private[this] def logHealthcheck(name: CheckName, check: HealthcheckResult): UIO[Unit] = {
+    val msg = s"${name.value}: ${check.msg}"
+    check match {
+      case _: Critical => logger.error(msg)
+      case _: Warning  => logger.warn(msg)
+      case _: Ok       => logger.debug(msg)
+    }
+  }
+}
+
+final object CheckCoreNumber extends Check {
+  def name: CheckName = CheckName("CPU Cores available")
+  def run: IOResult[HealthcheckResult] = for {
+    availableCores <- IOResult.effect(getRuntime.availableProcessors)
+  } yield {
+    availableCores match {
+      case i if i <= 0 => Critical("No Core available")
+      case 1 => Warning(s"Only one cores available")
+      case n => Ok(s"${n} cores available")
+    }
+  }
+}
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
@@ -714,6 +714,19 @@ object UserApi extends ApiModuleProvider[UserApi] {
   def endpoints = ca.mrvisser.sealerate.values[UserApi].toList.sortBy( _.z )
 }
 
+sealed trait HealthcheckApi extends EndpointSchema with GeneralApi with SortIndex
+object HealthcheckApi extends ApiModuleProvider[HealthcheckApi] {
+
+  // This endpoint run all checks to return the result
+  final case object GetHealthcheckResult extends HealthcheckApi with ZeroParam with StartsAtVersion12 with SortIndex { val z = implicitly[Line].value
+    val description    = "Result of a health check run"
+    val (action, path) = GET / "healthcheck"
+  }
+
+  def endpoints = ca.mrvisser.sealerate.values[HealthcheckApi].toList.sortBy( _.z )
+
+}
+
 /*
  * All API.
  */
@@ -731,6 +744,7 @@ object AllApi {
     RuleApi.endpoints :::
     InventoryApi.endpoints :::
     InfoApi.endpoints :::
+    HealthcheckApi.endpoints :::
     // UserApi is not declared here, it will be contributed by plugin
     Nil
 }
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestDataSerializer.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestDataSerializer.scala
@@ -37,34 +37,38 @@
 
 package com.normation.rudder.rest
 
-import com.normation.rudder.domain.workflows._
-import net.liftweb.common._
-import com.normation.rudder.domain.policies._
-import net.liftweb.json._
 import com.normation.cfclerk.domain._
-import com.normation.rudder.services.modification.DiffService
 import com.normation.cfclerk.services.TechniqueRepository
-import com.normation.rudder.domain.nodes._
-import com.normation.rudder.domain.queries.Query
 import com.normation.inventory.domain.NodeId
-import com.normation.rudder.domain.parameters._
 import com.normation.inventory.domain._
-import com.normation.rudder.domain.servers.Srv
-import com.normation.rudder.rule.category.RuleCategory
-import com.normation.rudder.rule.category.RuleCategoryId
-import com.normation.rudder.repository.FullNodeGroupCategory
-import com.normation.rudder.repository.FullActiveTechnique
 import com.normation.rudder.api.ApiAccount
-import net.liftweb.json.JsonDSL._
-import com.normation.rudder.web.components.DateFormaterService
-import com.normation.rudder.rest.data._
-import com.normation.rudder.api.ApiAccountKind.{PublicApi => PublicApiAccount}
-import com.normation.rudder.api.ApiAccountKind.User
 import com.normation.rudder.api.ApiAccountKind.System
-import com.normation.rudder.api.ApiAuthorization.{None => NoAccess}
+import com.normation.rudder.api.ApiAccountKind.User
+import com.normation.rudder.api.ApiAccountKind.{PublicApi => PublicApiAccount}
+import com.normation.rudder.api.ApiAuthorization.ACL
 import com.normation.rudder.api.ApiAuthorization.RO
 import com.normation.rudder.api.ApiAuthorization.RW
-import com.normation.rudder.api.ApiAuthorization.ACL
+import com.normation.rudder.api.ApiAuthorization.{None => NoAccess}
+import com.normation.rudder.domain.nodes._
+import com.normation.rudder.domain.parameters._
+import com.normation.rudder.domain.policies._
+import com.normation.rudder.domain.queries.Query
+import com.normation.rudder.domain.servers.Srv
+import com.normation.rudder.domain.workflows._
+import com.normation.rudder.repository.FullActiveTechnique
+import com.normation.rudder.repository.FullNodeGroupCategory
+import com.normation.rudder.rest.data._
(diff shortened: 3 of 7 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,58 @@
...
+  , msg   : String
+  , status: Status
+)
+
+class HealthcheckService(checks: List[Check]) {
+
+   def runAll: ZIO[Any, Nothing, List[HealthcheckResult]]  = ZIO.foreach(checks) { c =>
+      for {
+        res <- c.run.catchAll { err =>
+                 HealthcheckResult(c.name, s"A fatal error was encountered when running check '${c.name}': ${err.fullMsg}", Critical)
```
ElaadF, line 31: I get a `RudderError => HealthcheckResult` instead of `RudderError => ZIO[R,E,A] `

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,58 @@
...
+trait Check {
+  def name: String
+  def run: IOResult[HealthcheckResult]
+}
+
+final case class HealthcheckResult(
+    name  : String
+  , msg   : String
+  , status: Status
+)
```
fanf, line 24: Is there an advantage of that over:
```
sealed trait HealthcheckResult { def name; def msg }
object HealthcheckResult {
  final case class OK(val name: String, val msg: String) extends HealthcheckResult
  final case class Warning(val name: String, val msg: String) extends HealthcheckResult
  final case class Critical(val name: String, val msg: String) extends HealthcheckResult
}
``` ?

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala [hunk 3]
```diff
@@ -0,0 +1,58 @@
...
+
+import java.lang.Runtime.getRuntime
+
+import com.normation.errors.IOResult
+import com.normation.rudder.domain.logger.{HealthcheckLoggerPure => logger}
+import zio.UIO
+import zio.ZIO
+
+sealed trait Status
+case object Ok extends Status
```
fanf, line 11: please use format
```
sealed trait Status
final object Status { 
  final case object Ok extends Status 
...
}
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,58 @@
...
+        case 1           => HealthcheckResult(name, s"Only one cores available", Warning)
+        case n if n >= 2 => HealthcheckResult(name, s"Only one cores available", Warning)
+      }
+    }
+  }
+
+  private[this] def logHealthcheck(check: HealthcheckResult): UIO[Unit] = {
+    check.status match {
+      case Critical => logger.error(s"`${check.name}` check has detected a critical issue: ${check.msg}")
+      case Warning =>  logger.warn(s"`${check.name}` check has detected a waning issue: ${check.msg}")
```
fanf, line 54: wa_r_ning

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
fanf: look at catchAll sig, maybe you need a `.succeed`, or maybe it's not catchAll but an other one that allows to change an error to a success. You need something with signature `E => ZIO[R, E1, B]`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/HealthcheckApi.scala
```diff
@@ -0,0 +1,55 @@
...
+      result match {
+        case Full(jsons) =>
+          RestUtils.toJsonResponse(None, JArray(jsons))
+        case eb: EmptyBox =>
+          val message = (eb ?~ (s"Error when trying to run healthcheck")).msg
+          RestUtils.toJsonError(None, message)("getHealthcheck",true)
+      }
+    }
+
+  }la poucave
```
fanf, line 54: "oups"

fanf (changes requested):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
VinceMacBuche: You need to encapsulare your HealthCheckResult in a zio 

just call succeed function at the end

```
                 HealthcheckResult(c.name, s"A fatal error was encountered when running check '${c.name}': ${err.fullMsg}", Critical).succeed
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
VinceMacBuche: I think we may put the Check object inside the CheckResult so we could use it's name parameter direclty without errors

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,58 @@
...
+case object Ok extends Status
+case object Warning extends Status
+case object Critical extends Status
+
+trait Check {
+  def name: String
+  def run: IOResult[HealthcheckResult]
+}
+
+final case class HealthcheckResult(
```
VinceMacBuche, line 20: Y do think we need the value not just a message. You will need to represent it correctly in your dashboard... But maybe they are juste green checks / red cross for now ...

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
fanf: `Check` is not a structure, it's a service with a name. It can't be passed around in structure, that's forbiden. 
We can add a `CheckName(name: String)`, have `object CheckCoreNumber extends Check(CheckName("CPU Cores available"))` and pass that name object around

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
fanf: Yes, just green / yellow / red + explanation message (that will likely contains the expected value and the actual value). But not need for different vizualization for now.

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
ElaadF: Not sure to understand the utility of `CheckName` just replace the `name: String` in `Check` ?

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,61 @@
...
+class HealthcheckService(checks: List[Check]) {
+
+   def runAll: ZIO[Any, Nothing, List[HealthcheckResult]]  = ZIO.foreach(checks) { c =>
+      for {
+        res <- c.run.catchAll { err =>
+                 HealthcheckResult.Critical(
+                   s"A fatal error was encountered when running check '${c.name}': ${err.fullMsg}"
+                 ).succeed
+               }
+        _   <- logHealthcheck(res)
```
fanf, line 37: you need to pass `${c.name}`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,61 @@
...
+      } yield res
+    }
+
+  private[this] def logHealthcheck(check: HealthcheckResult): UIO[Unit] = {
+    check match {
+      case c: Critical => logger.error(s"`Healthcheck has detected a critical issue: ${check.msg}")
+      case w: Warning =>  logger.warn(s"Healthcheck check has detected a warning issue: ${check.msg}")
+      case o: Ok =>       logger.debug(s"Healthcheck check has passed: ${check.msg}")
+    }
```
fanf, line 46: I think this can be more succint and precise:
- we already know it's healthcheck because of the name of the logger, 
- we already know the level (it's in the log),
- we don't have the probe name, it should be added

I think it can be just, for all logs: `${name}: ${check.msg}` and just change the level, so:

```
  private[this] def logHealthcheck(name: String, check: HealthcheckResult): UIO[Unit] = {
    val msg = s"${name}: ${check.msg}"
    check match {
      case _: Critical => logger.error(msg)
      case _: Warning  => logger.warn(msg)
      case _: Ok       => logger.debug(msg)
    }
}
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,61 @@
...
+
+final object CheckCoreNumber extends Check{
+  def name: CheckName = CheckName("CPU Cores available")
+  def run: IOResult[HealthcheckResult] = for {
+    availableCores <- IOResult.effect(getRuntime.availableProcessors)
+  } yield {
+    availableCores match {
+      case 0           => Critical("No Core available")
+      case 1           => Warning(s"Only one cores available")
+      case n if n >= 2 => Ok(s"${n} cores available")
```
fanf, line 58: your test is not total here, it should be: 
```
    availableCores match {
      case n if n < 1 => Critical("No Core available")
      case 1          => Warning(s"Only one cores available")
      case _          => Ok(s"${n} cores available")
```

fanf: In a second time, we will need to work on messages, because we need to provide user with actionnable corrective actions. 
And perhaps we will want to have to level (one for the "chek tab", on for details).
And perhaps, we will want to externalize this messages, but perhaps not.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestDataSerializer.scala
```diff
@@ -549,6 +555,16 @@ final case class RestDataSerializerImpl (
     )
   }
 
+  def serializeHealthcheckResult(check: HealthcheckResult): JValue = {
+    val status = check match {
+      case c:Critical => "Critical"
+      case o:Ok => "Ok"
```
fanf, line 561: if you don't use a variable, don't bind it to a name (so here: `case _:OK =>` ...)

fanf (changes requested):

ElaadF:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -714,6 +714,18 @@ object UserApi extends ApiModuleProvider[UserApi] {
   def endpoints = ca.mrvisser.sealerate.values[UserApi].toList.sortBy( _.z )
 }
 
+sealed trait HealthcheckApi extends EndpointSchema with GeneralApi with SortIndex
+object HealthcheckApi extends ApiModuleProvider[HealthcheckApi] {
+
+  final case object GetHealthcheckResult extends HealthcheckApi with ZeroParam with StartsAtVersion2 with SortIndex { val z = implicitly[Line].value
```
ElaadF, line 720: `StartAtVersion2` should be `StartsAtVersion15` ?

ElaadF:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -714,6 +714,19 @@ object UserApi extends ApiModuleProvider[UserApi] {
   def endpoints = ca.mrvisser.sealerate.values[UserApi].toList.sortBy( _.z )
 }
 
+sealed trait HealthcheckApi extends EndpointSchema with GeneralApi with SortIndex
+object HealthcheckApi extends ApiModuleProvider[HealthcheckApi] {
+
+  // This endpoint run all checks to return the result
+  final case object GetHealthcheckResult extends HealthcheckApi with ZeroParam with StartsAtVersion15 with SortIndex { val z = implicitly[Line].value
```
fanf, line 721: current version is 12

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -0,0 +1,62 @@
...
+  }
+}
+
+final object CheckCoreNumber extends Check {
+  def name: CheckName = CheckName("CPU Cores available")
+  def run: IOResult[HealthcheckResult] = for {
+    availableCores <- IOResult.effect(getRuntime.availableProcessors)
+  } yield {
+    availableCores match {
+      case 0 => Critical("No Core available")
```
fanf, line 57: it should really be `case i if i <= 0 =>` because 0 core is an error (it can't be otherwise), and in case of error, you can't, make any assumption about what will be returned beside what the type contraints (an error is often a negative number in C-like semantic).

ElaadF:
Commit modified

PR #3272, opened 2020-10-19
Fixes #18346: Make the group option to include the Rudder server or not in the group more clear

fanf:
https://issues.rudder.io/issues/18346

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/server/server_details.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/server/server_details.html
@@ -18,7 +18,7 @@
                 <span class="ion ion-checkmark-round check-icon"></span>
               </label>
               <label for="typeQuery" class="form-control">
-                Include Rudder server components
+                Include Rudder root server components <span class="tooltipable glyphicon glyphicon-question-sign" title="Node(s) with Rudder web application or Rudder database"></span>
               </label>
             </div>
           </li>


webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/server/server_details.html
```diff
@@ -18,7 +18,7 @@
                 <span class="ion ion-checkmark-round check-icon"></span>
               </label>
               <label for="typeQuery" class="form-control">
-                Include Rudder server components
+                Include Rudder root server
```
ncharles, line 21: It is not only root, but all node with a server component: so nodes having rudder-reports, rudder-db, rudder-webapp or rudder-server-root
But not the relays

Telling root is not correct - maybe something in the line
"Include Rudder server components (web interface, database)"

ncharles (changes requested):
It is not only root server

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/server/server_details.html
fanf: nobody knows what "server components" are (I didn't think it's documented anywhere). I really think it should say "root". 

```
Include Rudder root server components (nodes with Rudder application or database)
```

fanf: "application" is maybe too broad, though

fanf:
```
Include Rudder root server components (nodes with Rudder web application or database)
```

fanf: image

fanf:
PR updated with a new commit

ncharles (approved):

fanf:
OK, squash merging this PR

PR #3285, opened 2020-10-21
Fixes #18381: outcome should be enums rather than hardcoded

gpoblon:
https://issues.rudder.io/issues/18381

### DIFF
--- a/rudder-lang/src/technique/outcome.rs
+++ b/rudder-lang/src/technique/outcome.rs
@@ -1,0 +1,80 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// SPDX-FileCopyrightText: 2019-2020 Normation SAS
+
+use crate::error::*;
+use std::{fmt, str::FromStr};
+use strum::{EnumIter, IntoEnumIterator};
+
+pub enum ReturnConditionOutcome {
+    Kept,
+    Repaired,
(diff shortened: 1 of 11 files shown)


gpoblon:
Commit modified

rudder-lang/src/technique/outcome.rs
```diff
@@ -0,0 +1,75 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
+// SPDX-FileCopyrightText: 2019-2020 Normation SAS
+
+use crate::error::*;
+use std::{fmt, str::FromStr};
+use strum::{EnumIter, IntoEnumIterator};
+
+#[derive(EnumIter, PartialEq, Debug)]
+pub enum ConditionOutcome {
```
peckpeck, line 9: Literal outcomes as used in return statements should only be Repaired, Kept or Error

gpoblon:
Commit modified

peckpeck:
I see a commit modified message, but no change, could you do fixups during review so that I can easily see dif of changes

gpoblon:
yes I just rebased the branch and started looking into a fix but I'd need to discuss about it with you

gpoblon:
yes I just rebased the branch and started looking into a fix but I'd need to discuss about it with you

PR #3293, opened 2020-10-22
Fixes #18395: Update dependencies for the webapp

ncharles:
https://issues.rudder.io/issues/18395

### DIFF
--- a/webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
+++ b/webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
@@ -50,7 +50,7 @@ import com.normation.zio.ZioRuntime
 import zio._
 import zio.syntax._
 import com.normation.zio._
-import org.apache.commons.lang.StringUtils
+import org.apache.commons.lang3.StringUtils
 
 import scala.collection.immutable.ArraySeq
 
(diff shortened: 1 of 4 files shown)


ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FillTemplatesService.scala
```diff
@@ -50,7 +50,7 @@ import com.normation.zio.ZioRuntime
 import zio._
 import zio.syntax._
 import com.normation.zio._
-import org.apache.commons.lang.StringUtils
+import org.apache.commons.lang3.StringUtils
```
fanf, line 53: Do you remember why we need to use that here, and only here? Perf reason ?

ncharles: yes, see https://github.com/Normation/rudder/pull/3293/files#diff-5e75cf601f6a1cca6e6f37ba80ee2658d103fe0a3f44fb412967f256dfd43b43L240

ncharles:
PR updated with a new commit

fanf (approved):

fanf (approved):

fanf:
OK, squash merging this PR

PR #3300, opened 2020-10-23
Fixes #17666: Node status API must not return an error when the node is not here

fanf:
https://issues.rudder.io/issues/17666

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
@@ -238,6 +238,10 @@ object NodeApi extends ApiModuleProvider[NodeApi] {
     val description = "List all accepted nodes with configurable details level"
     val (action, path)  = GET / "nodes"
   }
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -238,6 +238,10 @@ object NodeApi extends ApiModuleProvider[NodeApi] {
     val description = "List all accepted nodes with configurable details level"
     val (action, path)  = GET / "nodes"
   }
+  final case object GetNodesStatus extends NodeApi with ZeroParam with StartsAtVersion12 with SortIndex { val z = implicitly[Line].value
```
VinceMacBuche, line 241: Should we increase the version here ?? 12 is version 6.1

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
fanf: done

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #3319, opened 2020-11-03
Fixes #18466: Inherited node prop arrays are replaced not merged contrary to doc

fanf:
https://issues.rudder.io/issues/18466

I have a second idea with a modifier that can be appended to node properties to specify the inheritance behavior for object, arrays and string independently.

As it is a significant change, it should certainly add in 6.2.

PR #3335, opened 2020-11-04
Fixes #18466: Inherited node prop arrays are replaced not merged contrary to doc

fanf:
https://issues.rudder.io/issues/18466

Replacing previous PR: https://github.com/Normation/rudder/pull/3319

Add a "inheritMode" optionnal attribute in all properties (global param, group, node). Only global param has an UI (because it's likely where people want to change that, and because I have 0 idea about how to do one for group/node properties). 

inheritMode is a 3 char strings:
- 1: behavior for json object; m = merge (default), o = override
- 2: behavior for array: o = override (defaut), a = append, p = prepend
- 3: behavior for strings: o = override (defaut), a = append, p = prepend

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/ldap/rudder.schema
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/ldap/rudder.schema
@@ -356,6 +356,13 @@ attributetype ( RudderAttributes:307
   SUBSTR caseIgnoreSubstringsMatch
   SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 
+attributetype ( RudderAttributes:308
+  NAME 'inheritMode'
+  DESC 'inheritance mode like merge, override, etc for properties and parameters object/array/string'
+  EQUALITY caseExactMatch
+  SUBSTR caseIgnoreSubstringsMatch
+  SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
+
 #### Nodes
 
 attributetype ( RudderAttributes:351
@@ -540,7 +547,8 @@ objectclass ( RudderObjectClasses:120
   SUP top
   STRUCTURAL
   MUST ( parameterName )
-  MAY ( parameterValue $ description $ propertyProvider $ overridable ) )
+  MAY ( parameterValue $ description $ propertyProvider $
+        overridable $ inheritMode ) )
 
 #
 # Application properties
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
@@ -77,6 +77,61 @@ import net.liftweb.json.JsonDSL._
  */
 final case class PropertyProvider(value: String) extends AnyVal
 
+
+/*
+ * An inherit mode is three chars, one for object, one for array, one for string like: 'moo'
+ * The first char is for objects and can be: o = override, m = merge (default)
+ * The second char is for arrays and can be: o = override (default), a = append, p = prepend
+ * The third is for strings and can be:      o = override (default), a = append, p = prepend
+ */
+final case class InheritMode(
+    forObject: InheritMode.ObjectMode
+  , forArray : InheritMode.ArrayMode
+  , forString: InheritMode.StringMode
+) {
+  def value = s"${forObject.value}${forArray.value}${forString.value}"
+}
+final object InheritMode {
+  sealed trait ObjectMode { def value: Char }
+  final case object ObjectMode {
+    final case object Override extends ObjectMode { override val value = 'o' }
+    final case object Merge    extends ObjectMode { override val value = 'm' }
+
+    def all = ca.mrvisser.sealerate.values[ObjectMode].toList
+    def parse(c: Char) = all.find(c == _.value)
+  }
+  sealed trait ArrayMode { def value: Char }
+  final case object ArrayMode {
+    final case object Override extends ArrayMode { override val value = 'o' }
+    final case object Append   extends ArrayMode { override val value = 'a' }
+    final case object Prepend  extends ArrayMode { override val value = 'p' }
+
+    def all = ca.mrvisser.sealerate.values[ArrayMode].toList
+    def parse(c: Char) = all.find(c == _.value)
+  }
+  sealed trait StringMode { def value: Char }
+  final case object StringMode {
+    final case object Override extends StringMode { override val value = 'o' }
+    final case object Append   extends StringMode { override val value = 'a' }
+    final case object Prepend  extends StringMode { override val value = 'p' }
+
+    def all = ca.mrvisser.sealerate.values[StringMode].toList
+    def parse(c: Char) = all.find(c == _.value)
+  }
+
+  def parseString(s: String): PureResult[InheritMode] = s.toList match {
+    case obj :: arr :: str :: Nil =>
+      (ObjectMode.parse(obj), ArrayMode.parse(arr), StringMode.parse(str)) match {
+        case (Some(o), Some(a), Some(s)) => Right(InheritMode(o, a, s))
+        case _                           => Left(Inconsistency(s"Impossible to parse string as inherit mode option, expecting: [mo][oap][oap] but got: ${s}"))
+      }
+    case _                        =>
+                                            Left(Inconsistency(s"Impossible to parse string as inherit mode option, expecting: [mo][oap][oap] but got: ${s}"))
+  }
+
+  val Default = InheritMode(ObjectMode.Merge, ArrayMode.Override, StringMode.Override)
+}
+
 object PropertyProvider {
   /*
    * System property provider. These properties should never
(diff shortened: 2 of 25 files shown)


VinceMacBuche:
There is a wip commit that should be prefixed by fixup!

webapp/sources/rudder/rudder-core/src/main/resources/ldap/rudder.schema
```diff
@@ -356,6 +356,13 @@ attributetype ( RudderAttributes:307
   SUBSTR caseIgnoreSubstringsMatch
   SYNTAX 1.3.6.1.4.1.1466.115.121.1.15 )
 
+attributetype ( RudderAttributes:308
+  NAME 'inheritMode'
+  DESC 'inheritance mode like merge, overrid, etc for properties and parameters object/array/string'
```
VinceMacBuche, line 361:
```suggestion
  DESC 'inheritance mode like merge, override, etc for properties and parameters object/array/string'
```

You can accept suggestion by prefixingby fixup! when accepting

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -77,6 +77,62 @@ import net.liftweb.json.JsonDSL._
  */
 final case class PropertyProvider(value: String) extends AnyVal
 
+
+/*
+ * An inherit mode is three char, one for object, one for array, one for string like (?moo)
```
VinceMacBuche, line 82:
```suggestion
 * An inherit mode is three chars, one for object, one for array, one for string like (moo)
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -77,6 +77,62 @@ import net.liftweb.json.JsonDSL._
...
+ * The first char is for objects and can be: d = o = override, m = merge (default)
```
VinceMacBuche, line 83:
```suggestion
 * The first char is for objects and can be: o = override, m = merge (default)
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -77,6 +77,62 @@ import net.liftweb.json.JsonDSL._
...
+ * The second char is for arrays and can be: d = o = override (default), a = append, p = prepend
```
VinceMacBuche, line 84:
```suggestion
 * The second char is for arrays and can be: o = override (default), a = append, p = prepend
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -77,6 +77,62 @@ import net.liftweb.json.JsonDSL._
...
+  final case object StringMode {
+    final case object Override extends StringMode { override val value = 'o' }
+    final case object Append   extends StringMode { override val value = 'a' }
+    final case object Prepend  extends StringMode { override val value = 'p' }
+
+    def all = ca.mrvisser.sealerate.values[StringMode].toList
+    def parse(c: Char) = all.find(c == _.value)
+  }
+
+  def parseString(s: String): Option[InheritMode] = s.toList match {
```
VinceMacBuche, line 122: I think it should be a PureResult and give errors when parsing is not OK

Maybe also treat the case when there is more then 3 characters ? But maybe it's better to have only 3 chars as valid case, so maybe it's good like this, just discussing

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -77,6 +77,62 @@ import net.liftweb.json.JsonDSL._
  */
 final case class PropertyProvider(value: String) extends AnyVal
 
+
+/*
+ * An inherit mode is three char, one for object, one for array, one for string like (?moo)
+ * The first char is for objects and can be: d = o = override, m = merge (default)
+ * The second char is for arrays and can be: d = o = override (default), a = append, p = prepend
+ * The third is for strings and can ve:      d = o = override (default), a = append, p = prepend
```
VinceMacBuche, line 85:
```suggestion
 * The third is for strings and can ve:      o = override (default), a = append, p = prepend
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -77,6 +77,62 @@ import net.liftweb.json.JsonDSL._
...
+    case obj :: arr :: str :: Nil =>
+      (ObjectMode.parse(obj), ArrayMode.parse(arr), StringMode.parse(str)) match {
+        case (Some(o), Some(a), Some(s)) => Some(InheritMode(o, a, s))
+        case _                           => None
+      }
+    case _                        =>
+                                            None
+  }
+
+  val optionStringLenght = 6
```
VinceMacBuche, line 132: I think it's unused, and Lenght is spelled Length ;)

webapp/sources/rudder/rudder-core/src/main/resources/ldap/rudder.schema
fanf: It will need a rebase in all case

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
fanf: I'm not sure. It complicates a lot the usage.
And no, the only legit case if these three chars, so it's ok to return none in other cases

webapp/sources/rudder/rudder-core/src/main/resources/ldap/rudder.schema
VinceMacBuche: Non it will create a new commit and it will be rebased if you add fixup! to the commit message

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -194,24 +269,77 @@ object GenericProperty {
...
   /**
    * Merge two properties. newProp values will win.
    * You should have check before that "name" and "provider" are OK.
    */
   def mergeConfig(oldProp: Config, newProp: Config): Config = {
-    newProp.withFallback(oldProp)
+    val mode = (GenericProperty.getMode(oldProp), GenericProperty.getMode(newProp)) match {
+      case (None, None)                                   => None
+      case (None, Some(mode))                             => Some(mode)
+      case (Some(mode), None | Some(InheritMode.Default)) => Some(mode)
```
VinceMacBuche, line 331: You should not get oldProperty mode if someone has selected the same mode than default

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -194,24 +269,77 @@ object GenericProperty {
...
+        }
+      case (_,OBJECT) => newValue //override
+    }
   }
 
   /**
    * Merge two properties. newProp values will win.
    * You should have check before that "name" and "provider" are OK.
    */
   def mergeConfig(oldProp: Config, newProp: Config): Config = {
```
VinceMacBuche, line 327: The name of variables is quite confusing to me in fact, old means parent, it may not be the time to rename but it quite confusing to me

webapp/sources/rudder/rudder-core/src/main/resources/ldap/rudder.schema
fanf: No, I mean that I will have to do a manual rebase because there is already existing commit that are not fixup

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
fanf: Hum. Yes, you are right, it's a sequel of the past when it was not in option!

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
fanf: I tried parent/child before and it was way worse. old/new can be consistent whatever the origin, is just "what you had before, and what you want to add/merge".

VinceMacBuche (approved):

PR #3346, opened 2020-11-07
Fixes #18529: If LDAP server does not support subtree deletion, we get error when entry does not exists

fanf:
https://issues.rudder.io/issues/18529


fanf:
PR updated with a new commit

ncharles (approved):

fanf:
OK, squash merging this PR

PR #3351, opened 2020-11-09
Fixes #18534: Skip non-standard moint FS for free-space check

ElaadF:
https://issues.rudder.io/issues/18534

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
@@ -46,16 +46,16 @@ object HealthcheckUtils {
 
 class HealthcheckService(checks: List[Check]) {
 
-   def runAll: ZIO[Any, Nothing, List[HealthcheckResult]]  = ZIO.foreach(checks) { c =>
-     for {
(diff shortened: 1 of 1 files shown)


fanf:
No, you need to first filter to keep only lines starting with "/" and then keep the original logic of looking for "/var" or use "/"

fanf (changes requested):
No, you need to first filter to keep only lines starting with "/" and then keep the original logic of looking for "/var" or use "/"

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/healthcheck/HealthcheckService.scala
```diff
@@ -96,7 +96,7 @@ final object CheckFreeSpace extends Check {
 
   def run: IOResult[HealthcheckResult] = {
     val file          = root / "proc" / "mounts"
-    val mountsContent = file.lines.map(x => x.split(" ")(1)).toList
+    val mountsContent = file.lines.filter(_(0) == '/').map(x => x.split(" ")(1)).toList
```
fanf, line 99: can you had a comment to explain? With a link to the bug for future reference ?

fanf (changes requested):

ElaadF:
Commit modified

fanf (approved):

PR #3358, opened 2020-11-12
Fixes #18557: Adapt rudder package to support new versionning scheme

VinceMacBuche:
https://issues.rudder.io/issues/18557


relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
```diff
@@ -496,9 +500,14 @@ def list_plugin_name():
 
 p = Popen("rudder agent version", shell=True, stdout=PIPE)
 line = p.communicate()[0]
-m = re.match(r'Rudder agent (\d+\.\d+)\..*', line.decode('utf-8'))
+print(line.decode('utf-8'))
+m = re.match(r'Rudder agent (((\d+\.\d+)\.\d+)(\.((beta|rc)\d+))?)\..*?', line.decode('utf-8'))
```
amousset, line 504: knowing it comes from the "rudder agent version" command which takes it from the package manager (hence different formatting depending on the platform), which got the information dynamically at build time by a Makefile edition with sed, coming from version.sh file makes me kind of sad.

amousset: bit that's life. I can live with it.

VinceMacBuche: I can do otherwise if you want :) but don't know yet !! We could put it in rudder package at build time ...

VinceMacBuche: put a file containing the value in rudder agent at build time ?

VinceMacBuche: Would remove some steps in the 'rudder agent version' horror show

amousset: As it is new code that only runs on the relay or root we could use the version from relayd (`/opt/rudder/bin/rudder-relayd --version`) which is more consistent and self-contained. I'm not sure if it's actually better though.

VinceMacBuche:
Commit modified

amousset (approved):

fanf:
OK, merging this PR

PR #3364, opened 2020-11-13
Fixes #18508: Missing a PATCH API for allowed networks

fanf:
https://issues.rudder.io/issues/18508

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
@@ -69,12 +69,7 @@ import net.liftweb.common.Failure
 import net.liftweb.common.Full
 import net.liftweb.http.LiftResponse
 import net.liftweb.http.Req
-import net.liftweb.json.JNothing
-import net.liftweb.json.JsonAST.JBool
-import net.liftweb.json.JsonAST.JInt
-import net.liftweb.json.JsonAST.JString
-import net.liftweb.json.JsonAST.JValue
-import net.liftweb.json.JsonAST._
+import net.liftweb.json._
 import net.liftweb.json.JsonDSL._
 import com.normation.box._
 import com.normation.errors._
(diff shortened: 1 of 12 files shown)


amousset:
This will require a doc update too.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
```diff
@@ -903,6 +899,75 @@ final case object RestContinueGenerationOnError extends RestBooleanSetting {
...
+                      req.json.getOrElse(JNothing) \ "allowed_networks"
+                    } else {
+                      req.params.get("allowed_networks").flatMap(_.headOption.flatMap(parseOpt)).getOrElse(JNothing)
+                    }
+        // lift is dumb and have zero problem not erroring on extraction from JNothing
+        msg      = s"""Impossible to parse allowed networks diff json. Expected {"allowed_networks": {"add":["192.168.2.0/24", ...]",""" +
+                   s""" "delete":["192.168.1.0/24", ...]"}}, got: ${if(json == JNothing) "nothing" else compactRender(json)}"""
+        _        <- if(json == JNothing) Failure(msg) else Full(())
+        diff     <- try { Full(json.extract[AllowedNetDiff]) } catch { case ex => Failure(msg) }
+        _        =  println("extracted: " + diff)
```
VinceMacBuche, line 959: Println to remove

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/SettingsApi.scala
```diff
@@ -903,6 +899,74 @@ final case object RestContinueGenerationOnError extends RestBooleanSetting {
...
+        JArray(res.map(JString).toList)
+      }
+      RestUtils.response(restExtractorService, "settings", Some(id))(result, req, s"Error when trying to modify allowed networks for policy server '${id}'")
+    }
+  }
 }
 
 
+// for lift-json extraction, must be top-level
+final case class AllowedNetDiff(add: List[String], delete: List[String])
```
VinceMacBuche, line 971: iirc, with Lists you have to have 'add' and 'delete' defined in your json with empty arrays. if they were options they could be optionnal. or you could use our extractors that don't rely on extract

VinceMacBuche: I would have done likewise: https://github.com/fanf/rudder/pull/11/files

fanf: the problem is that lift `extract` method doesn't fail on JNothing, certainly for good reason, but not expected. 
I prefect to use `extract` because it will much simpler to switch to a different json lib if we take the habit to defined `case class` for API and use such methods.

fanf:
PR updated with a new commit

api-doc/.gitignore
```diff
@@ -1,2 +0,0 @@
-/build
```
amousset, line 1: we should keep this

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #3367, opened 2020-11-15
Fixes #16773: Batch of new nodes can overflow rudder server with inventories

fanf:
https://issues.rudder.io/issues/16773

Add a `saveInventoryBuffer` that stays between inotify watcher (in charge of the complexe logic of making sense of inotify events and waiting for signature) and the part that [parse inventory, check signature, forward to backend] and is heavy on CPU and RAM. 
That buffer can be big (chose 1024 without much rationnal safe that it's big and a power of 2, as asked in ZQueue comment).

The main benefits are: 
- info about the file to process are cheap (a string with name, 3 functions to open stream or test if the file exists)
- we can deduplicate files while they are in the buffer super easily (and so avoid starvation case where some node goes made and send a lot of inventory). 

Only the code path for inventories coming from file system has the buffer, because for REST, the API await the pre-processing status. So we need to wait for signature check etc, and so that part must be synchronous. A better API would be to just move uploaded inventory/signature to `/var/rudder/inventories/incoming` (or a dedicated incoming directory) and from that point, let them follow the usual path - but it's an API change which will need a major version for change. 

Other important code changes: 
- in `InventoryFileWatcher.scala`, `sendToProcessor` was renamed to `sendToProcessorBlocking` and moved out of `processFile`. `processFile` now send files to process to the buffer. 
- in `InventoryProcessor.scala`, the backend waiting queue, `queue` was renamed to `blockingQueue` and as the name implies, it is now a BLOCKING queue. Its default size is reduced in `configurations.properties.sample` because now, the only interest of a big queue would be for people using A LOT rest endpoint, 
- a latch which dismiss new incoming requests when one is pending is added in front of `blockingQueue` in the REST API code path to keep previous, non-blocking behavior.

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
@@ -57,7 +57,6 @@ import zio.syntax._
 import zio.duration._
 import com.normation.zio._
 import org.joda.time.DateTime
-import zio.blocking.Blocking
 import zio.clock.Clock
 
 final class Watchers(incoming: FileMonitor, updates: FileMonitor) {
@@ -240,7 +239,6 @@ class InventoryFileWatcher(
   val failed   = File(failedInventoryPath)
   logDirPerm(failed, "Failed")
 
-  implicit val scheduler = ZioRuntime.unsafeRun(ZIO.access[Blocking](_.get.blockingExecutor.asEC).provide(ZioRuntime.environment))
   // the cron that look for old files (or existing ones on startup/watcher restart)
   val cronForMissed = new SchedulerMissedNotify(
       incoming :: updated :: Nil
@@ -249,7 +247,7 @@ class InventoryFileWatcher(
     , ZioRuntime.environment
   )
   // service that will actually process files (queue for event from inotify/cron for missed files)
-  val fileProcessor = new ProcessFile(inventoryProcessor.saveInventory, received, failed, waitForSig, sigExtension)
+  val fileProcessor = new ProcessFile(inventoryProcessor, received, failed, waitForSig, sigExtension)
   // That reference holds watcher instance and scheduler fiber to be able to stop them if asked
   val ref = RefM.make(Option.empty[(Watchers,Fiber[Nothing, Nothing])]).runNow
   def startWatcher() = semaphore.withPermit(
@@ -309,6 +307,7 @@ final case class SaveInventoryInfo(
     fileName          : String
   , inventoryStream   : () => InputStream
   , optSignatureStream: Option[() => InputStream]
+  , testExits         : UIO[Boolean]
 )
 
 /*
@@ -317,11 +316,11 @@ final case class SaveInventoryInfo(
  * For that, it needs signature and inventory file.
  */
 class ProcessFile(
-    saveInventory: SaveInventoryInfo => IOResult[InventoryProcessStatus]
-  , received     : File
-  , failed       : File
-  , waitForSig   : Duration
-  , sigExtension : String // signature file extension, most likely ".sign"
+    inventoryProcessor: InventoryProcessor
+  , received          : File
+  , failed            : File
+  , waitForSig        : Duration
+  , sigExtension      : String // signature file extension, most likely ".sign"
 ) {
   val sign = if(sigExtension.charAt(0) == '.') sigExtension else "."+sigExtension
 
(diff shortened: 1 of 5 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
```diff
@@ -350,12 +349,40 @@ class ProcessFile(
...
+   * take a file, we  de-duplicate files in buffer queue with the same name (ie: remove them).
+   * It likely means that we add several "whatch even: add" for that file (either copied several time, touched,
+   * or inotify event emission not what we thought)
+   */
+  val saveInventoryBufferProcessing = {
+    for {
+      inv <- saveInventoryBuffer.take
+      // deduplicate. TakeAll is not blocking
+      all <- saveInventoryBuffer.takeAll
+      _   <- saveInventoryBuffer.offerAll(all.filterNot { case (f, _) => inv._1.name == f.name })
```
ncharles, line 377: isn't there a risk that we loose inventory ?
Doc says:
   * For Sliding Queue: uses `Sliding` Strategy
   * If there is room in the queue, it places the values otherwise it removes the old elements and
   * enqueues the new ones. Always returns true.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
```diff
@@ -350,12 +349,40 @@ class ProcessFile(
...
+
+  /*
+   * We have a second queue that is used as buffer for the step between "file ok to be processed"
+   * and "saving file" because "saving file imply parsing XML and doing other memory and CPU
+   * inefficient.
+   * The buffer also allowed to filter inventory for identical nodes to avoid parsing old/not up to date
+   * inventory.
+   * As only a small structure is saved (SaveInventory is a name and two function to create stream from file), buffer can be big.
+   * When queue is full, we prefer to drop old inventories to new ones. In the worst case, they will be catched up by the
+   * scheduler.
```
ncharles, line 362: it can still be surprising. Why not using a bounded queue?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
ncharles: and it contradict the comment up there: if we reoffer old inventories, they become new

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
```diff
@@ -391,17 +418,57 @@ class ProcessFile(
...
+   * This is a blocking action, only call it with a non-blocking buffer in front of it.
+   */
+
+  protected def sendToProcessorBlocking(inventory: File, signature: Option[File]): UIO[Unit] = {
+    // we don't manage race condition very well, so we have cases where
+    // we can have two things trying to move
+    def safeMove[T](chunk: =>T): IOResult[Unit] = {
+      Task.effect{chunk ; ()}.catchAll {
+        case ex: NoSuchFileException => // ignore
+          InventoryProcessingLogger.debug(s"Ignored exception '${ex.getClass.getSimpleName} ${ex.getMessage}'. The was file was correctly handled.")
```
ncharles, line 442: The error message is not very explicit
"The file '${inventory.pathAsString}' was correctly handled" maybe?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
fanf: I prefer to have a non blocking, RAM-bounbed structure. If there is a bug somewhere in the code (which is far from impossible for concurrent and async processing which involves inotify and moving files around. A blocking structure can lead to thread accumulation or even deadlock, while here, in the worst case we loose an event (and so an inventory processing). But it's easy to set up workaround for that, which is: `SchedulerMissedNotify`.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryFileWatcher.scala
fanf: Arf :)

fanf:
PR updated with a new commit

ncharles (approved):

fanf:
OK, squash merging this PR

PR #3404, opened 2020-11-22
Fixes #18618: Pretty-print json properties in nodes list

VinceMacBuche:
https://issues.rudder.io/issues/18618


fanf:
I think we must do as it is done in node properties to:
1/ be consistent between views,
2/ have a good solution for huge JSON. It's especially important in list of nodes since it's likely that a huge json for a node property will be huge on several/most nodes.

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
It is now displayed like node properties in tab node properties

The problem with huge json should be handled by pagination i think ... or maybe we should go totally async for properties (and never send properties with details request, just send request properties after global request is ok, like we do when we add a property)

fanf:
For huge JSON, I was just thinking about the display problem, not the load one: I don't want to have a page of 100 nodes with 300 lines of JSON each always open. So you correction addresses that case correctly from my point of view. And yes, for perf problem, we will see :)

fanf (approved):

fanf:
OK, squash merging this PR

PR #3416, opened 2020-11-24
Fixes #18681: Inherited properties always displayed in the column, even if checkbox unchecked

VinceMacBuche:
https://issues.rudder.io/issues/18681


fanf:
I think that doesn't handle the case where you add two times the same property, one with inherited, the other with non inherited mode. 
In that case, after a reload (not just after adding the column), both column will show inherited values

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

fanf (approved):

VinceMacBuche:
OK, squash merging this PR

PR #3420, opened 2020-11-24
Fixes #18682: Error code should be 404 and not 500 when a node is not found for details API

fanf:
https://issues.rudder.io/issues/18682


VinceMacBuche (approved):

VinceMacBuche:
I guess it's almost the same case for all other API (Rules Directives ...)

fanf:
Hum, you're right. It's just more visible on node (because people are more likely to ask for node by id).

fanf:
OK, merging this PR

PR #3423, opened 2020-11-24
Fixes #18686: Reporting error when using twice "	File from local source" with parameter and without parameter in rudder 6.2

ncharles:
https://issues.rudder.io/issues/18686


ncharles:
so far, the unit test reproducing the issue

ncharles:
PR replaced by https://github.com/Normation/rudder/pull/3424

PR #3425, opened 2020-11-25
Fixes #18694: Rudder-pkg fails at upgrade when no newer version of a plugin are available

Fdall:
https://issues.rudder.io/issues/18694

### DIFF
--- a/relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkg.py
+++ b/relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkg.py
@@ -348,22 +348,23 @@ def update():
 def upgrade_all(mode):
     for p in utils.DB["plugins"].keys():
         currentVersion = rpkg.PluginVersion(utils.DB["plugins"][p]["version"])
+        latestVersion = currentVersion
         pkgs = plugin.Plugin(p)
         pkgs.getAvailablePackages()
         if mode == "nightly":
             latest_packages = pkgs.getLatestCompatibleNightly()
             if (latest_packages is None):
-                print("No newer %s compatible versions found for the plugin %s"%(mode, p))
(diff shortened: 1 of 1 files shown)


relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkg.py
```diff
@@ -350,20 +350,21 @@ def upgrade_all(mode):
         currentVersion = rpkg.PluginVersion(utils.DB["plugins"][p]["version"])
         pkgs = plugin.Plugin(p)
         pkgs.getAvailablePackages()
+        latestversion = None
```
VinceMacBuche, line 353: I would have put currentVersion in latestVersion, so we don't have to check if it is None

relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkg.py
```diff
@@ -350,20 +350,21 @@ def upgrade_all(mode):
...
         if mode == "nightly":
             latest_packages = pkgs.getLatestCompatibleNightly()
             if (latest_packages is None):
-                print("No newer %s compatible versions found for the plugin %s"%(mode, p))
+                logging.error("No newer %s compatible versions found for the plugin %s"%(mode, p))
```
VinceMacBuche, line 357: I don't think that's an error but an info and maybe we can get rid of that message here because it will be also printed at the end by a logging.info

VinceMacBuche: I think we should only do the else par (with is not None in if)

Fdall:
PR updated with a new commit

Fdall:
PR updated with a new commit

VinceMacBuche (approved):

Fdall:
OK, squash merging this PR

PR #3430, opened 2020-11-30
Fixes #18714: When there is an error after saving a directive, impossible to scroll down directive details

RaphaelGauthier:
https://issues.rudder.io/issues/18714

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/style/rudder/rudder-directives.css
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/style/rudder/rudder-directives.css
@@ -43,7 +43,8 @@
   height: calc(100VH - 80px);
 }
 #policyConfiguration,
-#policyConfiguration > form{
+#policyConfiguration > form,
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/webapp/style/rudder/rudder-directives.css
```diff
@@ -43,7 +43,8 @@
   height: calc(100VH - 80px);
 }
 #policyConfiguration,
-#policyConfiguration > form{
+#policyConfiguration > form,
+#policyConfiguration form > form{
```
VinceMacBuche, line 47: is it impossible to have two forms ??

VinceMacBuche: no it's ok to have several forms ! i guess we are not replacing correct element here ...

VinceMacBuche (approved):

PR #3457, opened 2021-01-04
Fixes #18772: Create a parser for encrypted properties feature

ElaadF:
https://issues.rudder.io/issues/18772

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/PropertyEngineService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/PropertyEngineService.scala
@@ -1,0 +1,44 @@
+package com.normation.rudder.services.nodes
+
+import com.normation.errors._
+import com.normation.zio._
+import zio.Ref
+import zio.UIO
+
+final case class EngineOption(name: String, value: String)
+
+trait RudderPropertyEngine {
+  def name: String
+
+  def process(namespace: List[String], opt: Option[List[EngineOption]]): IOResult[String]
+}
+
+trait PropertyEngineService {
+  def process(engineName: String, namespace: List[String], opt: Option[List[EngineOption]]): IOResult[String]
+
+  def addEngine(engine: RudderPropertyEngine): UIO[Unit]
+}
+
+class PropertyEngineServiceImpl(listOfEngine: List[RudderPropertyEngine]) extends PropertyEngineService {
+
+  private[this] val engines: Ref[Map[String, RudderPropertyEngine]] = (for {
+    v <- Ref.make[Map[String, RudderPropertyEngine]](listOfEngine.map(e => e.name.toLowerCase -> e).toMap)
+  } yield v).runNow
+
+  override def process(engine: String, namespace: List[String], opt: Option[List[EngineOption]]): IOResult[String] = {
+    for {
+      engineMap            <- engines.get
+      engineProp           <- engineMap.get(engine.toLowerCase)
+                                .notOptional(s"Engine '${engine}' not found. Can not be expanded")
+      interpolatedValueRes <- engineProp.process(namespace, opt)
+    } yield {
+      interpolatedValueRes
+    }
+  }
+
+  def addEngine(engine: RudderPropertyEngine): UIO[Unit] = {
+    for {
+      _ <- engines.update(_ + (engine.name.toLowerCase -> engine))
+    } yield ()
+  }
+}
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -40,62 +40,66 @@ package com.normation.rudder.services.policies
 import java.nio.charset.StandardCharsets
 import java.util.concurrent.TimeUnit
 import better.files.File
-import org.joda.time.DateTime
-import com.normation.rudder.domain.Constants
-import com.normation.rudder.domain.nodes.NodeInfo
-import com.normation.rudder.domain.policies._
-import com.normation.rudder.repository._
-import com.normation.rudder.services.eventlog.HistorizationService
-import com.normation.rudder.services.nodes.NodeInfoService
-import com.normation.rudder.services.policies.nodeconfig.NodeConfigurationHash
-import com.normation.utils.Control._
-import net.liftweb.common._
-import com.normation.inventory.services.core.ReadOnlyFullInventoryRepository
-import com.normation.inventory.domain.AcceptedInventory
-import com.normation.inventory.domain.NodeInventory
-import com.normation.rudder.domain.parameters.GlobalParameter
-import com.normation.rudder.domain.reports.NodeExpectedReports
-import com.normation.inventory.domain.NodeId
-import com.normation.rudder.domain.reports.NodeConfigId
-import com.normation.rudder.reports.ComplianceMode
-import com.normation.rudder.reports.ComplianceModeService
-import com.normation.rudder.reports.AgentRunIntervalService
-import com.normation.rudder.reports.AgentRunInterval
-import com.normation.rudder.domain.logger.ComplianceDebugLogger
-import com.normation.rudder.services.reports.{CacheComplianceQueueAction, CachedFindRuleNodeStatusReports}
-import com.normation.rudder.services.policies.write.PolicyWriterService
-import com.normation.rudder.reports.GlobalComplianceMode
-import com.normation.rudder.domain.appconfig.FeatureSwitch
-import com.normation.inventory.domain.AixOS
-import com.normation.rudder.batch.UpdateDynamicGroups
-import com.normation.rudder.domain.reports.NodeModeConfig
-import com.normation.rudder.reports.HeartbeatConfiguration
-import com.normation.rudder.hooks.RunHooks
-import com.normation.rudder.hooks.HookEnvPairs
-import com.normation.rudder.domain.nodes.NodeState
-import com.normation.rudder.services.policies.nodeconfig.NodeConfigurationHashRepository
-import com.normation.rudder.domain.reports.DirectiveExpectedReports
-import com.normation.rudder.domain.reports.ComponentExpectedReport
-import com.normation.rudder.domain.reports.RuleExpectedReports
-import com.normation.rudder.domain.logger.TimingDebugLogger
-import com.normation.rudder.domain.logger.PolicyGenerationLogger
 import cats.data.NonEmptyList
+import com.normation.box._
+import com.normation.errors._
+import com.normation.inventory.domain.AcceptedInventory
+import com.normation.inventory.domain.AixOS
 import com.normation.inventory.domain.MemorySize
+import com.normation.inventory.domain.NodeId
+import com.normation.inventory.domain.NodeInventory
+import com.normation.inventory.services.core.ReadOnlyFullInventoryRepository
+import com.normation.rudder.batch.UpdateDynamicGroups
+import com.normation.rudder.domain.Constants
+import com.normation.rudder.domain.appconfig.FeatureSwitch
+import com.normation.rudder.domain.logger.ComplianceDebugLogger
+import com.normation.rudder.domain.logger.PolicyGenerationLogger
+import com.normation.rudder.domain.logger.PolicyGenerationLoggerPure
+import com.normation.rudder.domain.logger.TimingDebugLogger
+import com.normation.rudder.domain.nodes.CompareProperties
+import com.normation.rudder.domain.nodes.GenericProperty
+import com.normation.rudder.domain.nodes.GenericProperty.StringToConfigValue
+import com.normation.rudder.domain.nodes.NodeInfo
+import com.normation.rudder.domain.nodes.NodeProperty
+import com.normation.rudder.domain.nodes.NodeState
+import com.normation.rudder.domain.parameters.GlobalParameter
+import com.normation.rudder.domain.policies._
+import com.normation.rudder.domain.reports.ComponentExpectedReport
+import com.normation.rudder.domain.reports.DirectiveExpectedReports
+import com.normation.rudder.domain.reports.NodeConfigId
+import com.normation.rudder.domain.reports.NodeExpectedReports
+import com.normation.rudder.domain.reports.NodeModeConfig
 import com.normation.rudder.domain.reports.OverridenPolicy
+import com.normation.rudder.domain.reports.RuleExpectedReports
+import com.normation.rudder.hooks.HookEnvPairs
 import com.normation.rudder.hooks.Hooks
 import com.normation.rudder.hooks.HooksLogger
+import com.normation.rudder.hooks.RunHooks
+import com.normation.rudder.reports.AgentRunInterval
+import com.normation.rudder.reports.AgentRunIntervalService
+import com.normation.rudder.reports.ComplianceMode
+import com.normation.rudder.reports.ComplianceModeService
+import com.normation.rudder.reports.GlobalComplianceMode
+import com.normation.rudder.reports.HeartbeatConfiguration
+import com.normation.rudder.repository._
+import com.normation.rudder.services.eventlog.HistorizationService
+import com.normation.rudder.services.nodes.MergeNodeProperties
+import com.normation.rudder.services.nodes.NodeInfoService
+import com.normation.rudder.services.policies.nodeconfig.NodeConfigurationHash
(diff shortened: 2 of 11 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -172,6 +173,7 @@ object PropertyParserTokens {
   final case class Param(path: List[String])       extends AnyVal with Interpolation
   //here, we keep the case as it is given
   final case class Property(path: List[String], opt: Option[PropertyOption]) extends Interpolation
+  final case class RudderEngine(engine: String, method: List[String], param: Map[String, ConfigValue] ) extends Interpolation
```
ElaadF, line 176: Not sure about the `param: Map[String, ConfigValue]`, can we only use `ConfigValue` ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -422,6 +424,7 @@ class InterpolatedValueCompilerImpl extends InterpolatedValueCompiler {
       case NonRudderVar(s)    => s"$${${s}}"
       case NodeAccessor(path) => s"$${rudder.node.${path.mkString(".")}}"
       case Param(name)         => s"$${rudder.param.${name}}"
+      case RudderEngine(e, m, p) => s"$${rudder.engine[${e}${m.mkString("][")}(${p.head._1})}"
```
ElaadF, line 427: Cf `param` implementation above for `${p.head._1}`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: yes, it shoudl be only config value

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
fanf: I think it should be an error, because it won't be a valid format (cfengine or dsc). 
We will need to decide what to do with that. Perhaps use something easy to debug like: 
```
s"[missing value for: $${rudder.engine[${e}${m.mkString("][")}](${p.toCorrectSerialisationMethodWeWantHere})}]"
```
(also you were missing a closing `]`

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
ElaadF: Wouldn't it more robust if we create :
```scala
trait Engine {
   def name: String
   def compute: String //logic here to get value
 }
 ```
instead of `engine: String` in `RudderEngine`

(Not sure if it will be expandable for plugins with the `trait Engine`

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala [hunk 3]
```diff
@@ -855,8 +905,24 @@ trait PromiseGeneration_BuildNodeContext {
...
+            , nodeParam.map{case(k, g) => (k, g.value)}.toMap
+          )
           _            =  {timeNanoMergeProp = timeNanoMergeProp + System.nanoTime - timeMerge}
-          nodeInfo     =  info.modify(_.node.properties).setTo(mergedProps.map(_.prop))
+          propsCompiled = mergedProps.map { p =>
+                            p.copy(prop = NodeProperty(p.prop.config.getString("name"), parseJValue(p.prop.toJson, contextEngine).toString.toConfigValue, None, None))
+                          }
+          _ <- effectUioUnit {mergedProps.foreach(p => println(p.prop.config))}.toBox
+          _ <- effectUioUnit(
+            println(s"======== JSON INTERPOLLATION : ${parseJValue(mergedProps(2).prop.toJson.\("value"), contextEngine).toString}")
```
fanf, line 923: that will need to be removed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -236,6 +239,7 @@ trait AnalyseInterpolation[T, I <: GenericInterpolationContext[T]] {
       case NonRudderVar(s)     => Right(s"$${${s}}")
       case NodeAccessor(path)  => getNodeAccessorTarget(context, path)
       case Param(path)         => getRudderGlobalParam(context, path)
+      case RudderEngine(e, m, p) => Right("encrypted-password-toto") // TODO remove only testing purpose replace by match case and engine's logic
```
fanf, line 242: you will need to create a service that is able to look-up engines and pass them the method and param, ie something like:

```
trait RudderPropertyEngine { def name: String, def process(namespace: List[String], parameters: ConfigValue): IOResult[String] } 
class RudderPropertyEngineService(listOfEngine: List[RudderPropertyEngine]) {
  val engines = listOfEngine.map(e => (e.name.toLowerCase, e)

  def process(engine: String, namespace: List[String], param: ConfigValue): IOResult[String] = engines.get(engine.toLowerCase).map(e => e.process(namespace, param)).notOptionnal(s"Engine '${engine}' not found. Parameter can not be expanded: ${param}")
}
```

And then, for test, you can implement a false encrypted password engine with mock values.

fanf (commented):
We need tests cases beyond just the parser, else it's very hard to see/understand what's going on.

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
PR updated with a new commit
Architecure V1: service in InterpolatedValueCompiler

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -212,30 +219,32 @@ trait AnalyseInterpolation[T, I <: GenericInterpolationContext[T]] {
...
    * unknown parameter, etc)
    */
-  def analyse(context: I, token:Token): PureResult[String] = {
+  def analyse(context: I, token:Token, propertyEngineService: PropertyEngineService): PureResult[String] = {
     token match {
       case CharSeq(s)          => Right(s)
       case NonRudderVar(s)     => Right(s"$${${s}}")
       case NodeAccessor(path)  => getNodeAccessorTarget(context, path)
       case Param(path)         => getRudderGlobalParam(context, path)
+      case RudderEngine(e, m, p) => Right(ZioRuntime.runNow(propertyEngineService.process(e, m, p)))
```
fanf, line 247: You don't have the right to do that, you are lying to code caller: the code it not pure, it actually runs a (side!) effect, and says it's always OK - here, an error will lead to an exception. 
You should have a `expandWithPropertyEngine(context: I, engineName: String (= e), nameSpace: List[String] (= m), params: ConfigValue (= p)): IOResult[String]`.
It will taint the whole thing to `IOResult`, which means that we go from "we have something that is pure/whitout side effects" to "only gods know what happen. Can block, call network, etc". 
We don't like `IOResult`, they are too powerfull. But most of the time, we don't have a choice.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndGlobalParameterLookup.scala
```diff
@@ -249,6 +271,59 @@ class TestNodeAndGlobalParameterLookup extends Specification {
       test(all(_), """${something . cfengine}""", List(CharSeq("${something . cfengine"), CharSeq("}")))
     }
 
+    "parse a valid engine" in {
+      test(
+        all(_)
+        , """${rudder.engine[foo]("tst")}"""
```
fanf, line 277: I'm not sure we want quote around "tst". But it can wait for the moment where we harden syntaxe & extend to be able to pass json

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -397,18 +406,18 @@ object AnalyseNodeInterpolation extends AnalyseInterpolation[ConfigValue, Interp
   }
 }
 
-class InterpolatedValueCompilerImpl extends InterpolatedValueCompiler {
+class InterpolatedValueCompilerImpl(propertyEngineService: PropertyEngineService) extends InterpolatedValueCompiler {
```
fanf, line 409: you correctly pass it as an argument here, but it should not be used in `parse` method but in the implementation of `expandWithPropertyEngine(context: I, engineName: String (= e), nameSpace: List[String] (= m), params: ConfigValue (= p)): IOResult[String]`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -422,6 +431,7 @@ class InterpolatedValueCompilerImpl extends InterpolatedValueCompiler {
       case NonRudderVar(s)    => s"$${${s}}"
       case NodeAccessor(path) => s"$${rudder.node.${path.mkString(".")}}"
       case Param(name)         => s"$${rudder.param.${name}}"
+      case RudderEngine(e, m, p) => s"[missing value for: $${rudder.engine[${e}${m.mkString("][")}](${p.render()})}]"
```
fanf, line 434: Please add a comment "should not happen since a non expanded engine property should lead to an error" or something alike

fanf (commented):
Some remark, but it's going in a good direction!

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
VinceMacBuche: Not sure we want to push everything to ZIO World here, we could but i think we can keep to PureResult here....

So either we are going like François said and we go to IOResult all the way down (until we need to ZIORuntime.runNow) but I think we may keep to pure Result here by running it into an either

like this ZioRuntime.runNow(io.either) 


But in all cases, like François said you can't put it in a Right like that here,

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
VinceMacBuche: Btw if you want to log, either use a logger or do 

_ = println(youmessage)

You don't neet it to be in a zio

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -212,30 +219,32 @@ trait AnalyseInterpolation[T, I <: GenericInterpolationContext[T]] {
...
+        case (Right(str), token) => analyse(context, token, propertyEngineService).map(s => (str + s))
         case (Left(err) , _    ) => Left(err)
       }
     }
 
     build _
   }
 
-  /*
+  /*u4i
```
VinceMacBuche, line 234: Comment to fix here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
VinceMacBuche: Maybe the signature should be 

def expandRudderEngineProperty (context : I, rudderEngine : RudderEngine) : IOResult/PureResult[String] 

to express that this function should only be used with a RudderEngine property, and not something else

ElaadF:
PR updated with a new commit

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndGlobalParameterLookup.scala
ElaadF: Do we want : `${engine.X[premiere-value][sous-champ-value] | option1 = xxx }` ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -172,6 +177,7 @@ object PropertyParserTokens {
   final case class Param(path: List[String])       extends AnyVal with Interpolation
   //here, we keep the case as it is given
   final case class Property(path: List[String], opt: Option[PropertyOption]) extends Interpolation
+  final case class RudderEngine(engine: String, method: List[String], param:ConfigValue ) extends Interpolation
```
ElaadF, line 180: If we choose `${engine.name[param1][param2] | option1 = xxx, option2 = xxx }` format should we add a `Option[List[String]]`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/PropertyEngineService.scala
```diff
@@ -0,0 +1,52 @@
...
+    } yield ()
+  }
+}
+
+class SecretEngine extends RudderPropertyEngine {
+  override def name: String = "secret"
+
+  override def process(namespace: List[String], parameters: ConfigValue): IOResult[String] = "encrypted-string-test" .succeed //TODO add the logic here
+
+}
```
ElaadF, line 52: Since the secret variable feature will be in a plugin, we should remove this

ElaadF:
PR updated with a new commit

ElaadF:
PR updated with a new commit

ElaadF:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -833,8 +852,24 @@ trait PromiseGeneration_BuildNodeContext {
...
+            , nodeContextBefore
+            , nodeParam.map{case(k, g) => (k, g.value)}.toMap
+          )
           _            =  {timeNanoMergeProp = timeNanoMergeProp + System.nanoTime - timeMerge}
-          nodeInfo     =  info.modify(_.node.properties).setTo(mergedProps.map(_.prop))
+          propsCompiled <- ZIO.foreach(mergedProps) { p =>
+                            for {
+                              x <- parseJValue(p.prop.toJson, contextEngine)
+                            } yield {
+                              p.copy(prop = NodeProperty(p.prop.config.getString("name"), x.toString.toConfigValue, None, None))
```
ncharles, line 869: There is a problem here
value of mergedProps is
List(Global(Quoted("### Managed by Rudder, edit with care ###")))), NodePropertyHierarchy(NodeProperty({
    # String: 1
    "name" : "test",
    # hardcoded value
    "provider" : "inherited",
    # String: 1
    "value" : "test"

propsCompiled is
List(Global(Quoted("### Managed by Rudder, edit with care ###")))), NodePropertyHierarchy(NodeProperty({
    # hardcoded value
    "name" : "test",
    # hardcoded value
    "value" : "JObject(List(JField(name,JString(test)), JField(value,JString(test)), JField(provider,JString(inherited)), JField(inheritMode,JNothing)))"
}

PR #3461, opened 2021-01-06
Fixes #18800: fix invalid escape in parameter strings

gpoblon:
https://issues.rudder.io/issues/18800


gpoblon:
a part of this fix comes from an upper-level issue which should be resolved. Therefore this PR should not be reviewed until my next commit

gpoblon:
PR updated with a new commit

rudder-lang/src/parser.rs
```diff
@@ -379,6 +379,21 @@ fn enum_not_expression(i: PInput) -> PResult<PEnumExpressionPart> {
     )(i)
 }
 
+/// An unescaped string is a literal string delimited by '"""'.
```
peckpeck, line 382: Why move this ? and is it just a move ?

peckpeck (approved):

gpoblon:
OK, merging this PR

gpoblon:
OK, merging this PR

gpoblon:
OK, merging this PR

rudder-lang/src/technique.rs
```diff
@@ -433,14 +433,14 @@ impl Parameter {
                 self.value
             )));
         }
-        Ok((format!("{:#?}", self.value), None))
+        Ok((format!("\"\"\"{}\"\"\"", self.value), None))
```
VinceMacBuche, line 436: Here all parameters in rl files are unescaped strings, is that what we want ?? Don't we want to keep an unescaped (triple double quotes) string if we got one and and escaped (single double quote string if we got one (

PR #3471, opened 2021-01-12
Fixes #12937: When receiving an inventory, the merge_uuid part can be painfully slow

fanf:
https://issues.rudder.io/issues/12937


fanf:
This code is totally broken: for whatever reason, I thought that software RDN was (name, version), not a random uuid

PR #3480, opened 2021-01-20
Fixes #18845: Add a settings to delay start of policy generation

VinceMacBuche:
https://issues.rudder.io/issues/18845

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
@@ -52,10 +52,11 @@ import com.normation.eventlog.EventLogDetails
 
 import scala.xml.NodeSeq
 import com.normation.eventlog.ModificationId
+
 import java.io.File
-
-
 import com.normation.rudder.domain.logger.PolicyLogger
+
+import scala.concurrent.duration.Duration
 
 sealed trait StartDeploymentMessage
 
@@ -66,6 +67,31 @@ final case class AutomaticStartDeployment(modId: ModificationId, actor:EventActo
 //ask for a new deployment - manual deployment (human clicked on "regenerate now"
 //actor: the actor who asked for the deployment
 final case class ManualStartDeployment(modId: ModificationId, actor:EventActor, reason:String) extends HashcodeCaching with StartDeploymentMessage
+
+
+sealed trait PolicyGenerationTrigger
+object PolicyGenerationTrigger {
+  final case object AllGeneration extends PolicyGenerationTrigger
+  final case object OnlyManualGeneration extends PolicyGenerationTrigger
+  final case object NoGeneration extends PolicyGenerationTrigger
+
+  def apply(value : String) : Box[PolicyGenerationTrigger] = {
+
+    value.toLowerCase() match {
+      case "all" | "allgeneration" => Full(AllGeneration)
+      case "none" | "nogeneration" => Full(NoGeneration)
+      case "manual" | "onlymanual" | "onlymanualgeneration" => Full(OnlyManualGeneration)
+      case _ => Failure(s"'${value}' is not a valid generation policy")
+    }
+  }
+  def serialize(generationPolicy: PolicyGenerationTrigger) : String = {
+    generationPolicy match {
+      case AllGeneration => "all"
+      case NoGeneration => "none"
+      case OnlyManualGeneration => "onlyManual"
+    }
+  }
+}
 
 /**
  * State of the deployment agent.
@@ -112,6 +138,8 @@ final class AsyncDeploymentActor(
     deploymentService: PromiseGenerationService
   , eventLogger:EventLogDeploymentService
   , deploymentStatusSerialisation : DeploymentStatusSerialisation
+  , getGenerationDelay : () => Box[Duration]
+  , deploymentPolicy: () => Box[PolicyGenerationTrigger]
 ) extends LiftActor with ListenerManager with AsyncDeploymentAgent {
 
   deploymentManager =>
@@ -216,6 +244,16 @@ final class AsyncDeploymentActor(
       }
     }
   }
+
+  private[this] def logTriggerError(v : Box[PolicyGenerationTrigger]) = {
+    v match {
+      case eb: EmptyBox =>
+        val fail = eb ?~! "Policy generation: An error occurred while getting generation policy, Starting policy generation even if not authorized"
+        logger.error(fail.messageChain)
+      case _ => ()
+    }
+
+  }
   override protected def lowPriority = {
 
     //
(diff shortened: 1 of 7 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -309,9 +309,21 @@ trait PromiseGenerationService {
...
     PolicyLogger.debug(s"Policy generation parallelism set to: ${maxParallelism} (change with REST API settings parameter 'rudder_generation_max_parallelism')")
     PolicyLogger.debug(s"Policy generation JS evaluation of directive parameter timeout: ${jsTimeout} s (change with REST API settings parameter 'rudder_generation_jsTimeout')")
     PolicyLogger.debug(s"Policy generation continues on NodeConfigurations evaluation: ${generationContinueOnError} (change with REST API settings parameter 'rudder_generation_continue_on_error')")
+    if (delay.toMillis == 0L) {
+      PolicyLogger.debug(s"Policy generation will now start!")
+    } else {
+      PolicyLogger.debug(s"Policy generation will now hang for ${delay.toString} before effective start")
+    }
+
+    Thread.sleep(delay.toMillis)
```
fanf, line 326: I don't think we want it to be here. It should be out of generation duration at least, and I think from a code point of view, it should be out of the `deploy` method. Can it be just before its call?

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -373,6 +375,16 @@ final class AsyncDeploymentActor(
...
         logger.trace("Policy updater Agent: start to update policies")
         try {
+          val delay = getGenerationDelay() match {
+            case Full(delay) => delay
+            case eb : EmptyBox => Duration("Os")
+          }
+          if (delay.toMillis != 0L) {
+            logger.debug(s"Policy generation will start in ${delay.toString}")
+          }
+          logger.debug(s"Policy generation will starts now!")
```
fanf, line 385: that log needs to be after the sleep

fanf:
You can self-merge after moving log line

fanf:
We should add a case to disable policy generation completly: 

- `rudder.generation.enable`: boolean

And the behavior is: if generation is disabled, then skip policy generation and log a big info message.

VinceMacBuche:
I think we want to disable automatic generation and allow manual ones

fanf:
Good idea. It could be a tri-state: 

- none, 
- only manual, 
- all

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -227,32 +254,43 @@ final class AsyncDeploymentActor(
...
+        case Full(GenerationPolicy.NoGeneration | GenerationPolicy.OnlyManualGeneration) =>
+          logger.trace("Policy generation: Due to policy generation policy, no automatic policy generation was triggered ")
 
+        case v @ (Full(GenerationPolicy.AllGeneration) | _: EmptyBox) =>
+          v match {
+            case eb: EmptyBox =>
+              val fail = eb ?~! "Policy generation: An error occurred while getting generation policy, Starting automatic policy generation even if not authorized"
+              logger.error(fail.messageChain)
+            case _ => ()
+          }
```
fanf, line 267: can you put all of that in a:
 ```
def logErrorGettingGenerationPolicy(v: Box[GenerationPolicy]): Unit = {
          v match {
            case eb: EmptyBox =>
              val fail = eb ?~! "Policy generation: An error occurred while getting generation policy, Starting automatic policy generation even if not authorized"
              logger.error(fail.messageChain)
            case _ => ()
}
```

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala [hunk 4]
```diff
@@ -67,6 +68,30 @@ final case class AutomaticStartDeployment(modId: ModificationId, actor:EventActo
 //actor: the actor who asked for the deployment
 final case class ManualStartDeployment(modId: ModificationId, actor:EventActor, reason:String) extends HashcodeCaching with StartDeploymentMessage
 
+
+sealed  trait GenerationPolicy
```
fanf, line 72: I find GenerationPolicy extremely misleading, my brain absolutly wants to read "policy generation" and interpret it like that. 
Pehaps use: `GenerationTriggerEvent` (all, onlyManual, none) ?

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -67,6 +68,31 @@ final case class AutomaticStartDeployment(modId: ModificationId, actor:EventActo
...
+object PolicyGenerationTrigger {
+  final case object AllGeneration extends PolicyGenerationTrigger
+  final case object OnlyManualGeneration extends PolicyGenerationTrigger
+  final case object NoGeneration extends PolicyGenerationTrigger
+
+  def apply(value : String) : Box[PolicyGenerationTrigger] = {
+
+    value.toLowerCase() match {
+      case "all" | "allgeneration" => Full(AllGeneration)
+      case "none" | "nogeneration" => Full(NoGeneration)
```
fanf, line 82: when we are in "none" mode, it's a bit surprising to have nothing done when you click on "regenerate all". Perhaps we should disable that button, or just add an explanation text below "status" in the drop down: "generation mode: none ; delay: 5 minutes" with a long explanation on mouse over.

fanf: perhaps we can had that message only in 6.2 to minimize changes in 5.0/6.1 (in these versions, it will be documented only in settings API)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
fanf: s/starts/start

fanf (commented):
`CheckInventoryUpdate` should be an automatic update, not a manual one.

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
VinceMacBuche: I opened https://issues.rudder.io/issues/18896 in 6.2

fanf (approved):

fanf:
OK, squash merging this PR

PR #3493, opened 2021-02-02
Fixes #18878: handle rudderlang types expressions in conditions

gpoblon:
https://issues.rudder.io/issues/18878


gpoblon:
We decided that expressions should stick to handling enum and not rl types

PR #3494, opened 2021-02-03
Fixes #18879: API v14 with zio-json in place of lift-json

fanf:
https://issues.rudder.io/issues/18879

This PR goal is to demonstrate that we can build API without using `lift-json` at all with the long term goal to be able to remove "lift-json" as a dependency because it is on the path of either removing `lift` or removing `scalac` from the classpath, it is higly inefficient(`json` parsing is becoming one the main CPU/Memory usage of rudder, and `lift-json` is for something on that) and it may be unsafe (requiring runtime reflexion to do things). 

This pull request introduces the use of `zio-json` for several API. It leads to addition of a lot of tests in these API to ensure a consistency in behavior and it is built to try to be the less disruptive possible for old (up to v13) API version. 
So most of the code is built to not touch at all (and so not even refactor) existing v13, but just add new code. 

It put in light that there is A LOT of inconsistancies in our existing API responses, even if we put a lot of care in them - typically, we are not consistant at all about the layout of success answer (regaring the presence of an `id` for action on one object (whose usefulness is doubtfull), the presence of an array for data (always/only for bulk operation), etc. 

General feeling: `zio-json` use `case classes` to describe (and enforce) the API, for input and output. So there is a lot of dedicated case classes to write, sometime varying just in details. It needs a bit of time to find what works, but once the general organization is found, things are quite regular. The hardest part is to create the test plateform to be sure that we keep the same inconsistencies (or that we change them only willingly). And the addition of `chimney` (see below) helps reduce boilerplate.

For the big changes: 

- `RuleApi` (and the other ported API: `GlobalParameter`, `Directive`, `Technique`, `Group`) has a new service (`RuleApiServiceV14`) with all the code logic, and I duplicated each object for API (with a choice based on the version)

- `JsonSchema.json` is the place where all the new code is going. There is 3 main things: 

1/ `RudderJsonResponse` is where the object for generic response are defined with all the methods to facilitate its construction. You can see how it is used in `RuleApi` V14 objects.  

2/ JsonResponseObjects (`JR*`:`JRRule` etc) and JsonQueryObjects (`JQ*`:`JQRule` etc) is where all JSON REST objects (for query and responses) are defined. As `zio-json` is totally based on mapping that to/from string, they are very important since they actually define THE API. 

3/ `implicits` where zio `JsonEncoder` and `JsonDecoder` are defined. That's where the logic (most of it is automatic) to encode/decode json is done.

This PR also normalize the parallel between JSON and `form-encoded` requests. Now, they take the same parameters, and use the same decoders for complex ones (like directive parameter, rule target, etc). 

There are also the following small changes:
- we give access by default to a `DefaultParam` structure (`prettify`, `changeRequest` name and description, `reason`) that are not specific to one API request (and so, we factor-out the corresponding boilerplaty code); same for request actor. 
- we add a notion of "dataContainer" in `EndpointDefinition` to know how to name the array in JSON responses with multiple objects
- I cleaned-up `TechniqueVersion` which was over-complicated for nothing and made things harder than they should have
- this PR introduce the use of `chimney`, and utility lib that allows to copy data between similar case classes, and patch case classes with corresponding case class with optional fields. Not mandatory, but clean some of our boilerplate on these exact parts. 

Finally, this PR brings a LOT of unit tests for API. 
For that, we use YAML files under `src/test/resources/api`, one by API endpoint to test (by convention), where we describe both the requests and the expected responses. Both JSON and form-encoded requests can be done. 
For the back-end, we reimplemented all repository in a pure in-memory fashion, and created a coherent set of data, so that we can really play all the business logic, up to policy generation (which is not done). This is the most time-consuming part, but also the one that already helped discover tens of bugs. 
Plus, it allows to play request with both v13 and v14 of the API and check what changed, and if it's OK.  
All the repositories/mock code is in `MockServices.scala`, the test environment init is in `RestTestSetUp.scala`. A lot of it was backported to 6.1 release

This is a good time to discuss internal API / naming / etc


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
```diff
@@ -75,11 +75,14 @@ class RoLDAPRuleRepository(
...
    */
-  def get(id:RuleId) : IOResult[Rule]  = {
+  def getOpt(id:RuleId) : IOResult[Option[Rule]]  = {
     ruleMutex.readLock(for {
       con     <- ldap
-      crEntry <- con.get(rudderDit.RULES.configRuleDN(id.value)).notOptional(s"Rule with id '${id.value}' was not found")
-      rule    <- mapper.entry2Rule(crEntry).toIO.chainError("Error when transforming LDAP entry into a rule for id %s. Entry: %s".format(id, crEntry))
+      crEntry <- con.get(rudderDit.RULES.configRuleDN(id.value))
+      rule    <- crEntry match {
+                   case None    => None.succeed
```
ncharles, line 83: this changes the semantic, and should at least be updated in the comment
(but I find this much easier to understand)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -68,28 +68,35 @@ object ComplianceApi extends ApiModuleProvider[ComplianceApi] {
...
   }
   final case object GetNodeComplianceId extends ComplianceApi with OneParam with StartsAtVersion7 with SortIndex { val z = implicitly[Line].value
     val description = "Get compliance information for the given node"
     val (action, path)  = GET / "compliance" / "nodes" / "{id}"
+    val dataContainer = Some("nodes")
   }
   final case object GetGlobalCompliance extends ComplianceApi with ZeroParam with StartsAtVersion10 with SortIndex { val z = implicitly[Line].value
     val description = "Get the global compliance (alike what one has on Rudder main dashboard)"
     val (action, path)  = GET / "compliance"
+    val dataContainer = Some("gobalCompliance")
```
ncharles, line 91: typo: globalCompliance

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/JsonSchema.scala
```diff
@@ -0,0 +1,598 @@
...
+    , longDescription : String
+    , directives      : List[String] //directives ids
+    , targets         : List[JRRuleTarget]
+    , enabled         : Boolean
+    , system          : Boolean
+    , tags            : List[Map[String, String]]
+  )
+  object JRRule {
+    // create an empty json rule with just ID set
+    def empty(id: String) = JRRule(None, id, "", "", "", "", Nil, Nil, false, false, Nil)
```
ncharles, line 235: why do we need an empty one ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPRuleRepository.scala
fanf: I could let the `def get = getOpt.notOptionnal(...)`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/JsonSchema.scala
fanf: It's in the case of deletion, when the rule is absent, we don't have more info, only the id.

fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
PR rebased

fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/nodes/Properties.scala
```diff
@@ -446,6 +471,33 @@ object GenericProperty {
       case JArray(arr)      => ConfigValueFactory.fromIterable(arr.map(x => fromJsonValue(x)).asJava)
     }
   }
+  def fromZioJson(value: zio.json.ast.Json): ConfigValue = {
```
fanf, line 474: `zio-json` AST, used to parse a generic config value (that can be any json) into a property value.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/JsonSchema.scala
```diff
@@ -0,0 +1,1220 @@
+/*
```
VinceMacBuche, line 1: This file needs to be splitted. I see 4 files (at least)

* JsonResponse.scala: containing everything related to making a response (abstract and lift one)
* JsonResponseObjects.scala: containing all classes that represents our domain andsend as response(this file will be sponsored by JR Ewing from 'Dallas' TV soap )
* JsonEncoders.scala: containing all data encoders
* JsonQueryObjects: containg all classe that represents all data received for our queries

webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/angular/nodeProperties.js
```diff
@@ -131,7 +131,7 @@ nodePropertiesApp.controller('nodePropertiesCtrl', function ($scope, $http, DTOp
     var getUrlAPI = contextPath + '/secure/api/'+ objectName +'s/' + nodeId + '/displayInheritedProperties';
     $scope.fetchProperties = function() {
       return $http.get(getUrlAPI).success( function (result) {
-        $scope.properties = result.data[0].properties
+        $scope.properties = result.data[objectName +'s'][0].properties
```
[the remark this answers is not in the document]
fanf: it was broken and part of the normalization effort.

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/JsonSchema.scala
VinceMacBuche: I'm not sure about the advantage of all being in the same file, but if you think so, we will see :)

amousset:
Tests fail on this one. Do you know why the execution right regularly disappears from hooks?

```
[2021-05-31T08:16:05.848Z] + webapp/sources/rudder/rudder-core/src/test/resources/hooks.d/test-hooks.sh

[2021-05-31T08:16:05.848Z] Hooks is not executable: ./webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-ready/10-cf-promise-check

script returned exit code 1
```

VinceMacBuche (approved):

PR #3499, opened 2021-02-03
Fixes #18877: "Skipped" Directives on Rules Screen

fanf:
https://issues.rudder.io/issues/18877

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
@@ -526,14 +526,14 @@ object ComplianceData extends Loggable {
 
     //we can have rules with only overriden reports, so we just prepend them. When
     //a rule is defined for that id, it will override that default.
-    val overridesRules = overridesByRules.view.mapValues(_ => AggregatedStatusReport(Nil))
+    val overridesRules = overridesByRules.view.mapValues(_ => AggregatedStatusReport(Nil)).toMap
 
     val ruleComplianceLine = for {
       (ruleId, aggregate) <- (overridesRules ++ report.byRules)
       rule                <- rules.find( _.id == ruleId )
     } yield {
       val nodeMode = allNodeInfos.get(nodeId).flatMap { _.policyMode }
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -526,14 +526,14 @@ object ComplianceData extends Loggable {
 
     //we can have rules with only overriden reports, so we just prepend them. When
     //a rule is defined for that id, it will override that default.
-    val overridesRules = overridesByRules.view.mapValues(_ => AggregatedStatusReport(Nil))
+    val overridesRules = overridesByRules.view.mapValues(_ => AggregatedStatusReport(Nil)).toMap
```
fanf, line 529: This is important, it was the cause of directive duplication in node's compliance screen

VinceMacBuche (approved):

ncharles:
Why keep a view here if you .toMap it?

Le 3 février 2021 23:31:05 GMT+01:00, Francois Armand <notifications@github.com> a écrit :
>@fanf commented on this pull request.
>
>
>
>> @@ -526,14 +526,14 @@ object ComplianceData extends Loggable {
> 
>//we can have rules with only overriden reports, so we just prepend
>them. When
>     //a rule is defined for that id, it will override that default.
>-    val overridesRules = overridesByRules.view.mapValues(_ =>
>AggregatedStatusReport(Nil))
>+    val overridesRules = overridesByRules.view.mapValues(_ =>
>AggregatedStatusReport(Nil)).toMap
>
>This is important, it was the cause of directive duplication in node's
>compliance screen
>
>-- 
>You are receiving this because you are subscribed to this thread.
>https://github.com/Normation/rudder/pull/3499#pullrequestreview-582862830

-- 
Envoyé de mon appareil Android avec Courriel K-9 Mail. Veuillez excuser ma brièveté.

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
ncharles: Why keeping the view if there's a toMap right after?

PR #3503, opened 2021-02-14
Fixes #18900: Missing exception details in change request update

fanf:
https://issues.rudder.io/issues/18900


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
```diff
@@ -538,6 +538,7 @@ class ModificationValidationPopup(
             case eb:EmptyBox =>
               val e = eb ?~! "Error when moving the group (no change request was created)"
               //early return here
+              logger.error(e.messageChain)
               e.chain.foreach { ex =>
```
fanf, line 542: the `foreach` here only tests the `e` parent (ie `eb`) exists and is non-empty and display its `messageChain` but doesn't log it (I'm not sure why, but I prefer not to change that).

VinceMacBuche (approved):

PR #3505, opened 2021-02-16
Fixes #18902: Error message when editing properties in the interface

RaphaelGauthier:
https://issues.rudder.io/issues/18902

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/angular/nodeProperties.js
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/angular/nodeProperties.js
@@ -163,7 +163,6 @@ nodePropertiesApp.controller('nodePropertiesCtrl', function ($scope, $http, DTOp
       if(!$scope.alreadyUsed){
         $http.post($scope.urlAPI, data).then(function successCallback(response) {
           $scope.errorSaving = false;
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/angular/nodeProperties.js
```diff
@@ -163,7 +163,6 @@ nodePropertiesApp.controller('nodePropertiesCtrl', function ($scope, $http, DTOp
       if(!$scope.alreadyUsed){
         $http.post($scope.urlAPI, data).then(function successCallback(response) {
           $scope.errorSaving = false;
-          $scope.properties.push(propertyToSave);
```
RaphaelGauthier, line 166: The problem cames from the fact that the list of properties is updated twice when a property is added/modified.

This line is not necessary since we call `$scope.fetchProperties()` in a second step, which will fetch the list of all properties (plus the one we just added/modified) and update **$scope.properties** with the result.

fanf: why is not always failing ?

RaphaelGauthier:
PR replaced by https://github.com/Normation/rudder/pull/3510

PR #3513, opened 2021-02-19
Fixes #18926: nodes API with include managementTechnologyDetails leads to error 500 response

fanf:
https://issues.rudder.io/issues/18926

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
+++ b/webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
@@ -935,12 +935,17 @@ class InventoryMapper(
       agentNames         <- {
                               val agents = entry.valuesFor(A_AGENTS_NAME).toSeq.map(Some(_))
                               val agentWithKeys = agents.zipAll(publicKeys, None,None).filter(_._1.isDefined)
-                              ZIO.foreach(agentWithKeys) {
-                                case (Some(agent), key) =>
-                                  AgentInfoSerialisation.parseCompatNonJson(agent,key).chainError(s"Error when parsing agent security token '${agent}'")
-                                case (None, key)        =>
-                                  InventoryMappingRudderError.MissingMandatory("Error when parsing agent security token: agent is undefined").fail
-                              }
+                              ZIO.foreach(agentWithKeys) { case (opt, key) => ((opt, key) match {
+                                  case (Some(agent), key) =>
+                                    AgentInfoSerialisation.parseCompatNonJson(agent,key).chainError(s"Error when parsing agent security token '${agent}'")
+                                  case (None, key)        =>
+                                    InventoryMappingRudderError.MissingMandatory("Error when parsing agent security token: agent is undefined").fail
+                                }).foldM(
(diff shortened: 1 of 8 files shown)


webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
```diff
@@ -935,12 +935,17 @@ class InventoryMapper(
       agentNames         <- {
                               val agents = entry.valuesFor(A_AGENTS_NAME).toSeq.map(Some(_))
                               val agentWithKeys = agents.zipAll(publicKeys, None,None).filter(_._1.isDefined)
-                              ZIO.foreach(agentWithKeys) {
-                                case (Some(agent), key) =>
-                                  AgentInfoSerialisation.parseCompatNonJson(agent,key).chainError(s"Error when parsing agent security token '${agent}'")
-                                case (None, key)        =>
-                                  InventoryMappingRudderError.MissingMandatory("Error when parsing agent security token: agent is undefined").fail
-                              }
+                              ZIO.foreach(agentWithKeys) { case (opt, key) => ((opt, key) match {
```
ncharles, line 938: why do you need the extra  case?

fanf: we use a syntax shortcut in methods accepting function as argument: you can create a partial function with the following syntax `{ case x => ... }` and so use it directly as argument, what we regularly do for `map` and alike function - and in that case, we are careful to use a partial function which is actually total by not imposing restriction on the `x`:
```
myUserList.map { case User(name, age) => name.capitalize }
```

But that really syntax for partial function, so you can put as many guard as you want in the pattern, use as many pattern as you want, and even don't make it total (and be sad when the pattern match exception happens):

```
myUserList.map { 
  case User(name, age) if age > 17 => 
  case User(name, age) if age > 5  => 
// and pattern match exception when age <= 6, but YOLO!
}
```

So in our case, since I wanted to have a different logic for the two cases of `Option`:

```
ZIO.foreach(agentWithKeys) { 
  case (Some(agent), key) => ...
  case (None, key)        =>
}
```

But then, I needed to apply a `fold` on result, and I don't know of a syntax that allow to do that with several pattern match in the partial function syntact, so I reverted to the usual pattern match syntax and only one branch for the partial function: 


```
ZIO.foreach(agentWithKeys) { case (opt, key) =>  //that's for the function arg of `foreach`
  (opt, key) match { // that's a standard pattern matching 
    case (Some(agent), key) => ...
    case (None, key)        =>
  }
}
```

And with the fold: 
```
ZIO.foreach(agentWithKeys) { case (opt, key) =>  //that's for the function arg of `foreach`
  ((opt, key) match { // that's a standard pattern matching 
    case (Some(agent), key) => ...
    case (None, key)        =>
  }).fold(....)
}
```

ncharles: Thank you for the time spent for the superb explanation of why adding the extra case, i learnt something

ncharles (approved):

fanf:
OK, merging this PR

PR #3521, opened 2021-03-02
Fixes #18959: Inherited node properties are not returned in API

fanf:
https://issues.rudder.io/issues/18959

### DIFF
--- a/webapp/sources/api-doc/components/schemas/node-inherited-properties.yml
+++ b/webapp/sources/api-doc/components/schemas/node-inherited-properties.yml
@@ -1,0 +1,68 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2020 Normation SAS
+type: object
+required:
+  - id
+  - properties
+properties:
+  id:
+    type: string
+    description: Unique id of the node
+    format: uuid (or "root")
+    example: 9a1773c9-0889-40b6-be89-f6504443ac1b
+  properties:
+    type: array
+    description: Node properties (either set by user or filled by third party sources)
+    items:
+      type: object
+      required:
+        - name
+        - value
+      properties:
+        name:
+          type: string
+          description: Property name
+          example: datacenter
+        value:
+          format: string or JSON
+          example: AMS2
+          description: Resolved (ie, with inheritance and overriding done) property value (can be a string or JSON object)
+        provider:
+          type: string
+          description: Property provider (if the property is not a simple node property)
+          enum:
+          - inherited
+          - overridden
+          - plugin name like datasources
+        hierarchy:
+          type: array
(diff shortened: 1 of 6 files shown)


VinceMacBuche (approved):

webapp/sources/api-doc/components/schemas/node-inherited-properties.yml
```diff
@@ -0,0 +1,44 @@
...
+          description: Property provider (if the property is not a simple node property)
+          enum:
+          - inherited
+          - overridden
+          - plugin name like datasources
+        hierarchy:
+          type: string
+          description: An HTML description of the inheritance hierarchy for that property
+          example: "<p>from <b>Global Parameter</b>:<pre>\"start;\"</pre></p>"
+        origval:
```
amousset, line 41: original?

webapp/sources/api-doc/components/schemas/node-inherited-properties.yml
```diff
@@ -0,0 +1,44 @@
...
+          type: string
+          description: Property provider (if the property is not a simple node property)
+          enum:
+          - inherited
+          - overridden
+          - plugin name like datasources
+        hierarchy:
+          type: string
+          description: An HTML description of the inheritance hierarchy for that property
+          example: "<p>from <b>Global Parameter</b>:<pre>\"start;\"</pre></p>"
```
amousset, line 40: do we really want to promise to never change the html layout? I'm not sure it should be part of the public API.

fanf: Well, no. But it's what was done and is already in the wild. I can emphasis that it's a non structured format. I can remove it from API too, but I can't replace it easely/quickly.

fanf: Nonetheless, perhaps it's the correct solution

amousset: I think I'd prefer to remove it from the API and add a structured format in the future

fanf:
PR updated with a new commit

fanf:
Changed the HTML rendering with a json structure and updated doc accordingly

webapp/sources/api-doc/components/schemas/node-inherited-properties.yml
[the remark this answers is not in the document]
amousset: @fanf I don't understand what's this field from the doc.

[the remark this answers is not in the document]
fanf: it's a description of the inheritance chain for that property, so that an user can see from where the value comes. It contains the list of groups and/or global values that leads to the node property overriden value.

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #3523, opened 2021-03-04
Test dyn group improvement

ncharles:


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
```diff
@@ -112,18 +119,33 @@ class DynGroupServiceImpl(
...
       def numberOfQuery(group : NodeGroup) = group.query.map( _.criteria.filter(_.objectType.objectType == "group").size).getOrElse(0)
-      dyngroupIds.sortBy(numberOfQuery)
+      dyngroups.sortBy(numberOfQuery)
     }
   }.toBox
 
+  def getAllDynGroupsWithandWithoutDependencies(): Box[(Seq[NodeGroupId], Seq[NodeGroupId])] = {
+    (for {
+      con         <- ldap
+      entries     <- con.searchSub(rudderDit.GROUP.dn, dynGroupFilter, dynGroupAttrs:_*)
```
ncharles, line 137: here we should not take the nodeids, it would be even faster
it takes several seconds to fetch data when there is a large number of groupes

ncharles: ha, it is already not fetching it

ncharles:
replaced by  https://github.com/Normation/rudder/pull/3577

PR #3532, opened 2021-03-09
Fixes #18995: Missing mandatory directive parameter doesn't fail policy generation

fanf:
https://issues.rudder.io/issues/18995


fanf:
The proposed PR leads to error like that: 
image

fanf:
Actually, we don't need to deconstruct and rebuild map param for each nodes!

fanf:
PR updated with a new commit

ncharles (approved):

PR #3534, opened 2021-03-10
Fixes #18972: API to use secret database 

ElaadF:
https://issues.rudder.io/issues/18972

PR #3536, opened 2021-03-10
Ust 18972/api to use secret database

ElaadF:


ElaadF:
Wrong branch go here : https://github.com/Normation/rudder/pull/3534

PR #3551, opened 2021-03-19
Fixes #18999: when we install a plugin, if the dependency (package manager) is not met, it still tries to install it and fails

ncharles:
https://issues.rudder.io/issues/18999

### DIFF
--- a/relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
+++ b/relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
@@ -16,6 +16,17 @@ except Exception:
 
 logger = logging.getLogger("rudder-pkg")
 
+try:
+    import rpm
+except:
+    logger.debug("No rpm python lib found")
+
+try:
+    import apt
+except:
+    logger.debug("No apt python lib found")
+
+
 # See global variables at the end of the file
 """ Get Terminal width """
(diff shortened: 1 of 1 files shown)


ncharles:
Commit modified

relay/sources/rudder-pkg/lib/rudder-pkg/rudderPkgUtils.py
```diff
@@ -307,7 +318,36 @@ def install_dependencies(metadata):
             logger.warning("The binary " + executable + " was not found on the system, you must install it before installing " + metadata['name'])
             return False
       else:
```
Fdall, line 320: You should add a case in the same way that the "binary" type was defined. I am pretty sure we want to have typed dependencies and so, an "apt" and a "rpm" type of dependencies..

Fdall (changes requested):

ncharles:
PR updated with a new commit

Fdall (approved):

fanf:
OK, squash merging this PR

fanf:
OK, merging this PR

fanf:
OK, merging this PR

fanf:
OK, merging this PR

peckpeck:
OK, merging this PR

VinceMacBuche:
OK, merging this PR

PR #3553, opened 2021-03-23
Fixes #19061: Use rudderc to compile technique from the editor instead of rudder logic

VinceMacBuche:
https://issues.rudder.io/issues/19061

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -79,7 +79,6 @@ import com.normation.rudder.services.workflows.WorkflowLevelService
 import com.normation.utils.Control
 import net.liftweb.common.Box
 import net.liftweb.common.EmptyBox
-import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.hooks.Cmd
 import com.normation.rudder.hooks.RunNuCommand
 import com.normation.errors.RudderError
@@ -111,7 +110,6 @@ class TechniqueWriter (
   , basePath            : String
   , parameterTypeService: ParameterTypeService
   , techniqueSerializer : TechniqueSerializer
-  , doRudderLangTest    : Boolean
 ) {
 
   private[this] val agentSpecific = new ClassicTechniqueWriter(basePath, parameterTypeService) :: new DSCTechniqueWriter(basePath, translater, parameterTypeService) :: Nil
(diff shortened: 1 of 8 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -330,9 +350,12 @@ class TechniqueWriter (
     }
   }
 
+  def techniquePath(technique : Technique) = s"techniques/${technique.category}/${technique.bundleName.value}/${technique.version.value}"
```
fanf, line 353: What happens with category if it's a sub category? Does is contains the full relative path?
Also, I think we are going to have problem by calling `Technique` both `com.normation.rudder.ncf.Technique` and `com.normation.cfclerk.domain.Technique`. Can we rename the first one `EditorTechnique` ? (or something letting us know that it's not the same).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -268,36 +266,58 @@ class TechniqueWriter (
...
-      updateResources  = technique.ressources.collect{case r if r.state != ResourceFileState.Deleted => r.copy(state = ResourceFileState.Untouched) }
-      techniqueWithResourceUpdated = technique.copy(ressources = updateResources)
+      updateResources = technique.ressources.collect{case r if r.state != ResourceFileState.Deleted => r.copy(state = ResourceFileState.Untouched) }
+                         techniqueWithResourceUpdated = technique.copy(ressources = updateResources)
       json       <- writeJson(techniqueWithResourceUpdated)
+      agentFiles <- callRudderc(technique).catchAll(e =>
+        for {
+          _ <- ApplicationLoggerPure.info(s"An error occurred when compiling technique '${technique.name}' (id : '${technique.bundleName}') with rudderc, error details will be displayed below, falling back to old saving process")
+          _ <- ApplicationLoggerPure.error(s"Rudderc error when compiling technique '${technique.name}' (id : '${technique.bundleName}'): ${e.fullMsg}")
+          - <-  writeAgentFiles (technique, methods, modId, committer)
```
fanf, line 315: Nice :)
Perhaps we should store these rudderc exception in dedicated, independendant log files, for ex in `/var/log/rudder/rudderc/failures/YYYY-MM-dd-hhmmss-techniqueName.log` with the context that lead to the error:
- json file
- if it exists, `cf` `dsc` file
- the error log

That would allow users to send us these reports (or even to get them automatically with the support script)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -268,36 +266,58 @@ class TechniqueWriter (
...
     }
   }
 
-  def runRudderLangTestLoop(technique : Technique): Unit = {
-    System.getenv().asScala.get("PATH") match {
-      case Some(path: String) => RunNuCommand.run(Cmd("/opt/rudder/share/rudder-lang/tools/tester.sh", technique.bundleName.value :: technique.category :: Nil, Map("PATH" -> path))).either.runNow match {
-        case Left(error: Error) => ApplicationLogger.error(error.getMessage)
-        case Left(error: RudderError) => ApplicationLogger.error(error.msg)
-        case Right(_)  => ApplicationLogger.info(s"rudder-lang tester successfully looped for technique ${technique.bundleName.value}")
+  def callRudderc(technique : Technique) = {
```
fanf, line 274: I believe that needs to be an independant service, with constructor argument:
- config path file
- rudderc binary path
- target base path

Else we won't be able to test properties on the outputs at all.

fanf: Regarding testing, we will need to embed a `rudderc` at some point - but I wonder where is the limit between testing `rudderc` and testing integration of `rudderc` into rudder with the specific things for it.

fanf: (we need at least to test the general output layout, path, etc)

fanf (commented):
Some change should be done to allow testing and better debugging. Perhaps they can be done in a second PR, but I fear we won't do them at all in that case. I let you decide what you think is the best (you can selfmerge in the corresponding case).

VinceMacBuche:
PR rebased

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
```diff
@@ -74,7 +74,7 @@ import com.normation.rudder.ncf.MethodCall
 import com.normation.rudder.ncf.ParameterId
 import com.normation.rudder.ncf.MethodParameter
 import com.normation.rudder.ncf.TechniqueParameter
-import com.normation.rudder.ncf.{Technique => NcfTechnique}
+import com.normation.rudder.ncf.{EditorTechnique => NcfTechnique}
```
fanf, line 77: you should use `EditorTechnique` everywhere here, not `NcfTechnique`

fanf (commented):
Just a s/NcfTechnique/EditorTechnique and you can merge it.

VinceMacBuche:
PR updated with a new commit

PR #3570, opened 2021-04-01
Fixes #19114: Overridden directives in the same rule are missing (not even "skipped")

fanf:
https://issues.rudder.io/issues/19114

So, the problem was that we were filtering the overriden directive globally, and not by rule, so we had only one overridden for all overrides. 

So now, we filter rule by rule (in a smarter way, and at a place where we can do test, even better) and don't filter with the big axe when building the list of directive lines.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
@@ -66,14 +66,16 @@ import zio._
 object ReportingServiceUtils {
 
   /*
-   * Build rule status reports from node reports, decide=ing which directive should be "skipped"
+   * Build rule status reports from node reports, deciding which directives should be "skipped"
    */
   def buildRuleStatusReport(ruleId: RuleId, nodeReports: Map[NodeId, NodeStatusReport]): RuleStatusReport = {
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -66,14 +66,16 @@ import zio._
...
-   * Build rule status reports from node reports, decide=ing which directive should be "skipped"
+   * Build rule status reports from node reports, deciding which directives should be "skipped"
    */
   def buildRuleStatusReport(ruleId: RuleId, nodeReports: Map[NodeId, NodeStatusReport]): RuleStatusReport = {
     val toKeep = nodeReports.values.flatMap( _.reports ).filter(_.ruleId == ruleId).toList
     // we don't keep overrides for a directive which is already in "toKeep" or that don't target that rule
     val toKeepDir = toKeep.map(_.directives.keySet).toSet.flatten
     val overrides = nodeReports.values.flatMap( _.overrides.filterNot(r => r.policy.ruleId != ruleId || toKeepDir.contains(r.policy.directiveId))).toList.distinct
-    RuleStatusReport(ruleId, toKeep, overrides)
+    // and we must make overrides unique - ie, we don't keep overridden that are overridden by directive themselve in the overridden list
```
fanf, line 76: yep, wasn't able to make it clearer :)

PR #3577, opened 2021-04-04
Fixes #18981: Slow computation of dynamic groups on large system

ncharles:
https://issues.rudder.io/issues/18981

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
@@ -49,7 +49,9 @@ import com.normation.utils.StringUuidGenerator
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.logger.ScheduledJobLogger
 import org.joda.time.format.ISODateTimeFormat
-
+import com.normation.box._
+import com.normation.errors._
+import com.normation.rudder.utils.ParseMaxParallelism
 //Message to send to the updater manager to start a new update of all dynamic groups or get results
 sealed trait GroupUpdateMessage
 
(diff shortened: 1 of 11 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/appconfig/ConfigService.scala
```diff
@@ -441,6 +445,7 @@ class LDAPBasedConfigService(
        rudder.generation.delay=0s
        rudder.generation.trigger=all
        node.accept.duplicated.hostname=false
+       rudder.compute.dyngroups.max.parallelism=1
```
fanf, line 448: I think that for consistancy sack, it should be like `rudder.generation.max.parallelism`, ie accepting either an integer like here, or `xF`, with `F` a float and `x` the character `x`, meaning `F time the number of CPU available`. The same parsing logic can be used (and can even be factored out in an utility object, which would also ensure to always return an integer at least equals to `1`).
And with that, setting it to `x0.5` (ie, `number CPU / 2`) seems to be a good default.

fanf (commented):
Appart for some little enhancement in comments, this seems to be a good PR with limited risk and clear gain. Nice !

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/appconfig/ConfigService.scala
ncharles: it's not really generation, as it is to compute dynamic groups (so also done outside of policy generation)

i'll change the xF

ncharles: I'm not sure i'd like to change the default to something larger than 1, as it depends on the size of the LDAP connection pool which is 2 by default

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
```diff
@@ -227,16 +229,17 @@ class UpdateDynamicGroups(
...
-              error.rootExceptionCause.foreach(ex =>
-                logger.error("Exception was:", ex)
-              )
-            case Full(diff) =>
+          eitherRes match {
+            case Left(e) =>
+              val error = (e.fullMsg + s"Error when updating dynamic group '${id.value}'")
+              logger.error(error)
+             // e.fullMsg.foreach(ex =>
+             //   logger.error("Exception was:", ex)
```
fanf, line 240: this comment should be removed

ncharles:
PR updated with a new commit

fanf (approved):

fanf:
OK, squash merging this PR

PR #3586, opened 2021-04-14
Fixes #19146: in some case inverting a request leads to an error and 0 results

fanf:
https://issues.rudder.io/issues/19146

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -349,21 +349,33 @@ class InternalLDAPQueryProcessor(
     // Now, groups resulting set of same DnType according to composition
     // Returns None if the resulting composition (using AND) leads to an empty set of Nodes, Some filtering otherwise
     def dnMapSets(normalizedQuery: LDAPNodeQuery, dnMapMapSets: Map[DnType, Map[String,Set[DN]]]) : Option[Map[DnType, Set[DN]]] = {
-      val mapSet = dnMapMapSets map { case (dnType, dnMapSet) =>
-        val dnSet:Set[DN] = normalizedQuery.composition match {
-          case Or  => dnMapSet.foldLeft(Set[DN]())( _ union _._2 )
+      dnMapMapSets.foldLeft(Map[DnType, Set[DN]]().some){
+        // We got a None we can skip further folds
+        case (None, _ ) => None
+        case (Some(map),(dnType, dnMapSet)) =>
+        // Compute composition of all dn lists as an option, so that we can have the empty case as a None (see And branch)
+        val dnSet: Option[Set[DN]] = normalizedQuery.composition match {
+          case Or  => dnMapSet.foldLeft(Set[DN]().some){
+            case (Some(a),b) =>  Some(a union b._2 )
+            case (None,b)    =>  Some(b._2)
+          }
           case And =>
-            val s = if(dnMapSet.isEmpty) Set[DN]() else dnMapSet.foldLeft(dnMapSet.head._2)( _ intersect _._2 )
-            // Here if s is empty, it means we are ANDing with an empty Set, so we could simply drop all computation from there
-            if(s.isEmpty && normalizedQuery.transform != ResultTransformation.Invert) {
-              logPure.logEffect.debug(s"[${debugId}] `-> early stop query (non-inverted empty sub-query)")
-              return None
+            if (dnMapSet.isEmpty) {
+              None
+            } else {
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -414,58 +424,51 @@ class InternalLDAPQueryProcessor(
...
+            fss     = filterSeqSet(dms)
+            blf     <- buildLastFilter(nq, fss)
+
+            // for convenience
+            (finalLdapFilter, finalSpecialFilters) = blf
+
+            //final query, add "match only server id" filter if needed
+            rt      = nodeObjectTypes.copy(filter = finalLdapFilter)
+            _       <- logPure.debug(s"[${debugId}] |- (final query) ${rt}")
+            entries <- for {
```
VinceMacBuche, line 447: We don't need the for here if i read correctly

fanf: right, it was simplified enought :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -414,58 +424,51 @@ class InternalLDAPQueryProcessor(
...
       }
+      inverted <- query.transform match {
+                  case ResultTransformation.Identity => results.succeed
+                  case ResultTransformation.Invert   =>
+                    for {
+                      _      <- logPure.debug(s"[${debugId}] |- (need to get all nodeIds for inversion) ")
+                      allIds <- executeQuery(nodeObjectTypes.baseDn, nodeObjectTypes.scope, nodeObjectTypes.objectFilter, Some(ALL), Set(), Set(), nq.composition, debugId)
+                      ids    =  results.flatMap(x => x(A_NODE_UUID)).toSet
+                      res    =  allIds.collect { case e if(!ids.contains(e.value_!(A_NODE_UUID))) => e }
+                      _      <- logPure.debug(s"[${debugId}] |- (invert) entries after inversion: ${res.size}")
```
VinceMacBuche, line 464: why a collect and not a filter or a filterNot ?

VinceMacBuche: ie allIds.filterNot(e => ids.contains(e.value_!(A_NODE_UUID)))

fanf: I didn't change that. I imagine it used to be comething else in the right side of the `=>`

VinceMacBuche: That was added in a previous PR when you added the invert logic but it did not botherer me first time, and here yes, but that's not important

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -350,20 +350,30 @@ class InternalLDAPQueryProcessor(
...
+            case (Some(a),b) =>  Some(a union b._2 )
+            case (None,b)    =>  Some(b._2)
+          }
           case And =>
-            val s = if(dnMapSet.isEmpty) Set[DN]() else dnMapSet.foldLeft(dnMapSet.head._2)( _ intersect _._2 )
-            // Here if s is empty, it means we are ANDing with an empty Set, so we could simply drop all computation from there
-            if(s.isEmpty && normalizedQuery.transform != ResultTransformation.Invert) {
-              logPure.logEffect.debug(s"[${debugId}] `-> early stop query (non-inverted empty sub-query)")
-              return None
+            dnMapSet.toList match {
```
ncharles, line 359: i'm concerned about the convertion from map to list, and then from map to list to map again ( mapSet.toList.sequence.map(_.toMap) ) 
cost of convertion can be prohibitive with thousands of nodes

i don't have a better proposal though :/

VinceMacBuche: I think i fixed it here in this PR: https://github.com/fanf/rudder/pull/13 

I never convert the original map, and only fold in an optionnal result map

fanf:
PR rebased

fanf:
PR rebased

VinceMacBuche (approved):

PR #3588, opened 2021-04-15
Fixes #19150: Add option to disable email notifications for change-request

ElaadF:
https://issues.rudder.io/issues/19150


ElaadF:
We want to keep the plugin independent as possible from Rudder main project, the option should only be present in the plugin

PR #3590, opened 2021-04-16
Fixes #19160: Query type constructor broke scale out relay plugin

VinceMacBuche:
https://issues.rudder.io/issues/19160


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -947,6 +947,14 @@ object ResultTransformation {
   }
 }
 
+object Query{
```
fanf, line 950: Please add a comment "necessary for 6.2 binary compat, need to be removed in 7.0 to avoid ambiguities in constructor"

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
OK, squash merging this PR

PR #3598, opened 2021-04-26
Fixes #19182: Change validation details page is not accessible anymore when a Node group query has been modified

VinceMacBuche:
https://issues.rudder.io/issues/19182

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -951,7 +951,7 @@ object ResultTransformation {
 // This trait and the apply method are necessary to allow backward compatibilty in 6.2.5 for plugin scale-out relay due to changes to apply method in https://issues.rudder.io/issues/19138
 // This can be removed in 7.0, going back to a case class Query with 4 fields
 // The issue was that we added a new field to a case class and this breaks binary compatibility
-sealed trait QueryTrait {
+sealed trait Query {
   def  returnType : QueryReturnType  //"node" or "node and policy servers"
   def  composition: CriterionComposition
   def  criteria   : List[CriterionLine] //list of all criteria to be matched by returned values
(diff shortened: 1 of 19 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -1005,26 +1005,31 @@ sealed trait QueryTrait {
     override def hashCode() = returnType.hashCode * 17 + composition.hashCode * 3 + transform.hashCode * 11 + criteria.toSet.hashCode * 7
 }
 
-final case class Query(
+final case class OldQuery(
```
fanf, line 1008: this renaming will be binary incompatible no? It will break backs the other plugin.

fanf (changes requested):
I think that PR is not what we want. 
We need to at least correct the `equals` method, it will lead to problem at some point. But for the other changes, I think we should try to see if we can do a release of change validation that is compatible with both version.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
VinceMacBuche: It won't because we have an apply ... I think it still could break if there was some pattern patching though due to different unapply method

fanf: But in the previous case, an apply in the object wasn't enought to keep binary compat. Why it is, here?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -990,7 +990,7 @@ sealed trait QueryTrait {
 
     override def equals(other:Any) : Boolean = {
       other match {
-        case a :  QueryTrait => //criteria order does not matter
+        case a :  Query => //criteria order does not matter
           this.returnType == a.returnType &&
           this.composition == a.composition &&
           this.transform == a.transform &&
```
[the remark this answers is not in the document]
VinceMacBuche: It's ok ! we already compare the transform parameter (which is the invert ;) )

VinceMacBuche:
I'm closing this pr

PR #3602, opened 2021-04-28
Fixes #19186: Webapp tests fail if repository path contains an '@'

fanf:
https://issues.rudder.io/issues/19186

This PR correct the title problem by adding `.toURI` before getPath. 
It also clean-up other log output in tests so that we can find problem more quickly in output.


fanf:
PR rebased

VinceMacBuche (approved):

PR #3623, opened 2021-05-05
Fixes #14753: Warning: Nashorn engine is planned to be removed from a future JDK release 

fanf:
https://issues.rudder.io/issues/14753
This PR replace the use of `nashorn` js engine, removed in JVM 12, by `graalvm`.
There is two commit depending on how far we want to go. 

The first commit is the minimal change possible: we use `graalvm` in `nashorn compatible mode`, which give us few changes, but few gain too: we can't benefits from the improve security of graalvm model, nor use code caching/warmup.

The second is the use of `graalvm` as it should: 
- we use it directly, not behind a `ScriptEngine`
- we use it's security layers, which allows to remove all the Java `Policy` brittle horror, 
- we can use one `Engine` / several `Context` (one by eval), which ensure no code sharing between eval, but cache & warmup sharing, 

We still need to keep our execution threadpool to manage the force-stop of too-long running script - GraalVM has a "sandbox" option, but only for enterprise edition. 

In that second case, the performance gain is substantial: `x20`, from `~10s` to `~450ms` for a 5000 steps loop which just write `processing $i`.


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestJsEngine.scala
```diff
@@ -161,39 +157,48 @@ class TestJsEngine extends Specification {
       } must beEqualTo(Right(true))
     }
 
-    "not be able to access FS with safeExec" in {
```
VinceMacBuche, line 164: Should be removed

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #3637, opened 2021-05-10
Fixes #19248: Empty /var/rudder/policy-generation-info/node-configuration-hashes.json after a policy generation that changed nothing

fanf:
https://issues.rudder.io/issues/19248

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
@@ -543,11 +543,17 @@ class FileBasedNodeConfigurationHashRepository(path: String) extends NodeConfigu
   }
 
   override def save(hashes: Set[NodeConfigurationHash]): Box[Set[NodeId]] = {
-    timeLog(s => s"Updating node configuration hashes took ${s}")(semaphore.withPermits(1) {
-      for {
-        _       <- nonAtomicWrite(NodeConfigurationHashes(hashes.toList.sortBy(_.id.value)))
-      } yield hashes.map(_.id)
-    }).toBox
+    val nodeIds = hashes.map(_.id)
+    timeLog(s => s"Updating node configuration hashes took ${s}")(
+      if(hashes.size == 0) nodeIds.succeed
+      else semaphore.withPermits(1) {
+        for {
+          current  <- nonAtomicRead()
+          filtered =  current.hashes.filterNot(h => nodeIds.contains(h.id))
+          updated  =  (filtered ++ hashes).sortBy(_.id.value)
+          _        <- nonAtomicWrite(NodeConfigurationHashes(updated))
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepositoryTest.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  val h3_0 = NodeConfigurationHash(NodeId("node3"), d1, 0, 0, 0, Set(
+    PolicyHash(PolicyId(RuleId("r3"), DirectiveId("d3"), TechniqueVersion("1.0")), 0)
+  ))
+
+  val s1 = Set(h1_0, h0_0, h2_0)
+
+  sequential
+
+  "when nothing was written, we should get an empty succe" >> {
```
ncharles, line 108:
```suggestion
  "when nothing was written, we should get an empty success" >> {
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
```diff
@@ -543,11 +543,17 @@ class FileBasedNodeConfigurationHashRepository(path: String) extends NodeConfigu
...
-        _       <- nonAtomicWrite(NodeConfigurationHashes(hashes.toList.sortBy(_.id.value)))
-      } yield hashes.map(_.id)
-    }).toBox
+    val nodeIds = hashes.map(_.id)
+    timeLog(s => s"Updating node configuration hashes took ${s}")(
+      if(hashes.size == 0) nodeIds.succeed
+      else semaphore.withPermits(1) {
+        for {
+          current  <- nonAtomicRead()
+          filtered =  current.hashes.filterNot(h => nodeIds.contains(h.id))
```
ncharles, line 552: i'm unsure tht you need this one: you are traversing current.hashes , and n times the nodeIds

doing simply (filtered ++ hashes).toList.sortBy(_.id.value) would yield the same result

ncharles (commented):
i *think* the filterNot is overcomplicating the method

fanf:
PR updated with a new commit

ncharles (approved):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
fanf: no, `filtered` is a list, not a map, and `hashes` is not a map either, so I will have several node config hashes for the same node Id.  
I would need to do `current.hashes.map(h => (h.id, h)).toMap`, same for `hashes`, so two traverse, and one more (more coslty) for the map merge.

fanf:
OK, squash merging this PR

PR #3650, opened 2021-05-27
Fixes #19323: Be able to group reporting and methods so that we have clearer techniques and a better reporting

VinceMacBuche:
https://issues.rudder.io/issues/19323


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -136,10 +139,6 @@ final case class SectionSpec(
   private def recCloneMultivalued: Either[LoadTechniqueError, SectionSpec] = {
     val multivaluedChildren = children.toList.traverse { child => child match {
       case s: SectionSpec =>
-        if (s.isMultivalued) LoadTechniqueError.Consistancy(
```
VinceMacBuche, line 139: I think this isnot relevant anymore since it could prevent usage of group reporting

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
fanf: I think you are right, cc @ncharles

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -437,3 +436,20 @@ object DisplayPriority {
     }
   }
 }
+
+sealed  trait CompositionRule
```
fanf, line 440: Ho, it's THAT rule, not the other one. 
I think we need to change that name, it is extremely misleading.
`GroupedComponentComplianceAlgo` ? Or if the algo can be used for other elements than for `GroupedComponents`, it can be:  `ComplianceComputingAlgo`, `ComplianceSumAlgo`, `ComplianceReduceAlgo`, ...

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -1421,35 +1424,54 @@ object RuleExpectedReportBuilder extends Loggable {
...
-        None
+        childExpectedReports
       }
-    }.toList
+    }
+
+
+    val allComponents = technique.rootSection.children.collect{case c : SectionSpec => c }.flatMap(sectionToExpectedReports).toList
 
+    logger.info("allComponent")
```
fanf, line 1465: remaining debug logger

fanf (commented):
WIP, fastly broken

webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/ViewBlock.elm
```diff
@@ -0,0 +1,417 @@
...
+                  --|> addAction ("click", GenerateId (\s -> CloneMethod block (CallId s)))
+                  |> addAttributeList
+                     [ type_ "button", title "Clone this method", attribute "data-toggle" "tooltip"
+                     , attribute "data-trigger" "hover", attribute "data-container" "body", attribute "data-placement" "left"
+                     , attribute "data-html" "true", attribute "data-delay" """'{"show":"400", "hide":"100"}'"""
+                     ]
+                  |> appendChild cloneIcon
+    removeIcon = element "i" |> addClass "fa fa-times-circle"
+    removeButton = element "button"
+                  |> addClass "text-danger method-action tooltip-bs"
```
fanf, line 266: missing `title`

fanf: The remove button does not work for components in groups

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -437,3 +436,20 @@ object DisplayPriority {
...
+case object WorstReport extends CompositionRule
+case object SumReport extends CompositionRule
+case class ComponentReport(component : String) extends CompositionRule
+
+object CompositionRule {
```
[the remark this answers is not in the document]
VinceMacBuche: Renamed into "ReportingLogic"

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -130,14 +134,24 @@ final case class DirectiveExpectedReports (
 /**
  * The Cardinality is per Component
  */
-final case class ComponentExpectedReport(
+trait ComponentExpectedReport {
+  def componentName : String
+}
+
+final case class GroupComponentExpectedReport (
```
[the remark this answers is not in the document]
VinceMacBuche: I Renamed all "GroupComponent" into "Block" and "UniqueComponent" into "Value" , which contains ComponentValues 

I think i want to get rid of componentValues in the future and go for a one entity with an id regrouping the Component and the Value

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -437,3 +436,20 @@ object DisplayPriority {
     }
   }
 }
+
+sealed  trait CompositionRule
+case object WorstReport extends CompositionRule
+case object SumReport extends CompositionRule
+case class ComponentReport(component : String) extends CompositionRule
```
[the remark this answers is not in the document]
VinceMacBuche: Component name, and use the compliance of all reports matching this component name

Like previous comment, now that we do have an id for all method call we should change the reporting to get the id now and not matching component name and value (I'm started a doc on the subject, It will be shared on Slack)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -307,9 +363,34 @@ final case class ComponentStatusReport(
...
+          val newValues = ComponentValueStatusReport.merge(r.flatMap( _.componentValues.values))
+          Some(UniqueComponentStatusReport(cptName, newValues))
+      }
+      val groupComponent= reports.collect{ case c : GroupComponentStatusReport => c }.toList match {
+        case Nil => None
+        case r =>
+          val compositionRule = r.map(_.compositionRule).reduce(
+            (a,b) => (a,b) match {
+                       case(WorstReport, _ ) => WorstReport
+                       case(SumReport, _ ) => SumReport
```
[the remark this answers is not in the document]
VinceMacBuche: Because i was dumb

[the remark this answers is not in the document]
VinceMacBuche: The rule is also arbitrary, I thought worst report to be with the most priority would be the best ... but in real this won't happen much ...

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
VinceMacBuche: oops

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/SectionSpecParser.scala
```diff
@@ -69,16 +68,6 @@ class SectionSpecParser(variableParser:VariableSpecParser) extends Loggable {
...
 
-          /*
-           * check that all variable and section names are unique
-           */
-          val allVarUniqueName = checkUniqueness(variableNames) {
-            "At least two variables have the same name (case unsensitive), what is forbiden: "
-          }
-          val allSectionUniqueName = checkUniqueness(root.getAllSections.map(_.name)) {
-            "At least two sections have the same name (case unsensitive), what is forbiden: "
-          }
```
fanf, line 80: these restriction can be lift because we now have an unique identifier for each component ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -278,11 +280,60 @@ object DirectiveStatusReport {
...
+    subComponents.find(_.componentName == componentName).toList :::
+    subComponents.collect{case g :BlockStatusReport => g}.flatMap(_.findChildren(componentName))
+  }
+  def compliance: ComplianceLevel = {
+    import ReportingLogic._
+    reportingLogic match {
+      case WorstReport =>
+        ComplianceLevel.compute(ReportType.getWorseType(subComponents.map(_.status)) :: Nil)
+      case SumReport => ComplianceLevel.sum(subComponents.map(_.compliance))
+      case FocusReport(component) => ComplianceLevel.sum(findChildren(component).map(_.compliance))
```
fanf, line 306: I think that if the component is not found, it should not be 0 or whatever `ComplianceLevel.sum(Nil)` does, but it should be a missing report error.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -278,11 +280,60 @@ object DirectiveStatusReport {
...
+      case SumReport =>
+        ReportType.getWorseType(subComponents.map(_.status))
+      case FocusReport(component) => ReportType.getWorseType(findChildren(component).map(_.status))
+    }
+  }
+}
+final case class ValueStatusReport  (
     componentName      : String
-    //only one ComponentValueStatusReport by value
+    //only one ComponentValueStatusReport by valuex.
```
fanf, line 334: typo

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -437,3 +436,20 @@ object DisplayPriority {
...
+  final case class FocusReport(component : String) extends ReportingLogic
+  def apply(value : String) : PureResult[ReportingLogic] = {
+    value match {
+      case "worst" => Right(WorstReport)
+      case "sum" => Right(SumReport)
+      case s"focus:${a}" => Right(FocusReport(a))
+      case _ => Left(Unexpected(s"Value '${value}' is not a valid reporting composition rule."))
+    }
+  }
+}
```
fanf, line 455: I would prefer that each `ReportingLogic` has a `name` attribute, and that name is used everywhere when the string is needed. If we want to ensure consistency. 
So something like: 
```
sealed  trait ReportingLogic { def name: Sring}
final case object WorstReport extends ReportingLogic { val name = "worst" }
final case class FocusReport(component : String) extends ReportingLogic { val name = "focus:${component}" }
etc
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -246,25 +257,39 @@ object ExpectedReportsSerialisation {
...
+      case c: ValueExpectedReport =>
+        (("componentName" -> c.componentName)
+          ~ ("values" -> c.componentsValues)
+          ~ ("unexpanded" -> c.unexpandedComponentsValues)
+          )
+      case c: BlockExpectedReport =>
+        import ReportingLogic._
+        (("componentName" -> c.componentName)
+          ~ ("composition" -> (c.reportingLogic match {
+          case WorstReport => "worst"
```
fanf, line 272: `c.reportingLogic.name`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
```diff
@@ -323,6 +333,30 @@ class TechniqueSerializer(parameterTypeService: ParameterTypeService) {
...
+        case b : MethodBlock => serializeMethodBlock(b)
+      }
+    }
+
+    def serializeCompositionRule(reportingLogic: ReportingLogic) :JValue = {
+      import ReportingLogic._
+      reportingLogic match {
+        case WorstReport =>  ("type"-> "worst")
+        case SumReport =>  ("type"-> "sum")
+        case FocusReport(component)  => ("type"  -> "focus") ~ ("value" -> component)
```
fanf, line 348: ok, so you need to split `name` and `serialize` in `ReportingLogic` (with by default `val serialize = name` and for focus report something different)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -213,10 +213,32 @@ class TechniqueWriter (
...
+        res <- calls.toList.traverse(block =>
+          for {
+            childs <- reportingSections(block.calls)
+          } yield {
+
+            import com.normation.cfclerk.domain.ReportingLogic._
+            val reportingLogic = block.reportingLogic match {
+              case WorstReport => "worst"
+              case SumReport => "sum"
+              case FocusReport(component) => s"focus:${component}"
```
fanf, line 228: `block.reportingLogic.serialize`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -939,25 +935,29 @@ final case class ContextForNoAnswer(
...
+                                         })
+                                     }
+
 
                                    val t5 = System.nanoTime
                                    u4 += t5-t4
 
 
-                                   (missing, unexpected, expected)
+                                   ( unexpected, expected)
```
fanf, line 954: I don't get why you don't have missing at all here: the case is still possible if some reports were lost (are more likely because of bad reporting logic somewhere)

fanf (commented):
Reviewed ~1/3 :) Some tiyn things to update for now, overall very good

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
VinceMacBuche: They are already treated  in another branch, i got duplicated missing if i keep it

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
VinceMacBuche: yes the difference is here: https://github.com/Normation/rudder/pull/3650/files/b5ee112bc450916897f1e3cbd6a032ff6feb5b21#diff-32bcd76fdbbd93efed1323a8e48c79320607d7ef97094e819936a74e6fab88feR938 

Previously we treated missing reports in its own place, so we were parsing all reports to check missing components so we produced our missing reports ,then when doing expected we were filtering missing key ( by using okKeys) 

in expected we now look for all expected keys, and when treating a component if we don't have reports we produce a missing

So we were going through all expected reports twice so find all missing and now this is done in one go

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionRepository.scala
```diff
@@ -120,6 +120,9 @@ class CachedReportsExecutionRepository(
 
   override def getNodesLastRun(nodeIds: Set[NodeId]): Box[Map[NodeId, Option[AgentRunWithNodeConfig]]] = semaphore.withPermit(IOResult.effect {
     val n1 = System.currentTimeMillis
+
+    TimingDebugLogger.info(s"cache last run ; ${cache}")
+    TimingDebugLogger.info(s"node id diff last run ; ${nodeIds.diff(cache.keySet)}")
```
fanf, line 125: i think it's debug info, not timing here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -418,35 +469,41 @@ class ClassicTechniqueWriter(basePath : String, parameterTypeService: ParameterT
...
-
-        s"""    "${promiser}" usebundle => ${reportingContext(method, classParameterValue)},
-           |     ${promiser.map(_ => ' ')}         if => concat("${condition}");
-           |    "${promiser}" usebundle => ${method.methodId.value}(${args}),
-           |     ${promiser.map(_ => ' ')}         if => concat("${condition}");""".stripMargin('|')
-
-      }).mkString("\n")
+          }  yield {
+            val condition = canonifyCondition(call, parentBlocks)
+            val promiser = call.id
```
fanf, line 491: I don't understand where the ID is generated.

webapp/sources/rudder/rudder-web/src/main/elm/build-app.sh
```diff
@@ -6,6 +6,6 @@ ELM_DIR="$( cd "$( dirname "$0" )" && pwd )"
 PROJECTS=("notifications" "healthcheck" "editor" "onboarding")
 for PROJECT in ${PROJECTS[*]}; do
   cd ${ELM_DIR}/${PROJECT}
-  elm make sources/rudder-${PROJECT}.elm --output=generated/rudder-${PROJECT}.js --optimize
+  elm make sources/rudder-${PROJECT}.elm --output=generated/rudder-${PROJECT}.js
```
fanf, line 9: why removing the `--optimize`?

webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/MethodElemUtils.elm
```diff
@@ -0,0 +1,64 @@
...
+    Block _ b -> List.concatMap getAllCalls b.calls
+    _ -> []
+
+setIdRec : String -> List MethodElem -> (List MethodElem, Bool)
+setIdRec  newId elems =
+  case elems of
+    [] -> ([], False)
+    head :: tail ->
+      if ((getId head ) == CallId "") then
+        ((setId (CallId newId) head ) :: tail, True)
```
fanf, line 19: ok, it's here that we generate block id!

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
fanf: ok, it's in `webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/MethodElemUtils.elm`

webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/MethodElemUtils.elm
fanf: But I still don't understand what happens for existing techniques, or imported one without an id

webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/MethodConditions.elm
```diff
@@ -212,3 +213,153 @@ parseOs os =
                           Nothing -> Nothing
                           x       -> Just (AIX { version = x })
     _ -> Nothing
+
+
+noVersion = {major = Nothing, minor = Nothing }
+
+osList : List (Maybe OS)
```
fanf, line 220: I think that that list of OS etc should be provided from the server (even if it's `json` in a var in the page). 
It feels like we want to be able to update that list ni parallele with the list of supported OS from inventory (which should also be taken from a config file, but that's an other story). Else we risk to forket to update that part when updating supported os on the other side.

fanf (commented):
scala code reviewed, it seems ok.

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -417,9 +443,22 @@ object ExpectedReportsSerialisation {
         // , (json \ "cardinality") // #10625: ignore cardinality
         , (json \ "values").extractOpt[List[String]]
         , (json \ "unexpanded").extractOpt[List[String]]
+        , (json \ "subComponents") match {
```
VinceMacBuche, line 446: Extractor and serializer don't have se same names ... I need to take care of this

webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/MethodElemUtils.elm
VinceMacBuche: existing techniques are migrated by Rudder, When we parse technique json on migration, we support missing id field by generating and id in case this case https://github.com/Normation/rudder/blob/e821a7e044d5aad0afbbbe1fbfcae8941ae5c4b4/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala#L1228

I need to look for imported techniques but either it fails if id are missing or it generates one UI side (my guess is an error)

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

language/src/parser.rs
```diff
@@ -1040,6 +1040,23 @@ fn pstate_declaration(i: PInput) -> PResult<PStateDeclaration> {
...
 }
 
+/// A State Declaration is a given required state on a given resource
+#[derive(Debug, PartialEq)]
+pub struct PBlockDeclaration<'src> {
+    pub metadata: Vec<PMetadata<'src>>,
+    pub childs: Vec<PStatement<'src>>,
+}
+fn pblock_declaration(i: PInput) -> PResult<PBlockDeclaration> {
+    wsequence!(
```
peckpeck, line 1050: there is no need for wsequence in single element match

language/src/parser.rs
```diff
@@ -1040,6 +1040,23 @@ fn pstate_declaration(i: PInput) -> PResult<PStateDeclaration> {
...
+        {
+            childs: delimited_parser("{", |j| many0(pstatement)(j), "}");
```
peckpeck, line 1052: `|j| many0(pstatement)(j)` is equivalent to : `many0(pstatement)`

language/src/ir/resource.rs
```diff
@@ -364,6 +364,13 @@ pub struct StateDeclaration<'src> {
     pub outcome: Option<Token<'src>>,
 }
 
+/// A Block Declaration is a given required state on a given resource
+#[derive(Debug, PartialEq)]
+pub struct BlockDeclaration<'src> {
+    pub metadata: TomlMap<String, TomlValue>,
+    pub childs: Vec<Statement<'src>>,
```
peckpeck, line 371: children

VinceMacBuche:
PR updated with a new commit

language/src/parser.rs
VinceMacBuche: that does not compile with the following error

 ```the trait bound `impl std::ops::Fn<(nom_locate::LocatedSpan<&str, &str>,)>: std::marker::Copy` is not satisfied

the trait `std::marker::Copy` is not implemented for `impl std::ops::Fn<(nom_locate::LocatedSpan<&str, &str>,)>`rustc(E0277)
baseparsers.rs(196, 8): required by a bound in this
baseparsers.rs(202, 8): required by this bound in `parser::baseparsers::delimited_parser`
parser.rs(1052, 33): the trait `std::marker::Copy` is not implemented for `impl std::ops::Fn<(nom_locate::LocatedSpan<&str, &str>,)>```

I'll keep it like that atm

VinceMacBuche:
PR updated with a new commit

language/src/parser.rs
peckpeck: OK :shrug:

PR #3665, opened 2021-06-04
Fixes #19394: When a node is totally deleted, cache in NodeInfoService can't know about it

ncharles:
https://issues.rudder.io/issues/19394

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
@@ -194,6 +194,7 @@ class WoLDAPNodeRepository(
     })
   }
 
+  // this method is unused
   override def deleteNode(node: Node, modId: ModificationId, actor: EventActor, reason: Option[String]): IOResult[Node] = {
     val entry = mapper.nodeToEntry(node)
     nodeLibMutex.writeLock(for {
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
@@ -272,13 +272,21 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
   override def getNumberOfManagedNodes: Int = nodeCache.map(_.managedNodes).getOrElse(0)
 
   /*
-     * Our cache
-     */
+   * Our cache
+   */
   private[this] var nodeCache = Option.empty[LocalNodeInfoCache]
 
   // we need modifyTimestamp to search for update and entryCSN to remove already processed entries
   private[this] val searchAttributes = nodeInfoAttributes :+ A_MOD_TIMESTAMP :+ "entryCSN"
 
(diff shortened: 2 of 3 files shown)


ncharles:
need extensive testing

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
```diff
@@ -199,6 +201,7 @@ class WoLDAPNodeRepository(
     nodeLibMutex.writeLock(for {
       con <- ldap
       _   <- con.delete(entry.dn).chainError(s"Error when trying to delete node '${node.id.value}'")
+      _   <- nodeInfoService.removeNodeFromCache(node.id)
```
ncharles, line 204: ok, this change doesn't do anything: this method is unused

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/RemoveNodeService.scala
```diff
@@ -513,6 +514,18 @@ class RemoveNodeServiceImpl(
 }
 
 
+class RemoveFromCache(nodeInfoService: NodeInfoService) extends PostNodeDeleteAction {
+  override def run(nodeId: NodeId, mode: DeleteMode, info: Option[NodeInfo], status: Set[InventoryStatus]): UIO[Unit] = {
+    if (mode == DeleteMode.Erase) {
+      NodeLoggerPure.Delete.debug(s"  - delete node from NodeInfoService Cache'${nodeId.value}'") *>
```
fanf, line 520: I think we should do that in all cases and remove the logic that looks in the `removed` branch in the cache `isUpToDate` method

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -159,6 +159,10 @@ trait NodeInfoService {
    */
   def getAllSystemNodeIds() : Box[Seq[NodeId]]
 
+  /**
+   * Remove a node from cache. It cannot fail - if node is not in cache, it is a success
+   */
+  def removeNodeFromCache(nodeId: NodeId): IOResult[Unit]
```
fanf, line 165: it should only be in `NodeInfoServiceCached` version, or in a separated trait. It has nothing to do with the main service interface

ncharles: indeed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/RemoveNodeService.scala
ncharles: I think that removing the check in removed branch would fail, as NodeInfoService also looks in this branch for getDeletedNodeInfos (even if not using the cache for that); at least not in this PR

ncharles:
PR updated with a new commit

fanf (approved):

ncharles:
OK, squash merging this PR

ncharles:
OK, merging this PR

PR #3675, opened 2021-06-08
Fixes #19414: Inventory with updated properties doesn't always start a policy generation

fanf:
https://issues.rudder.io/issues/19414

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
@@ -46,10 +46,14 @@ import com.normation.rudder.domain.logger.ScheduledJobLoggerPure
 import zio._
 import zio.duration._
 import com.normation.zio._
+import org.joda.time.DateTime
 
 /**
- * A naive scheduler which checks every N seconds if inventories are updated.
+ * A scheduler which checks every N seconds if inventories are updated.
  * If so, I will trigger a promise generation.
+ *
+ * Note that without this scheduler, if nobody needs the nodeInfoCache, it will never
+ * be updated.
  */
 class CheckInventoryUpdate(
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
```diff
@@ -58,6 +62,11 @@ class CheckInventoryUpdate(
   , updateInterval      : Duration
 ) {
 
+  // we need to store the time of last modification we saw so that is someone else update the cache
+  // (for example, a user is using the UI), then even if the cache is upToDate, we now we have to start
+  // a policy generation.
+  val lastUpdate = Ref.make(new DateTime()).runNow
```
ncharles, line 68: it may miss the first update with that if the NodeInfoCache is init before the CheckInventoryUpdate. We should use new DateTime(0)

ncharles (changes requested):
i think the init condition is not valid

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CheckInventoryUpdate.scala
fanf: I was not sure. Using epoch means that we will *always* trigger a policy generation when the scheduler first run. But we already have one around start time, which mean we will always have two. 
Not sure what the best here.

ncharles: if we don't do that, and there was a change before the scheduler kicked in, it will not apply until a new change is made. I'd rather play safe here

fanf:
PR updated with a new commit

ncharles (approved):

fanf:
OK, squash merging this PR

PR #3690, opened 2021-06-22
Fixes #19470: Relayd does not start since some keys are missing from its logging.conf file

Fdall:
https://issues.rudder.io/issues/19470


amousset:
It should start with default values, it is a bug in relayd

PR #3713, opened 2021-07-09
Fixes #19492: Add policy server certificate information to policies

fanf:
https://issues.rudder.io/issues/19492

Main changes: 
- output for the root server hash in nodelist is now base64 encoded and starts with `sha256//` in place of being hex encoded and start by `sha256:`
- renamed `CfengineKey` utility class to `NodeKeyHash`, since it deals with all kind of hashes
- refactor a bit key crypto processing to factor out common parts, even if bouncy castle and java api try hard to make that difficult, 
- add `POLICY_SERVER_KEY_HASH` which holds the base64 encoded string. This one also has the prefix `sha256//` - is it what we want ?

Also, in node details, we are displaying certificate "fingerprint", which is yet something different: `SHA1 Fingerprint: </label> <samp>{SHA1.hash(cert.getEncoded).grouped(2).mkString(":")}`. Do we want to also display the base64 string, since it's the one that will be usefull for debugging ? Also, shouldn't we have in `rudder agent` a command (in `check` ?) that display the base64 cert string with the correct openssl incantation ?

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
@@ -174,6 +174,11 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                            , multivalued = false
                                            , constraint = Constraint(mayBeEmpty=true)
                         )
+    , SystemVariableSpec(
+                       "POLICY_SERVER_KEY_HASH" , "Base64 encoding of SHA-256 DER encoding of KEY of the policy server prefixed with sha256//"
+                                           , multivalued = false
+                                           , constraint = Constraint(mayBeEmpty=true)
+                        )
       // end
       //
       // The following variables contains information about all the nodes managed by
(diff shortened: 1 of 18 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
```diff
@@ -174,6 +174,11 @@ class SystemVariableSpecServiceImpl extends SystemVariableSpecService {
                                            , multivalued = false
                                            , constraint = Constraint(mayBeEmpty=true)
                         )
+    , SystemVariableSpec(
+                       "POLICY_SERVER_KEY_HASH" , "Base64 encoding of SHA-256 DER encoding of KEY of the policy server"
```
amousset, line 178: it should have the sha256 prefix

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/SystemVariableSpecServiceImpl.scala
fanf: it has (last fixup). I can precise it here too

fanf:
PR updated with a new commit

ncharles (approved):

fanf:
OK, squash merging this PR

fanf:
OK, squash merging this PR

fanf:
OK, merging this PR

PR #3715, opened 2021-07-12
Fixes #19529: Add root.pem and policy-server.pem in node inputs 

fanf:
https://issues.rudder.io/issues/19529

Add `certs/root.pem` and `certs/policy-server.pem` in agent generated policies. 
The main thing that is a bit problematic is that `policy-server.pem` is a symlink to `root.pem` when it's the same content, and that our test lib for generated share use `FileUtils.copy` for directory copy, which does not know anything about symlink. And specs2 does error if we have a file and a symlink. So we need to ignore that file, which is a bit sad.

Most of the changes are in test config file, since we now have a certificate for root security token, it has repercussion all around.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -150,6 +152,31 @@ class PrepareTemplateVariablesImpl(
...
     )
+
+    /*
+     * Policy server certificates for the node
+     */
+    val policyServerCertificates = {
+      def getCertifiate(nodeId: NodeId): IOResult[String] = {
+        allNodeConfigs.get(nodeId) match {
+          case None    =>
+            "".succeed.tap(_ => PolicyGenerationLoggerPure.error(s"Policy server '${nodeId.value}' was not found during generation: can not provide its certificate"))
```
ncharles, line 163: is it a succeed?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -150,6 +152,31 @@ class PrepareTemplateVariablesImpl(
       , agentNodeConfig.config.nodeInfo.isPolicyServer
       , agentNodeConfig.config.nodeInfo.serverRoles
     )
+
+    /*
+     * Policy server certificates for the node
+     */
+    val policyServerCertificates = {
+      def getCertifiate(nodeId: NodeId): IOResult[String] = {
```
ncharles, line 160:
```suggestion
      def getCertificate(nodeId: NodeId): IOResult[String] = {
```

ncharles (changes requested):
i wonder why policyServerCertificates is always a succeeed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
fanf: it uses to be like that because we chose to not fail policy generation for a missing/bad key or certificate. 
I think we may be able to revisit that assumption, perhaps for 7.1, if we have node by node generation by default, and after cleaning code thanks to the fact that we only have certificate for everything in 7.0.

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
fanf: I added a comment

ncharles (approved):

PR #3717, opened 2021-07-15
Fixes #18730: Move allowed network data to settings 

fanf:
https://issues.rudder.io/issues/18730

This PR moved `allowed networks` from the system directive value to a real data storage backend. A setting was chosen for that (see rationnal in ticket). 

The new data structure is more than just storing allowed networks for each policy server, it allows to be extended in the futur with more information (like specific syncho for nodes below it, or whatever).

To be able to test that and keep atomicity of change, we added a little diff command dsl that can be interpreted on policy servers structure and update it according to commands. That part is bit long but quite systematic. 

The json serialisation is done with `zio.json`, which make me realize that all our nice convention works well for things in `rudder-rest`, but not for thing in `rudder-core` (starting with: we don't have `rudder-rest` when in `rudder-core`). As it's the first case of it, the convention are a bit ad-hoc, we should change them in the futur. 

I also renamed all "authorized networks" into "allowed networks" (appart in event log) to be consistent with UI and API result.
I did it now, since it breaks API (and so, a 7.0 is a good time for that).

We reuse existing system variable `AUTHORIZED_NETWORKS` but renamed it `ALLOWED_NETWORKS` for consistency.

For some reason, the bit of code:

```
    policy_server::
      "acl" slist => {
        "127.0.0.0/8" , "::1", @{sys.ip_addresses},
<<<<< that part <<<<
        host2ip("${server_info.policy_server}"), # the policy server can connect to a relay
        "192.168.12.0/24","192.168.49.0/24","127.0.0.1/24",
>>>>>> here >>>>>
    };
```
Was generated of all nodes, even the one not policy server. It may be an artifact of the mock service for "getAllowedNetwork" which was alway returning the same list of ip, whatever the node in parameter. 
Now, we only have some policy server defined, and so it is empty on regular node. 

Finaly, it seems that we have a very old bug with `rudder_syslog_protocol` when we need to default to the HOCON default config, because in it, we can't have both `"rudder_syslog_protocol` and `"rudder_syslog_protocol_default` (the first would be node, not a leaf in its json representation). So it fails. 
I needed to change that for the tests, and so added `_transport` to the first key (which of course would break migration from a set up which used it). Since we are removing syslog in 7.0, it is certainly not very important. Still, a very old silent bug...


fanf:
PR rebased

fanf:
PR rebased

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/test/resources/configuration-repository/expected-share/node-cfe-with-500-directives/rules/cfengine-community/rudder.json
```diff
@@ -3,11 +3,7 @@
   "AGENT_RUN_SCHEDULE":"\"Min01\", \"Min06\", \"Min11\", \"Min16\", \"Min21\", \"Min26\", \"Min31\", \"Min36\", \"Min41\", \"Min46\", \"Min51\", \"Min56\"",
   "AGENT_RUN_SPLAYTIME":"4",
   "AGENT_TYPE":"cfengine-community",
-  "AUTHORIZED_NETWORKS":[
-    "192.168.12.0/24",
-    "192.168.49.0/24",
-    "127.0.0.1/24"
-  ],
+  "AUTHORIZED_NETWORKS":[],
```
ElaadF, line 6: Do we keep `AUTHORIZED_NETWORKS` instead of `ALLOWED_NETWOKS` ?

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -545,18 +550,22 @@ class TestNodeConfiguration(prefixTestResources: String = "") {
...
-    override def deleteRelaySystemObjectsPure(policyServerId: NodeId): IOResult[Unit] = ???
-    override def updateAuthorizedNetworks(policyServerId: NodeId, addNetworks: Seq[String], deleteNetwork: Seq[String], modId: ModificationId, actor: EventActor): Box[Seq[String]] = ???
+  val policyServerManagement = new PolicyServerManagementService() {
+    override def getPolicyServers(): IOResult[PolicyServers] = {
+      PolicyServers(
+        PolicyServer(Constants.ROOT_POLICY_SERVER_ID, List("192.168.12.0/24", "192.168.49.0/24", "127.0.0.1/24").map(s => AllowedNetwork(s, s"name for " + s)))
+        , Nil
+      ).succeed
+    }
+    override def updatePolicyServers(commands: List[PolicyServersUpdateCommand], modId: ModificationId, actor: EventActor): IOResult[PolicyServers] = ???
```
ElaadF, line 560: I don't understand the purpose of `???`

webapp/sources/rudder/rudder-core/src/test/resources/configuration-repository/expected-share/node-cfe-with-500-directives/rules/cfengine-community/rudder.json
fanf: We decided to change

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
fanf: it's a placeholder that compiles but throws a "non implemented exception" if that code path is reached. It is usefull when you want to just implement some parts of a service for a test, but some methods are not needed in that test. 

It should never appears elsewhere than in test, and even in tests, it may mean that there is not a good separation of concerns between services, and that some services are too big and should be split over (but then, you are in usual balance research between atomicity and ease of use). So keeps eyes open when you see that, it may be the next refactoring.

fanf:
PR rebased

VinceMacBuche (approved):

PR #3721, opened 2021-07-19
Fixes #19538: Or selector for property Name=Value selector returns 0 elements

fanf:
https://issues.rudder.io/issues/19538

Correct two bugs:
- the filter was incorrect when there wasn't any filters for`Or` case. openldap just returns nothing in that case, which was by luck what we want, but any composition with that filter lead to the same empty result (bad for or)
- in the case where LDAP queries didn't return any nodes, if we are in `Or` case, we STILL need to do the search on all nodes for the properties case

Plus add debugging info to help spot other such errors and move the query logic for properties in the same file than other to make it more easier to look at.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -172,13 +172,15 @@ sealed trait CriterionType extends ComparatorList {
 // a case class that allows to precompute some parts of the NodeInfo matcher which are indep from the
 // the node.
 trait NodeInfoMatcher {
+  def debugString: String
   def matches(node: NodeInfo): Boolean
 }
 
 object NodeInfoMatcher {
   // default builder: it will evaluated each time, sufficiant if all parts of the matcher uses NodeInfo
-  def apply(f: NodeInfo => Boolean): NodeInfoMatcher = {
+  def apply(s: String, f: NodeInfo => Boolean): NodeInfoMatcher = {
     new NodeInfoMatcher {
+      override val debugString: String = s
       override def matches(node: NodeInfo): Boolean = f(node)
     }
   }
@@ -206,8 +208,8 @@ final case object NodeStateComparator extends NodeCriterionType {
 
   override def matches(comparator: CriterionComparator, value: String): NodeInfoMatcher = {
     comparator match {
-      case Equals => NodeInfoMatcher( (node: NodeInfo) => node.state.name == value )
-      case _      => NodeInfoMatcher( (node: NodeInfo) => node.state.name != value )
+      case Equals => NodeInfoMatcher(s"Prop equals '${value}'", (node: NodeInfo) => node.state.name == value )
+      case _      => NodeInfoMatcher(s"Prop not equals '${value}'", (node: NodeInfo) => node.state.name != value )
     }
   }
 
@@ -287,6 +289,7 @@ final case class NodePropertyComparator(ldapAttr: String) extends NodeCriterionT
                               } catch { //malformed patterned should not be saved, but never let an exception be silent
                                 case ex: PatternSyntaxException => false
                               }
+                            override val debugString = s"Prop matches '${value}'"
                             override def matches(node: NodeInfo): Boolean = node.properties.exists(predicat)
                           }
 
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -172,13 +172,15 @@ sealed trait CriterionType extends ComparatorList {
 // a case class that allows to precompute some parts of the NodeInfo matcher which are indep from the
 // the node.
 trait NodeInfoMatcher {
+  def debugString: String
```
fanf, line 175: This one is added to be able to debug the part with node properties filtering: 
```
12:49:26.662 [main]            DEBUG com.normation.rudder.services.queries.AcceptedNodesLDAPQueryProcessor - LDAP result: 0 entries obtained in 291ms for query { returnType:'NodeReturnType' with 'Or' criteria [serializedNodeProperty.name.value eq foo=bar] }
12:49:27.008 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node6: NOK for [Prop name=value equals'foo=bar']
12:49:27.008 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node7: NOK for [Prop name=value equals'foo=bar']
12:49:27.009 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node3: NOK for [Prop name=value equals'foo=bar']
12:49:27.009 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node0: NOK for [Prop name=value equals'foo=bar']
12:49:27.009 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node4: NOK for [Prop name=value equals'foo=bar']
12:49:27.009 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node2: NOK for [Prop name=value equals'foo=bar']
12:49:27.009 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node5: NOK for [Prop name=value equals'foo=bar']
12:49:27.010 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - node1: OK for [Prop name=value equals'foo=bar']
12:49:27.010 [zio-default-blocking-1-2117173674] TRACE com.normation.rudder.services.queries - root: NOK for [Prop name=value equals'foo=bar']
12:49:27.027 [main]            DEBUG com.normation.rudder.services.queries.AcceptedNodesLDAPQueryProcessor - [post-filter:rudderNode] Found 1 nodes when filtering for info service existence and properties (366 ms)````

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala [hunk 2]
```diff
@@ -871,33 +871,13 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
...
-      } else {
-        val validPredicates = predicates.reduceLeft(combine).matches(nodeInfo)
-        comp(contains,validPredicates)
-      }
-    }
-
-    // if nodeIds is empty, return an empty set
-    if (!nodeIds.isEmpty) {
+    // if nodeIds is empty and composition is and, return an empty set; with or, we need to run it in all cases
+    if (nodeIds.isEmpty && composition == And) {
```
fanf, line 875: the second problem was around that check

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -386,7 +433,8 @@ class InternalLDAPQueryProcessor(
       }.map { case (ldapFilters, specialFilters) =>
         val finalLdapFilter = normalizedQuery.composition match {
           case Or  =>
-            Some(OR( (filterSeqSet.foldLeft(ldapFilters)( _ union _ )).toSeq:_*))
```
fanf, line 389: that used to create a syntaxically false filter that trigger an ldap error (server side) and make open ldap returns nothings

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/queries/TestQueryProcessor.scala
```diff
@@ -800,9 +802,13 @@ class TestQueryProcessor extends Loggable {
       """).openOrThrowException("For tests"),
       s(1) :: Nil)
 
-    testQueries(q1 :: q2 :: q3 :: q4 :: q5 ::  q6 :: q7 :: q8 :: Nil, true)
+    val andQueries = q1 :: q2 :: q3 :: q4 :: q5 ::  q6 :: q7 :: q8 :: Nil
+    // And and Or must yield same results when there is only one criteria for node prop, see: #19538
+    val orQueries = andQueries.map(_.modify(_.query.composition).setTo(Or))
+    testQueries(andQueries ::: orQueries, false)
```
fanf, line 808: add the same check for and and or queries on properties (there is only one criteria, it leads to the same results)

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -386,7 +433,8 @@ class InternalLDAPQueryProcessor(
...
+            if( filterSeqSet.size+ldapFilters.size <= 0) Some(NOT(HAS(A_OC))) // means that we don't have any ldap filter, only extended one, so select no nodes, see #19538
```
ncharles, line 436: can it be negative? shouldn't it be
if( filterSeqSet.isEmpty && ldapFilters.isEmpty) Some(NOT(HAS(A_OC)))

fanf: it could have been the `isEmpty` version. The perf of that, at that point, does matter (unlikelly it's an hot spot for anything)

PR #3724, opened 2021-07-24
Fixes #19558: Add a way to disable/enable reporting of a method in technique editor

VinceMacBuche:
https://issues.rudder.io/issues/19558


fanf (approved):

fanf:
I wonder if it wouldn't be better with "enable_reporting", default to true, to avoid double negation, but not strong feeling

PR #3728, opened 2021-07-26
Fixes #19151: add caching for NodeExpectedReports

ncharles:
https://issues.rudder.io/issues/19151

This PR adds a cache for NodeExpectedReports. This cache works as a map of [NodeId, Option]. The cache is inited with all nodes, and None for each NodeExpectedReports. 
Then once we search for *current* expected reports, cache is explored, and fetched expected reports are added to the cache.
When a policy generation happens, it also update the current expected reports for modified nodes

The rational is that it is *really* expensive to fetch these from the database, and keeping them in the cache does improve perfs.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
@@ -41,7 +41,6 @@ import com.normation.inventory.domain._
 import org.joda.time.DateTime
 import com.normation.ldap.sdk.LDAPConnectionProvider
 import com.normation.rudder.domain.NodeDit
-import net.liftweb.common._
 import com.normation.rudder.domain.nodes.{Node, NodeInfo}
 import com.normation.rudder.domain.RudderLDAPConstants._
 import com.normation.inventory.ldap.core.LDAPConstants._
@@ -62,7 +61,6 @@ import com.normation.ldap.sdk.LDAPIOResult._
 import zio._
 import zio.syntax._
 import com.normation.errors._
-import com.normation.box._
 import com.normation.zio._
 import com.normation.ldap.sdk.syntax._
 import com.normation.rudder.domain.logger.NodeLoggerPure
@@ -117,17 +115,12 @@ trait NodeInfoService {
    * Retrieve minimal information needed for the node info, used (only) by the
    * LDAP QueryProcessor.
    */
-  def getLDAPNodeInfo(nodeIds: Set[NodeId], predicates: Seq[NodeInfoMatcher], composition: CriterionComposition) : Box[Set[LDAPNodeInfo]]
+  def getLDAPNodeInfo(nodeIds: Set[NodeId], predicates: Seq[NodeInfoMatcher], composition: CriterionComposition) : IOResult[Set[LDAPNodeInfo]]
 
   /**
    * Return a NodeInfo from a NodeId. First check the ou=Node, then fetch the other data
-   * @param nodeId
-   * @return
    */
-  def getNodeInfo(nodeId: NodeId) : Box[Option[NodeInfo]]
-
-  // Pure version of getNodeInfo
-  def getNodeInfoPure(nodeId: NodeId) : IOResult[Option[NodeInfo]]
+  def getNodeInfo(nodeId: NodeId) : IOResult[Option[NodeInfo]]
 
   /**
    * Return the number of managed (ie non policy server, no rudder role )nodes.
@@ -144,7 +137,13 @@ trait NodeInfoService {
    * missing (but the corresponding nodeInfos won't be present)
    * So it is possible that getAllIds.size > getAll.size
    */
-  def getAll() : Box[Map[NodeId, NodeInfo]]
+  def getAll() : IOResult[Map[NodeId, NodeInfo]]
+
+  /**
+   * Get all nodes id
+   */
+
+  def getAllNodesIds(): IOResult[Set[NodeId]]
 
   /**
    * Get all nodes.
@@ -153,29 +152,22 @@ trait NodeInfoService {
    * missing (but the corresponding nodeInfos won't be present)
    * So it is possible that getAllIds.size > getAll.size
    */
-  def getAllNodes() : Box[Map[NodeId, Node]]
+  def getAllNodes() : IOResult[Map[NodeId, Node]]
 
   /**
    * Get all systen node ids, for example
    * policy server node ids.
    * @return
    */
-  def getAllSystemNodeIds() : Box[Seq[NodeId]]
+  def getAllSystemNodeIds() : IOResult[Seq[NodeId]]
 
   /**
    * Getting something like a nodeinfo for pending / deleted nodes
    */
-  def getPendingNodeInfos(): Box[Map[NodeId, NodeInfo]]
-  def getPendingNodeInfo(nodeId: NodeId): Box[Option[NodeInfo]] = {
-    getPendingNodeInfoPure(nodeId).toBox
-  }
-  def getPendingNodeInfoPure(nodeId: NodeId): IOResult[Option[NodeInfo]]
-
-  def getDeletedNodeInfos(): Box[Map[NodeId, NodeInfo]]
-  def getDeletedNodeInfo(nodeId: NodeId): Box[Option[NodeInfo]] = {
-    getDeletedNodeInfoPure(nodeId).toBox
-  }
-  def getDeletedNodeInfoPure(nodeId: NodeId): IOResult[Option[NodeInfo]]
+  def getPendingNodeInfos(): IOResult[Map[NodeId, NodeInfo]]
+  def getPendingNodeInfo(nodeId: NodeId): IOResult[Option[NodeInfo]]
+  def getDeletedNodeInfos(): IOResult[Map[NodeId, NodeInfo]]
+  def getDeletedNodeInfo(nodeId: NodeId): IOResult[Option[NodeInfo]]
 }
 
 object NodeInfoService {
@@ -809,10 +801,6 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
     }
   }
 
-  override final def getPendingNodeInfos(): Box[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(PendingInventory).toBox
-
-  override final def getDeletedNodeInfos(): Box[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(RemovedInventory).toBox
-
   private[this] def getNotAcceptedNodeInfo(nodeId: NodeId, status: InventoryStatus): IOResult[Option[NodeInfo]] = {
     val dit = status match {
       case AcceptedInventory => inventoryDit
@@ -843,8 +831,10 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
     }
   }
 
-  override final def getPendingNodeInfoPure(nodeId: NodeId): IOResult[Option[NodeInfo]] = getNotAcceptedNodeInfo(nodeId, PendingInventory)
-  override final def getDeletedNodeInfoPure(nodeId: NodeId): IOResult[Option[NodeInfo]] = getNotAcceptedNodeInfo(nodeId, RemovedInventory)
+  override final def getPendingNodeInfos(): IOResult[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(PendingInventory)
+  override final def getDeletedNodeInfos(): IOResult[Map[NodeId, NodeInfo]] = getNotAcceptedNodeDataFromBackend(RemovedInventory)
+  override final def getPendingNodeInfo(nodeId: NodeId): IOResult[Option[NodeInfo]] = getNotAcceptedNodeInfo(nodeId, PendingInventory)
+  override final def getDeletedNodeInfo(nodeId: NodeId): IOResult[Option[NodeInfo]] = getNotAcceptedNodeInfo(nodeId, RemovedInventory)
 
   /**
    * Clear cache.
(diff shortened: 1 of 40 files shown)


ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,352 @@
...
+      mergedEntries
+    })
+    )
+  }
+  /**
+   * get the current expected reports
+   */
+  def getCurrentExpectedReports(nodeIds: Set[NodeId]): Box[Map[NodeId, Option[NodeExpectedReports]]] = {
+    // First, get all nodes from cache (even the none)
+    val dataFromCache = cache.filter{  case (nodeId, _) => nodeIds.contains(nodeId) }
```
ncharles, line 244: i'm unsure if this should be in a semaphore or not

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -96,6 +97,7 @@ object CacheComplianceQueueAction {
   final case class RemoveNodeInCache      (nodeId: NodeId                                          ) extends CacheComplianceQueueAction
   final case class InitializeCompliance   (nodeId: NodeId, nodeCompliance: Option[NodeStatusReport]) extends CacheComplianceQueueAction // do we need this?
   final case class UpdateCompliance       (nodeId: NodeId, nodeCompliance: NodeStatusReport        ) extends CacheComplianceQueueAction
+  final case class InitNodeConfiguration  (nodeId: NodeId, nodeConfiguration: NodeExpectedReports  ) // unsure about keeping this one
```
ncharles, line 100: it's unused, i' pretty sure i should not keep it

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -121,13 +124,15 @@ class ReportingServiceImpl(
...
   , val nodeInfoService                 : NodeInfoService
+  , val nodeConfigService               : NodeConfigurationService
   , val batchSize                       : Int
   , val complianceRepository            : ComplianceRepository
 ) extends ReportingService with RuleOrNodeReportingServiceImpl with CachedFindRuleNodeStatusReports {
   val confExpectedRepo = defaultFindRuleNodeStatusReports.confExpectedRepo
   val directivesRepo   = defaultFindRuleNodeStatusReports.directivesRepo
   val rulesRepo        = defaultFindRuleNodeStatusReports.rulesRepo
 
+  val nodeConfigrationService = nodeConfigService
```
ncharles, line 135: THIS is bothering me. i need to register the nodeConfigrationService because it in the CachedFindRuleNodeStatusReports

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,352 @@
...
+
+  /**
+   * Do something with the action we received
+   */
+  private[this] def performAction(action: CacheComplianceQueueAction): IOResult[Unit] = {
+    import CacheComplianceQueueAction._
+    // in a semaphore
+    semaphore.withPermit(
+       action match {
+                          case insert: InsertNodeInCache => IOResult.effectNonBlocking {  cache = cache + (insert.nodeId -> None) }
```
ncharles, line 179: i'm missing the call for that, when adding a node

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,352 @@
...
+ * if an entry exists but without anything, then it will query the database
+ * if an entry exists but wthout the right nodeconfigid, it will query the database (but not update the cache)
+ */
+trait NodeConfigurationService {
+  /**
+   * retrieve expected reports by config version
+   */
+  def findNodeExpectedReports(
+         nodeConfigIds: Set[NodeAndConfigId]
+      ): Box[Map[NodeAndConfigId, Option[NodeExpectedReports]]]
```
fanf, line 69: Can you return `IOResult` in place of Box` ? We try to get ride of them, and `IOResult` compose better

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -146,6 +146,12 @@ trait NodeInfoService {
    */
   def getAll() : Box[Map[NodeId, NodeInfo]]
 
+  /**
+   * Get all nodes id
+   */
+
+  def getAllNodeIds(): Box[Set[NodeId]]
```
fanf, line 153: prefer use of `IOResult`, `Box` is deprecated in rudder

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: clean it, it's cognitively more costly to have something and wonder why it's there than needing to add it if needed later ont

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -2331,11 +2338,17 @@ object RudderConfig extends Loggable {
   ////////////////////// Snippet plugins & extension register //////////////////////
   lazy val snippetExtensionRegister: SnippetExtensionRegister = new SnippetExtensionRegisterImpl()
 
+  private[this] lazy val cachedNodeConfigurationService = {
+    val cached = new CachedNodeConfigurationService(findExpectedRepo, nodeInfoServiceImpl)
+    cached.init().runNow
+    cached
+  }
+
   /*
```
fanf, line 2347: is it possible to completly hide findExpectedRepo from public use with a facade that delegate to either it or cache ? We would be sure to never call the bad service

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
fanf: perhaps you need to create a facade/new trait on top of `FindExpectedReportRepository` and `NodeConfigurationService` to only expose what is needed?

fanf (commented):
I think the general logic is ok. Please, limit the use of Box to compat (and if possible, remove some more of them). I'm don't understand fully enought all the usage to be sure it's sound. If you can, you should hide from public use (ie, in `RudderConfig`) the non-cached version of the feature, to avoid people (in plugins etc, for ex) to not take benefits of them

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
ncharles: it is a private[this] - i'm unsure how to do better than that (unless putting a big comment around it to not use it)

ncharles: or it is used twice, once per cache, so I could instanciate one per cache, overhead should be minimal

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
ncharles: Good idea !

ncharles: actually, it's never used, so I removed it

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,350 @@
...
+  private[this] def performAction(action: CacheComplianceQueueAction): IOResult[Unit] = {
+    import CacheComplianceQueueAction._
+    // in a semaphore
+    semaphore.withPermit(
+       action match {
+                          case insert: InsertNodeInCache => IOResult.effectNonBlocking {  cache = cache + (insert.nodeId -> None) }
+                          case delete: RemoveNodeInCache => IOResult.effectNonBlocking {  cache = cache.removed(delete.nodeId) }
+                          case update: UpdateNodeConfiguration => IOResult.effectNonBlocking {  cache = cache + (update.nodeId -> Some(update.nodeConfiguration)) }
+                          case something =>
+                            Inconsistency(s"NodeConfiguration service cache received unknown command : ${something}").fail
```
fanf, line 181: I'm going to try to make a subtype for action to be sure that that property holds

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,350 @@
...
+  // we have the begin date of the nodeexpectedreport that my offer a way to ensure that we don't replace
+  // a value with something older
+
+  // Merge the data with the cache
+  // Strategy:
+  // * if nodeId is not in cache, remove the nodeId
+  // * if the nodeId is in cache, and value is None, augment cache
+  // * if the nodeid is in cache, and value is not None, pick the most recent value to augment cache
+  // Returns the values merged with cache
+  private[this] def mergeDataFromDBWithCache(fromDb: Map[NodeId, Option[NodeExpectedReports]]):IOResult[ Map[NodeId, Option[NodeExpectedReports]] ] = {
```
fanf, line 212: I'm not sure that that function should be split appart from the "getCurrentExpectedReports" part. It makes harder to see what need to be atomic, and it hides the fact that nodes "fromDb" should all have beed not initialized in cache (since it's called for that reason). 
Let me try to merge them

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,350 @@
...
+    // In a semaphore, nothing should change the cache
+    semaphore.withPermit(
+      IOResult.effect({
+      val mergedEntries = fromDb.map {  case (nodeId, expFromDb) => cache.get(nodeId) match {
+        case None => None // nodeid not in cache, dropping
```
fanf, line 217: for example, here I don't understand how the nodeId could be not in cache, since we restricted the call to fromDb to only node in cache but not initialized

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,350 @@
...
+        case Some(value) => value match { // node id in cache
+          case None => Some((nodeId -> expFromDb)) // with no value, the one from DB is better (even if none)
+          case Some(entry) => // now we must compare with the one from DB
+            expFromDb match {
+              case None => Some((nodeId -> value))
+              case Some(db) if db.beginDate.isAfter(entry.beginDate) => Some((nodeId -> expFromDb)) // data fom db is newer
```
fanf, line 223: and here, only node with "none" in cache were called for db, so that should not happen

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
ncharles: i *think* it needs to be appart. It'll be probably used in other methods

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
ncharles: adding/removing nodes can happen at the same time
also, a node may exist without compliance (ignored, or policy not yet generated)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
ncharles: query can take several seconds, cache might have been updated in between with a fresh nodeconfiguration pushed by policy generation

fanf: I have an alternative proposal: https://github.com/ncharles/rudder/pull/10
Cache should not be updated in between, that doesn't make sense to look for empty nodeIds and then consider that that data is not the truth - or it was not usefull to check for that at first. I can't see how the DB query can't be in the semaphore if you need consistency (for ex, if you don't want to miss the fact that a cache reset was done in between).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
fanf: For the case without compliance, it's ok: it was none, it's still none after. 
For the case "adding/removing nodes can happen at the same time", I don't understand: here, we are just augmenting the cache knwon entries, we don't remove entries. And there's a semaphore to avoid side effects from `performAction`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
fanf: but for now, it is not, and it hads substential complexity.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
fanf: It works (with some boilerplate), so we can make it checkable: https://github.com/ncharles/rudder/pull/10

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -0,0 +1,318 @@
...
+   */
+  def getCurrentExpectedReports(nodeIds: Set[NodeId]): IOResult[Map[NodeId, Option[NodeExpectedReports]]] = {
+    // add logging, to ensure that semaphoring the whole is not too blocking
+    logger.trace(s"Calling getCurrentExpectedReports - before semaphore")
+    val before_semaphoreTime = System.currentTimeMillis
+
+    // In a semaphore, nothing should change the cache
+    semaphore.withPermit(for {
+      timeInSemaphore <- currentTimeMillis
+      _               = logger.trace(s"Entered the semaphore after ${timeInSemaphore - before_semaphoreTime} ms")
```
fanf, line 213: it needs to be `<- logger ...`

ncharles:
PR updated with a new commit

fanf (approved):

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -180,19 +182,19 @@ class RuleGrid(
       val start = System.currentTimeMillis
 
       ( for {
-          rules           <- roRuleRepository.getAll(false).toBox.map { allRules => onlyRules match {
```
ncharles, line 183: this causes https://issues.rudder.io/issues/19651

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -1370,8 +1371,9 @@ trait PromiseGeneration_setExpectedReports extends PromiseGenerationService {
     }.toList
   }
 
-  override def invalidateComplianceCache(actions: Seq[(NodeId, CacheComplianceQueueAction)]): Unit = {
-    complianceCache.invalidateWithAction(actions).runNow
+  override def invalidateComplianceCache(actions: Seq[(NodeId, CacheExpectedReportAction)]): IOResult[Unit] = {
+    cachedNodeConfigurationService.invalidateWithAction(actions) *>
```
ncharles, line 1375: this is probably not doing what we want

PR #3729, opened 2021-07-26
Fixes #19567: Rudder fails to stop on boot error

fanf:
https://issues.rudder.io/issues/19567


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -1187,10 +1189,9 @@ object RudderConfig extends Loggable {
    * lazy and will only be initialised on the first call to one
    * of its (public) methods.
    */
-  def init() : Unit = {
+  def init() : IO[SystemError, Unit] = {
 
-    // this one must be in a fork thread pool
-    ZioRuntime.internal.unsafeRunAsync(IOResult.effect {
```
fanf, line 1193: the "async" made exception disappear and not be catched, ever. Not sure what was  happening. I failed to understood why it needed to be more on an other threadpool than what is provided by `IOResult.effect`. It works on one cpu, with plugins with the patch, too.

ncharles (approved):

fanf:
OK, merging this PR

PR #3734, opened 2021-07-27
Fixes #19576: Add system configuration object for scale-out-relay plugin

fanf:
https://issues.rudder.io/issues/19576

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
@@ -44,6 +44,8 @@ import com.normation.eventlog.EventActor
 import com.normation.eventlog.ModificationId
 import ca.mrvisser.sealerate
 import cats.data.NonEmptyList
+import com.normation.cfclerk.domain.TechniqueName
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
```diff
@@ -416,3 +438,136 @@ object RelaySynchronizationMethod {
...
+      , _isEnabled = true
+      , isSystem = true
+      , Tags(Set.empty)
+    )
+  }
+
+  val relayTechniques = List("server-common", "rudder-service-apache","rudder-service-relayd")
+  val rootTechniques = List("rudder-service-postgresql","rudder-service-slapd","rudder-service-webapp")
+
+  def directiveCommonHasPolicyServer(nodeId: NodeId, hostname: String, policyServerId: NodeId) = {
```
ncharles, line 475: this signature is super confusing: i would expect the policyServerId to be the final policyServerId (so the relay id) - i think it needs either a comment or better method naming

ncharles (changes requested):
i would really enjoy a comment

fanf:
PR updated with a new commit

ncharles (approved):

PR #3741, opened 2021-07-28
Fixes #19593: Add a link to the local doc in healtcheck for file descriptor

ElaadF:
https://issues.rudder.io/issues/19593


fanf:
We need to support html or markdow + rendering to finish than one.

fanf:
I'm closing this one: we will need to take it back from scratch once we better know what we want to do exactly

PR #3743, opened 2021-07-28
Fixes #19600: error getting techniques when coming back on technique page

fanf:
https://issues.rudder.io/issues/19600


webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/JsonDecoder.elm
```diff
@@ -82,7 +82,7 @@ decodeMethodCall =
     |> required "parameters"  (list decodeCallParameter )
     |> required "class_context"  (map parseCondition string)
     |> required "component"  string
-    |> required "disableReporting" bool
+    |> optional "disableReporting" bool False
```
fanf, line 85: that was the main problem: a new field was added, but old techniques don't have it.

amousset (approved):

PR #3759, opened 2021-07-30
Fixes #19639: Updgrade zio-json to last version for rudder 7.0

fanf:
https://issues.rudder.io/issues/19639

- *lot* of security / decoding problem fixed
- amazing perf boost thanks to jsoniter author, 
- availability of an `Json` ast from/to which we can parse objects. Much more flexible for complexe cases.


fanf:
PR updated with a new commit

ncharles (approved):

PR #3805, opened 2021-08-10
Fixes #19751: cannot execute command because of parent ticket - revert nuprocess to 2.0.1

ncharles:
https://issues.rudder.io/issues/19751


ncharles:
Commit modified

amousset (approved):

PR #3856, opened 2021-08-27
Fixes #19786: When we delete a directive, it's not removed from the rules

fanf:
https://issues.rudder.io/issues/19786

The problem was that by changing the serialisation format in LDAP for directive ids in a rule id (from `ab459975-7116-40ad-aab7-6479694ff606` to `{"id":"ab459975-7116-40ad-aab7-6479694ff606"}` (with optionnaly `, "revision":"xxx"`), we broken LDAP search on directive IDs in rules used for cleaning-up thing when a directive was deleted. 

So I reverted to the same format that used for displaying ID (ie: `uuid+revision`). 
I also changed that format for the API case, because it will likely result in the same kind of breakages. 

But now, I'm sad because we have two formats for object ID: 
- when we talk about the object own ID, it's two fields (both in LDAP and in API): `{... "id":"xxx", "revision":... }`. 
- when we talk about IDs of objects referenced by the current object (like directives in rules), we use a string serialisation `uuid+id`

I'm not sure what to do, but we need to choose before final.


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #3861, opened 2021-09-01
Fixes #19897: Display result of search with thousands of nodes is really slow on 6.2 & 7.0

ncharles:
https://issues.rudder.io/issues/19897

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
@@ -121,13 +121,14 @@ trait NodeInfoService {
 
   /**
    * Return a NodeInfo from a NodeId. First check the ou=Node, then fetch the other data
-   * @param nodeId
-   * @return
    */
   def getNodeInfo(nodeId: NodeId) : Box[Option[NodeInfo]]
 
   // Pure version of getNodeInfo
   def getNodeInfoPure(nodeId: NodeId) : IOResult[Option[NodeInfo]]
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -121,14 +121,17 @@ trait NodeInfoService {
...
-   * @return
    */
   def getNodeInfo(nodeId: NodeId) : Box[Option[NodeInfo]]
 
   // Pure version of getNodeInfo
   def getNodeInfoPure(nodeId: NodeId) : IOResult[Option[NodeInfo]]
 
+  def getNodeInfos(nodeIds: Set[NodeId]) : Box[Set[NodeInfo]]
+
+  def getNodeInfosPure(nodeIds: Set[NodeId]): IOResult[Set[NodeInfo]]
```
fanf, line 132: you don't need to make a "pure" version if there was no existing method without pure. We did that to ensure binary compat. 
So you can just do one ` def getNodeInfos(nodeIds: Set[NodeId]) : IOResult[Set[NodeInfo]] ` and use `toBox` when needed, since it's the general direction we are set on.

fanf: (plus having explicit `toBox` help find code which need to be ported to pure zio code)

ncharles:
PR updated with a new commit

fanf (approved):

ncharles:
OK, merging this PR

PR #3862, opened 2021-09-01
Fixes #19899: correct escaping for technique editor in 7.0

ncharles:
https://issues.rudder.io/issues/19899

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
@@ -188,7 +188,7 @@ object ParameterType {
       (paramType,agentType) match {
         case (Raw,_) => Right(value)
         case (StringParameter | HereString, AgentType.CfeCommunity | AgentType.CfeEnterprise) =>
-          Right(s""""${value.replaceAll("""(?<!\\)"""", """\\"""").replaceAll("""(?<!\\)\\(?!(\\|"|'))""","""\\\\""")}"""")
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
```diff
@@ -188,7 +188,7 @@ object ParameterType {
       (paramType,agentType) match {
         case (Raw,_) => Right(value)
         case (StringParameter | HereString, AgentType.CfeCommunity | AgentType.CfeEnterprise) =>
-          Right(s""""${value.replaceAll("""(?<!\\)"""", """\\"""").replaceAll("""(?<!\\)\\(?!(\\|"|'))""","""\\\\""")}"""")
+          Right(s""""${value.replaceAll("""\\""", """\\\\""").replaceAll(""""""" , """\\"""" )}"""")
```
VinceMacBuche, line 191: maybe for more readability we can so it in replaceall

value.replaceAll("(\\|\")", "\\$1")

or 
value.replaceAll("""(\\|")""", "\\$1")

VinceMacBuche: github escapes my \ ... i had to put four 4 to display 2 ....

ncharles: is it more readable ?

VinceMacBuche (approved):

PR #3867, opened 2021-09-03
Fixes #19881: Saving groups in the rules UI is broken

fanf:
https://issues.rudder.io/issues/19881


webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/JsonDecoder.elm
```diff
@@ -205,12 +205,12 @@ decodeTargets =
    , D.lazy (\_ -> decodeOr)
    , map (\s ->
            if String.startsWith "group:" s then
-             NodeGroupId s
```
VinceMacBuche, line 208: You could have used String.dropLeft

VinceMacBuche (approved):

PR #3869, opened 2021-09-05
Fixes #19915: Improve rust build times

amousset:
https://issues.rudder.io/issues/19915

### DIFF
--- a/rust.makefile
+++ b/rust.makefile
@@ -48,6 +48,8 @@ version:
 	@echo "RUSTC_WRAPPER=$${RUSTC_WRAPPER}"
 	sccache --show-stats || true
 
+# https://matklad.github.io/2021/09/04/fast-rust-builds.html#ci-workflow
+build: CARGO_INCREMENTAL=0
 build: version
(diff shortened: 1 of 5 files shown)


amousset:
`debug = 0` seems to make a ~10% difference locally, but could be more on slow io.

rust.makefile
```diff
@@ -21,6 +21,9 @@ else ifneq ($(REDHATOS),)
 PKG_INSTALLER := $(YUM)
 endif
 
+# https://matklad.github.io/2021/09/04/fast-rust-builds.html#ci-workflow
+export CARGO_INCREMENTAL = 0
```
peckpeck, line 25: doesn't this slow down local developper build ?

amousset: I never use this makefile for local dev, but it seems to be used for rudder-lang testing so we need to do it differently indeed. I'll update the PR to pass the parameter through jenkins job directly.

amousset:
PR updated with a new commit

peckpeck (approved):

amousset:
OK, squash merging this PR

amousset:
OK, merging this PR

PR #3872, opened 2021-09-07
Fixes #19926: Support reporting logic metadata in language

peckpeck:
https://issues.rudder.io/issues/19926

### DIFF
--- a/language/src/technique.rs
+++ b/language/src/technique.rs
@@ -12,15 +12,20 @@ use crate::{
 use colored::Colorize;
 use lazy_static::lazy_static;
 use regex::{Captures, Regex};
+use serde::de::{MapAccess, Visitor};
+use serde::ser::SerializeStruct;
 use serde::{
     de::{self, Deserializer},
     Deserialize, Serialize, Serializer,
 };
+use std::collections::HashMap;
+use std::fmt::{Display, Formatter};
 use std::{
     convert::TryFrom,
     fmt,
     str::{self, FromStr},
 };
+use toml::Value as TomlValue;
 use uuid::Uuid;
 
 // Techniques are limited subsets of CFEngine in JSON representation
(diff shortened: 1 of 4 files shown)


peckpeck:
PR updated with a new commit

peckpeck:
PR updated with a new commit

language/src/technique.rs
```diff
@@ -200,8 +202,36 @@ enum MethodElem {
     MethodBlock(MethodBlock),
 }
 
+#[derive(Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
+enum ReportingLogicType {
+    Focus,
```
amousset, line 208: what does the "focus" type does?

peckpeck: Ask the author :-)
I means : "look and the value element to extract the reporting of a single sub item"

amousset: so this one needs to be put on one of the states inside of the technique?

peckpeck: This one is the technique editor representation of this information, but you are right, it should be a metadata of a state inside the block in the language

amousset: LGTM with a comment explaining the focus case requires additional stuff

VinceMacBuche: I think it should be represented differently.

Focus should always have a "value" parameter and the other should never have one. I think their type should represent that

VinceMacBuche: More like 
enum ReportingLogic {
    Focus { value : String },
   Worst,
  Sum
}

peckpeck: indeed

peckpeck: but this woudn't match the json structure

VinceMacBuche: I think it should match the semantic of the program we want not the json structure, we don't want a value for Sum or Worst,and we always want one for Focus (where missing Value is an error and not optionnal)

peckpeck:
done

VinceMacBuche (approved):

peckpeck:
OK, squash merging this PR

peckpeck:
OK, merging this PR

PR #3874, opened 2021-09-08
Fixes #19860: On fresh install the login page refresh so fast that we can not copy/paste the command to create a new user

ncharles:
https://issues.rudder.io/issues/19860


fanf:
Given the addition of the clipboard copy button, I think we should let a shorter refresh rate

fanf:
Closing it thanks to the change in UI.

PR #3875, opened 2021-09-08
Fixes #19872: Rules can't be accessed directly by url, all redirection broken

RaphaelGauthier:
https://issues.rudder.io/issues/19872

So, there is apparently no turnkey solution to set up a URL system the way we want it. Here is my workaround, which I think can be highly improved: 

I used 2 new ports:
  - `pushUrl` which updates the url based on what happens in the Elm app (e.g. when a rule is selected):
<pre>
OpenRuleDetails rId ->
      (model, Cmd.batch [(getRuleDetails model rId), (pushUrl ((getUrl model) ++ "/rule/" ++rId.value))])
</pre>      
  - `onUrlChange` which updates the Elm app when the page loads and when the url is changed. This port calls the `UrlChanged` function, which parses the url and determines what action to take based on the result. (All the Url rewriting logic is done on the Elm side.
  
I also added redirections to avoid 404 errors when the url looks like **[...]/secure/configurationManager/ruleManagement/rule/1234-5678-1234-5678.**  

I think this workaround could be more robust. For example by giving a type to the URLs to avoid manipulating strings. 
And I didn't deal with the case of Rule / Category creation, for the moment, the url is not updated in these cases. But for now I'm waiting for your feedback on this!

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ApiCalls.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ApiCalls.elm
@@ -104,6 +104,22 @@ getRuleDetails model ruleId =
   in
     send GetRuleDetailsResult req
 
+getRulesCategoryDetails : Model -> String -> Cmd Msg
+getRulesCategoryDetails model catId =
+  let
+    req =
+      request
+        { method          = "GET"
+        , headers         = []
+        , url             = getUrl model ("/rules/categories/" ++ catId)
+        , body            = emptyBody
+        , expect          = expectJson decodeGetCategoryDetails
+        , timeout         = Nothing
+        , withCredentials = False
+        }
+  in
+    send GetCategoryDetailsResult req
+
 getRulesCompliance : Model -> Cmd Msg
 getRulesCompliance model =
   let
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
@@ -87,7 +87,7 @@ class LinkUtil (
     RedirectTo(baseTargetLink(target))
 
   def baseRuleLink(id:RuleId) =
-    s"""/secure/configurationManager/ruleManagement#{"ruleId":"${id.value}"}"""
+    s"""/secure/configurationManager/ruleManagement/rule/${id.value}"""
 
   def ruleLink(id:RuleId) =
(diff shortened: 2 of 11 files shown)


RaphaelGauthier:
I need to do the same thing for categories before merging this PR

VinceMacBuche:
You need a function to set hash when selecting a rule so that the url can be shared

fanf:
Actually, if we can not use js+hashes but url rewrite, it would be great because js+hashes breaks on login redirection, and now navigator better supports it. And hashes are sooooo 2010...

RaphaelGauthier:
PR rebased

RaphaelGauthier:
So, there is apparently no turnkey solution to set up a URL system the way we want it. Here is my workaround, which I think can be highly improved: 

I used 2 new ports:
  - `pushUrl` which updates the url based on what happens in the Elm app (e.g. when a rule is selected):
<pre>
OpenRuleDetails rId ->
      (model, Cmd.batch [(getRuleDetails model rId), (pushUrl ((getUrl model) ++ "/rule/" ++rId.value))])
</pre>      
  - `onUrlChange` which updates the Elm app when the page loads and when the url is changed. This port calls the `UrlChanged` function, which parses the url and determines what action to take based on the result. (All the Url rewriting logic is done on the Elm side.
  
I also added redirections to avoid 404 errors when the url looks like **[...]/secure/configurationManager/ruleManagement/rule/1234-5678-1234-5678.**  

I think this workaround could be more robust. For example by giving a type to the URLs to avoid manipulating strings. 
And I didn't deal with the case of Rule / Category creation, for the moment, the url is not updated in these cases. But for now I'm waiting for your feedback on this!

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ApiCalls.elm
```diff
@@ -104,6 +104,22 @@ getRuleDetails model ruleId =
   in
     send GetRuleDetailsResult req
 
+getRulesCategoryDetails : Model -> String -> Cmd Msg
```
VinceMacBuche, line 107: Why do you changes on RUles Category here ?

RaphaelGauthier: I forgot to specify this change. For several reasons:
- To have the same behavior as when selecting a rule
- To be sure to be up to date when opening the details of a category
- And above all, to be able to retrieve the information of a category only from an ID (since this is the only information we get when we parse the URL)

RaphaelGauthier:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #3876, opened 2021-09-08
Fixes #19740: When 2 changes are close in time,  expected report on the compliance page don't show the new config

ncharles:
https://issues.rudder.io/issues/19740

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
@@ -80,6 +80,11 @@ trait NodeConfigurationService {
   def findNodesApplyingRule(ruleId: RuleId): IOResult[Set[NodeId]]
 }
 
+
+trait NewExpectedReportsAvailableHook {
+  def newExpectedReports(action: CacheExpectedReportAction): IOResult[Unit]
+}
+
 class CachedNodeConfigurationService(
     val confExpectedRepo: FindExpectedReportRepository
   , val nodeInfoService : NodeInfoService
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeConfigurationService.scala
```diff
@@ -90,6 +90,20 @@ class CachedNodeConfigurationService(
...
+   * configs are changed. However, the compliance cache needs this present class
+   * So it's a stackoverflow
+   * Workaround is to define once the compliance cache is defined the compliance cache
+   * in the configuration cache.
+   * Introduced in https://issues.rudder.io/issues/19740
+   */
+  var complianceCache : Option[CachedFindRuleNodeStatusReports] = None
+
+  def setComplianceCache(cachedCompliance: CachedFindRuleNodeStatusReports) = {
+    complianceCache = Some(cachedCompliance)
```
fanf, line 105: There's a way to make it clear on what the dependency is, and limite the exosed API to only what you need to expose (ie the react to `new expected reports available`). It needs a bit of boilerplate to specify that API thought:

- create a trait `NewExpectedReportsAvailableHook { def newExpectedReports(nodeId: NodeId): UIO[Unit] }` (I also like to add a "name" property so that latter on, when we want to add logs, we can use that, and it forces people to document at least in that name why that hook exists)
- in current class, in place of the `var complianceCache`, add a `var hooks = List[NewExpectedReportsAvailableHook]` + a method to add one (`def addHook(hook: ...Hook): Unit = { hooks = hook :: hooks}`). It also ca be a ref in place of a var, and exposed directly or hidden behind the same kind of add method.
- make CachedNodeConfigurationService implements `NewExpectedReportsAvailableHook` and implement the `newExpectedReport` method for it. 

With that, you minimize what needs to be know between the two interdependant services, and when at some point we switch to an internal messaging hub, it will be clear what messages need to be produced and where they are consumed (and even if we never do that, the API is clear and the loop dependency broken, since now it is on a message, and not an internal implementation)

fanf (commented):
There is a proposition to make the circular dependency better

ncharles:
PR updated with a new commit

fanf (approved):

PR #3881, opened 2021-09-13
Fixes #19869: Persist revisions of node inventories in fact-repository

fanf:
https://issues.rudder.io/issues/19869


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
```diff
@@ -0,0 +1,602 @@
...
+
+    def update(ruleCategory: RuleCategory) = {
+      ruleCategory.using(this).ignoreRedundantPatcherFields.patch
+    }
+  }
+
+
+  final case class JQDirectiveSectionVar(
+      name : String
+    , value: String
```
ncharles, line 113: we used to have null value and "" value, having sometimes different or similar behaviour.
shouldn't the value be an option for empty value that are not mandatory?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
```diff
@@ -0,0 +1,602 @@
...
+  case class PatchDirective(
+      id              : Option[DirectiveId]
+    , techniqueVersion: Option[TechniqueVersion]
+    , parameters      : Option[Map[String, Seq[String]]]
+    , name            : Option[String]
+    , shortDescription: Option[String]
+    , policyMode      : Option[Option[PolicyMode]]
+    , longDescription : Option[String]
+    , priority        : Option[Int]
+    , _isEnabled      : Option[Boolean]
```
ncharles, line 142: why the _ ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/readme.adoc
```diff
@@ -0,0 +1,22 @@
...
+- renaming a field (even optionnal): the new name would be looked up in old files, leading to incorrect data if it was present and non empty under the old name.
+- changing the type or the multiplicity of a field
+
+All these changes can be workarouned with adaptators (ex: default value for the new mandatory field, looking-up the old name, etc) and that's why we need a fileformat.
+
+These changes don't need a file format increment:
+- removing a mandatory field or transforming one to optionnal.
+  This is ok since it means that old serialized objects will have it, just that
+  now it is not required anymore.
+- removing or adding optionnal fields
```
ncharles, line 22: thank you for thie readme

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
fanf: I think we just don't have that level of precision from the user input. The parsing is done afterward, here the object is just a direct translation of json. It means that for optionnal values, the user can't do:
```
{"vars": [ {"var": { "name":"var1", "value":"val1" }, {"var": { "name":"var2" }}]}
```
And should do: 
```
{"vars": [ {"var": { "name":"var1", "value":"val1" }}]}
```
(and var2 will be missing on parsing, and it will be set to None in the domain object). 

If you know about a precise use case, I can add it in test in an other ticket.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
fanf: it's the name of the field in the target object, the patch lib look-up field by name by default

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactRepository.scala
```diff
@@ -0,0 +1,241 @@
...
+                .modify(_.machine.each.processors).using(_.sortBy(_.name))
+                .modify(_.machine.each.slots).using(_.sortBy(_.name))
+                .modify(_.machine.each.sounds).using(_.sortBy(_.name))
+                .modify(_.machine.each.storages).using(_.sortBy(_.name))
+                .modify(_.machine.each.videos).using(_.sortBy(_.name))
+
+
+    val json = FullDetailLevel.toJson(nodeInfo, inventory.node.main.status, None, Some(inv), software.sortBy(_.name.getOrElse("")))
+    // add entity type and file format at the begining
+    ("entity" -> entity ) ~ ( "fileFormat" -> fileFormat ) ~ json
```
VinceMacBuche, line 159: I think you want renderJson in the trait that adds entity and fileFormat
and a toJson that renders the json   

something in the trait 

``` 
def renderJson(b : B) : IOResult[String] = {
  val json = ("entity" -> entity ) ~ ( "fileFormat" -> fileFormat ) ~ toJson(b)
  IOResult.effect(prettyRender(json))
} ```

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactRepository.scala [hunk 5]
```diff
@@ -0,0 +1,241 @@
...
+import org.eclipse.jgit.lib.PersonIdent
+
+import zio._
+import zio.syntax._
+import com.normation.errors._
+
+/*
+ * write node facts
+ */
+trait NodeFactRepository {
```
VinceMacBuche, line 60: We may want to do a FactRepository[A,B] that goes well with the SerializeFacts trait

VinceMacBuche: at least persist

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryProcessor.scala
```diff
@@ -348,12 +348,15 @@ class InventoryProcessor(
       saved <- inventorySaver.save(inventory).chainError("Can't merge inventory inventory in LDAP directory, aborting").either
```
VinceMacBuche, line 348: inventory inventory, remove one :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/InventoryProcessor.scala
```diff
@@ -348,12 +348,15 @@ class InventoryProcessor(
...
       _     <- saved match {
                  case Left(err) =>
-                   InventoryProcessingLogger.error(s"Error when trying to process inventory: ${err.fullMsg}")
-                 case Right(inventory) =>
+                   InventoryProcessingLogger.error(s"Error when trying to process report: ${err.fullMsg}")
```
VinceMacBuche, line 351: I think you don't want to rename to report

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactRepository.scala
fanf: I wanted to hide the `Json` liftweb type to allow using either that one or zio-json one (which doesn't have an intermediate representation). But I agree that entity and fileformat are generic. I'm not sure how to make it generic, so I stuck with the simple thing for now. I will add comment to explain.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactRepository.scala
fanf: yeah, but here, too, I'm not sure how to do it, b/c I'm not sure that we want to have only one big fact repository with all fact, and thus, it should have one persist by type (and so param is at persist level), or if we want one small repo by fact (like what we have for configuration). The second solution may feel simpler, but it leads to a LOT of duplication of code, and maintenance is not easy. I will add comment, too.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
ncharles: no, tht's fine like that: if it's the user input, it is as it should be. Thank you for the explaination

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
ncharles: :+1:

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/RestDataSerializer.scala
```diff
@@ -35,7 +35,7 @@
 *************************************************************************************
 */
 
-package com.normation.rudder.rest
+package com.normation.rudder.apidata
```
ncharles, line 38: this breaks changevalidation plugin

PR #3887, opened 2021-09-15
Fixes #19752: error when using "group_absent" in a technique in 7.0

ncharles:
https://issues.rudder.io/issues/19752


amousset (changes requested):
should be name, could probably be renamed elsewhere

ncharles:
> should be name, could probably be renamed elsewhere

why ?

PR #3893, opened 2021-09-21
Fixes #19980: spurious "connection_read(9): no connection!" in /var/log/rudder/ldap/slapd.log

ncharles:
https://issues.rudder.io/issues/19980

### DIFF
--- a/webapp/sources/scala-ldap/src/main/scala/com/normation/ldap/sdk/LDAPIOResult.scala
+++ b/webapp/sources/scala-ldap/src/main/scala/com/normation/ldap/sdk/LDAPIOResult.scala
@@ -34,7 +34,8 @@ sealed trait LDAPRudderError extends RudderError {
 object LDAPRudderError {
   // errors due to some LDAPException
   final case class BackendException(msg: String, cause: Throwable)  extends LDAPRudderError {
-    override def fullMsg: String = super.fullMsg + s"; cause was: ${RudderError.formatException(cause)}"
+    override def fullMsg: String = super.fullMsg + s"; cause was: ${RudderError.formatException(cause)} ;" +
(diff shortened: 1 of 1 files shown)


webapp/sources/scala-ldap/src/main/scala/com/normation/ldap/sdk/LDAPIOResult.scala
```diff
@@ -34,7 +34,8 @@ sealed trait LDAPRudderError extends RudderError {
 object LDAPRudderError {
   // errors due to some LDAPException
   final case class BackendException(msg: String, cause: Throwable)  extends LDAPRudderError {
-    override def fullMsg: String = super.fullMsg + s"; cause was: ${RudderError.formatException(cause)}"
+    override def fullMsg: String = super.fullMsg + s"; cause was: ${RudderError.formatException(cause)} ;" +
+      s"you probably need to increase value ldap.maxPoolSize property in configuration file, please check the documentation"
```
VinceMacBuche, line 38: maybe we can put a link to the documentation ? maybe the embedded one ?

ncharles: i'm unsure we will effectively maintain this link in an error message :/

fanf: I think it's better with the link. It's better for the user, and even if it becomes incorrect at some point, it helps the user going at the right place. 
Not sure if we have the base url for the local doc though.

ncharles:
PR updated with a new commit

fanf (approved):

fanf:
OK, squash merging this PR

PR #3894, opened 2021-09-21
Fixes #19981: Display the name of the groups of a rule instead of their id

RaphaelGauthier:
https://issues.rudder.io/issues/19981

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewRuleDetails.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewRuleDetails.elm
@@ -98,6 +98,10 @@ editionTemplate model details =
     nbInclude = case Maybe.Extra.unwrap [] .targets originRule of
       [Composition (Or i) (Or e)] -> List.length i
       targets -> List.length targets
+
+    nbDirectives = case originRule of
+      Just oR -> String.fromInt (List.length oR.directives)
+      Nothing -> "0"
   in
     div [class "main-container"]
     [ div [class "main-header "]
@@ -131,7 +135,7 @@ editionTemplate model details =
           [ a[onClick (UpdateRuleForm {details | tab = Directives})]
             [ text "Directives"
             , span[class "badge badge-secondary badge-resources tooltip-bs"]
-              [ span [class "nb-resources"] [ text (String.fromInt(List.length rule.directives))]
+              [ span [class "nb-resources"] [ text nbDirectives]
               , ( if diffDirectivesPos /= 0 then span [class "nb-resources new"] [ text (String.fromInt diffDirectivesPos)] else text "")
               , ( if diffDirectivesNeg /= 0 then span [class "nb-resources del"] [ text (String.fromInt diffDirectivesNeg)] else text "")
               ]
--- a/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewTabContent.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewTabContent.elm
@@ -387,9 +387,12 @@ tabContent model details =
             in
               span [class ("rudder-label label-sm label-" ++ policyMode)][]
 
-          buildIncludeList : RuleTarget -> Html Msg
-          buildIncludeList ruleTarget =
+          buildIncludeList : Bool -> RuleTarget -> Html Msg
+          buildIncludeList includeBool ruleTarget =
+
             let
+              groupsList = getAllElems model.groupsTree
+
               id = case ruleTarget of
                 NodeGroupId groupId -> groupId
                 Composition _ _ -> "compo"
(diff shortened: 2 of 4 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewRuleDetails.elm
```diff
@@ -98,6 +98,10 @@ editionTemplate model details =
     nbInclude = case Maybe.Extra.unwrap [] .targets originRule of
       [Composition (Or i) (Or e)] -> List.length i
       targets -> List.length targets
+
+    nbDirectives = case originRule of
+      Just oR -> String.fromInt (List.length oR.directives)
```
VinceMacBuche, line 103: Why the number from the original Rule and not for the current Rule ? does not that number need to change if you add directives ? Same question for group though

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewTabContent.elm
```diff
@@ -444,33 +450,52 @@ tabContent model details =
...
-                      , div [class "treeActions-container"]
-                        [ span [class "treeActions"][ span [class "tooltipable fa action-icon accept", onClick (SelectGroup (NodeGroupId item.id) True)][]]
-                        , span [class "treeActions"][ span [class "tooltipable fa action-icon except", onClick (SelectGroup (NodeGroupId item.id) False)][]]
-                        ]
+                let
+                  checkIncludeOrExclude : List RuleTarget -> Bool
+                  checkIncludeOrExclude lst = lst
+                    |> List.map (\t -> case t of
+                       NodeGroupId groupId -> groupId
+                       Composition _ _ -> "compo"
```
VinceMacBuche, line 458: I Think this need to be a recursive function and check in And/Or/Composition Element to see if the item is inside

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewRuleDetails.elm
RaphaelGauthier: Because I think we should display the number of directives/groups that are actually applied to the rule, and this should not change until we save our changes. We keep track of the number of directives we've added / removed in another badge next to it.

I could be wrong though.

RaphaelGauthier:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewTabContent.elm
```diff
@@ -452,16 +452,20 @@ tabContent model details =
...
-                       And _ -> "and"
-                    )
-                    |> List.member item.id
+                  checkIncludeOrExclude lst =
+                    let
+                      getTargetIds : List RuleTarget -> List String
+                      getTargetIds listTargets =
+                        List.concatMap (\target -> case target of
+                           NodeGroupId groupId -> [groupId]
+                           Composition i e -> getTargetIds [i, e]
```
RaphaelGauthier, line 461: Should it be `Composition i e -> getTargetIds [ i ]` when we call `checkIncludeOrExclude includedTargets`,
and `Composition i e -> getTargetIds [ e ]` when we call `checkExcludeOrExclude excludedTargets`, or is it correct like this ?

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewRuleDetails.elm
VinceMacBuche: Ha yes totally !

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewTabContent.elm
VinceMacBuche: no it's targets form i minus targets from e

a composition is a list of included targets minus items from excluded targets

you should do something 

 let included = getTargets [ i ] 
 excluded = getTargets [e] 
in included |>
      List.filter (i -> List.member i excluded) 

so that member from excluded  are removed from included

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewTabContent.elm
```diff
@@ -444,33 +450,56 @@ tabContent model details =
...
+                           Special spe -> [spe]
+                           Node node -> [node]
+                           Or t  -> getTargetIds t
+                           And t -> getTargetIds t
```
VinceMacBuche, line 465: And is more complicated (or is okay) and you need to keep targets that are member of all targets list in t

But you may not have to do it right know never mind (maybe put a comment here that And is not ok as of now)

RaphaelGauthier:
Commit modified

VinceMacBuche (approved):

PR #3897, opened 2021-09-24
Fixes #19585: webapp is slow to detect inventories received in /var/rudder/inventories/accepted-nodes-updates/

ncharles:
https://issues.rudder.io/issues/19585


ncharles:
server don't see inventory with that...

fanf:
Super seeded by #4281

PR #3899, opened 2021-09-27
Fixes #20006: Store processes of an inventory after storing the inventory

ncharles:
https://issues.rudder.io/issues/20006

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-provisioning-core/src/main/scala/com/normation/inventory/ldap/provisioning/DefaultReportSaver.scala
+++ b/webapp/sources/ldap-inventory/inventory-provisioning-core/src/main/scala/com/normation/inventory/ldap/provisioning/DefaultReportSaver.scala
@@ -47,6 +47,7 @@ import com.normation.inventory.ldap.core._
 import com.normation.ldap.sdk.RwLDAPConnection
 import com.normation.zio._
 import zio._
+import zio.syntax.ToZio
 
 /**
  * Post-commit convention:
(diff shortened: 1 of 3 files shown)


webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
```diff
@@ -775,6 +773,16 @@ class InventoryMapper(
         win.setOpt(productId, A_WIN_ID, { x: String => x })
         win
     }
+    root
+  }
+
+  def treeFromNode(server:NodeInventory) : LDAPTree = {
```
fanf, line 779: it needs at least a comment, perhaps even a parameter `includeProcess: Boolean = false`. Else it's sure that one day we will wonder why our nodes are missing process in LDAP

webapp/sources/ldap-inventory/inventory-provisioning-core/src/main/scala/com/normation/inventory/ldap/provisioning/DefaultReportSaver.scala
```diff
@@ -83,11 +83,15 @@ class DefaultReportSaver(
         t3  <- currentTimeMillis
         _   <- InventoryProcessingLogger.timing.trace(s"Saving node: ${t3 - t2} ms")
 
-        d3  <- ZIO.foreach(report.vms) { x =>con.saveTree(mapper.treeFromMachine(x), deleteRemoved = true) }
+        d3  <- con.save(mapper.processesFromNode(report.node), removeMissingAttributes = false)
```
fanf, line 86: I *think* (not sure) that we could even ignore errors on that one (or at least timeout errors) (with a `catchAll(err => log(err) *> return an empty list of modif for that part)`

ncharles:
PR updated with a new commit

webapp/sources/ldap-inventory/inventory-provisioning-core/src/main/scala/com/normation/inventory/ldap/provisioning/DefaultReportSaver.scala
fanf: ha

fanf: oh it's outdated :)

ncharles: i found a solution when i pressed enter :)
i need to check it works

fanf (approved):

fanf:
OK, squash merging this PR

PR #3900, opened 2021-09-27
Fixes #19988: Number of CPU is wrongly reported

ElaadF:
https://issues.rudder.io/issues/19988


ElaadF:
There is maybe a another problem in the API that seems to return only one CPU (with the quantity) even if there are several

fanf (approved):

fanf:
OK, merging this PR

PR #3910, opened 2021-10-04
Fixes #20041: rudder-lang in 7.0 should use ncf 7.0

amousset:
https://issues.rudder.io/issues/20041


fanf (approved):

peckpeck:
we will have this problem for every release

PR #3922, opened 2021-10-08
Fixes #20100: Rudder package optional args are not clear enough

Fdall:
https://issues.rudder.io/issues/20100


Fdall:
Never mind, looks like I was not up to date with the current options

PR #3923, opened 2021-10-10
Fixes #20065: After migrating from 6.2 to 7.0, techniques that were totally valid may become invalid because of name collision in rudderc

peckpeck:
https://issues.rudder.io/issues/20065


peckpeck:
PR updated with a new commit

PR #3926, opened 2021-10-11
Fixes #20102: parent ticket totally disable the wizard

ncharles:
https://issues.rudder.io/issues/20102


ncharles:
PR updated with a new commit

PR #3931, opened 2021-10-12
Fixes #20112: Missing API and data for Rule page: special targets and node compliance

VinceMacBuche:
https://issues.rudder.io/issues/20112


fanf:
You need to update rest tests (api_groups.yml misses the new targets element, see https://ci.normation.com/jenkins/job/rudder/job/PR-3931/2/console) but otherwise, it seems to be ok

PR #3937, opened 2021-10-13
Fixes #20099: Add a feature switch for rudderc use during generation

fanf:
https://issues.rudder.io/issues/20099

Add the possibility to restrict rudderc enabled target via API (and only API in that PR, we will need an other one for UI, but at least it does not block beta2 anymore). 

Use by default only "cfengine" target. 

Test both API and writing files.


fanf:
PR updated with a new commit

PR #3938, opened 2021-10-14
Fixes #20130: rudder language generate_lib should use the existing ncf python library instead of implement itself the ncf methods parsing

Fdall:
https://issues.rudder.io/issues/20130


peckpeck:
tests must be updated

peckpeck:
you need to self merge

PR #3943, opened 2021-10-17
Fixes #19769: Cannot save a technique with a block with reporting focused on one method

VinceMacBuche:
https://issues.rudder.io/issues/19769


VinceMacBuche:
Commit modified

PR #3946, opened 2021-10-18
Fixes #20146: disable variable methods that don't work with language until we find a proper solution

peckpeck:
https://issues.rudder.io/issues/20146


peckpeck:
PR updated with a new commit

peckpeck:
OK, squash merging this PR

PR #3987, opened 2021-11-09
Fixes #20253: improve compliance computation efficiency

ncharles:
https://issues.rudder.io/issues/20253


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -358,12 +360,21 @@ object ExecutionBatch extends Loggable {
     Pattern.compile("""\Q"""+ x.replaceAll(replaceCFEngineVars, """\\E.*\\Q""") + """\E""")
   }
 
-  final def checkExpectedVariable(expected : String,effective : String) : Boolean = {
-    val isVar = matchCFEngineVars.pattern.matcher(expected).matches()
-    if (isVar) { // If this is not a var, there isn't anything to replace.
-      replaceCFEngineVars(expected).matcher(effective).matches()
+  final def checkExpectedVariable(expected: String, effective: String) : Boolean = {
+    if (expected == effective) {
```
ncharles, line 364: this does save roughly 10% in test "performance for mergeCompareByRule"

PR #3992, opened 2021-11-10
Fixes #19888: technique cannot be compile if name contains \\"x 

peckpeck:
https://issues.rudder.io/issues/19888


amousset (approved):

amousset:
Could be good to add this to a test case

PR #3993, opened 2021-11-16
Fixes #20264: Allow to customize in which menu plugins are set

VinceMacBuche:
https://issues.rudder.io/issues/20264

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -356,28 +356,28 @@ class Boot extends Loggable {
 
     // All the following is related to the sitemap
     val nodeManagerMenu =
-      (Menu("NodeManagerHome", <i class="fa fa-sitemap"></i> ++ <span>Node management</span>: NodeSeq) /
+      (Menu(MenuUtils.nodeManagementMenu, <i class="fa fa-sitemap"></i> ++ <span>Node management</span>: NodeSeq) /
         "secure" / "nodeManager" / "index"  >> TestAccess( ()
             => userIsAllowed("/secure/index",AuthorizationType.Node.Read) )).submenus (
 
-          Menu("List Nodes", <span>Nodes</span>) /
+          Menu("110-nodes", <span>Nodes</span>) /
             "secure" / "nodeManager" / "nodes"
             >> LocGroup("nodeGroup")
 
-        , Menu("SearchNodes", <span>Node search</span>) /
+        , Menu("120-search-nodes", <span>Node search</span>) /
             "secure" / "nodeManager" / "searchNodes"
             >> LocGroup("nodeGroup")
 
-        , Menu("NodeDetails", <span>Node details</span>) /
+        , Menu("120-node-details", <span>Node details</span>) /
             "secure" / "nodeManager" / "node"
             >> LocGroup("nodeGroup")
             >> Hidden
 
-        , Menu("ManageNewNode", <span>Pending nodes</span>) /
+        , Menu("130-pending-nodes", <span>Pending nodes</span>) /
             "secure" / "nodeManager" / "manageNewNode"
             >>  LocGroup("nodeGroup")
 
-        , Menu("Groups", <span>Groups</span>) /
+        , Menu("140-groups", <span>Groups</span>) /
             "secure" / "nodeManager" / "groups"
             >> LocGroup("groupGroup")
             >> TestAccess( () => userIsAllowed("/secure/index",AuthorizationType.Group.Read ) )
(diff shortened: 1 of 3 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -356,28 +356,28 @@ class Boot extends Loggable {
 
     // All the following is related to the sitemap
     val nodeManagerMenu =
-      (Menu("NodeManagerHome", <i class="fa fa-sitemap"></i> ++ <span>Node management</span>: NodeSeq) /
+      (Menu("10", <i class="fa fa-sitemap"></i> ++ <span>Node management</span>: NodeSeq) /
```
fanf, line 359: I think we should keep the number only as a prefix, like what you did for plugin below, because: 
- it will make debugging in some cases easier (IIRC, we have a map of menu id -> menu at some places, and having only "10" won't help much)
- having globally unique name is less risky if at some point we make a map of all menu items (including sub menu).

fanf: these name will become part of the api (since they can be used by plugin), we need to normalize them, I propose: 
- two digits followed by  `-`
- lower case, no space, `-` between words

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -386,91 +386,91 @@ class Boot extends Loggable {
...
             => userIsAllowed("/secure/index",AuthorizationType.Configuration.Read) )).submenus (
 
-          Menu(name+"RuleManagement", <span>Rules</span>) /
+          Menu("10", <span>Rules</span>) /
             "secure" / (name+"Manager") / "ruleManagement"
             >> LocGroup(name+"Group")
             >> TestAccess( () => userIsAllowed("/secure/index",AuthorizationType.Rule.Read ) )
 
-        , Menu("Rule Details", <span>Rule</span>) /
+        , Menu("10", <span>Rule</span>) /
```
fanf, line 398: I don't think we should authorize same menu idea, I'm pretty sure it will create unspecified behavior

fanf: (even for hidden menu)

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
fanf: OK, using a name is perhaps not super futur proof, because it will be hard to rename that id. I'm not sur it is of importance (we can still rename between major version, since plugin are not compatible between them).

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -490,24 +490,24 @@ class Boot extends Loggable {
...
+        , Menu("30", <span>Health check</span>) /
             "secure" / "utilities" / "healthcheck"
             >> LocGroup("utilitiesGroup")
             >> TestAccess ( () => userIsAllowed("/secure/index",AuthorizationType.Administration.Read) )
       )
     }
 
     val rootMenu = List(
-        Menu("Dashboard", <i class="fa fa-dashboard"></i> ++ <span>Dashboard</span>: NodeSeq) / "secure" / "index"
+        Menu("0", <i class="fa fa-dashboard"></i> ++ <span>Dashboard</span>: NodeSeq) / "secure" / "index"
```
fanf, line 510: use `00` here, for consistancy

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -386,91 +386,91 @@ class Boot extends Loggable {
...
-        , Menu("apiManagement", <span>API accounts</span>) /
+        , Menu("50", <span>API accounts</span>) /
             "secure" / "administration" / "apiManagement"
             >> LocGroup("administrationGroup")
             >> TestAccess ( () => userIsAllowed("/secure/administration/policyServerManagement",AuthorizationType.Administration.Write) )
       )
 
     def pluginsMenu = {
-      (Menu("PluginsHome", <i class="fa fa-puzzle-piece"></i> ++ <span>Plugins</span> ++ <span data-lift="PluginExpirationInfo.renderIcon"></span>: NodeSeq) /
+      (Menu("50 - plugins", <i class="fa fa-puzzle-piece"></i> ++ <span>Plugins</span> ++ <span data-lift="PluginExpirationInfo.renderIcon"></span>: NodeSeq) /
```
fanf, line 468: you should not use that with the `toIntOption`. But I'm not sure we should convert to int at all.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -386,91 +386,91 @@ class Boot extends Loggable {
...
         "secure" / "plugins" / "index"
         >> LocGroup("pluginsGroup")
         >> TestAccess ( () => userIsAllowed("/secure/index", AuthorizationType.Administration.Read)
       )) submenus (
-          Menu("pluginInformation", <span>Plugin information</span>) /
+          Menu("0", <span>Plugin information</span>) /
```
fanf, line 473: I think I would use `01` here to let the possibility to have things before in the future, in the middle of a release cycle.

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
```diff
@@ -121,14 +121,26 @@ trait DefaultPluginDef extends RudderPluginDef {
...
-          case m@Menu(l, _* ) if(l.name == "PluginsHome") =>
-            Menu(l , (m.kids.toSeq :+ menu):_* )
+        val (parentName, updatedMenu) = pluginMenuParent match {
+          case None => ("90", menus)
+          case Some(parentMenu) =>
+            val name = parentMenu.loc.name
+            val menu = if (menus.exists (_.loc.name == name )) {
+               menus
+            } else {
+              (parentMenu :: menus).sortBy(_.loc.name.toIntOption.getOrElse(Int.MaxValue))
```
fanf, line 137: I think you can use the same logic without the `toInt`

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):

fanf:
OK, merging this PR

PR #3997, opened 2021-11-18
Fixes #20273: dashboard only shows the compliance of the nodes that really answer when logging on the test platform in 7.0

ncharles:
https://issues.rudder.io/issues/20273

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -746,12 +746,12 @@ final case class ContextForNoAnswer(
             )
         }
 
-      case NoReportInInterval(expectedConfig, _) =>
+      case NoReportInInterval(expectedConfig, expirationTime) =>
         ComplianceDebugLogger.node(nodeId).trace(s"Node didn't received reports recently, status depend of the compliance mode and previous report status")
         buildRuleNodeStatusReport(
-            //these reports don't really expires - without change, it will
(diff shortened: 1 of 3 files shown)


ncharles:
actually, the problem sometime still happen

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -746,12 +746,12 @@ final case class ContextForNoAnswer(
...
         }
 
-      case NoReportInInterval(expectedConfig, _) =>
+      case NoReportInInterval(expectedConfig, expirationTime) =>
         ComplianceDebugLogger.node(nodeId).trace(s"Node didn't received reports recently, status depend of the compliance mode and previous report status")
         buildRuleNodeStatusReport(
             //these reports don't really expires - without change, it will
             //always be the same.
-            MergeInfo(nodeId, None, Some(expectedConfig.nodeConfigId), END_OF_TIME)
+            MergeInfo(nodeId, None, Some(expectedConfig.nodeConfigId), expirationTime)
```
fanf, line 754: that seems to contradict the comment just above, what changed in the logic that implies that change in behaviour?

ncharles: ha, indeed the comment is wrong

ncharles:
PR updated with a new commit

fanf (approved):

fanf:
OK, squash merging this PR

PR #4001, opened 2021-11-22
Fixes #20278: Error when creating a rule in 7.0

VinceMacBuche:
https://issues.rudder.io/issues/20278

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
@@ -739,9 +739,16 @@ class RuleApiService14 (
                   , params.reason
                 )
       id       <- workflow.startWorkflow(cr, actor, params.reason)
+      directiveLib <- readDirectives.getFullDirectiveLibrary().toBox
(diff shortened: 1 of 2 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -739,9 +739,16 @@ class RuleApiService14 (
                   , params.reason
                 )
       id       <- workflow.startWorkflow(cr, actor, params.reason)
+      directiveLib <- readDirectives.getFullDirectiveLibrary().toBox
+      groupLib <- readGroup.getFullGroupLibrary().toBox
+      nodesLib <- readNodes.getAll().toBox
+      globalMode  <- getGlobalPolicyMode().toBox
+
     } yield {
+      val status = getRuleApplicationStatus(change.newRule, groupLib, directiveLib, nodesLib, globalMode)
```
fanf, line 748: we should really keep the rule application status somewhere in cache, it does not make any sens to compute it each time when we can track each things that can make it change. 

Perhaps a first step is to extract a method that does the status just from the ruleid so that it is easier to centralize a cache (not for now, but for 7.1)

fanf (approved):

PR #4031, opened 2021-11-30
Fixes #20342: Add a worst-case, weighted 1 for block compliance mode

fanf:
https://issues.rudder.io/issues/20342

Implemented with a sub-select for worst case: 

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
@@ -444,30 +444,49 @@ object DisplayPriority {
   }
 }
 
+/*
+ * Compliance reporting logic, especially for block components. We have 4 kind of computation available:
+ * - `focus`:
+ *    reporting will be the exact same as the compliance of one of the direct subelements.
+ *    If the element is missing, and error will be reported, weighted 1.
+ * - `weighted`
+ *    compliance value will be the weighted contribution of all direct sub-elements
+ * - `worst case (weight = 1)`
+ *    compliance value will have the same level as the worst case, and will have weight 1
+ * - `worst case (weight = sum all)`
+ *    compliance value will have the same level as the worst case, and will have weight equals
+ *    to the sum of weight of all sub components.
+ */
 sealed  trait ReportingLogic {
   def value : String
 }
 
 object ReportingLogic {
 
-  final case object WorstReport extends ReportingLogic {
-    val value = "worst"
-  }
-  final case object SumReport extends ReportingLogic {
-    val value = "sum"
+  object FocusReport {
+    val key = "focus"
   }
   final case class FocusReport(component : String) extends ReportingLogic {
     val value = s"${FocusReport.key}:${component}"
   }
-  object FocusReport {
-    val key = "focus"
+  final case object WorstReportWeightedOne extends ReportingLogic {
+    val value = "worst-case-weighted-one"
   }
-  def apply(value : String) : PureResult[ReportingLogic] = {
-    value match {
-      case WorstReport.value => Right(WorstReport)
-      case SumReport.value => Right(SumReport)
-      case s"${FocusReport.key}:${a}" => Right(FocusReport(a))
-      case _ => Left(Unexpected(s"Value '${value}' is not a valid reporting composition rule."))
+  final case object WorstReportWeightedSum extends ReportingLogic {
+    val value = "worst-case-weighted-sum"
+  }
+  final case object WeightedReport extends ReportingLogic {
+    val value = "weighted"
+  }
+  def parse(value : String, defaultFocusKey: String = "") : PureResult[ReportingLogic] = {
+    value.toLowerCase match {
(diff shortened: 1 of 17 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -444,30 +444,49 @@ object DisplayPriority {
...
-    val value = "sum"
+  object FocusReport {
+    val key = "focus"
   }
   final case class FocusReport(component : String) extends ReportingLogic {
     val value = s"${FocusReport.key}:${component}"
   }
-  object FocusReport {
-    val key = "focus"
+  final case object WorstReportWeightedOne extends ReportingLogic {
```
VinceMacBuche, line 472: I think it should have been like focus report

final case class WorstReport ( weight : ReportWeight )  {
  val value = "worst:${weight.value}
}
sealed trait ReportWeight
final case object OneReport { val value = "one" }
final case object AllReports { val value = "all" }

webapp/sources/rudder/rudder-web/src/main/elm/editor/sources/ViewBlock.elm
```diff
@@ -207,14 +205,15 @@ showBlockTab model parentId block uiInfo techniqueUi =
       let
         compositionText  = (\reportingLogic ->
                                case reportingLogic of
-                                 WorstReport -> "Worst report"
-                                 SumReport -> "Sum of reports"
+                                 WorstReportWeightedSum -> "Worst report (weighted sum)"
```
VinceMacBuche, line 208: Should only be one entry with Worst report, and a second select to chose the weight, default all

fanf: That would make two clicks, which is worse than one no? Or a "select with sub-select opening on mouseover the choice" ? (but I don't know how to do that)

VinceMacBuche: The cost of understanding what the two option means and in what they differ is costier to me

VinceMacBuche: To me it's easier to understand if there is less option, and a to think about the wieght after

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
fanf: I think the case are different: for focus, we don't know at compile time what will be the cases, so we should not use the same syntax (if somewhere latter, we want to make it configurable, we will need to change logic at several place to adapt to the new variability, it makes sens to update api format at that time).

cc @amousset what do you thing (and impact on rudder lang) ?

If we go for the proposed syntaxt, it should not be `worst:one` and `worst:all` because it makes it hard to understand out of context. I propose to keep: `worst:weighted-one` and `worst:weighted-sum`, so we would have:

```
- weighted
- focus:sub-component-id
- worst:weighted-one
- worst:weighted-sum
```

I still can add an trait level for WorstReport/Report, there's one place where we do something for worse case.

amousset: I don't have strong opinion about this one, and it does not really have impact on the language, it's just a metadata. I think that the weight and the outcome source are two different axes. We could image a focus method with a weight of 10.

Here we are addressing the most likely cases with "source+weight" bundle that makes most sense:

* weighted outcome + sum weight = weighted
* focus outcome + sum weight = focus
* worst outcome + 1 weight = worst-case-weighted-one
* worst outcome  + sum weight = worst-case-weighted-sum

So I don't really think it makes sense to treat the weight as a property of the "worst" case.

VinceMacBuche: And on the UI side, whould you rather have 4 options ? or Three (one worst + a second select for weight )

I don't think a weighted outcome could have a different weight than a sum weight. And in focus outcome, in weight 1 you would have to chose on which report (there may be several reports matching the component, choose the worst one ? you can't have a mean in this case)

VinceMacBuche: And btw, if you think both this should be the way, then let's go for this structure.

fanf:
summing up everything:
- we keep 4 states at the same level: 
``` 
    weighted outcome + sum weight = weighted
    focus outcome + sum weight = focus
    worst outcome + 1 weight = worst-case-weighted-one
    worst outcome + sum weight = worst-case-weighted-sum
```

For the UI part, we only show 3 options in first select (`weighted`, `focus`, `worst`) and a second select for `focus` (select focus on what) and for `worst` (weight of worst is 1 or is the sum of components)

VinceMacBuche (approved):

PR #4017, opened 2021-11-24
Fixes #20310: Method getByRulesCompliance used by API is highly inefficient

ncharles:
https://issues.rudder.io/issues/20310

PR #4033, opened 2021-11-30
Fixes #20302: Repair compliance display in Rules page

VinceMacBuche:
https://issues.rudder.io/issues/20302

* rule compliance was not aware of blocks, so client side was failing on that case => added all the decoding logic for blocks in elm, + rewrite lots of structures to be able to deal with that.

* compliance for nodes from rule was done in client side, but was not always consistant with backend => done in backend

* rule list was abysmally slow => first call for API only get minimal info so that the list is quick, and compliance details are loaded when the user click on rule details

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
@@ -83,7 +83,9 @@ final case class ByRuleRuleCompliance(
   , compliance     : ComplianceLevel
   , mode           : ComplianceModeName
   , directives     : Seq[ByRuleDirectiveCompliance]
-)
+) {
+  lazy val nodes = ByRuleByNodeCompliance.fromDirective(directives).toSeq
+}
 
 final case class ByRuleDirectiveCompliance(
     id        : DirectiveId
@@ -91,6 +93,7 @@ final case class ByRuleDirectiveCompliance(
   , compliance: ComplianceLevel
   , components: Seq[ByRuleComponentCompliance]
 )
+
 
 sealed trait ByRuleComponentCompliance {
   def name : String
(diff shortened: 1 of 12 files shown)


VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -112,9 +115,84 @@ final case class ByRuleValueCompliance(
...
+final case class ByRuleByNodeByDirectiveByValueCompliance(
+  name      : String
+  , compliance: ComplianceLevel
+  , values     : Seq[ComponentValueStatusReport]
+) extends ByRuleByNodeByDirectiveByComponentCompliance
+
+object ByRuleByNodeCompliance {
+  def fromDirective(directives : Seq[ByRuleDirectiveCompliance]) = {
+    final case class TmpStruct (directiveId: DirectiveId, blocks: List[String], component:  String, node : ByRuleNodeCompliance)
+    def recurseComponent( byRuleComponentCompliance: ByRuleComponentCompliance): Seq[(NodeId,ByRuleByNodeByDirectiveByComponentCompliance)] = {
```
ncharles, line 156: a bit of comments here would be much appreciated when we'll have to delve into it in the future

VinceMacBuche: Yes I will add them after I tried this in real life, it compiles, and the logic in my head is ok, but I need to confront it in real life (the PR is still a wip)

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -83,7 +83,9 @@ final case class ByRuleRuleCompliance(
   , compliance     : ComplianceLevel
   , mode           : ComplianceModeName
   , directives     : Seq[ByRuleDirectiveCompliance]
-)
+) {
+  val nodes = ByRuleByNodeCompliance.fromDirective(directives).toSeq
```
ncharles, line 87:
```suggestion
  lazy val nodes = ByRuleByNodeCompliance.fromDirective(directives).toSeq
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -313,6 +313,7 @@ class ComplianceAPIService(
                   ByRuleNodeCompliance(
                     nodeId
                     , nodeInfos.get(nodeId).map(_.hostname).getOrElse("Unknown node")
+                    , ComplianceLevel.sum(uniqueComponents.find(_._1 == nodeId).map(_._2.compliance).toList)
```
ncharles, line 316: why not using the components ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -313,6 +313,7 @@ class ComplianceAPIService(
...
                     , uniqueComponents.toSeq.sortBy(_._2.componentName).flatMap(_._2.componentValues.values)
```
ncharles, line 317: components

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

ncharles:
i've been testing this PR, and unfortunately, i still don't have compliance on the rule page:(

ncharles:
by backporting https://github.com/Normation/rudder/pull/4017 in it, i get result on the page, but it is really slow

Biggest perf offenders:
  | https://192.168.43.2/rudder/secure/api/groups/tree : 3,7s waiting, 3,7s receiving (24MB) : for each groups there are all nodes included. We 10k nodes, that's a lot of data. I'm unsure why we would need that when on the rule list page
  | https://192.168.43.2/rudder/secure/api/rules/tree : 12,46s waiting, 11ms receiving (471kb); a lot of details about the rules, but no idea why it's so slow
  | https://192.168.43.2/rudder/secure/api/nodes : 4.2s waiting, 290ms receiving (10.57MB); a lot of details about the nodes, unsure why we would need that. I'm unsure why we would need that when on the rule list page
  | https://192.168.43.2/rudder/secure/compliance/rules?level=1 : 5.5s waiting, 0ms receiving (20kB)

VinceMacBuche:
Commit modified

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
ncharles:
```suggestion
                    , ComplianceLevel.sum(components.find(_._1 == nodeId).map(_._2.compliance).toList)
```

webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewRulesTable.elm
```diff
@@ -48,12 +48,13 @@ getSortFunction model r1 r2 =
                   else
                     c.compliance
               Nothing -> -2.0
-          r1Compliance = getCompliance (getRuleCompliance model r1.id)
-          r2Compliance = getCompliance (getRuleCompliance model r2.id)
+          --r1Compliance = getCompliance (getRuleCompliance model r1.id)
+          --r2Compliance = getCompliance (getRuleCompliance model r2.id)
```
ncharles, line 52: can you remove the commented code?

ncharles (changes requested):
I don't understand it all, but there is at least the compliance computation that as an error (see suggestion)
Also, i don't see why the nodelist is loaded when arriving on the rule page and not on a rule detail ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -336,6 +336,7 @@ class ComplianceAPIService(
                     ByRuleNodeCompliance(
                       nodeId
                       , nodeInfos.get(nodeId).map(_.hostname).getOrElse("Unknown node")
+                      , ComplianceLevel.sum(components.find(_._1 == nodeId).map(_._2.compliance).toList)
```
ncharles, line 339: i'm surprised by this line
the .find().map() returns an option. the toList hides it

fanf: if there is 0 component (none found), it seems correct to return "n/a". 
But I'm unsure why we can't have several component here. 
And it could be done in one pass with `collectFirst { case (id, v) if(id == nodeId) => v.compliance }`

OK, it was a map before being "toSeq"

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -336,6 +336,7 @@ class ComplianceAPIService(
                     ByRuleNodeCompliance(
                       nodeId
```
fanf, line 337: All the part in the yield is getting quite complicated (and where the actual business logic lies). 

Can you create a pure function that takes in argument all the IO parts obtained in the for loop and with the logic, so that it can be tested in isolation ?

Actually: I fear that just integration tests on these API will be hard, but perhaps my fear is misplaced and it's better to start with test at that level, even if harder to get right, they will be more meaningfull (and we need them in all cases)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
fanf: is it time, now :)

fanf (commented):
Can you:
- add a description in the pr header to explain the big changes, main implementation choices, and what was touchy or hard (a good start would be to tell us what was broken and how it's repaired here)
- tests, because, well we need them on API, but also because it would be easier to get onbouard with sample of exptected/rejected behavior

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #4035, opened 2021-12-01
Fixes #20350: gitgc is always run at start whatever its schedule is and run every 3min42

fanf:
https://issues.rudder.io/issues/20350

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/GitGC.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/GitGC.scala
@@ -104,8 +104,8 @@ class GitGC(
       logger.info(s"Disable automatic git-gc on ${gitRepo.rootDirectory.name} (schedule: '${DISABLED}')")
     case Some(cron) =>
       val schedule = cron.toSchedule
-      logger.info(s"Automatic git-gc starts on ${gitRepo.rootDirectory.name} (schedule: '${schedule.toString}')") *>
-      gitgc.repeat(schedule).unit
+      logger.info(s"Automatic git-gc starts on ${gitRepo.rootDirectory.name} (schedule 'sec min h dayMonth month DayWeek': '${cron.toString}')") *>
+      gitgc.schedule(schedule).unit
   }
 
   // start cron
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/GitGC.scala
```diff
@@ -104,8 +104,8 @@ class GitGC(
       logger.info(s"Disable automatic git-gc on ${gitRepo.rootDirectory.name} (schedule: '${DISABLED}')")
     case Some(cron) =>
       val schedule = cron.toSchedule
-      logger.info(s"Automatic git-gc starts on ${gitRepo.rootDirectory.name} (schedule: '${schedule.toString}')") *>
-      gitgc.repeat(schedule).unit
+      logger.info(s"Automatic git-gc starts on ${gitRepo.rootDirectory.name} (schedule 'sec min h dayMonth month DayWeek': '${cron.toString}')") *>
```
fanf, line 107: display the actual cron schedule, not a class instance name

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/GitGC.scala
```diff
@@ -104,8 +104,8 @@ class GitGC(
...
+      gitgc.schedule(schedule).unit
```
fanf, line 108: `repeat` semantic is `do that, then do it again on schedule`, while we want `do that on schedule`

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -392,7 +392,11 @@ object RudderConfig extends Loggable {
     config.getString("rudder.git.gc")
   } catch {
     // missing key, perhaps due to migration, use default
-    case ex: Exception => "0 42 3 * * ?"
+    case ex: Exception => {
+      val default = "0 42 3 * * ?"
+      logger.info(s"`rudder.git.gc` property is missing, using default schedule: ${default}")
```
fanf, line 397: let user be able to discover that he can change git gc schedule with that property

ncharles (approved):

PR #4036, opened 2021-12-01
Fixes #20351: System technique migration fails on old server

fanf:
https://issues.rudder.io/issues/20351

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/TechniqueAcceptationDatetimeUpdater.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/TechniqueAcceptationDatetimeUpdater.scala
@@ -243,9 +243,15 @@ class TechniqueAcceptationUpdater(
                                 // If an active technique, not system, still exists for that technique, disable it.
                                 // In the case of a system one, something is very broken. Don't disable it, as
                                 // It is likely to worsen things, but log an error.
+                                // Avoid writting the log for 'server-roles' and 'distributePolicy' - they are removed in 7.0 and up
                                 if(activeTechnique.isSystem) {
-                                  ApplicationLoggerPure.error(s"System technique '${name}' (${versions.map(_.debugString).mkString(",")})' is deleted in" +
-                                                             s"git base. This will likely cause grave problem. You should investigate.")
+                                  val removedIn7_0 = List("server-roles", "distributePolicy")
+                                  if(removedIn7_0.contains(name)) {
+                                    ApplicationLoggerPure.info(s"System technique '${name}' removed from git base according to Rudder 7.0 update")
+                                  } else {
+                                    ApplicationLoggerPure.error(s"System technique '${name}' (${versions.map(_.debugString).mkString(",")})' is deleted in " +
+                                                                s"git base. This will likely cause grave problem. You should investigate.")
+                                  }
 
                                 } else {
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
```diff
@@ -511,7 +511,8 @@ object PolicyServerConfigurationObjects {
     )
   }
 
-  val relayTechniques = List("server-common", "rudder-service-apache","rudder-service-relayd")
+  // we need to have "common" here, even if it's not server specific, else it leads to it being deleted during migration
+  val relayTechniques = List("common", "server-common", "rudder-service-apache","rudder-service-relayd")
```
fanf, line 515: this *might* be the cause of `common: deleted`, but I have no reproducible case for it.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckMigrateSystemTechniques7_0.scala
```diff
@@ -164,11 +167,11 @@ object SystemConfig {
 
   object v6_2 {
     final case class DirectiveCommon(policyServerId: NodeId, at: AT) {
-      val id = s"common-${policyServerId.value}"
+      val id = s"${at.id}-${policyServerId.value}"
       val dn = new DN(s"directiveId=${id},${at.dn.toString}")
     }
     final case class DirectiveDistributePolicy(policyServerId: NodeId, at: AT) {
-      val id = s"common-${policyServerId.value}"
+      val id = s"${at.id}-${policyServerId.value}"
```
fanf, line 174: this was the cause of the fatal error - missing required field

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
ncharles: adding common here breaks migration: it will be used in the rulePolicyServer constructor, and directiveId common-${nodeId.value} , which does not exists, so it breaks

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/TechniqueAcceptationDatetimeUpdater.scala
```diff
@@ -243,9 +243,15 @@ class TechniqueAcceptationUpdater(
                                 // If an active technique, not system, still exists for that technique, disable it.
                                 // In the case of a system one, something is very broken. Don't disable it, as
                                 // It is likely to worsen things, but log an error.
+                                // Avoid writting the log for 'server-roles' and 'distributePolicy' - they are removed in 7.0 and up
                                 if(activeTechnique.isSystem) {
-                                  ApplicationLoggerPure.error(s"System technique '${name}' (${versions.map(_.debugString).mkString(",")})' is deleted in" +
-                                                             s"git base. This will likely cause grave problem. You should investigate.")
+                                  val removedIn7_0 = List("server-roles", "distributePolicy")
+                                  if(removedIn7_0.contains(name)) {
```
ncharles, line 249: name is 'server-roles' (1.0)

ncharles: but i can't see why

PR #4038, opened 2021-12-02
Fixes #20359: techniques are seens as deleted even if there are there in Rudder 7.0

ncharles:
Fixes #20359: techniques are seens as deleted even if there are there in Rudder 7.0

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
@@ -511,8 +511,7 @@ object PolicyServerConfigurationObjects {
     )
   }
 
-  // we need to have "common" here, even if it's not server specific, else it leads to it being deleted during migration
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
```diff
@@ -512,7 +512,7 @@ object PolicyServerConfigurationObjects {
   }
 
   // we need to have "common" here, even if it's not server specific, else it leads to it being deleted during migration
```
amousset, line 514: ,could you also edit the comment

ncharles:
PR updated with a new commit

fanf:
this line was added to correct https://issues.rudder.io/issues/20351 so that will just revert the problm to previous bad case.

fanf:
let me look and find the root problem

fanf:
so ok, that's actually the correct correction. I added tests to check it always in PR: https://github.com/ncharles/rudder/pull/11

fanf (approved):

fanf:
OK, squash merging this PR

PR #4044, opened 2021-12-09
Fixes #20394: Remove metrics section from wizard

ElaadF:
https://issues.rudder.io/issues/20394
Before:
image
After:
image


ElaadF:
Note that the logic added for te metrics remain, I just remove the entry of the section

amousset (approved):

PR #4045, opened 2021-12-09
Fixes #20390: webapp doesn't start after upgrade from 6.2 to 7.0

ncharles:
https://issues.rudder.io/issues/20390


fanf:
I think we should address the problem in an other way: 
- by default, we should forbid system technique deletion (as it was previously to https://issues.rudder.io/issues/20091)
- with an exception for the two system techniques known to be deleted (`server-roles` and `distributePolicy`).

This should remove most problem related to random restarts

fanf:
By "forbid", I mean "not delete active technique when we get that git notification, as it is likely false"

fanf (approved):

PR #4058, opened 2021-12-15
Fixes #20440: Bring back recent changes to life.

VinceMacBuche:
https://issues.rudder.io/issues/20440

This PR add a minimal version of recent changes: 

- in rule list, it adds the number of recent changes in the last 6 hours for each rule
- in rule details, it adds a new tab for recent changes. The tab allows to choose among the last 12 interval of 6 hours (most recent chosen by default) and display the list of changes for that interval below. 

In the backend, it adds two *internal only* API endpoints: 

- `GET /changes`: get the number of changes for all rules in the last 6 hours, 
- `GET /changes/{ruleId}`: get the list of all recent changes for one rule (for the 12 most recent intervals of 6 hours)

In the front end, there is a new data type for recent changes, `RepairedReport`, and corresponding JSON decoder (for the number of changes by rule, there is no dedicated type). 

The code for the new recent changes tab is in `ViewRepairedReports.elm` (it's just a table for now).
`ViewTabContent.elm` gets a lot of changes, but mostly indent: look at it with `?w=1`: I added function for each tabs, so that logic is more cleanly separated (and of course the new tab).

In `ViewRulesTable.elm`, it adds the number of changes for each rule. 
It also correct a bug in compliance sorting.

Peek 2022-01-03 17-11

The code also normalize the format of URL in elm API call with an array format, from:
```
         , url             = getUrl model "/compliance/rules?level=1"
```

To:
```
         , url     = getUrl model [ "compliance", "rules"] [ int "level" 1 ]
```


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RecentChangesAPI.scala
```diff
@@ -0,0 +1,157 @@
...
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.DirectiveUid
+import zio.syntax.ToZio
+
+import scala.util.Random
+
+class RecentChangesAPI (
+   nodeChangesService: NodeChangesService
+) extends RestHelper with Loggable {
```
fanf, line 65: Can you use the endpoint definition that we usually use, not the liftweb specific one? (I understand it was for a quick test, but it needs to be changed for the final PR)

VinceMacBuche: We don't use our endrpoint definition for internal and it can't be integrated this way

fanf: yes we use it for internal API, see for example: `NodeDisplayInheritedProperties` in `NodeApi`. Perhaps we didn't ported all existing internal API the new (sort of) schema, but it should be done, at least for new ones. 

I don't see why we couldn't integrate that API this way, what is the technical blocking point ?

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):

PR #4066, opened 2021-12-21
Fixes #20462: Delay policy generation until rudder app is fully boot

fanf:
https://issues.rudder.io/issues/20462

That PR add a promise (ie something that can be done exactly once, and that other things can wait for it to be done) that is checked for completion before actually starting policy generation.

The promise is completed at end of `Boot.boot` method, ie once lift knows rudder is started.

To be able to use the `promise.await` in policy generation agent, I needed to transform the code into a ZIO effect.

Also remove a warning for missing pattern match which would lead to a runtime exception, likely introduced in pr #4065

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
@@ -43,23 +43,29 @@ import net.liftweb.actor._
 import org.joda.time._
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.services.policies.PromiseGenerationService
+
 import net.liftweb.http.ListenerManager
 import com.normation.eventlog.{EventActor, EventLog}
 import com.normation.rudder.domain.eventlog._
 import com.normation.rudder.services.marshalling.DeploymentStatusSerialisation
 import com.normation.rudder.services.eventlog.EventLogDeploymentService
+
 import net.liftweb.common._
 import com.normation.eventlog.EventLogDetails
 
 import scala.xml.NodeSeq
 import com.normation.eventlog.ModificationId
 
-import com.normation.errors.RudderError
 import com.normation.errors.Unexpected
 import com.normation.rudder.domain.logger.PolicyGenerationLogger
-import com.normation.zio._
+import com.normation.rudder.domain.logger.PolicyGenerationLoggerPure
 
+import com.normation.zio._
 import scala.concurrent.duration.Duration
+
+import zio._
+import zio.syntax._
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -381,39 +388,42 @@ final class AsyncDeploymentActor(
...
+        )
+      }
+
+      bootSemaphore.await *> managedErr
+    }
+
     override protected def messageHandler = {
       //
-      // Start a new deployment
+      // Start a new deployment. Wait for the
```
ncharles, line 423: the comment here is not complete

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -554,6 +554,10 @@ class Boot extends Loggable {
         ApplicationLogger.error(s"Error when trying to save the EventLog for application start: ${err.fullMsg}")
       case Right(_)  => ApplicationLogger.info("Application Rudder started")
     }
+
+    // release guard for promise generation awaiting end of boot
+    RudderConfig.policyGenerationBootGuard.succeed(()).runNow
```
ncharles, line 559: I would never have thought to put it there, i was more thinking to put it at the end of bootcheck. It is without doubt better the way you did

ncharles (changes requested):
You need to rebase the test, and correct a comment
The rest is :100:

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/NodeGroupRepository.scala
```diff
@@ -190,6 +190,7 @@ final case class FullNodeGroupCategory(
       case FullOtherTarget(t) => t match {
         case AllTarget => true
         case AllTargetExceptPolicyServers => !node.isPolicyServer
+        case AllPolicyServers => node.isPolicyServer
```
ncharles, line 193: this file should not change

fanf: it removes a warning missing pattern match (and the corresponding runtime exception) for a previous pr regarding migration. 
I'm adding a comment in head.

ncharles: ok

ncharles (approved):

PR #4077, opened 2021-12-29
Fixes #20474: error on system status tabs when there are missing reports

ncharles:
https://issues.rudder.io/issues/20474


ncharles:
Commit modified

ElaadF (approved):

PR #4086, opened 2022-01-05
Fixes #20513: Sometimes the webapp fails to generate the reporting file of new techniques

amousset:
https://issues.rudder.io/issues/20513

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -181,39 +181,13 @@ class RudderCRunner(
     } yield ()
   }
 
-  // Create an empty reporting bundle for compatibility with the fallback process
-  // Can be removed once rudderc is the only generation method
-  def emptyReportingFile(technique : EditorTechnique) = {
-    val bundleParams = if (technique.parameters.nonEmpty) technique.parameters.map(_.name.canonify).mkString("(",",",")") else ""
-    val content =
-      s"""bundle agent ${technique.bundleName.value}_rudder_reporting${bundleParams}
-         |{
-         |}"""
-    val reportingFile = File(outputPath) / "techniques"/ technique.category / technique.bundleName.value / technique.version.value / "rudder_reporting.cf"
-
-    for {
-      _ <- TechniqueWriterLoggerPure.debug(s"Creating empty reporting file for target '${technique.name}' in path ${reportingFile.path.toString}")
-      _ <- IOResult.effect(s"Could not write empty reporting Technique file '${technique.name}' in path ${reportingFile.path.toString}") {
-             reportingFile.createFileIfNotExists(true).write(content.stripMargin('|'))
-           }
-    } yield {
-      ()
-    }
-  }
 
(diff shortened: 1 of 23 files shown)


amousset:
This does not work

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -181,12 +181,40 @@ class RudderCRunner(
...
+         |}"""
+
+    val reportingFile = File(outputPath) / "rudder_reporting.cf"
+
+
+    TechniqueWriterLoggerPure.error(s"PLOUF target '${technique.name}' in path ${reportingFile.path.toString}")
+
+    IOResult.effect(s"Could not write empty reporting Technique file '${technique.name}' in path ${reportingFile.path.toString}") {
+      reportingFile.createFileIfNotExists(true).write(content.stripMargin('|'))
+      Seq(File(outputPath).relativize(reportingFile.path).toString)
```
fanf, line 201: this is not used, can be removed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -181,12 +181,40 @@ class RudderCRunner(
...
+    val bundleParams = if (technique.parameters.nonEmpty) technique.parameters.map(_.name.canonify).mkString("(",",",")") else ""
+    val content =
+      s"""bundle agent ${technique.bundleName.value}_rudder_reporting${bundleParams}
+         |{
+         |}"""
+
+    val reportingFile = File(outputPath) / "rudder_reporting.cf"
+
+
+    TechniqueWriterLoggerPure.error(s"PLOUF target '${technique.name}' in path ${reportingFile.path.toString}")
```
fanf, line 197: as that instuction is a pure value, it won't product side effect. If you want (you want) to include it in tha chain of instruction to be interpreted by the `IO` interpreter, you need to bundle it with the following one (which is returned) by adding a `*>` (which means "and pack it to next instruction, I don't care of its returned value"). 
You can also do: 
```
for {
  _ <- TechniqueWriterLoggerPure.error(s"PLOUF target '${technique.name}' in path ${reportingFile.path.toString}")
 _ <- IOResult.effect(s"Could....) { ... }
} yield () 
``` 
which would allows to use intermediate returned values if you needed them.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -181,12 +181,40 @@ class RudderCRunner(
...
+
+    IOResult.effect(s"Could not write empty reporting Technique file '${technique.name}' in path ${reportingFile.path.toString}") {
+      reportingFile.createFileIfNotExists(true).write(content.stripMargin('|'))
```
fanf, line 200: by default, `stripMargin` uses `|` , you can remove it (but it's ok to keep it too)

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

VinceMacBuche (approved):

amousset:
OK, squash merging this PR

PR #4092, opened 2022-01-07
Fixes #20535: We are recreating ldap object while we could duplicate them, leading to suboptimal perf

ncharles:
https://issues.rudder.io/issues/20535

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -513,7 +513,10 @@ class InternalLDAPQueryProcessor(
                       res
                     }
                   }
-      postFiltered = postFilterNode(inverted.distinctBy(_.dn), query.returnType, limitToNodeIds)
+      // distinctBy computes the hashcode of the object
+      // It is really expensive on LDAP entries.
+      // The dn string is already computed, so the toString is a good alternative (and as also unicity)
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -513,7 +513,10 @@ class InternalLDAPQueryProcessor(
                       res
                     }
                   }
-      postFiltered = postFilterNode(inverted.distinctBy(_.dn), query.returnType, limitToNodeIds)
+      // distinctBy computes the hashcode of the object
+      // It is really expensive on LDAP entries.
+      // The dn string is already computed, so the toString is a good alternative (and as also unicity)
```
fanf, line 518: has also

ncharles:
This changes saves ~2 seconds on group computation, going from 1m25-1m27 to 1m22-1m25, on load test platform on 7.0

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
fanf: ok, so the cost is because `.dn.hashcode` does `.dn.toNormalizeString.hashcode` and :
- we have normalize string by construction in rudder 
- but `toNormalizeString` is extremely costly: it goes through each `rdn`, check with schema, normalize it, and then hashcode 
- and we already have the string dn (always present in ldap entries)

fanf (approved):

PR #4098, opened 2022-01-11
Fixes #20373: compliance from API doesn't always adds up to 100%

fanf:
https://issues.rudder.io/issues/20373

I was not able to reproduce the mentionned cases in rudder (99.99% missing) but I *believe* it was something around 3 components having each 1 missing, and so each contributed 33.33% percent. 
Nonetheless, I was able to make a failing unit test with a compliance with 1 success, 1 na, 1 error that sum to 99.99% corrected by that PR, so I believe that that PR should also correct the composition since AFAIK (and found) compliance percent are not coupounded/summed, it's always the compliane level (which is good). 

So, this PR centralize all computing of percent in the `CompliancePercent` class, and in particular it adds a `CompliancePercent.fromLevels` method that build percents with a sum of 100 by construction. It also remove any percent computing from `ComplianceLevel`.

By construction, it also ensure that a compliance is alway AT LEAST 0.01%, even if it should be 0.0000001% and rounded to 0 to avoid having https://issues.rudder.io/issues/10773 coming back.

With that, there should be no other part in Rudder that computes compliance, and compliance should always be obtained from backend. 

The logic to sum to 100 is that only the "biggest" compliance rounding and the one of small compliance below threshold are adapted, and for that, we do: 

- biggest is defined by "with the more level, or in case of equality, the WORSE kind of report according to `ReportType` sorting"
- sort compliance level by number
- when total is 0 or that level is 0, return 0 (globaly or for that level)
- for all but the biggest, compute compliance, keep two digits, use min threshold if needed, 
- for the biggest, use "100 - sum of others"

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
@@ -37,6 +37,8 @@
 
 package com.normation.rudder.domain.reports
 
+import com.normation.rudder.domain.logger.ComplianceLogger
+
 import net.liftweb.http.js.JE
 import net.liftweb.http.js.JE.JsArray
 import net.liftweb.json.JsonAST.JObject
@@ -44,16 +46,32 @@ import net.liftweb.json.JsonAST.JValue
 import net.liftweb.json.JsonAST.JInt
 
 /**
- * That file define a "compliance level" object, which store
- * all the kind of reports we can get and compute percentage
- * on them.
+ * That file define a "compliance level" object, which store all the kind of reports we can get and
+ * compute percentage on them.
  *
- * This file also define simple addition on such compliance level.
+ * Percent are stored a double from 0 to 100 with two relevant digits so 12.34% is actually stored
+ * as Double(12.34) (not 0.1234)
+ *
+ * Since use only two digits in percent, we only have a 10e-4 precision, but we nonetheless NEVER EVER
+ * want to make 1 error among 20000 success reports disapear by being rounded to 0.
+ * So we accept that we have a minimum for all percent, `MIN_PC`, and whatever the real number, it
+ * will be returned as that minimum.
+ *
+ * For precise computation, you of course MUST use the level numbers, not the percent.
+ * It also mean that any transformation or computation on compliance must alway be done on level, never
+ * compliance percent, which are just a model for human convenience, but is false.
+ *
+ * This class should always be instanciated with  CompliancePercent.fromLevels` to ensure sum is 100%
+ * The rounding algorithm is:
+ * - sort levels by number of reports, less first, sum them to get total number of reports
+ * - for each level except the last one (most number):
+ *   - if there is 0 reports for that level, it's 0
+ *   - if the percent is LESS THAN `MIN_PC`, then it's MN
+ *   - else compute percent rounded to two digits
+ * - the last level is computed by: 100 - sum(other percent)
+ *
+ * In case of equality, `ReportType.level` is used for comparison.
  */
-
-
-//simple summary that only store percents - no intelligence at all
-//here, see ComplianceLevel to understand how percent are calculated.
 final case class CompliancePercent(
     pending           : Double = 0
   , success           : Double = 0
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -69,7 +74,114 @@ final case class CompliancePercent(
...
+      , c.auditNotApplicable
+      , c.nonCompliant
+      , c.auditError
+      , c.badPolicyMode
+    ).zipWithIndex
+
+    // sort smallest first, and then by worst type
+    levels.sortWith { case ((l1, i1), (l2, i2)) =>
+      if(l1 == l2)  {
+        WORSE_ORDER.getOrElse(i1, WORSE_ORDER.size) > WORSE_ORDER.getOrElse(i2, WORSE_ORDER.size)
```
ncharles, line 163: If levels were created with compliance levels from worse to best, and then zipWithIndexed, you could simply use
i1 > i2


this methods would be much easier, and skip the whole map key search

ncharles (changes requested):
This is an interesting change, however I'm worried about the maintenaibility of this.
I suggested a way to make it both faster and easier to read, what do you think?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -69,7 +74,114 @@ final case class CompliancePercent(
...
+   * Rules are:
+   * - always keeps at least 0.01% for any case
+   *   (given that we have 14 categories, it means that at worst, if 13 are at 0.1 in place of ~0%;
+   *   the last one will be false by 0.13%, which is ok)
+   * - the biggest value is rounded to compensate
+   * - in case of equality (ex: 1/3 success, 1/3 NA, 1/3 error), the biggest is
+   *   chosen accordingly to "getWorsteType" logic.
+   */
+  def fromLevels(c: ComplianceLevel): CompliancePercent = {
+    val MINPC = BigDecimal(0.01) // 0.01% since we keep two digits in percents
```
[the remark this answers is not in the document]
fanf: I think the compiler is smart and does that but I could, yes

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
fanf: I'm not sure I understand. Do you propose to change ComplianceLevel properties order? If not, can you explain again? If yes, I thought to that, but that order is used as an hypothesis for a lot of things, from test to backend to js, and I fear consequence would be far reaching. Plus, extension would mean changing relative order each time, so if we want to do that, we need to make the change index independent (ie forbids things like `ComplianceLevel(0,1,0)`, and it's  implicated with case classes)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -69,7 +74,114 @@ final case class CompliancePercent(
...
+   * returned a list of (level, index) where index correspond to the index of the
+   * level in the compliance array. The list is sorted in the relevant order so that
+   * the smaller level are at head, and in case of equality, the "worse order" order is
+   * kept (worse order last)
+   */
+  def sortLevels(c: ComplianceLevel): List[(Int, Int)] = {
+    // we want to compare accordingly to `ReportType.getWorsteType` but I don't see any
+    // way to do it directly since we don't use the same order in compliance.
+    // So we map index of a compliance element to it's worse type order and compare by index
+    val WORSE_ORDER = Seq(13, 4, 3, 12, 11, 2, 5, 6, 8, 0, 1, 9, 10, 7).zipWithIndex.toMap
```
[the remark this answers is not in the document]
fanf: I can put that in report level and use it for the "get worse" algo, it would simplify both classes.
Here we would have name in the seq (Success.index, ehctc)

And we don't need a map, we have the index of the element, we can just access by index

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
ncharles: no, not changing the order of properties, but the order in the list of levels, to have the live easily ordered. 
But now I understand that there is no translation from that to the levels in the case class, so it would only means moving it away.. which may or may not be better

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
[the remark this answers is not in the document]
fanf: I did that, it feels OK to maintain

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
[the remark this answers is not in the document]
fanf: done

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
fanf: OK, it was just displacing the problem. I did my other idea, and it works well.

fanf:
I had forgotten to `git add` the test file :'( 
Now corrected

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -141,21 +269,20 @@ final case class ComplianceLevel(
...
-  private def pc_for(i:Int, total:Int) : Double = if(total == 0) 0 else (i * 100 / BigDecimal(total)).setScale(2, BigDecimal.RoundingMode.HALF_UP).toDouble
 
   def compute(reports: Iterable[ReportType]): ComplianceLevel = {
     import ReportType._
     if(reports.isEmpty) {
       ComplianceLevel(notApplicable = 1)
     } else {
-      var notApplicable = 0
+      var pending = 0
       var success = 0
```
fanf, line 280: keep same order everywhere for maintenance

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/reports/TestComplianceLevel.scala
```diff
@@ -0,0 +1,96 @@
...
+package com.normation.rudder.domain.reports
+
+import org.junit.runner.RunWith
+import org.specs2.mutable.Specification
+import org.specs2.runner.JUnitRunner
+
+
+/**
+ * Test properties about Compliance Level,
+ * especially pourcent related (sum to 100, rounding error, etc)
```
ncharles, line 47:
```suggestion
 * especially percent related (sum to 100, rounding error, etc)
```

ncharles (approved):

amousset:
OK, squash merging this PR

PR #4103, opened 2022-01-12
Fixes #20490: Compliance by nodes in the Rules page shows the node id rather than it's name

VinceMacBuche:
https://issues.rudder.io/issues/20490


fanf (approved):

ncharles (approved):

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -190,7 +190,8 @@ object ByRuleByNodeCompliance {
             (nodeId, data) <- v.nodes.groupBy(_.id)
           } yield {
             // You get all reports that were for a Node matching our value, regroup these report for our node in the structure)
-            (nodeId, ByRuleByNodeByDirectiveByValueCompliance(v.name, ComplianceLevel.sum(data.map(_.compliance)),data.flatMap(_.values)))
+            // Node name should be the same for all nodes, we need to send it to upper structure, link it with id
+            ((nodeId, v.nodes.map(_.name).headOption.getOrElse(nodeId.value)), ByRuleByNodeByDirectiveByValueCompliance(v.name, ComplianceLevel.sum(data.map(_.compliance)),data.flatMap(_.values)))
```
ncharles, line 194: i think it should not be v.nodes.head, but the name corresponding to the nodeId

PR #4110, opened 2022-01-14
Fixes #20573: Compliance of rules should not be rounded to the nearest hundredth

RaphaelGauthier:
https://issues.rudder.io/issues/20573


fanf:
Alternative correction that uses the same rounding logic than in backend to ensure consistence: https://github.com/Normation/rudder/pull/4115

VinceMacBuche (approved):

PR #4115, opened 2022-01-17
Fixes #20573: Compliance of rules should not be rounded to the nearest hundredth

fanf:
https://issues.rudder.io/issues/20573

A pure backend correction of https://issues.rudder.io/issues/20573, should replace https://github.com/Normation/rudder/pull/4110 cc @RaphaelGauthier 

The idea is to be able to specify what precision we want for compliance percent in the API calls, and let the backend compute compliance with that value. Most of the pr is passing the `precision` parameter around. 

Note that we can't specify `precision` value for the  `lazy val pc` attribute in `ComplianceLevel` instance, so a default value of `2` is kept. That `pc` value does not seems to be used in the compliance displaying for UI, so it makes me wonder if we really need an heavy lazy val in each `ComplianceLevel` objects, perhaps we should only compute it when needed with `CompliancePercent.fromLevels(c, precision)` - cc @ncharles for that. 

The 2nd commit totaly remove `lazy val` from `ComplianceLevel` object. It will free 3 object alloc by compliance level object, which is not nothing given the number of short lived ones we have, and it seems that there is not many places where they are reused (and where it can't be computed before the different use place).

### DIFF
--- a/webapp/sources/api-doc/components/parameters/compliance-percent-precision.yml
+++ b/webapp/sources/api-doc/components/parameters/compliance-percent-precision.yml
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2020 Normation SAS
+name: "precision"
+in: query
+schema:
+  type: integer
+  example: 0
+  default: 2
+description: Number of digits after comma in compliance percent figures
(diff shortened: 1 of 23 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
Replaced by #4127

PR #4120, opened 2022-01-18
Fixes #20540: Unexpected reports when using nested blocks

VinceMacBuche:
https://issues.rudder.io/issues/20540

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
@@ -1094,12 +1094,12 @@ class ExecutionBatchTest extends Specification {
       res.compliance === ComplianceLevel(success = 1, unexpected = 2) // 2 unexpected because the whole "foo" becomes unexpected
     }
     "when allow duplicated, duplicate messages is ignored" in {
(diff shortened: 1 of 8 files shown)


webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -1095,11 +1095,11 @@ class ExecutionBatchTest extends Specification {
     }
     "when allow duplicated, duplicate messages is ignored" in {
       val res = ExecutionBatch.checkExpectedComponentWithReports(expectedComponent, duplicated, ReportType.Missing, PolicyMode.Enforce, UnexpectedReportInterpretation(Set(UnexpectedReportBehavior.AllowsDuplicate)))
-      res.compliance === ComplianceLevel(success = 2)
+      res.compliance === ComplianceLevel(success = 1, unexpected = 2)
```
fanf, line 1098: You need to delete that test, and the whole concept of `UnexpectedReportInterpretation(Set(UnexpectedReportBehavior.AllowsDuplicate))` I think, since it doesn't make sense without syslog (or is it also used to control the property for different unexpected, like the one produced by loops ? If so, only remove `UnexpectedReportBehavior.AllowsDuplicate` case)

fanf: Ah yes, it is, with `UnboundVarValues`, so keep `UnexpectedReportBehavior` but remove `AllowsDuplicate`

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #4127, opened 2022-01-19
Fixes #20607: Make possible to choose precision in compliance percent API

fanf:
https://issues.rudder.io/issues/20607

Superseed #4115 and keep #4110 for correction of https://issues.rudder.io/issues/20573

The idea is to be able to specify what precision we want for compliance percent in the API calls, and let the backend compute compliance with that value. Most of the pr is passing the `precision` parameter around. 

Note that we can't specify `precision` value for the  `lazy val pc` attribute in `ComplianceLevel` instance, so a default value of `2` is kept. That `pc` value does not seems to be used in the compliance displaying for UI, so it makes me wonder if we really need an heavy lazy val in each `ComplianceLevel` objects, perhaps we should only compute it when needed with `CompliancePercent.fromLevels(c, precision)` - cc @ncharles for that. 

The 2nd commit totaly remove `lazy val` from `ComplianceLevel` object. It will free 3 object alloc by compliance level object, which is not nothing given the number of short lived ones we have, and it seems that there is not many places where they are reused (and where it can't be computed before the different use place).

### DIFF
--- a/webapp/sources/api-doc/components/parameters/compliance-percent-precision.yml
+++ b/webapp/sources/api-doc/components/parameters/compliance-percent-precision.yml
@@ -1,0 +1,9 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2020 Normation SAS
+name: "precision"
+in: query
+schema:
+  type: integer
+  example: 0
+  default: 2
+description: Number of digits after comma in compliance percent figures
--- a/webapp/sources/api-doc/paths/compliance/global.yml
+++ b/webapp/sources/api-doc/paths/compliance/global.yml
@@ -4,6 +4,8 @@ get:
   summary: Global compliance
   description: Get current global compliance of a Rudder server
   operationId: getGlobalCompliance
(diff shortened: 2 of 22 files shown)


webapp/sources/api-doc/components/parameters/compliance-percent-precision.yml
```diff
@@ -0,0 +1,9 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2020 Normation SAS
+name: "precision"
+in: query
+schema:
+  type: integer
+  example: 0
+  default: 2
+description: Number of digits after comma in compliance percent figures
```
amousset, line 9: we can maybe precise here that a low value will never be zero and that is it not a simple rounding.

ncharles (approved):

ncharles:
This is a nice change. I'm indeed not sure of the impact of the lazy val - i did not notice anything significant in my tests - i'm still unsure if we can do better than that or not, but that will be for a next iteration, as perfs are satisfactory

PR #4138, opened 2022-01-24
Fixes #20638: Reccent changes count on rule tab

ElaadF:
https://issues.rudder.io/issues/20638

Maybe i'm wrong, is the count of recent changes should be the same as "changes"  count displayed in table rows ?

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewUtils.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewUtils.elm
@@ -538,3 +538,12 @@ generateLoadingList =
     ]
   , li[][i[][], span[][]]
   ]
+
+countRecentChanges: RuleId -> (Dict String (List Changes)) -> Float
+countRecentChanges rId changes =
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/ViewUtils.elm
```diff
@@ -538,3 +538,12 @@ generateLoadingList =
     ]
   , li[][i[][], span[][]]
   ]
+
+countRecentChanges: RuleId -> (Dict String (List Changes)) -> Float
+countRecentChanges rId changes =
+  case Dict.get rId.value changes of
+    Just cl ->
+      case List.Extra.last cl of
+        Just c -> c.changes
```
ElaadF, line 547: Why `changes` is a `Float`?

ncharles: it should be a long (or equivalent). There's no way it holds a fraction

VinceMacBuche (approved):

PR #4155, opened 2022-02-03
Fixes #20716: Improve dynamic group computation speed

ncharles:
https://issues.rudder.io/issues/20716

This is a WIP - the code is not finished, and may land in another branch

The goal here is to explore how to do dynamic group computation with minimal LDAP queries, which are really expensive
We have many data in the nodeinfocache, we can use them

As a first approach, I changed the ldap search on hostname to a nodeinfofilter (so that it's a *post* filter). The drawback is that when we only search on hostname, there is no filter to apply, so it fetches all nodes to do the search. Kind of kill the advantage of doing so, that's why i resolve to add a boolean that says: ok, don't compute anything, and take all from cache

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -51,7 +51,6 @@ import com.normation.rudder.domain.nodes.NodeGroupId
 import com.normation.rudder.domain.nodes.NodeInfo
 import com.normation.rudder.domain.nodes.NodeState
 import com.normation.rudder.domain.properties.NodeProperty
-import com.normation.rudder.domain.queries.OstypeComparator.osTypes
 import com.normation.rudder.services.queries._
 import com.normation.zio._
 import com.unboundid.ldap.sdk._
@@ -223,6 +222,7 @@ final case object NodeStateComparator extends NodeCriterionType {
 }
 
 final case object NodeOstypeComparator extends NodeCriterionType {
+  val osTypes = List("AIX", "BSD", "Linux", "Solaris", "Windows")
   override def comparators = Seq(Equals, NotEquals)
   override protected def validateSubCase(v:String,comparator:CriterionComparator) = {
     if(null == v || v.isEmpty) Left(Inconsistency("Empty string not allowed")) else Right(v)
@@ -237,7 +237,7 @@ final case object NodeOstypeComparator extends NodeCriterionType {
 
   override def toForm(value: String, func: String => Any, attrs: (String, String)*) : Elem =
     SHtml.select(
-      (osTypes map (e => (e,e))).toSeq
+        (osTypes map (e => (e,e))).toSeq
       , { if(osTypes.contains(value)) Full(value) else Empty}
       , func
       , attrs:_*
@@ -249,10 +249,10 @@ final case object NodeOsNameComparator extends NodeCriterionType {
   import net.liftweb.http.S
 
   val osNames = AixOS ::
-    BsdType.allKnownTypes.sortBy { _.name } :::
-    LinuxType.allKnownTypes.sortBy { _.name } :::
-    (SolarisOS :: Nil) :::
-    WindowsType.allKnownTypes
+                BsdType.allKnownTypes.sortBy { _.name } :::
+                LinuxType.allKnownTypes.sortBy { _.name } :::
+                (SolarisOS :: Nil) :::
+                WindowsType.allKnownTypes
 
 
   override def comparators = Seq(Equals, NotEquals)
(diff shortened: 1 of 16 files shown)


ncharles:
Impact on simple search on hostname is dramatic: from 500ms to nearly nothing when the search don't return any node (the display part is really expensive on 10 000 nodes, so i can only measure when nothing is returned)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -222,6 +222,31 @@ final case object NodeStateComparator extends NodeCriterionType {
 }
 
 
+final case object NodeStringComparator extends NodeCriterionType {
```
ncharles, line 225: I should make it a case class, with a parameter telling on which attribute of nodeinfo we should search

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -578,6 +578,10 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
     )
   }
 
+  def getAllNodesEntry() = {
```
ncharles, line 581: i'm not reall proud of this one

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -174,7 +174,7 @@ class AcceptedNodesLDAPQueryProcessor(
     val timePreCompute =  System.currentTimeMillis
 
     for {
-      res            <- processor.internalQueryProcessor(query,select,limitToNodeIds,debugId).toBox
+      res            <- processor.internalQueryProcessor(query,select,limitToNodeIds,debugId, nodeInfoService.getAllNodesEntry).toBox
```
ncharles, line 177: this should not always we fetched, only when there is no filter

ncharles: and we are on the acceptedInventoryDIT only

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -973,17 +991,20 @@ class InternalLDAPQueryProcessor(
...
       // at that point, it may happen that nodeFilters and otherFilters are empty
-      val mainFilters = if(groupedSetFilter.isEmpty) {
+      val (mainFilters, andAndEmpty) = if(groupedSetFilter.isEmpty) {
         query.composition match {
           // In that case, we add a "get all nodes" query and all filters will be done in node info.
-          case And => Some(Set[ExtendedFilter](LDAPFilter(BuildFilter.ALL)))
+          // we should have a specific case here, saying: it's empty, we are AND, so we AND with all the cache
+//          case And => (None, true)
+
+          case And => (Some(Set[ExtendedFilter](LDAPFilter(BuildFilter.ALL))), true)
```
ncharles, line 1000: Here for the test, it's always true, but it will cause many problems for the non accepted DIT

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
ncharles: However, the big trouble here is that it breaks the detection on which group a *pending* node will appear, as the pending node is not in cache

ncharles: and properties are not correctly used to detect in which group the node will be

ncharles:
The queryChecker need also to be corrected/modified to make use of NodeInfoFilter, using the getPendingNodeInfos of NodeInfoService

ncharles:
PR updated with a new commit

ncharles:
I managed, on my test bench, with the code at this point, to have a *2 perfs increase, from 2 minutes to 1 minute
I've been really surprised by the total time as most of my 497 query went from about 100-300ms to 30 ms
The sum of all the  `com.normation.rudder.services.queries.AcceptedNodesLDAPQueryProcessor - LDAP result: 10001 entries obtained in 236ms for query` went from 71507ms to 8935ms (so the LDAP part is nearly 10 times faster)
The sum of all the postfilter `filter:rudderNode] Found 15 nodes when filtering for info service existence and properties (294 ms)` has understandably went up, from 7636ms to 22636ms (*3) 


Investigating showed:
* the time spent computing the group is now 66850 ms
* the time spend computing and UPDATING the group is 86500 ms

so 1/3 of the time is to check that the group changed on ldap 
but thee are nearly 33s unaccounted for (8935 + 22636 << 66850 )

ncharles:
Ok, i strongly suspect that looking in the nodeinfoservicecached fo matching nodeinfo, to get to LDAPNodeInfo, to convert thme to QueryResult (same looking object) to put them in a Seq, to finally get the NodeId is probably node the finest move in processOnlyId
Same for Process
Both could simply return NodeInfo

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeGroupRepository.scala
```diff
@@ -752,28 +752,37 @@ class WoLDAPNodeGroupRepository(
         con          <- ldap
         existing     <- getSGEntry(con, nodeGroup.id).notOptional("Error when trying to check for existence of group with id %s. Can not update".format(nodeGroup.id))
         oldGroup     <- mapper.entry2NodeGroup(existing).toIO.chainError("Error when trying to check for the group %s".format(nodeGroup.id.value))
-        systemCheck  <- if(onlyUpdateNodes) {
-          oldGroup.succeed
-        } else (oldGroup.isSystem, systemCall) match {
-          case (true, false) => "System group '%s' (%s) can not be modified".format(oldGroup.name, oldGroup.id.value).fail
-          case (false, true) => "You can not modify a non system group (%s) with that method".format(oldGroup.name).fail
-          case _ => oldGroup.succeed
+        // check if old group is the same as new group
```
ncharles, line 755: this change make the overhead of updated vs computed go from 18s to 13s

ncharles:
PR updated with a new commit

ncharles:
Now, dynamic group update is 50s with this change on my test bench, so a bit better than *2 perf improvement
52s spend in updating, 40s in computing (still 12 seconds for updated that don't update anythingr) <- note it's ore than 50 because it's multi-threaded

Edit: the logger part add a massive overhead, removing the debug from com.normation.rudder.services.queries.AcceptedNodesLDAPQueryProcessor show 35s for dynamic group update, about *4 perf improvement

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
Tests now pass, the "or" part was not correctly handled
However perfs have plummeted to 1M15s
This need investigation

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -157,6 +147,7 @@ trait NodeInfoService {
    */
   def getAllNodes() : IOResult[Map[NodeId, Node]]
 
+  def getAllNodeInfos():IOResult[Seq[NodeInfo]]
```
ncharles, line 150: interestingly, having a Seq rather than a Set makes this much faster

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -867,15 +858,8 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
...
-    // if nodeIds is empty and composition is and, return an empty set; with or, we need to run it in all cases
-    if (nodeIds.isEmpty && composition == And) {
-      Set[LDAPNodeInfo]().succeed
-    } else {
-      withUpToDateCache(s"${nodeIds.size} ldap node info") { cache =>
-        PostFilterNodeFromInfoService.getLDAPNodeInfo(nodeIds, predicates, composition, cache).succeed
-      }
-    }
+  def getAllNodeInfos():IOResult[Seq[NodeInfo]] = withUpToDateCache("all nodeinfos") { cache =>
+    cache.view.values.map(_._2).toSeq.succeed
```
ncharles, line 862: here, with a toSet, this method takes 88ms, with a toSeq, it take 5 ms

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

PR #4164, opened 2022-02-09
Fixes #20716: Correct dynamic group computation onf properties and improve speed

ncharles:
https://issues.rudder.io/issues/20716

This PR corrects the computation of invertion on dynamic groups, that was failing due to the ordering of action (invert was before the properties computation)

It goes the extra mile to improve perf of dynamic group computation, by changing easy ldap search on properties available on nodeinfo to a nodeinfofilter (so that it's a post filter). The drawback is that when we only search on nodeinfo related info, there is no filter to apply, so it fetches all nodes to do the search, so it kind of kill the advantage of doing so, that's why i resolved to add a boolean that says: ok, don't compute anything, and take all from cache

Main changes in code
* in CMDBQuery, the filters on NodeInfo based on LDAP filter have been replaced to filters on NodeInfo ( `NodeCriterionType` ), as we have the NodeInfo available for free thanks to the cache. Not all filters have been changed because this is a change on patch version, and I didn't want to take too much risk for those that are not trivial
* `LDAPQueryProcessor` was significantly changed, with 
  * `internalQueryProcessor` having a lambda returning all nodes as a parameter, used when there are no LDAP filter applied, or an inversion, or an "or" query
  * postfilter on NodeInfo is done before inversion & filtering on server roles
  * `check` in `PendingNodesLDAPQueryChecker` uses the Lambda being the lis of pending nodes
  * `LdapQueryProcessorResult` class removed: it was not used in a meaningful way, and added only a layer of indirection, making it harder to understand
  * `PostFilterNodeFromInfoService` object has been changed, with `getLDAPNodeInfo` to return a Seq rather than a Set (`Set` of `NodeInfo` are realy expensive), and optimisations in place based on `And` or `Or` query types
* in `LDAPNodeGroupRepository.scala` a small change to check if the NodeGroup changed or not before going on with all other checks. As surprising as it sound, it saves ~ 15 seconds on my tests on 10 000 nodes and 464 groups, probably because of the mutex that's blocking
* Some cleaning in `NodeInfoService`, to remove unused method

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -222,7 +222,122 @@ final case object NodeStateComparator extends NodeCriterionType {
     )
 }
 
+final case object NodeOstypeComparator extends NodeCriterionType {
+  val osTypes = List("AIX", "BSD", "Linux", "Solaris", "Windows")
+  override def comparators = Seq(Equals, NotEquals)
+  override protected def validateSubCase(v:String,comparator:CriterionComparator) = {
+    if(null == v || v.isEmpty) Left(Inconsistency("Empty string not allowed")) else Right(v)
+  }
 
+  override def matches(comparator: CriterionComparator, value: String): NodeInfoMatcher = {
+    comparator match {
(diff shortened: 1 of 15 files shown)


ncharles:
replaces https://github.com/Normation/rudder/pull/4155

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

fanf:
LGTM. 
I tested on several test environment with/without the change, and observed no diff in group node list. 
The code is a bit simpler, and there is more tests. The top level description was quite helpful. This is a solid PR. Thanks!

PR #4172, opened 2022-02-13
Fixes #20758: Improve node computation

ncharles:
Based on https://github.com/Normation/rudder/pull/4164

Changing the types to remove most of the hashCode computation, most notably, using Set for NodeInfo is a really bad idea
With this (unfinished) code, computation of dynamic group goes from 1m15 (pr 4164) to 35s
Do not merge for now, this is only exploration

The only relevant thing is to return nodeId rather than NodeInfos now that we don't need serverrole info

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
@@ -114,6 +114,7 @@ trait NodeInfoService {
 
   def getNodeInfos(nodeIds: Set[NodeId]): IOResult[Set[NodeInfo]]
 
+  def getNodeInfosSeq(nodeIds: Seq[NodeId]): IOResult[Seq[NodeInfo]]
 
   /**
    * Return the number of managed (ie non policy server, no rudder role nodes.
@@ -874,6 +875,10 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
 
   def getNodeInfos(nodeIds: Set[NodeId]): IOResult[Set[NodeInfo]] = withUpToDateCache(s"${nodeIds.size} nodes infos") { cache =>
     cache.filter(x => nodeIds.contains(x._1)).values.map( _._2).toSet.succeed
+  }
+
+  def getNodeInfosSeq(nodeIds: Seq[NodeId]): IOResult[Seq[NodeInfo]] = withUpToDateCache(s"${nodeIds.size} nodes infos") { cache =>
+    nodeIds.map(id => cache.get(id).map(_._2)).flatten.succeed
   }
 }
 
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -163,7 +163,7 @@ class AcceptedNodesLDAPQueryProcessor(
       query:QueryTrait,
       select:Seq[String],
       limitToNodeIds:Option[Seq[NodeId]]
-  ) : Box[Seq[NodeInfo]] = {
+  ) : Box[Seq[NodeId]] = {
 
     val debugId = if(logger.isDebugEnabled) Helpers.nextNum else 0L
     val timePreCompute =  System.currentTimeMillis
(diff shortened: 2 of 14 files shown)


ncharles:
Do not review, still a wip

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -168,93 +160,64 @@ class AcceptedNodesLDAPQueryProcessor(
...
-      }
-
       //filter out Rudder server component if necessary
-
       query.returnType match {
         case NodeReturnType =>
             // we have a special case for the root node that always never to that group, even if some weird
             // scenario lead to the removal (or non addition) of them
-          val withoutServerRole = inNodes.filterNot { case QueryResult(e, inv, _) =>  e(A_NODE_UUID) == Some("root") }
+          val withoutServerRole = foundNodes.filterNot { case x: NodeId =>  x.value == "root" }
```
ncharles, line 178: It should no't be root, it will miss all relays!

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -294,23 +279,37 @@ object PostFilterNodeFromInfoService {
...
+        timeres        =  (System.currentTimeMillis - timePreCompute)
+        _              =  logger.debug(s"LDAP result: ${foundNodes.size} entries in pending nodes obtained in ${timeres}ms for query ${query.toString}")
       } yield {
-        ids
+        //filter out Rudder server component if necessary
+        (query.returnType match {
+          case NodeReturnType =>
+            // we have a special case for the root node that always never to that group, even if some weird
+            // scenario lead to the removal (or non addition) of them
+            foundNodes.filterNot { case x: NodeId =>  x.value == ("root") }
```
ncharles, line 310: same here, it cannot be only root

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
```diff
@@ -875,6 +876,10 @@ trait NodeInfoServiceCached extends NodeInfoService with NamedZioLogger with Cac
   def getNodeInfos(nodeIds: Set[NodeId]): IOResult[Set[NodeInfo]] = withUpToDateCache(s"${nodeIds.size} nodes infos") { cache =>
     cache.filter(x => nodeIds.contains(x._1)).values.map( _._2).toSet.succeed
   }
+
+  def getNodeInfosSeq(nodeIds: Seq[NodeId]): IOResult[Seq[NodeInfo]] = withUpToDateCache(s"${nodeIds.size} nodes infos") { cache =>
+    nodeIds.map(id => cache(id)).map(_._2).succeed
```
fanf, line 881: it could fail if some requested node id are not in cache (for some inconsistency from somewhere). 
What is the correct thing to do on missing nodeId ? Fail everything or ignore ? 
If fail, something like:
```withUpToDateCache(s"${nodeIds.size} nodes infos") { cache =>
ZIO.foreach(nodeIds) { id => cache.get(id).map(_._2).notOptionnal(s"node with id ${id.value} not in cache, likely an error")} }
```
If ignore:
```withUpToDateCache(s"${nodeIds.size} nodes infos") { cache =>
nodeIds.map(id => cache.get(id).map(_._2)).collect { case Some(info) => info }.succeed
}```

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala [hunk 4]
```diff
@@ -259,22 +262,20 @@ object PostFilterNodeFromInfoService {
...
+        foundNodeIds = foundNodeInfos.map(_.id).toSet
+        allNodesInfos.filter { nodeinfo => predicate(nodeinfo, predicates, composition, Some(foundNodeIds)) }.map(_.id)
       case And =>
         if (predicates.isEmpty) {
           // paththru
-          foundNodeInfos
+          foundNodeInfos.map(_.id)
         } else {
-          foundNodeInfos.filter { nodeinfo => predicate(nodeinfo, predicates, composition, None) }
+          foundNodeInfos.flatMap { case nodeinfo => if (predicate(nodeinfo, predicates, composition, None)) { Some(nodeinfo.id) } else None }
```
fanf, line 272: can be directly
```
foundNodeInfos.collect { case nodeinfo if (predicate(nodeinfo, predicates, composition, None)) => nodeinfo.id } 
```

fanf (commented):
Appart from the very small feedback, that seems great!
You can self merge once your change are commited

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
ncharles: actually, it could still be the filter

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NodeInfoService.scala
ncharles: ha, very good point. 
to be consistent, we should ignore, so the second approach is the right one. Thank you !

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
ncharles: the collect is better, it avoids one creation of collection

ncharles:
PR updated with a new commit

fanf (approved):

fanf:
OK, squash merging this PR

PR #4187, opened 2022-02-22
Fixes #20802: NodeAndPolicyServerReturnType name is misleading for search that include root server

ncharles:
https://issues.rudder.io/issues/20802

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
@@ -187,7 +187,7 @@ class AcceptedNodesLDAPQueryProcessor(
             }
           }
           withoutServerRole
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -308,8 +308,8 @@ class PendingNodesLDAPQueryChecker(
           case NodeReturnType =>
             // we have a special case for the root node that always never to that group, even if some weird
             // scenario lead to the removal (or non addition) of them
-            foundNodes.filterNot { case x: NodeInfo =>  x.node.id.value ==("root") || x.isPolicyServer }
```
fanf, line 311: I think it will change group composition when there is relays

fanf: ok, it is just in pending, a mis-correction in previous PR

fanf (approved):

PR #4190, opened 2022-02-23
Fixes #20808: Missing reports with a reportid are missing in compliance

fanf:
https://issues.rudder.io/issues/20808

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -46,6 +46,7 @@ import com.normation.rudder.domain.policies.DirectiveId
 import com.normation.rudder.domain.reports._
 import com.normation.rudder.reports._
 import com.normation.rudder.reports.execution.AgentRunId
+
 import net.liftweb.common.Loggable
 
 import java.util.regex.Pattern
@@ -53,7 +54,10 @@ import com.normation.rudder.domain.policies.PolicyMode
 import com.normation.rudder.domain.reports.ReportType.BadPolicyMode
 import com.normation.rudder.reports.execution.AgentRunWithNodeConfig
 import com.normation.rudder.domain.policies.RuleId
+
 import org.apache.commons.lang3.StringUtils
+
+import scala.annotation.tailrec
 
 /*
  *  we want to retrieve for each node the expected reports that matches it LAST
@@ -391,7 +395,7 @@ final case class ContextForNoAnswer(
    */
   private[this] def runType(traceMessage: String, runType: RunAndConfigInfo)(implicit nodeId: NodeId): RunAndConfigInfo = {
     val msg = if(traceMessage.trim.isEmpty) "" else ": " + traceMessage
-    ComplianceDebugLogger.node(nodeId).trace(s"Run config for node ${nodeId.value}: ${runType.logName} ${msg}")
+    ComplianceDebugLogger.node(nodeId).debug(s"Run config for node ${nodeId.value}: ${runType.logName} ${msg}")
     runType
   }
 
@@ -695,13 +699,13 @@ final case class ContextForNoAnswer(
     //only interesting reports: for that node, with a status
     val nodeStatusReports = agentExecutionReports.collect{ case r: ResultReports if(r.nodeId == nodeId) => r }
 
-    ComplianceDebugLogger.node(nodeId).trace(s"Computing compliance for node ${nodeId.value} with: [${runInfo.toLog}]")
+    ComplianceDebugLogger.node(nodeId).debug(s"Computing compliance for node ${nodeId.value} with: [${runInfo.toLog}]")
 
     val t1 = System.currentTimeMillis
     val ruleNodeStatusReports = runInfo match {
 
       case ReportsDisabledInInterval(expectedConfig, _) =>
-        ComplianceDebugLogger.node(nodeId).trace(s"Compliance mode is ${ReportsDisabled.name}, so we don't have to try to merge/compare with expected reports")
+        ComplianceDebugLogger.node(nodeId).debug(s"Compliance mode is ${ReportsDisabled.name}, so we don't have to try to merge/compare with expected reports")
         buildRuleNodeStatusReport(
             //these reports don't really expires - without change, it will
             //always be the same.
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -391,7 +395,7 @@ final case class ContextForNoAnswer(
    */
   private[this] def runType(traceMessage: String, runType: RunAndConfigInfo)(implicit nodeId: NodeId): RunAndConfigInfo = {
     val msg = if(traceMessage.trim.isEmpty) "" else ": " + traceMessage
-    ComplianceDebugLogger.node(nodeId).trace(s"Run config for node ${nodeId.value}: ${runType.logName} ${msg}")
+    ComplianceDebugLogger.node(nodeId).debug(s"Run config for node ${nodeId.value}: ${runType.logName} ${msg}")
```
fanf, line 398: the debug level wasn't used and I added logs that *really* need trace level

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -1201,129 +1211,144 @@ final case class ContextForNoAnswer(
     , policyMode              : PolicyMode //the one of the directive, or node, or global
     , unexpectedInterpretation: UnexpectedReportInterpretation
   ) : List[ComponentStatusReport] = {
+    // an utility class that store an expected value and the list of matching reports for it
```
fanf, line 1214: that's just moving things around

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -1201,129 +1211,144 @@ final case class ContextForNoAnswer(
...
-                      recPairReports(tail, newFreeValues, v :: newPairedValues, unexpected, mode)
-                  }
-              }
-          }
+        val (matchValue, matchId) = expectedComponent.componentsValues.partitionMap{
+          case e: ExpectedValueMatch => Left(e)
+          case e: ExpectedValueId    => Right(e)
         }
 
+        // 1. start with the easy case: expected reports with a report ID
```
fanf, line 1324: The interesting part is below: 
- the previous matching was missing the missing (ahah) because it loops over `matched` reports, ie the one we received for the reportId. That won't tell us which one we should have received
- it was not check for the consitancy between expect component/value and what is on the report. There is no reason that a report with the correct reportId has a bad component/value, but things happen, and the test were testing that already. 
- we also assume that the number of reports received, when it's at least one, is always the correct number to behave like the old merge (for iteration I presume)

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -1026,6 +1033,90 @@ class ExecutionBatchTest extends Specification {
...
+   *  reportid2 = 10957b22-20e5-43c3-8ef2-aa5b666b3bd9
+   *  value2 = /tmp/rudder-file-gm2
+   *  message2 = Presence of file /tmp/rudder-file-gm2 was correct
+   *  reportid3 = e0a496c2-14a5-4cd2-81bf-8d4630549b41
+   *  value3 = /tmp/rudder-file-gm3
+   *  message3 = Presence of file /tmp/rudder-file-gm3 was correct
+   *
+   *  t1 = 2022-02-23T11:40:30.000+01:00
+   *  t2 = 2022-02-23T11:40:23.000+01:00
+   *
```
fanf, line 1059: these logs are a bit verbose, but we can alsmost copy/past then in the test to check what is going wrong

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: you can switch until next comment

amousset (approved):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -1201,129 +1211,144 @@ final case class ContextForNoAnswer(
...
+          case ((acc, reports), expectedValueId) =>
+            val (matched, left) = reports.partition(_.reportId == expectedValueId.id)
+
+            // By construction, components and values in reports with the same report id should be OK.
+            // We still want to check quickly if it's not an error, ie component or value need to be the same,
+            // else it's unexpected, appart if they contains a variable.
+            // We also assume that we await AT LEAST one report, but that more reports is OK to account for iterators
+
+            val (ok, unexpected) = matched.foldLeft((List.empty[ResultReports], List.empty[ResultReports])) { case ((ok, unexp), next) =>
+              if(  ( next.component == expectedComponent.componentName || matchCFEngineVars.pattern.matcher(next.component).matches())
```
ncharles, line 1335: we need to advertise that the component name can use a variable

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -550,34 +550,41 @@ class ExecutionBatchTest extends Specification {
...
     val withGood = ExecutionBatch.checkExpectedComponentWithReports(expectedComponent, reports, ReportType.Missing, PolicyMode.Enforce, strictUnexpectedInterpretation).head
     val withBad  = ExecutionBatch.checkExpectedComponentWithReports(expectedComponent, badReports, Missing, PolicyMode.Enforce, strictUnexpectedInterpretation).head
+
+    val goodBlock = withGood.asInstanceOf[BlockStatusReport]
+    val componentValues = withGood.componentValues
+
+
     "return a success block " in {
-      withGood.compliance === ComplianceLevel(success = 1)
+      withGood.compliance === ComplianceLevel(repaired = 1)
```
ncharles, line 559: the title of the test is not correct

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
VinceMacBuche: we don't want to check if the comopnent name is valid, and i clearly wanted to have component with different value htat was something i expected so we can have reports on iterators and use the value of the iterator in the component/value

fanf: if so, we have to change expected reports, because it makes absolutly no sense to have an expected reports with a component name and accept other component name. We need to have `reportId` in the expected reports, and only use that

fanf: plus, iterator will be a variable no ? If so, they are accepted here

PR #4193, opened 2022-02-24
Fixes #16197: Serialization of NodeConfigurations is fairly expensive

fanf:
https://issues.rudder.io/issues/16197

Rewrite NodeExpectedReports json serialisation with zio-json. 
This is extremely boilerplaty because:
- I wanted to kept the `JsonNodeExpectedReports` which is the current API with "the world outside the serialisation logic"
- I needed to have something compatible with json from 7.0 and the new version (so that's 2x the code).

Appart the boilerplateness, it is fairly systematic. For each version, I:
- created case classes to match the exact names of the JSON
- for each of these case class, I define a `transform` method to its corresponding `JsonNodeExpectedReports` counterpart, and an implicit case class for the opposite operation
- then, for each case class, I create a JsonCodec to read from/to JSON. Most are just simple derivation, some are a bit more code because our json structure is not very nice for ADT (it misses a discriminant to answer "what case class of the ADT")

And it's almost all. 
There is a lot of code in tests, too, because well, I had to hand write JSON ADT to be sure the matching was OK.

The only a bit interesting part is the fallback logic: we try first version 7.1, then version 7.0 decoder (because we should converge with "everything in 7.1"). And we always encode in 7.1.

It's seems fast. 100k `parse(serialise)`  takes 19s for a 10kB json, so with my high precision measure, 0.2 ms for a `parse(serialise)`. 
Lift was taking 10s for 10k, so that's at least 5x better. 

The new version is takes around 50% less space. That should also help performances.


fanf:
PR rebased

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -199,350 +204,593 @@ final case class NodeConfigIdInfo(
...
+     *          "did": "9b0dc972-f4dc-4aaa-bae6-1189cf9074b6"
+     *       },{
+     *         "ob": {                                         // overridden by
+     *           "rid": "2278f76f-28d3-4326-8199-99561dd8c785",
+     *           "did" "093a494b-1073-49e9-bb1e-3c128c7f6a42"
+     *        }
+     *     ]
+     */
+
+    // name are false for startHour / splayHour and no override
```
ncharles, line 545: i don't understand this comment

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -199,350 +204,593 @@ final case class NodeConfigIdInfo(
...
+    }
+    
+    final case class JsonGlobalPolicyMode7_1(m: PolicyMode, o: PolicyModeOverrides) {
+      def transform = GlobalPolicyMode(m, o)
+    }
+    implicit class _JsonGlobalPolicyMode7_1(x: GlobalPolicyMode) {
+      def transform = JsonGlobalPolicyMode7_1(x.mode, x.overridable)
+    }
+
+    final case class JsonModes7_1(
```
ncharles, line 565: could you add in comment the meaning of each of the field name. It would improve readability and avoid typos

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -199,350 +204,593 @@ final case class NodeConfigIdInfo(
...
+        case a: BlockExpectedReport => a.transform
       }
     }
+    final case class JsonValueExpectedReport7_1(
+        vid: String
+      , vs: List[Either[List[String],JsonExpectedValueId7_1]]
+    ) extends JsonComponentExpectedReport7_1 {
+      def transform = ValueExpectedReport(vid, vs.map {
+        case Left(Nil)     => ExpectedValueMatch("None", "None")
+        case Left(a::Nil)  => ExpectedValueMatch(a,a)
```
ncharles, line 637: i'm unsure how we can have a Left ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -199,350 +204,593 @@ final case class NodeConfigIdInfo(
...
+        case a: ValueExpectedReport => a.transform
+        case a: BlockExpectedReport => a.transform
       }
     }
+    final case class JsonValueExpectedReport7_1(
+        vid: String
+      , vs: List[Either[List[String],JsonExpectedValueId7_1]]
+    ) extends JsonComponentExpectedReport7_1 {
+      def transform = ValueExpectedReport(vid, vs.map {
+        case Left(Nil)     => ExpectedValueMatch("None", "None")
```
ncharles, line 636: can you add in the comment that the Left is for compatibility with pre-7.1 ? I've been a bit at a lost here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
```diff
@@ -199,350 +204,593 @@ final case class NodeConfigIdInfo(
...
+    implicit class _JsonNodeExpecteReports7_1(x: JsonNodeExpectedReports) {
+      def transform = JsonNodeExpectedReports7_1(x.modes.transform, x.ruleExpectedReports.map(_.transform), x.overrides.map(_.transform))
+    }
+
+
+
+    ////////// json codec //////////
+
+    implicit val codecJsonGlobalPolicyMode7_1: JsonCodec[JsonGlobalPolicyMode7_1] = DeriveJsonCodec.gen
+    implicit val codecJsonAgentRun7_0: JsonCodec[JsonAgentRun7_1] = DeriveJsonCodec.gen
```
ncharles, line 704: is it a typo ?

ncharles (changes requested):
This looks great, and is quite clear
I'd just like a bit more comment on the case classes, so that we don't mix up field, and I *think* there is a 7_0 instead of a 7_1

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
fanf: ah, I copy/copied it. It's on the version 7 for agentrun

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ExpectedReports.scala
fanf: yep. The name does not matter, but it's quite misleading :)

fanf:
PR updated with a new commit

ncharles (approved):

PR #4194, opened 2022-02-25
Fixes #20828: Unexpected/missing for same component, same value

fanf:
https://issues.rudder.io/issues/20828

So, as explained in the ticket, the problem was that we were building expected reports at the end of generation, we were looking at possible variables for each component expecting reporting. 
But during Policy/PolicyDraft/etc generaly earlier, we missed to differentiate variables based on reportId. So, for us, all variable with the same component name were the same, and we kept just one (adapting cardinality I think, but actually we don't care much about cardinality in any case, because of iterators and the like). 
So, we ended with just ONE variable. 
And so, the expected component creation was looking around, correctly finding its variables with report id (because they come from elsewhere), but when it tries to reconciliate with the policy to get back the component value name (I'm not sure why we don't get him throught expected reports), it finds only one available, and take that component name value for all expected reports.

So, we end up with expected reports with all the same value name. And so, the actual reports get rejected. 

The main suprise that made me lose 2h is that the `reportId` is not defined at variable level in metadata (ie in technique `rootSection`, but in its parent section. So even after threading `reportId` everywhere, they still were `None` and so the problem was still there.


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #4208, opened 2022-03-14
Fixes #20886: Add name for spring security main auth configuration bean to be used by oauth2 

fanf:
https://issues.rudder.io/issues/20886

We need to set a name to the main http configuration filter chain so that we can change it programmatically in a plugin at runtime to add Oauthv2 filters


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #4215, opened 2022-03-22
Fixes #20926: Starting policy generation by hand fails if node-configuration-hashes.json 

fanf:
https://issues.rudder.io/issues/20926


fanf:
PR updated with a new commit

PR #4236, opened 2022-04-14
Fixes #20989: Serialization to JSON from compliance/rules api is really slow

ncharles:
https://issues.rudder.io/issues/20989

Perf of API for compliance is not so great: about 3s to convert to JSON the object
So I figured: hey, why not switch to zio-json.
Hundreds of lines later, I have exactly the same perf between both, and the zio-json doesn't work as expected:
instead of "complianceDetails":{"noReport":100.0} I have "complianceDetails":[{"status":"noReport","value":100.0}]

and there's a loop in the tree of result, and it fails with an error too long for my terminal

Timings:
v13
[2022-04-14 20:16:54+0000] TRACE debug_timing - getRuleCompliance - getting  ByRuleRuleCompliance in 795 ms
[2022-04-14 20:16:56+0000] TRACE debug_timing - getRuleCompliance - getting  rule.toJson in 2560 ms
[2022-04-14 20:16:56+0000] TRACE debug_timing - getRuleCompliance v13 - total time before reponse  in 3355 ms

v14
[2022-04-14 20:17:12+0000] TRACE debug_timing - getRuleCompliance - getting  ByRuleRuleCompliance in 825 ms
[2022-04-14 20:17:15+0000] TRACE debug_timing - getRuleCompliance V14 - making all ByRuleRuleCompliance in 3601 ms

I suspect doing all these object is pretty costly

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
@@ -435,6 +435,7 @@ trait RudderJsonDecoders {
  * We want to get ride of RestExtractorService but for now, we keep it for the parameter parts.
  */
 class ZioJsonExtractor(queryParser: CmdbQueryParser with JsonQueryLexer) {
+
   import JsonResponseObjects._
   import JsonQueryObjects._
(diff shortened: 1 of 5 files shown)


ncharles:
PR updated with a new commit

ncharles:
closing, as this change is not relevant to perf

PR #4237, opened 2022-04-16
Fixes #20998: Performance of compliance computation in ComplianceLevel can be improved

ncharles:
https://issues.rudder.io/issues/20998

This PR revisits the way ComplianceLevel are computed, to make it faster, and to fix issue in specific cases (ex,  200 succes, 199 repaired, 1 error, 1 missing, 1 unexpected , 1 non compliancere gives 47% success and 50% repaired with 0 precision)

All in all, the computation is ~4 times faster, mainly by using Long computation rather than BigDecimal
The drawback is that we can't go further than precision 5 (I can't seem to see which use case would need more than that), so computation are made using Long, by multiplying with the power of ten, and then truncating.

It also introduces a method to compute compliance directly without pending result, so that we can skip the duplication of ComplianceLevel and the creation of a CompliantPercent

It is *not* enough to make the display of the Nodes list instantaneous in the Rule detail page, but my tests show it's twice as fast as before. Unfortunately the way it's done computed twice by level the compliance (once without pending, and one with pending), and it's made for the Rule, and then each Directives and Nodes, and then for each components, etc etc.
The API answers on my test platform in 2.4s (from 4.5s before) - removing all compliance computation makes it answer in 1s (which is the cost of getting the Groups and Rules)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
@@ -37,8 +37,9 @@
 
 package com.normation.rudder.domain.reports
 
-import com.normation.rudder.domain.logger.ComplianceLogger
 import com.normation.rudder.domain.reports.ComplianceLevel.PERCENT_PRECISION
+import com.normation.rudder.domain.reports.CompliancePrecision.{Level0, Level2}
+import net.liftweb.common._
 import net.liftweb.http.js.JE
 import net.liftweb.http.js.JE.JsArray
 import net.liftweb.json.JsonAST.JObject
@@ -87,7 +88,7 @@ final case class CompliancePercent(
   , nonCompliant      : Double = 0
   , auditError        : Double = 0
   , badPolicyMode     : Double = 0
-)(val precision: Int = 0) {
+)(val precision: CompliancePrecision = Level0) {
   val compliance = success+repaired+notApplicable+compliant+auditNotApplicable
 }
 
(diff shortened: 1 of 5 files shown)


ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
i'm unsure this does anything meaningfull at this point, I have only one working test and no valid support of precision

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

fanf:
what s the status of that PR ? There is some commented out tests, should I still review it now ?

ncharles:
You can review it. The tests are commented because they fail, but they failed also before (well, they didn't exist)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -111,57 +110,111 @@ object CompliancePercent {
...
-    if(c.total == 0) {  // special case: let it be 0
+    val total = c.total
+    if (total == 0) { // special case: let it be 0
       CompliancePercent()(precision)
     } else {
-      val total = c.total // not zero
-      def pc_for(i:Int) : Double = {
-        if(i == 0) {
+      // these depends on the precision
+      val diviser = divisers(precision)
```
ncharles, line 135: @fanf do you know what's the cost of doing an assert ? I'll need to assert that prevision is <=5

fanf: create a case class precision with case object Level1 to Level5 and its value setted. It should not be perf impacting (object are global, and accessing a static field is very quick, and the traduction to level is only done at border like API so not important)

ncharles:
PR updated with a new commit

fanf (approved):

ncharles:
OK, squash merging this PR

ncharles:
OK, merging this PR

PR #4248, opened 2022-04-25
Fixes #21028: The way groups are fetched when responding to API compliance is inefficient

ncharles:
https://issues.rudder.io/issues/21028

This change makes the lock more tight on getting all groups, and introduce a method that only get the NodeGroupId, Set[NodeId] to make the fetching of nodes per group faster.
Impact is minimal (from 400-800ms to 300-500ms) but nonetheless wlecomed


VinceMacBuche (approved):

ncharles:
PR updated with a new commit

PR #4251, opened 2022-04-29
Fixes #17728: List of nodes by version does not include windows nodes

ElaadF:
https://issues.rudder.io/issues/17728
dsc_agent

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/homePage.js
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/homePage.js
@@ -217,11 +217,10 @@ function doughnutChart (id,data,count,colors) {
                   case 'nodeAgents':
                       query.query.where = [{
                           objectType: "software"
-                        , attribute : "cn"
-                        , comparator: "eq"
-                        , value     : "rudder-agent"
+												 , attribute : "cn"
+												 , comparator: "regex"
+												 , value     : "rudder-agent|Rudder agent \\(DSC\\)"
                       },{
-
                           objectType: "software"
                         , attribute : "softwareVersion"
                         , comparator: "regex"


webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/homePage.js
```diff
@@ -216,11 +216,6 @@ function doughnutChart (id,data,count,colors) {
                       break;
                   case 'nodeAgents':
                       query.query.where = [{
-                          objectType: "software"
-                        , attribute : "cn"
-                        , comparator: "eq"
-                        , value     : "rudder-agent"
```
ElaadF, line 222: I'm not sure if this is important, if I follow the logic `Rudder agent (DSC)` whould be use when a version contains a windows node (`rudder-agent OR Rudder agent (DSC)`)

amousset: the problem is that t will match any software in the same version, which is not that likely to happen but would be very surprising for the user.
Can't we use a regex to match both `rudder-agent` and `Rudder agent (DSC)` in the filter?

VinceMacBuche: We should look for the version we have in the "management technology" in out inventory (that's an entry we control and now we set Rudder version within, it was not the case in the old days

VinceMacBuche:
you can also do egexp on software name

something like ```rudder-agent|Rudder Agent (Dsc)``` may be tester

VinceMacBuche:
Rudder.*|rudder-agent as regex works fine

VinceMacBuche:
{"query":{"select":"nodeAndPolicyServer","composition":"And","where":[{"objectType":"software","attribute":"cn","comparator":"regex","value":"rudder-agent|Rudder.*"},{"objectType":"software","attribute":"softwareVersion","comparator":"regex","value":"6(.|~)2-1(.|~)22.*"}]}}

works on our lab

VinceMacBuche (approved):

PR #4257, opened 2022-05-05
Fixes #21068: Make user login not case sensitive 

ElaadF:
https://issues.rudder.io/issues/21068


ElaadF:
Not what I wanted to do...

PR #4258, opened 2022-05-05
Fixes #21068: Make user login not case sensitive

ElaadF:
https://issues.rudder.io/issues/21068
#### Case sensitivity 
- [x] Option parameter `case-sensitivity` in XML file
- [x] Log a warning when parameter `case-sensitivity` is missing or the value si different from `true` or `false` 
- [x] By default case-sensitivity is set to `true` even if the parameter is missing or unknown
- [x] Parsing XML file and setup parameter for case sensivity in `RudderConfig`
- [x] Find where login comparison is made to .lowercase the login send from the login form 
- [x] lowercase the stored user's login 
- [x] Login as `ToTO` and `toto` is working
 
#### Sanity check
- [x] Log warning message when case sensitivity is enable and usernames are similar
- [x] Log error message when there is user with same username (according to case-sensitivity)
- [x] Ignore users with same username
#### Documentation
- [ ] Add documentation (in User-management)
## Screenshots
admin_case

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -343,7 +343,8 @@ object LogFailedLogin {
 class RudderInMemoryUserDetailsService(val authConfigProvider: UserDetailListProvider) extends UserDetailsService {
   @throws(classOf[UsernameNotFoundException])
   override def loadUserByUsername(username:String) : RudderUserDetail = {
-    authConfigProvider.authConfig.users.getOrElse(username, throw new UsernameNotFoundException(s"User with username '${username}' was not found"))
+    val u = if(RudderConfig.rudderUsernameCaseSensitive) username else username.toLowerCase()
+    authConfigProvider.authConfig.users.getOrElse(u, throw new UsernameNotFoundException(s"User with username '${username}' was not found"))
   }
 }
 
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
@@ -50,7 +50,6 @@ import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.domain.logger.PluginLogger
 import com.normation.rudder.rest.RoleApiMapping
 import com.normation.rudder.web.services.RudderUserDetail
-
 import org.bouncycastle.util.encoders.Hex
 import org.springframework.security.core.GrantedAuthority
 import org.xml.sax.SAXParseException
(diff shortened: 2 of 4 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -512,10 +513,16 @@ class RestAuthenticationFilter(
   }
 
   private[this] def authenticate(userDetails: RudderUserDetail) : Unit = {
```
ElaadF, line 515: The lowercase here doesn't work in fact

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -212,6 +216,43 @@ final class FileUserDetailListProvider(roleApiMapping: RoleApiMapping, authorisa
...
+    (duplicatesLogin.isEmpty, RudderConfig.rudderUsernameCaseSensitive) match {
+      case (true, true) =>
+        val potentialDup = userLogins.map(_.toLowerCase()).groupBy(identity).collect { case (x, List(_,_,_*)) => x }
+        if(potentialDup.nonEmpty) {
+          ApplicationLogger.warn(s"Users with potential username collision if case sensitivity is deactivated: ${potentialDup.mkString(", ")}")
+          userDetails
+        } else {
+          userDetails
+        }
+      case (false, _) =>
```
ElaadF, line 250: This will never happen here because one duplicate seems to be removed before, for exemple if I have 2 user `toto` it seems that one of them will be ignored

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -317,7 +376,7 @@ object UserFileProcessing {
       //now, get users
       val users = ( (xml \ "user").toList.flatMap { node =>
        //for each node, check attribute name (mandatory), password  (mandatory) and role (optional)
-       (   node.attribute("name").map(_.toList.map(_.text))
+       (   node.attribute("name").map(_.toList.map(username => if (RudderConfig.rudderUsernameCaseSensitive) (username.text) else username.text.toLowerCase()))
```
VinceMacBuche, line 379: I think you should not lowercase here but in sanity log filter only sanityLoginFilter ( in fact you don't need to lowercase anymore after if you do it here, but you lost your case so and cannot display it in log messages... )

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -136,16 +135,52 @@ final case class UserDetailList(
...
+   *   - users with same username (according to the case sensitivity)
+   *     will be removed from the returned list
+   */
+  def sanityLoginFilter(userDetails: List[RudderUserDetail]): List[RudderUserDetail] = {
+    val userLogins = if(RudderConfig.rudderUsernameCaseSensitive){
+      userDetails.map(_.getUsername)
+    } else {
+      userDetails.map(_.getUsername.toLowerCase())
+    }
+    val duplicatesLogin = userLogins.groupBy(identity).collect { case (x, List(_,_,_*)) => x }.toList
```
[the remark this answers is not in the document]
VinceMacBuche: You will have a Map(String, List[RudderUserDetails])

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
VinceMacBuche: Ok the problem is that it may be complicated to mody the user account afterward, keep it  like this

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -289,6 +325,29 @@ object UserFileProcessing {
...
+        case e: Exception =>
+          Left(UserConfigFileError("User definitions: An error occured while parsing /opt/rudder/etc/rudder-users.xml. Logging in to the Rudder web interface will not be possible until this is fixed and the application restarted.", Some(e)))
+      }
+    }
+    for {
+      xml <- optXml
+      root = (xml \\ "authentication")
+    } yield {
+      (root(0) \ "@case-sensitivity").text.toLowerCase match {
+        case "true" => true
```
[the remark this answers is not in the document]
ElaadF: ?you want to applied case sensitivity on parameter `authentication` and `case-sensitivity` in the XML ?

VinceMacBuche (approved):

PR #4261, opened 2022-05-09
Fixes #21077: Don't display oauth/oidc client secret in logs

fanf:
https://issues.rudder.io/issues/21077

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
@@ -268,6 +268,12 @@ object RudderConfig extends Loggable {
   val hiddenRegisteredProperties = scala.collection.mutable.Buffer[String]()
   hiddenRegisteredProperties += "rudder.dir.licensesFolder"
 
+  // auth backend is init too late to have a chance to hide its values, which is a bit sad.
+  // We still need to make invisible all oauth/oidc client secret
+  hiddenRegisteredProperties ++= {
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -268,6 +268,12 @@ object RudderConfig extends Loggable {
   val hiddenRegisteredProperties = scala.collection.mutable.Buffer[String]()
   hiddenRegisteredProperties += "rudder.dir.licensesFolder"
 
+  // auth backend is init too late to have a chance to hide its values, which is a bit sad.
+  // We still need to make invisible all oauth/oidc client secret
+  hiddenRegisteredProperties ++= {
+    import scala.jdk.CollectionConverters._
+    config.entrySet().asScala.map(_.getKey).filter(s => s.startsWith("rudder.auth.oauth2.provider") && s.endsWith("client.secret"))
```
ncharles, line 275:
```suggestion
    config.entrySet().asScala.map(_.getKey).filter(s => s.startsWith("rudder.auth.") && s.endsWith("client.secret"))
```

so that others that oauth2 will be protected also ?

fanf: the namespace is the same for oauth and oidc (b/c oidc is a layer on top of oauth2). 
But yes, if we add other backend, like samlv2, we will need to do the same thing.

ncharles (approved):

PR #4265, opened 2022-05-10
Fixes #21091: Require Java 11

amousset:
https://issues.rudder.io/issues/21091

### DIFF
--- a/webapp/sources/pom.xml
+++ b/webapp/sources/pom.xml
@@ -158,8 +158,7 @@ limitations under the License.
             </compilerPlugin>
         </compilerPlugins>
           <args>
-            <!-- require java8, cf http://www.rudder-project.org/redmine/issues/9917 -->
-            <arg>-target:jvm-1.8</arg>
+            <arg>-target:11</arg>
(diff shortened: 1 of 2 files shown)


webapp/sources/pom.xml
```diff
@@ -239,9 +238,7 @@ limitations under the License.
         <artifactId>maven-compiler-plugin</artifactId>
         <version>3.7.0</version>
         <configuration>
-          <!-- require java8, cf http://www.rudder-project.org/redmine/issues/8963 -->
-          <source>1.8</source>
-          <target>1.8</target>
+          <release>11</release>
```
amousset, line 241: https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html says that:

> Starting JDK 9 [...]
> Unlike the [-source and -target options](https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-source-and-target.html), the compiler will detect and generate an error when using APIs that don't exist in previous releases of Java SE.

which sounds like something we want.

fanf (approved):

PR #4290, opened 2022-05-23
Fixes #21081: Adapt rule details to have the list of directives/nodes…

ncharles:
… that appear faster


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -530,6 +530,12 @@ object RuleApi extends ApiModuleProvider[RuleApi] {
     override def dataContainer = None
   }
 
+  // For the rule detail page
+  final case object GetRuleNodesAndDirectives extends RuleApi with OneParam with StartsAtVersion14 with SortIndex { val z = implicitly[Line].value
```
fanf, line 534: It must be defined out of RuleAPI to be made internal API only

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
```diff
@@ -265,6 +266,76 @@ object RuleTarget extends Loggable {
...
 
+  /**
+   * Return all node ids that match the set of target.
+   * allNodes pair is: (nodeid, isPolicyServer)
+   */
+  def getNodeIdsChunk(
+      targets : Set[RuleTarget]
+    , allNodes: Map[NodeId, Boolean /* isPolicyServer */]
+    , groups  : Map[NodeGroupId, Chunk[NodeId]]
+    , allNodesAreThere: Boolean = true  // if we are working on a subset of node, set to false
```
fanf, line 277: this parameter does not seems to be used and it seems quite dangerous in the public method API (in rec method, it's OK, it's an implementation detail). Can you remove it ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RulesInternalAPI.scala
```diff
@@ -0,0 +1,80 @@
...
+import zio.syntax._
+
+class RulesInternalApi(
+    restExtractorService  : RestExtractorService
+  , ruleInternalApiService: RuleInternalApiService
+)  extends LiftApiModuleProvider[API] {
+
+  import RestUtils._
+
+  def response(function: Box[JValue], req: Req, errorMessage: String, dataName: String)(implicit action: String): LiftResponse = {
```
fanf, line 28: I think you don't need that, it should not bu part of the parent trait so can be avoided

fanf (commented):
Some little changes, but it's really nice ! And adhoc api definition for internal need is rather clean and not too much verbose!

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
ncharles: it's the same as in here: https://github.com/Normation/rudder/pull/4290/files#diff-76a4f7665e1ec00753ea84eaab024d9eca6f32fed653555aba8cd9c70c14cb09L218
goal is to only use this method and phase out the getNodeIds that returns Set

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -640,6 +640,22 @@ class LDAPEntityMapper(
...
+  def entryToGroupNodeIdsChunk(e:LDAPEntry) : InventoryMappingPure[(NodeGroupId, Chunk[NodeId])] = {
+    if(e.isA(OC_RUDDER_NODE_GROUP)) {
+      for {
+        id <- e.required(A_NODE_GROUP_UUID)
+        nodeIds = e.valuesForChunk(A_NODE_UUID).map(x => NodeId(x))
+      } yield {
+        (NodeGroupId(id), nodeIds)
+      }
+    } else {
+      Thread.currentThread().getStackTrace.foreach(println)
```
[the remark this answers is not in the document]
ncharles: same behaviour as https://github.com/Normation/rudder/pull/4290/files#diff-392a77504d47edf5cb77a4792164ad3b4e876b1229638ef9eb71bbcd31c48656R639

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
fanf: I don't get it. The linked methods does not have the `allNodesAreThere` parameter: 
`def getNodeIdsChunk(allGroups: Map[NodeGroupId, Chunk[NodeId]], targets: Set[RuleTarget], allNodeInfos: Map[NodeId, NodeInfo]) : Chunk[NodeId] = ...`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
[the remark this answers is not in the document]
fanf: Well, OK, but it looks like an error in the linked place and should be removed too. If you don't see an obvious reason to keep them, can you remove them?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
ncharles: yes, but i want to replace getNodeIds by getNodeIdsChunk in the future, so having the same signature ease that

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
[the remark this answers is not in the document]
ncharles: i removed them

fanf (approved):
LGTM

PR #4301, opened 2022-05-25
Fixes #21180: Improve elm build consistency with plugins

amousset:
https://issues.rudder.io/issues/21180


VinceMacBuche (changes requested):
You need to rename in html files the path the js file is the generated file has changed

amousset:
I did not rename anything in Rudder, only in plugins

amousset:
But previous PR introduced a bug by adding `.min.`

VinceMacBuche (approved):

VinceMacBuche:
I misread the cp command, mon mauvais :felisk:

PR #4302, opened 2022-05-25
Fixes #21156: Race condition between elm builds

amousset:
https://issues.rudder.io/issues/21156

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/build-app.sh
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/build-app.sh
@@ -8,6 +8,8 @@ ELM_DIR="$( cd "$( dirname "$0" )" && pwd )"
 PROJECTS=("notifications")
 for PROJECT in ${PROJECTS}; do
   cd ${ELM_DIR}/${PROJECT}
+  # To make sure we build up-to-date stuff and avoid corruption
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/build-app.sh
```diff
@@ -8,6 +8,8 @@ ELM_DIR="$( cd "$( dirname "$0" )" && pwd )"
 PROJECTS=("notifications")
 for PROJECT in ${PROJECTS}; do
   cd ${ELM_DIR}/${PROJECT}
+  # To make sure we build up-to-date stuff and avoid corruption
+  rm -rf elm-stuff
```
VinceMacBuche, line 12: I don't want to download everything at each build when developing it's ok to me until 7.2, and in 7.2 don't do it in dev build

VinceMacBuche (approved):

webapp/sources/rudder/rudder-web/src/main/elm/build-app.sh
amousset: Yes I will do it only for release builds starting from 7.2 :+1:

amousset:
OK, merging this PR

PR #4317, opened 2022-06-09
Fixes #21241: API to fetch nodes + software times out on large instance

ncharles:
https://issues.rudder.io/issues/21241

This PR improves performance of node API when we fetch softwares, and overall some of the perfs for software related topic
However, it is not sufficient to make improvement enough to get many softwares out of Ruder with the API:
* getting inventory with software from 1 node takes 10s with this change (never ended before).
* getting inventories with software from 111 nodes don't finish

Adding an index on softwareId makes the inventory from 1 node respond in 1 second, and 20s for the 111 nodes


webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/ReadOnlySoftwareInventoryDAOImpl.scala
```diff
@@ -108,12 +133,25 @@ class ReadOnlySoftwareDAOImpl(
                                  s
                                }
                         } yield {
-                          (NodeId(id), vs)
+                          (NodeId(id), vs.toSeq)
```
ncharles, line 136: i wanted to use valuesForChunk but it's not there yet in 6.2

ncharles:
PR updated with a new commit

ncharles:
PR updated with a new commit

fanf (approved):

ncharles:
OK, squash merging this PR

ncharles:
OK, merging this PR

PR #4327, opened 2022-06-15
Fixes #21220: When upgrading to 7.1, techniques are not upgraded to the new format and reporting issue occur

VinceMacBuche:
https://issues.rudder.io/issues/21220

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
@@ -238,11 +238,11 @@ object JsonResponseObjects {
     def from (elem : MethodElem, methods : Map[BundleName,GenericMethod]): JRTechniqueElem = {
        elem match {
          case c : MethodCall =>
-           val renameParam = methods.get(c.methodId).map(_.renameParam).getOrElse(Nil)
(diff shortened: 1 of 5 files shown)


ncharles (approved):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -238,11 +238,11 @@ object JsonResponseObjects {
     def from (elem : MethodElem, methods : Map[BundleName,GenericMethod]): JRTechniqueElem = {
        elem match {
          case c : MethodCall =>
-           val renameParam = methods.get(c.methodId).map(_.renameParam).getOrElse(Nil)
+           val newCall = MethodCall.renameParams(c,methods)
```
fanf, line 241: newcall is not used here, is it a debug info remaining or a bug?

VinceMacBuche: Since techniques (and therefore method calls) are all read from json and now json is always ok (becasue we correct them on read now) and thus the renameParams does nothing really. I kept them to be 1000% sure but I'm 100% without them (in fact, I'm quite sure the renameParams sould only be done at the beginning) 

We can in all case, change add the usage of 'newCall' variable here to become 1000% sure again

PR #4334, opened 2022-06-17
Fixes #20603: Reports on method using iterator are wrong in the cli output

ncharles:
https://issues.rudder.io/issues/20603

This PR changes the way ncf techniques are generated to support (at least partially) iterators.
The main idea is to create a *bundle* that will set the context and call the effective generic method. This is necessary so that any iterator within the arguments of the generic method are not iterated once for the reporting context, and then once for the generic method

One bundle is created by generic method, named by the promiser of the generic method plus an increment id.

The resulting code looks like

```
    "id1_${report_data.directive_id}" usebundle => id1_0("Customized component", "${node.properties[apache_package_name]}", "id1", "${node.properties[apache_package_name]}", "2.2.11"),
                                             if => concat("(debian)");`
```
and the bundle doing the action being

```
bundle agent id1_0(arg_0, arg_1, arg_2, arg_3, arg_4) {
  methods:
    "id1_${report_data.directive_id}" usebundle => _method_reporting_context_v4("${arg_0}","${arg_1}","${arg_2}");
    "id1_${report_data.directive_id}" usebundle => package_install_version("${arg_3}","${arg_4}");

}
```

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -593,13 +593,114 @@ class ClassicTechniqueWriter(basePath : String, parameterTypeService: ParameterT
     s"""_method_reporting_context_v4("${component}", "${value}","${methodCall.id}")"""
   }
 
+  def reportingContextInBundle(args: Seq[String]) = {
+    s"_method_reporting_context_v4(${convertArgsToBundleCall(args)})"
+  }
+
+
+  def convertArgsToBundleCall(args:Seq[String]) : String = {
+    args.map(escapeCFEngineString(_)).map(""""${""" + _ + """}"""").mkString(",")
+  }
 
 
   def writeAgentFiles( technique : EditorTechnique, methods : Map[BundleName, GenericMethod] )  : IOResult[Seq[String]] = {
 
+    // increment of the bundle number in the technique, used by createCallingBundle
+    var bundleIncrement = 0
+
     val bundleParams = if (technique.parameters.nonEmpty) technique.parameters.map(_.name.canonify).mkString("(",",",")") else ""
 
-    def bundleMethodCall( parentBlocks : List[MethodBlock])(method : MethodElem) : List[String] = {
+    // generate a bundle which encapsulate the method_reporting_context + the actual method call
+    // and the method to call this bundle
+    // Params are:
+    // * condition when to call this bundle
+    // * the methodCall itself from the technique editor
+    // * the class parameter _value_
+    // * all the parameters already converted to cfengine format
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
I think the name of the parameter in the "id" method should me more explicit, ie the name of the method parameters for real parameter, and for the "reporting" value some more explicite name (component, value, id), it would be more explicit when debugging. (but here a problem on name colision, if there is a paramter that is name "id" or "value" 

Also the name og the bundle could be different like "call_<id>_index" but it's ok.

I think i'll merge this like this and we may do another pr (or not)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -593,13 +593,97 @@ class ClassicTechniqueWriter(basePath : String, parameterTypeService: ParameterT
...
+                          call.id :: Nil
+
+      // create the bundle arguments:
+      // there are 3 arguments corresponding to the reportingArgs, (nned to be quoted)
+      // the rest is for the methodArgs.
+      val bundleArguments = reportingArgs.map( x  => s""""${x}"""")  ::: params.toList
+
+      val argsList = bundleArguments.zipWithIndex.map {case (_, id) => "arg_" + id}
+
+      val bundleName = (call.id + "_" + bundleIncrement).replaceAll("-", "_")
```
ncharles, line 646: this may be not unique. We ought to prefix it with the technique name probably
Most likely technique_name_generic_method_name_incr would be more suited

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
VinceMacBuche: what would make this not unique ? rudder_reporting ?

VinceMacBuche:
If found several issues, especially the use of the wrong bundle (effective bundle and not log_na_rudder in rudder_reporting.cf) 

I fixed some of them in 
https://github.com/ncharles/rudder/pull/12

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
VinceMacBuche: we were having problems of non execution due to call.id being reused when the technique was applied twice, and the problem was that the promiser was not unique

VinceMacBuche: but the technique file will be created once

ncharles:
there is still an issue with log_na

VinceMacBuche (approved):

PR #4338, opened 2022-06-20
Fixes #21117: Moving create node plugin into Rudder

fanf:
https://issues.rudder.io/issues/21117

Add the node create API to Rudder. 
Also refactored out of InventoryParsing the OS detection part so that the code is not duplicated, it's already brittle enought like that, we really don't want to have an other place to change when we add support for new OS.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestUtils.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestUtils.scala
@@ -42,6 +42,7 @@ import com.normation.eventlog.ModificationId
 import com.normation.rudder.UserService
 import com.normation.rudder.api.ApiVersion
 import com.normation.utils.StringUuidGenerator
+
 import net.liftweb.common.Box
 import net.liftweb.common.EmptyBox
 import net.liftweb.common.Failure
@@ -49,11 +50,23 @@ import net.liftweb.common.Full
 import net.liftweb.common.Loggable
 import net.liftweb.http.Req
 import net.liftweb.http._
-import net.liftweb.http.js.JsExp
+import net.liftweb.http.provider.HTTPCookie
 import net.liftweb.json.JsonAST.RenderSettings
 import net.liftweb.json.JsonDSL._
 import net.liftweb.json._
 import net.liftweb.util.Helpers.tryo
+
+
+/*
+ * A JsonResponse able to keep pretty format is asked to.
+ * It's basically a copy/post of lift JsonResponse
+ */
+case class JsonResponsePrettify(json: JValue, headers: List[(String, String)], cookies: List[HTTPCookie], code: Int, prettify: Boolean) extends LiftResponse {
+  def toResponse = {
+    val bytes = (if(prettify) RestUtils.render(json) else compactRender(json)).getBytes("UTF-8")
+    InMemoryResponse(bytes, ("Content-Length", bytes.length.toString) :: ("Content-Type", "application/json; charset=utf-8") :: headers, cookies, code)
+  }
+}
 
 /**
  */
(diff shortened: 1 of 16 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestUtils.scala
```diff
@@ -145,12 +145,9 @@ object RestUtils extends Loggable {
...
                   ( "result" -> status.status ) ~
                   ( status.container   ->  message )
-    val content : JsExp = new JsExp {
-      lazy val toJsCmd = if(prettify) render(json) else compactRender(json)
-    }
-
-    JsonResponse(content,List(),List(), status.code)
+    val content = if(prettify) render(json) else compactRender(json)
 
+    InMemoryResponse(content.getBytes(),List(),List(), status.code)
```
VinceMacBuche, line 150: Why changing the response type ? I think it needs header stating its a json response

VinceMacBuche: JsonResponse is defining two headers: 


    InMemoryResponse(bytes, ("Content-Length", bytes.length.toString) :: ("Content-Type", "application/json; charset=utf-8") :: headers, cookies, code)

You should at least define them

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/MockServices.scala
```diff
@@ -41,22 +41,22 @@ import com.normation.GitVersion
 import com.normation.appconfig.ConfigRepository
 import com.normation.appconfig.GenericConfigService
 import com.normation.appconfig.ModifyGlobalPropertyInfo
-import com.normation.cfclerk.domain._
```
VinceMacBuche, line 44: Intellij redefined import witn *

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
```diff
@@ -37,67 +37,39 @@
...
-import java.io.ByteArrayInputStream
-import java.io.IOException
-import java.io.InputStream
-import java.io.OutputStream
-import java.io.PipedInputStream
-import java.io.PipedOutputStream
-import java.net.ConnectException
-import java.nio.charset.StandardCharsets
-import java.util.Arrays
+import com.normation.box.*
```
VinceMacBuche, line 44: Imports with *

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestUtils.scala
fanf: the answer was not really a json response: it was text but in a `JsExp` construct. We did that to allow formatting without having to change lift JsonResponse. 
This was breaking test that try to test the actual Json structure, since it was not a json structure but just formatted string. 
I will look for a better solution or if not, include the provided headers.

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestUtils.scala
fanf: I created a new `JsonResponse` derivative to handle the case

VinceMacBuche (approved):

fanf:
PR rebased

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

PR #4340, opened 2022-06-20
Fixes #21302: Wrong target for rudderc install

amousset:
https://issues.rudder.io/issues/21302


amousset:
PR updated with a new commit

PR #4345, opened 2022-06-22
Fixes #21247: Add API endpoint to export/import a rules, their directives, techniques, groups in an archive

fanf:
https://issues.rudder.io/issues/21247

Only the export part in that PR. 
The syntax for the API is: 
```
 Request format:
   ../archives/export/rules=rule_ids&directives=dir_ids&techniques=tech_ids&groups=group_ids&include=scope
 Where:
 - rule_ids = xxxx-xxxx-xxx-xxx[,other ids]
 - dir_ids = xxxx-xxxx-xxx-xxx[,other ids]
 - group_ids = xxxx-xxxx-xxx-xxx[,other ids]
 - tech_ids = techniqueName/1.0[,other tech ids]
- scope = all (default), none, directives, techniques (implies directive), groups
```
Ssytem items are filtered out, it seems that it would cause more pain an complication than the opposite (with in addition the case of special target to handle). 

Interesting parts:

- zip is an horrible format. 
- there is a feature switch to enable the feature in settings (`rudder_featureSwitch_archiveApi=enalbed,disabled`)
- we use item names and not UUID for file names (UUID is still of course in the serialized json), with a simple name mangling that replace non ascii chars to their ascii correspondance (as utf and java intended)
- revision works, you can ask for specific revision of items (but the `+` in url must be encoded as `%2B` because urls)

It needs https://github.com/Normation/rudder/pull/4343 to be merged before it

### DIFF
--- a/webapp/sources/api-doc/code_samples/curl/archives/export.sh
+++ b/webapp/sources/api-doc/code_samples/curl/archives/export.sh
@@ -1,0 +1,1 @@
+curl --header "X-API-Token: yourToken" https://rudder.example.com/rudder/api/latest/archives/export/rules=8e522a3e-aa45-43cd-bf33-4e8d04a566f6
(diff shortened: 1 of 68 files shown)


fanf:
PR updated with a new commit

PR #4347, opened 2022-06-23
Fixes #21322: Using '{' in node property lead to error

ElaadF:
https://issues.rudder.io/issues/21322


VinceMacBuche:
Fixed in #4348

PR #4348, opened 2022-06-23
Fixes #21322: Using '{' in node property lead to error

fanf:
https://issues.rudder.io/issues/21322

Sorting properties in returned JSON request is to ease tests and be consistent with what we do in 7.0 and latter.

The real change is just not serializing then parsing again the property when it's already a JValue.

Other changes are just for test infra.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
@@ -694,9 +694,8 @@ final case class RestExtractorService (
         }
         (for {
           _ <- PropertyParser.validPropertyName(nameValue)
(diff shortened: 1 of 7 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala
```diff
@@ -694,9 +694,8 @@ final case class RestExtractorService (
         }
         (for {
           _ <- PropertyParser.validPropertyName(nameValue)
-          p <- NodeProperty.parse(nameValue, GenericProperty.serializeJson(value), inheritMode, provider)
```
fanf, line 697: this is the heart of the correction

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #4349, opened 2022-06-24
Fixes #21175: Impossible to put a Rule in a subcategory

ElaadF:
https://issues.rudder.io/issues/21175
close this PR not the correct target branch, go here : https://github.com/Normation/rudder/pull/4355


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -944,7 +944,11 @@ class RuleApiService14 (
     def listCatAcc(categories: List[RuleCategory], acc: List[RuleCategoryId]) : List[RuleCategoryId] = {
       categories match {
         case Nil => acc
-        case c :: t => listCatAcc(t, c.id :: acc)
+        case c :: t =>
```
VinceMacBuche, line 947: you never explore the other subcategory if the head category is not empty

VinceMacBuche: should be something like this : 

val children =  listCatAcc(c.childs, c.id :: acc) 
listCatAcc(t, children)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -944,7 +944,11 @@ class RuleApiService14 (
     def listCatAcc(categories: List[RuleCategory], acc: List[RuleCategoryId]) : List[RuleCategoryId] = {
       categories match {
         case Nil => acc
-        case c :: t => listCatAcc(t, c.id :: acc)
```
fanf, line 947: We have the exact same code in 7.0, the bug was not present in that branch?

PR #4355, opened 2022-06-28
Fixes #21175: Impossible to put a Rule in a subcategory

ElaadF:
https://issues.rudder.io/issues/21175

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/RuleTests.scala
+++ b/webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/RuleTests.scala
@@ -1,0 +1,150 @@
+/*
+*************************************************************************************
+* Copyright 2022 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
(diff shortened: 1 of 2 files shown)


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/RuleTests.scala
```diff
@@ -0,0 +1,132 @@
...
+        , RuleCategoryId("cat1")
+        , RuleCategoryId("cat2")
+        , RuleCategoryId("cat3")
+        , RuleCategoryId("cat4")
+        , RuleCategoryId("subcat4")
+        , RuleCategoryId("subsubcat4")
+      )
+      restTestSetUp.ruleApiService14.listCategoriesId(root) shouldEqual(listCatIds)
+    }
+
```
fanf, line 106: you should add a test that you only get cat4 and children if starting from cat4 (locality) and that you correctly get nothing if non existing cat and that you only get one cat starting from cat1

fanf (commented):
LGTM modulo missing tests, you can merge after adding them

PR #4367, opened 2022-07-07
Fixes #21321: Add API endpoint to import an archive of rules/etc

fanf:
https://issues.rudder.io/issues/21247
 (before: https://issues.rudder.io/issues/21321)

This PR add the possiblity to export and import via upload zip archives to rudder. 

### Limitation are:

- we don't check much for now. Things that should be tested: technique and version used in directive, existence of provided group id for new groups, existence of groups/dierctives used in rule.
- the save is not even a bit atomic. We save technique, reload technique library, then directives, then groups, then rules. All that sould be done in one go, but we don't have underlying infrastructure yet. It might prove to be a showstopper that needs to be addressed before release (because here, we may start semi generation). An alternative would be to write everything in git, then "reload from master", but I have doubts making that the default behavior for archive upload, without someone to take care of the mess if something goes wrong (concurrent change for ex)

### Main arch change:
- created an `EventMetadata` to factor out `changeId, actor, option msg`. We should use it everywhere in the future, it's so much lighter. Perhaps find a shorter name
- some change in `ZioUtils#accumulate` to remove some dependencies on cats (I was needed some variation of what we had already)
- move the low level part of writting tehcniques (ie the commit part) in git archiver file, change a bit its signature
- update a lot of parts of mocking framework to make it usable with zip upload - also handle some cases were binary content were causing unhandled errors in our API

### For the export part in that PR
The syntax for the API is: 
```
 Request format:
   ../archives/export/rules=rule_ids&directives=dir_ids&techniques=tech_ids&groups=group_ids&include=scope
 Where:
 - rule_ids = xxxx-xxxx-xxx-xxx[,other ids]
 - dir_ids = xxxx-xxxx-xxx-xxx[,other ids]
 - group_ids = xxxx-xxxx-xxx-xxx[,other ids]
 - tech_ids = techniqueName/1.0[,other tech ids]
- scope = all (default), none, directives, techniques (implies directive), groups
```
Ssytem items are filtered out, it seems that it would cause more pain an complication than the opposite (with in addition the case of special target to handle). 

Interesting parts:

- zip is an horrible format. 
- there is a feature switch to enable the feature in settings (`rudder_featureSwitch_archiveApi=enalbed,disabled`)
- we use item names and not UUID for file names (UUID is still of course in the serialized json), with a simple name mangling that replace non ascii chars to their ascii correspondance (as utf and java intended)
- revision works, you can ask for specific revision of items (but the `+` in url must be encoded as `%2B` because urls)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
@@ -889,6 +889,35 @@ object UserApi extends ApiModuleProvider[UserApi] {
 }
 
 /*
+ * An API for import & export of archives of objects with their dependencies
+ */
+sealed trait ArchiveApi extends EndpointSchema with GeneralApi with SortIndex {
+  override def dataContainer: Option[String] = None
+}
+object ArchiveApi extends ApiModuleProvider[ArchiveApi] {
+  /*
+   * Request format:
+   *   ../archives/export?rules=rule_ids&directives=dir_ids&techniques=tech_ids&groups=group_ids&include=scope
+   * Where:
+   * - rule_ids = xxxx-xxxx-xxx-xxx[,other ids]
+   * - dir_ids = xxxx-xxxx-xxx-xxx[,other ids]
+   * - group_ids = xxxx-xxxx-xxx-xxx[,other ids]
+   * - tech_ids = techniqueName/1.0[,other tech ids]
+   * - scope = all (default), none, directives, techniques (implies directive), groups
+   */
+  final case object ExportSimple extends ArchiveApi with ZeroParam with StartsAtVersion15 with SortIndex {val z = implicitly[Line].value
+    val description    = "Export the list of objects with their dependencies"
+    val (action, path) = GET / "archives" / "export"
+  }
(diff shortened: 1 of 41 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -888,6 +888,35 @@ object UserApi extends ApiModuleProvider[UserApi] {
...
+/*
+ * An API for import & export of archives of objects with their dependencies
+ */
+sealed trait ArchiveApi extends EndpointSchema with GeneralApi with SortIndex {
+  override def dataContainer: Option[String] = None
+}
+object ArchiveApi extends ApiModuleProvider[ArchiveApi] {
+  /*
+   * Request format:
+   *   ../archives/export/rules=rule_ids&directives=dir_ids&techniques=tech_ids&groups=group_ids&include=scope
```
VinceMacBuche, line 900: format is /archives/export?rules=....

VinceMacBuche:
I think it's ok to me apart the comment with url format with a typo (/ instead of ? ) Do you want the documentation of the api to be done in another PR ?

fanf:
PR rebased

fanf:
It seems that I lose the documentation at some point, I think I forgot to add the changes :/
It was rather long to do, I will do it in an other pr with ticket: https://issues.rudder.io/issues/21394

PR #4376, opened 2022-07-19
Fixes #21264: Refresh Button in Rules menu under Directives Tab is not working

RaphaelGauthier:
https://issues.rudder.io/issues/21264

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/DataTypes.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/rules/sources/DataTypes.elm
@@ -342,3 +342,5 @@ type Msg
   | UpdateGroupFilters     Filters
   | GoTo                   String
   | FoldAllCategories      Filters
+  | RefreshComplianceTable RuleId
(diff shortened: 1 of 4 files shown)


fanf:
there is something not working when a group is modified, we don't see the little `+2-1` next to groups tab, and so we can still miss that we need to click "save". 

For the node list: as you told me, the list is not updated. But we have the correct count of nodes next to the tab. It seems that the problem is that we take the node list from the rule details, and since compliance is not yet ready for these nodes, we don't get them.

I think that we need to open an other ticket for that, we should get the actual real list of nodes, even if compliance is not ready, as soon as the rule is saved. 
In the meantime, we should change the text accordingly to what `nodesanddirectives` api returned:
image

Something like: "that rule is applied to N nodes. P node have compliance for it" (with N coming from the same place that the number next to the nodes tab, and P being the count in the list below).

fanf (approved):

RaphaelGauthier:
OK, merging this PR

RaphaelGauthier:
OK, merging this PR

PR #4380, opened 2022-07-21
Fixes #21445: 

amousset:
https://issues.rudder.io/issues/21445


webapp/sources/rudder/rudder-web/src/main/webapp/WEB-INF/web.xml
```diff
@@ -40,6 +40,7 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
   <session-config>
    <cookie-config>
     <http-only>true</http-only>
+    <comment>__SAME_SITE_LAX__</comment>
```
amousset, line 43: https://stackoverflow.com/a/59352696

could not find it Jetty documentation but seems to work as expected.

fanf (approved):

PR #4432, opened 2022-08-18
Fixes #21621: Campaign events are not put in queue again after a new handler service was added, hence they are not treated after boot

VinceMacBuche:
https://issues.rudder.io/issues/21621


VinceMacBuche:
Commit modified

amousset (approved):

PR #4435, opened 2022-08-23
Fixes #18575: when validating change in directive, a popup appear, but the background is only half grayed out

VinceMacBuche:
https://issues.rudder.io/issues/18575


amousset (approved):

VinceMacBuche:
OK, merging this PR

PR #4446, opened 2022-08-29
Fixes #21546: When editing a rule, the groups changes do not behave like the directives changes

ElaadF:
We only follow inclusive groups, adding et removing exluded groups does not influence the count
https://issues.rudder.io/issues/21546
count_group


ElaadF:
New PR here : https://github.com/Normation/rudder/pull/4454

PR #4447, opened 2022-08-29
Fixes #21637: Popups in rules page are missing separation from content

ElaadF:
Still missing the animation, but this is a quick fix

https://issues.rudder.io/issues/21637
image
image
image


ElaadF:
new PR here : https://github.com/Normation/rudder/pull/4453

PR #4448, opened 2022-08-29
Fixes #21490: when we select "Display rule compliance columns " option in the settings, display on the directive page is not great

ElaadF:
https://issues.rudder.io/issues/21490
image


ElaadF:
New PR here : https://github.com/Normation/rudder/pull/4452

PR #4452, opened 2022-08-30
Fixes #21490: when we select "Display rule compliance columns " option in the settings, display on the directive page is not great

ElaadF:
https://issues.rudder.io/issues/21490
image


VinceMacBuche:
The commit name is not correct

ElaadF:
I'm closing this PR because the problem is more complexe, it affect the table when we resize the window

PR #4458, opened 2022-08-30
Fixes #21667: Rudder agent version is not detected for windows on pending nodes in 7.2

ncharles:
https://issues.rudder.io/issues/21667


fanf (approved):

ncharles:
> I think we want to change the name of the software within rudder agent windows, we don't want to link it to dsc

that's the name used within the inventory, to detect the software rudder agent, not the displayed name

VinceMacBuche (approved):

ncharles:
it's a really big change to change on agent side: we'll need to support old and new format, and we're too late in the development cycle to do that

VinceMacBuche:
I don't say when we should change it

PR #4467, opened 2022-09-02
Fixes #21698: Add reason field to disabled field in campains and campain events

VinceMacBuche:
https://issues.rudder.io/issues/21698

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
@@ -401,7 +401,7 @@ CREATE TABLE CampaignEvents (
   campaignId   text
 , eventid      text PRIMARY KEY
 , name         text
-, state        text
+, state        json
 , startDate    timestamp with time zone NOT NULL
 , endDate      timestamp with time zone NOT NULL
 , campaignType text
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignEventRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignEventRepository.scala
@@ -38,9 +38,9 @@
 package com.normation.rudder.campaigns
 
 import com.normation.errors.IOResult
-import com.normation.rudder.campaigns.CampaignEventState.Scheduled
 import com.normation.rudder.db.Doobie
 import doobie.Fragments
+import doobie.Meta
 import doobie.Read
 import doobie.Write
 import doobie.implicits._
@@ -48,8 +48,6 @@ import doobie.implicits.javasql._
 import doobie.implicits.toSqlInterpolator
 import org.joda.time.DateTime
 import zio.interop.catz._
-
-import java.sql.Timestamp
 
 
 trait CampaignEventRepository {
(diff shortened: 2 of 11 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignEventRepository.scala
```diff
@@ -83,7 +85,7 @@ class CampaignEventRepositoryImpl(doobie: Doobie, campaignSerializer: CampaignSe
             CampaignEventId(d._1)
           , CampaignId(d._2)
           , d._3
-          , CampaignEventState.parse(d._4).getOrElse(Scheduled)
+          , d._4.fromJson[CampaignEventState].getOrElse(CampaignEventState.parse(d._4).getOrElse(Scheduled))
```
fanf, line 88: I think the default in case of problem when unserializing should be either an error or `Skipped("error when reading from base even state: ...")`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -175,14 +192,14 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
...
-            case Finished =>
+            case Finished|Skipped(_)  =>
               for {
                 campaign <- campaignRepo.get(event.campaignId)
                 up <-  scheduleCampaignEvent(campaign)
               } yield {
                 up
               }
-            case Scheduled|Running|Skipped =>
+            case Scheduled|Running =>
```
fanf, line 202: we are going to need to draw the transition table for that, I'm sure we are going to misse edge else.

fanf (commented):
Some little things to change, you can self merge after

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignEventRepository.scala
VinceMacBuche: In fact i should even remove the first "getOrElse" ! nice idea the Skpped for error case

VinceMacBuche: I kept it first getOrElse to migrate my dev env but in fact i should not keep it

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -401,7 +401,7 @@ CREATE TABLE CampaignEvents (
   campaignId   text
 , eventid      text PRIMARY KEY
 , name         text
-, state        text
+, state        json
```
fanf, line 404: I think we should keep a simple enumeration for that. And in fact, I find that having a case class for state makes everything much more complicated. I would prefer to keep the state and its informations/details in a separated columns.

In fact, we will likely need a "summary" or whatever the name columns, json, that give meta info about the state without having to make queries to other tables. That columns was in spec, but was not implemented yet. 
`
That column would be json, dependent of `state` and `campaignType` (perhaps). For `skipped`, it would give the reason. For `success`, a summary of things that worked ("5423 noes updates"). For error, some statistics about what happened. Etc. 

So, I will validate it like that now, but I would like to have it thought again before release

fanf (approved):
We will need to come back to design choices here before release, validating for rc

PR #4472, opened 2022-09-07
Fixes #8022: Allow users to specify node hostname (FQDN)

fanf:
https://issues.rudder.io/issues/8022

Allow to override hostname of a node with an inventory hook that define the key `rudder_override_hostname`

If that key is defined, it is prefered as other places (`<RUDDER><HOSTNAME>` or else `<OPERATINGSYSTEM><FQDN>`) as the hostname for the node. In that case, an new inventory hook key is also defined with the original node hostname (`rudder_original_hostname`).

All check on hostname validity are now done in the same place (in `FusionInventoryParser`) which simplify consistency (and exposed a bad hostname in one of our test inventories).

Overall, the change is very simple and seems to be low risk, since behavior is not changed when custom properties with the magic key are not defined.


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4477, opened 2022-09-12
Fixes #21371: No link to directives in a rule if the rule is not applied to some nodes

ElaadF:
https://issues.rudder.io/issues/21371
Need to validate in a proper environment, seems to be working in a dev environment


ElaadF:
Commit modified

VinceMacBuche (approved):

PR #4480, opened 2022-09-15
Fixes #21766: Reschedule correclty campaign events when a campaign changes

VinceMacBuche:
https://issues.rudder.io/issues/21766

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
@@ -81,7 +81,7 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
   def saveCampaign(c : Campaign) = {
     for {
       _ <- campaignRepo.save(c)
-      _ <- scheduleCampaignEvent(c)
+      _ <- scheduleCampaignEvent(c, DateTime.now())
     } yield {
       c
     }
@@ -195,7 +195,7 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
             case Finished|Skipped(_)  =>
               for {
                 campaign <- campaignRepo.get(event.campaignId)
-                up <-  scheduleCampaignEvent(campaign)
+                up <-  scheduleCampaignEvent(campaign, newCampaign.start)
               } yield {
                 up
               }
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -296,11 +296,13 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
     }
   }
 
-  def scheduleCampaignEvent(campaign: Campaign) : IOResult[CampaignEvent] = {
+  def scheduleCampaignEvent(campaign: Campaign, date : DateTime) : IOResult[CampaignEvent] = {
     val now = DateTime.now()
```
fanf, line 300: I think it should use the provided date here, not now

fanf:
OK, for the general approach for that PR, modulo the `date` argument not used in `scheduleCampaignEvent`. 

For the future, we want to be able to uncouple the scheduling logic, which only deals with date and waiting, and the resulting action that deal with repository requests, or specific campaign handler. 

The alarm point for me are:
- we have query on repositories at several points. We sould not. The scheduler should just take a decision and give the answer.
- the time-related logic is complicated and will become more complicated (more states than just running, whatever). It's a state-machine, and for now we can't test it because the receiving part are too complicated.

So, I think we can uncouple the code with these guiding lines: 
- make extremely easy to test the scheduler in isolation, with as little code dep as possible (not at all would be even better): no repository call in the scheduler and we must be able to change date for tests: that gives us guiding lines to what should be passed in input parameters, and what need to be either given as output parameter or call to an external service to do the logic
- after the waiting, we don't have other operation: we just queue back an event that will be processed and starts the processing if needed. Even better: the waiting is not part of the scheduler, the scheduler just call a trait method "wait for XXX and re-queue"

With that, I thing the scheduler can be autonomous and simply tested. 
The part that processed resulting action from the scheduler can also be tested in isolation, avoiding to have bugs triggered by combination of scheduler+action processing.
And we still need integration test with both parts to check that we correctly weave the two together.

fanf (commented):
The new parameter Date seems to not be used scheduleCampaignEvent

VinceMacBuche:
Commit modified

VinceMacBuche:
> The new parameter Date seems to not be used scheduleCampaignEvent

I did not commit my last changes .........................

fanf (approved):

PR #4482, opened 2022-09-15
Fixes #21755: Impossible to move groups

ElaadF:
https://issues.rudder.io/issues/21755

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
@@ -222,10 +222,15 @@ class NodeGroupForm(
       & "group-save" #> { if (CurrentUser.checkRights(AuthorizationType.Group.Edit))
                     <span class="save-tooltip-container">
                       {
-                        SHtml.ajaxButton(<span>Save <i class="fa fa-download"></i></span>, onSubmit _)  %  ("id" -> saveButtonId) % ("class" -> " btn btn-success btn-icon")
+                        SHtml.ajaxOnSubmit(onSubmit _)(
+                          <button class="btn btn-success btn-icon ui-button ui-corner-all ui-widget" id={saveButtonId}>
+                            <span>Save <i class="fa fa-download"></i></span>
+                          </button>
+                        )
                       }
                       <span class="save-tooltip bsTooltip" title="Your 'Save' button is disabled, it means that you have updated the query without Searching for new Nodes. Please click on 'Search' to enable saving again"></span>
                     </span>
+
                    else NodeSeq.Empty
                 }
       & "group-delete" #> SHtml.ajaxButton(<span>Delete <i class="fa fa-times-circle"></i></span>, () => onSubmitDelete(), ("class" -> " btn btn-danger btn-icon"))


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
```diff
@@ -222,7 +222,7 @@ class NodeGroupForm(
       & "group-save" #> { if (CurrentUser.checkRights(AuthorizationType.Group.Edit))
                     <span class="save-tooltip-container">
                       {
-                        SHtml.ajaxButton(<span>Save <i class="fa fa-download"></i></span>, onSubmit _)  %  ("id" -> saveButtonId) % ("class" -> " btn btn-success btn-icon")
+                        SHtml.ajaxSubmit("Save", onSubmit _)  %  ("id" -> saveButtonId) % ("class" -> " btn btn-success")
```
ElaadF, line 225: Not an easy one for me, I'm still not so sure about how it works, but it seems to resolve the bug
This is how it was done in 7.0

ElaadF, line 225: Not an easy one for me, I'm still not so sure about how it works, but it seems to resolve the bug
This is how it was done in 7.0

VinceMacBuche: You need to keep the element as before, just replace ajaxButton by ajaxSubmit

VinceMacBuche: Submit, applies all changes in the form before executing it, Ajaxbutton executes it without the changes from the form. so it explains why the changes were not kept

ElaadF: W

> You need to keep the element as before, just replace ajaxButton by ajaxSubmit

We can't give HTML to `ajaxSubmit` like `ajaxButton` (for the icon), we need to find another way to add the dowload icon at the end of the button

ElaadF:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
```diff
@@ -222,10 +222,12 @@ class NodeGroupForm(
       & "group-save" #> { if (CurrentUser.checkRights(AuthorizationType.Group.Edit))
                     <span class="save-tooltip-container">
                       {
-                        SHtml.ajaxButton(<span>Save <i class="fa fa-download"></i></span>, onSubmit _)  %  ("id" -> saveButtonId) % ("class" -> " btn btn-success btn-icon")
+                          <button class="btn btn-success btn-icon ui-button ui-corner-all ui-widget" id="groupSaveButtonId"> <span>Save <i class="fa fa-download"></i></span></button>
+                          SHtml.ajaxOnSubmit(onSubmit _)
```
ElaadF, line 226: I don't understand how `ajaxOnSubmit` works, I get this in UI :
image

ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
```diff
@@ -222,10 +222,15 @@ class NodeGroupForm(
       & "group-save" #> { if (CurrentUser.checkRights(AuthorizationType.Group.Edit))
                     <span class="save-tooltip-container">
                       {
-                        SHtml.ajaxButton(<span>Save <i class="fa fa-download"></i></span>, onSubmit _)  %  ("id" -> saveButtonId) % ("class" -> " btn btn-success btn-icon")
+                        SHtml.ajaxOnSubmit(onSubmit _)(
+                          <button class="btn btn-success btn-icon ui-button ui-corner-all ui-widget" id="groupSaveButtonId">
```
VinceMacBuche, line 226: I think you should try to give the id attribute from the saveButtonId val

try something like id="${saveButtonId}"

If saveButtonId was only used here, you may delete it i would be quite happy too

ElaadF: It's used in other places

VinceMacBuche (approved):

ElaadF:
PR rebased

PR #4486, opened 2022-09-19
Fixes #21784: add a default value for syslog facility output for agent and server compontent

ncharles:
https://issues.rudder.io/issues/21784

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
@@ -8,6 +8,9 @@
     },
     "server": {
       "cf_serverd_bind_address": "::"
+    },
+    "log": {
+      "syslog_facility": "LOG_LOCAL3"
     }
   },
   "provider": "system"


webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
```diff
@@ -8,7 +8,8 @@
     },
     "server": {
       "cf_serverd_bind_address": "::"
-    }
+    },
+    "syslog_facility": "LOG_LOCAL3"
```
amousset, line 12:
```suggestion
    "agent_syslog_facility": "LOG_LOCAL3"
```

amousset: IIRC it is only a setting for the agent

fanf: if it's for agent only, it should be `"agent":  { "syslog_facility" : "LOG_LOCAL3" }`

If it's for both, it seems worthy to have a "log" category, it seems common enough to be useful. I'm not sure why but I would prefer to minimize number of top level props.

ncharles: it's for both cf-agent and cf-serverd

fanf: so, for now we will put it in log, and we will try to split / make it more granular if we want/can in further releases:
```
{ ...
"log": { "syslog_facility" : "LOG_LOCAL3" }
...
}
```

ncharles:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/rudder-system-global-parameter.conf
ncharles: Pr updated

fanf (approved):

fanf:
OK, squash merging this PR

PR #4487, opened 2022-09-19
Fixes #21783: Switch to jsonb for campaign event state

VinceMacBuche:
https://issues.rudder.io/issues/21783

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/Migration/dbMigration-7.1-7.2-create-campaign-table.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/Migration/dbMigration-7.1-7.2-create-campaign-table.sql
@@ -43,9 +43,10 @@ CREATE TABLE CampaignEvents (
   campaignId   text
 , eventid      text PRIMARY KEY
 , name         text
-, state        json
+, state        jsonb
 , startDate    timestamp with time zone NOT NULL
 , endDate      timestamp with time zone NOT NULL
 , campaignType text
 );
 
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/resources/Migration/dbMigration-7.1-7.2-create-campaign-table.sql
```diff
@@ -43,9 +43,10 @@ CREATE TABLE CampaignEvents (
   campaignId   text
 , eventid      text PRIMARY KEY
 , name         text
-, state        json
+, state        jsonb
```
amousset, line 46: how will it work for systems with the rc installed? Do we need to delete the table?

fanf: no, we "just" need a migration script that says: 
```
ALTER TABLE CampaignEvents
  ALTER COLUMN state
  SET DATA TYPE jsonb
  USING state::jsonb;
```

fanf: ok, so we are choosing to not support update from rc (only one user with only test data).
We will need to give that user a process to migrate ("before upgrade, remove plugin, delete table `CampaignEvents`, delete plugin data directory in /var/rudder/..., do upgrade, install plugin")

fanf (approved):

PR #4489, opened 2022-09-19
Fixes #21730: on a fresh 7.1 install, the number of group displayed for "Global configuration of all nodes" is invalid

ElaadF:
https://issues.rudder.io/issues/21730


fanf:
The rule format used to be ok, and we should still support it.
That being said, I'm not sure about when we switch and if it's really relevant to continue to support it.

For the short run, your change is OK. I will note for 7.3 to deprecated (or remove) the support for old format.

fanf (approved):

PR #4493, opened 2022-09-20
Fixes #21792: Switch campaign schedule from a duration to a end time

VinceMacBuche:
https://issues.rudder.io/issues/21792

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
@@ -45,7 +45,6 @@ import zio.json.jsonDiscriminator
 import zio.json.jsonField
 import zio.json.jsonHint
 
-import scala.concurrent.duration.Duration
 
 
 trait Campaign {
@@ -61,7 +60,6 @@ case class CampaignInfo (
   , description : String
   , status : CampaignStatus
   , schedule : CampaignSchedule
-  , duration: Duration
 )
 
 case class CampaignId (value : String, rev: Revision = GitVersion.DEFAULT_REV) {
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
```diff
@@ -127,18 +126,26 @@ case object Sunday extends DayOfWeek{
   val value = 7
 }
 
+case class DayTime (
+   day : DayOfWeek
+ , hour : Int
+ , minute : Int
+) {
+  val realHour = hour % 24
+  val realMinute = minute % 60
```
fanf, line 135: shouldn't day of week get a %7 too ?

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
fanf: ok, ignore, it's not an int, I'm not sure what I saw there

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -225,33 +225,36 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
     def start() = loop().forever
   }
 
-  def nextCampaignDate(schedule : CampaignSchedule, date : DateTime) : IOResult[DateTime] = {
+  def nextDateFromDayTime(date : DateTime, start : DayTime) : DateTime = {
```
fanf, line 228: this will need comments and unit tests, I spent quite some time understanding the logic here (if I got it: "date" is before; start, increment by one week but get all other time info from "start". I'm not sure about what happen if date is several weeks behind start, perhaps it should not happen ?)

fanf (approved):

VinceMacBuche:
Commit modified

PR #4496, opened 2022-09-20
Fixes #21795: Add security headers in Lift config

amousset:
https://issues.rudder.io/issues/21795

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -244,7 +244,7 @@ class Boot extends Loggable {
 
     //exclude Rudder doc from context-path rewriting
     LiftRules.excludePathFromContextPathRewriting.default.set(() => (path:String) => {
-      val noRedirectPaths= "/rudder-doc" :: "/ncf" :: "/ncf-builder" :: Nil
+      val noRedirectPaths= "/rudder-doc" :: Nil
       noRedirectPaths.exists(path.startsWith)
     })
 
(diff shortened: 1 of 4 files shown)


amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -244,7 +244,7 @@ class Boot extends Loggable {
 
     //exclude Rudder doc from context-path rewriting
     LiftRules.excludePathFromContextPathRewriting.default.set(() => (path:String) => {
-      val noRedirectPaths= "/rudder-doc" :: "/ncf" :: "/ncf-builder" :: Nil
+      val noRedirectPaths= "/rudder-doc" :: Nil
```
fanf, line 247: I think that's a bug and should be removed for all 7.x branches

fanf (approved):
Seems good, but we will need to test for real. I don't know if we can unit test that with lift test facilities

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
amousset: It is a leftover that does not exist anymore, not really a bug per se IMHO

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
@fanf added a setting for HSTS as it is not always safe to enable.

amousset:
I still need to be able to test these.

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

PR #4502, opened 2022-09-22
Fixes #21809:  Dont redirect other tabs to login form when session cookie is replaced - add a notification

amousset:
https://issues.rudder.io/issues/21809

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -38,7 +38,7 @@
 package bootstrap.liftweb
 
 import net.liftweb.http._
-import net.liftweb.http.js.JsCmd
+import net.liftweb.http.js.JE.JsRaw
 import net.liftweb.common._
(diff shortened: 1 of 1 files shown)


amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -360,7 +360,7 @@ class Boot extends Loggable {
     // Do nothing instead, as at allows to keep open tabs context until we get the new cookie
     // This does not affect security as it is only a redirection anyway and did not change
     // the session itself.
-    LiftRules.noCometSessionCmd.default.set(() => JsCmd.unitToJsCmd(()))
+    LiftRules.noCometSessionCmd.default.set(() => JsRaw(s"createErrorNotification('You have been signed out. Please reload the page to sign in again.')").cmd)
```
fanf, line 363: We might check in the future if we want to do the same thing for `noAjaxSessionCmd`

amousset:
OK, squash merging this PR

PR #4507, opened 2022-09-22
Fixes #21825: Port code to ZIO2

fanf:
https://issues.rudder.io/issues/21825

As explained in the ticket, we have a very useful migration from where you copy&past things: https://zio.dev/guides/migrate/zio-2.x-migration-guide/

Main anoying/unintersting changes: 
- there is a bug in zio-json (https://github.com/zio/zio-json/issues/622) that forces us to not use codec (see [ExpectedReports.scala](https://github.com/Normation/rudder/pull/4507/files#diff-223450414fa9cd87b265b4f5bc93a25dae3bd733e366c8008208e5360195a750))
- lots of renaming, the most pervasive being `.effect` to `.attempt`. Good to understand, bad for future merge. For consistency, I also renamed `IOResult.effect` into `IOResult.attemp`
- lots of others little renaming things like no more `UIO.unit` but `ZIO.unit`, `RefM` into `Ref.Synchronized`, all `.xxxM` methods into `.xxxZIO` methods, etc.
- A bit more involving, no more `Managed` but `ZIO.scope(...).flatMap(...)` and no more `bracket` but `ZIO.acquireReleaseWith` 
- `zio` package as a `System` object, so we need to `import zio.{System => _, _}` to avoid conflict with `java.lang.System`
- no more `zio.duration._` package, the implicits for `.seconds` etc are in `zio._`
- parallelism in traverse is not `.foreachParN(...)(N)` anymore but `.foreachPar(...).withParallelism(N)`
- queue method getAll etc returns `Chunk` so it's simpler to use chunk elsewhere around them

More interesting:

- `untraced` is not needed anymore: trace information are compile-time now and have almost not overhead at runtime
- no need to pass `clock` around, they are now part of `ZIO` based effects always here. It's also much more easier to access it for sleeping (see for ex [Scheduler.scala](https://github.com/Normation/rudder/pull/4507/files#diff-c618252426fbe1b029c1c9b066644a78efde2f0936a1d5aa24e7111d825b64d5)

The amazing

- they corrected https://github.com/zio/zio/issues/1275 which means that WE DON'T HAVE TO CARE ANYMORE IF THINGS BLOCK!
ZIO will try to do everything non-blocking and migrate the blocking effects in a dedicated threadpool. We can advise it, but it learn by itself what methods block and what don't! 
So we gain:
- no more deadlock due to blocking effects that we didn't know about (rare case now for us, since by default, we told that everything was blocking -see https://github.com/Normation/rudder/pull/4507/files#diff-d4590d1d8c70916376ea07260ff5f905f68f324e53b5298c5e9038fbe913fc15R95)
- no more special case for hotpoint, like [LDAPConnectionProvider.scala](https://github.com/Normation/rudder/pull/4507/files#diff-b9abdc511c6ce9db807dbbb8310bd3a91b65cbd6c62b083bf0ea7fc05d9e2e14)
- hopefully more consistant performance after a warm-up

And finally, the rewrite of [CronParser.scala](https://github.com/Normation/rudder/pull/4507/files#diff-a5d9d49bde6a00fd487a2447f2b62195b0649fbd5b7f15348ed0ca4211bdcc19) was kindly provided on ZIO chan.

We were lucky to not be using the IoC/`Has` framework which changed A LOT. Streams, too. All in all, the port is rather simple and for the hardest part, the community was extremelly helpfull.

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/AgentTypes.scala
+++ b/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/AgentTypes.scala
@@ -232,7 +232,7 @@ object AgentInfoSerialisation {
    */
   def parseJson(s: String, optToken : Option[String]): IOResult[AgentInfo] = {
     for {
-      json      <- IO.effect { parse(s) } mapError  { ex => InventoryError.Deserialisation(s"Can not parse agent info: ${ex.getMessage }", ex) }
+      json      <- ZIO.attempt { parse(s) } mapError  { ex => InventoryError.Deserialisation(s"Can not parse agent info: ${ex.getMessage }", ex) }
       agentType <- (json \ "agentType") match {
                      case JString(tpe) => AgentType.fromValue(tpe).toIO
                      case JNothing     => InventoryError.SecurityToken("No value defined for security token").fail
@@ -246,7 +246,7 @@ object AgentInfoSerialisation {
                 case JObject(json) => parseSecurityToken(agentType, json, optToken)
                 case _             => parseSecurityToken(agentType, JNothing, optToken)
               }).toIO
-      capabilities <- IOResult.effect(json \ "capabilities" match {
+      capabilities <- IOResult.attempt(json \ "capabilities" match {
                         case JArray(capa) => capa.flatMap(c => c match {
                           case JString(s) => Some(AgentCapability(s))
                           case _          => None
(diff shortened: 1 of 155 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

ncharles:
Tested on the load platform. No obvious perf improvment or degradation; policy generation seems ~5% faster, dynamic group computation time has less variability

fanf:
> Tested on the load platform. No obvious perf improvment or degradation; policy generation seems ~5% faster, dynamic group computation time has less variability

amazing! (it is, because I removed all the hand-optimisation of effect/effectBlocking cherry picking, especially the one on `LDAPConnection` without which we were able to run rudder on the load plateform)

fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
PR rebased

VinceMacBuche (approved):

PR #4519, opened 2022-09-30
Fixes #21891: A namevirtualhost is mandatory when splitting vhost

peckpeck:
https://issues.rudder.io/issues/21891

### DIFF
--- a/relay/sources/apache/rudder-vhost.conf
+++ b/relay/sources/apache/rudder-vhost.conf
@@ -30,12 +30,12 @@
 </VirtualHost>
 
 # #############################################################################
-# # To use a different certificate/port/configuration for Web/API access and 
+# # To use a different certificate/port/configuration for Web/API access and
 # # and node communication, comment the default virtual host and uncomment and
 # # modify the following configuration.
 # #
 # # This is useful for:
-# # 
+# #
 # # * using a different certificate for the web UI/API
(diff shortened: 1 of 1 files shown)


fanf (approved):

relay/sources/apache/rudder-vhost.conf
```diff
@@ -46,13 +46,17 @@
 # # If you change the port for one of the virtual hosts you will likely need to add
 # # Listen 8443
 #
+# # If you already have this you can remove it
+# NameVirtualHost *:443
```
amousset, line 50: this has no effect on apache 2.4

amousset: And I think we don't have 2.2 anywhere

amousset:
```
In 2.3.11 and later, any time an IP address and port combination is used in multiple virtual hosts, name-based virtual hosting is automatically enabled for that address.

This directive currently has no effect.
```

https://httpd.apache.org/docs/2.4/mod/core.html#namevirtualhost

amousset: and we already successfully split the virtual host without it, so I don't think it changes something.

peckpeck:
Let's close it and reopen if we encouter it again

PR #4552, opened 2022-10-26
Fixes #21963: Make rudderc technique able to be used by the technique editor

amousset:


policies/rudderc/src/backends/metadata.rs
```diff
@@ -4,87 +4,298 @@
...
+impl Section {
+    fn from(resources: Vec<ResourceKind>) -> Vec<Self> {
+        // recursive function to handle blocks. Block recursion levels shouldn't be a
+        // stack overflow threat.
+        resources
+            .into_iter()
+            .map(|r| match r {
+                ResourceKind::Block(b) => Section {
+                    name: b.name,
+                    id: b.id,
```
Fdall, line 178: Why is this the only id that is not cloned?

amousset: Because it is moved (as we do not reuse it after)

Fdall (approved):

amousset:
OK, squash merging this PR

amousset:
OK, merging this PR

PR #4563, opened 2022-11-10
Fixes #14371: At least on windows, technique editor parameters can not handle "_" char

VinceMacBuche:
https://issues.rudder.io/issues/14371

So, the problem is that the parameter name of the technique is:
- "transformed" to be a valide power shell name in the list of parameters (ie the name of the parameter is changed to removed "_"), 
- not transformed when call with `${}`
The mismatch between declaration and call-site lead to the problem. 

Plus, it seems that we don't need a "power shell valid name" for the parameter name, and that "_" are ok.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
@@ -1120,7 +1120,7 @@ class DSCTechniqueWriter(
       case params =>
         params
           .map(p => s"""      [parameter(Mandatory=$$true)]
-             |      [string]$$${p.name.validDscName},""")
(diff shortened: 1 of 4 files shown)


VinceMacBuche:
Commit modified

fanf:
we need unit tests on that, else we will get into the same problem on the next code change (or do the opposite and forget why we did that)

VinceMacBuche:
There is some unit test on the generated technique, what other tests do you need ?

fanf:
Not sur. Ideally something that test the invariant we are correcting on that PR, ie that the param as the same name than what is used in ${..}. 
Perhaps looking for the variable name in the metadata.xml/technique code and checking it's the same? 

I don't know, just here the test don't test that our change makes sens, it only tests that something changed.

VinceMacBuche:
OK, merging this PR

PR #4564, opened 2022-11-14
Fixes #21982: There are no visual indication that a field contains multiline text in technique editor

RaphaelGauthier:
https://issues.rudder.io/issues/21982


RaphaelGauthier:
tech-editor-auto-height-input

PR #4582, opened 2022-11-23
Fixes #22129: Id to start looking for report json should not be zero but highest id.

VinceMacBuche:
https://issues.rudder.io/issues/22129

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/JSONReportsHandler.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/JSONReportsHandler.scala
@@ -70,12 +70,15 @@ case class JSONReportsAnalyser(reportsRepository: ReportsRepository, propRepo: R
     handlers.map(_.handle).fold(base) { case (a, b) => b orElse a }(report)
   }
   def loop = {
+    val highestId = reportsRepository.getHighestId().getOrElse(0L)
     for {
-      lowerId <- propRepo.getReportHandlerLastId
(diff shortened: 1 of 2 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/JSONReportsHandler.scala
```diff
@@ -70,12 +70,15 @@ case class JSONReportsAnalyser(reportsRepository: ReportsRepository, propRepo: R
     handlers.map(_.handle).fold(base) { case (a, b) => b orElse a }(report)
   }
   def loop = {
+    val highestId = reportsRepository.getHighestId().getOrElse(0)
     for {
-      lowerId <- propRepo.getReportHandlerLastId
-      reports <- reportsRepository.getReportsByKindBetween(lowerId.getOrElse(0), None, 1000, List(Reports.REPORT_JSON)).toIO
+      optLowerId <- propRepo.getReportHandlerLastId
+      lowerId     = optLowerId.getOrElse(highestId)
+      reports    <- reportsRepository.getReportsByKindBetween(lowerId, None, 1000, List(Reports.REPORT_JSON)).toIO
```
ncharles, line 77: if a json report arrive between here and l73, it will be fetched here and at next run.

VinceMacBuche: nope, if fetched here it will be treated below and put in the property as max id and we will start from its id

fanf (commented):
LGTM to me, can be self merged once the tests pass

VinceMacBuche:
Commit modified

PR #4602, opened 2022-12-07
Fixes #22178: Cannot deploy changes on a group where properties are defined with a validated user

VinceMacBuche:
https://issues.rudder.io/issues/22178


fanf (approved):
It needs formating but else LGTM

VinceMacBuche:
Commit modified

PR #4603, opened 2022-12-07
Migration to Scala 3

WojciechMazur:
This PR contains a mostly complete migration from Scala 2.13 to Scala 3 (3.2.1 current latest)
General status: 
* All projects main/test sources compile with Scala 3
* There are 3 failing test suites - I'm not sure if those are semantic changes in the produced code or problems with the local environment: 
  - `com.normation.rudder.rest.TestRestFromFileDef`
  - `com.normation.rudder.services.policies.write.WriteSystemTechniquesTest` 
  - `com.normation.rudder.services.policies.write.WriteSystemTechniques500Test` 
   
During the migration we've spotted some bugs in compiler. Most of them were mitigated to allow for compilation. Bug fixes for them would probably be shipped in either 3.2.2 or 3.3: 
* https://github.com/lampepfl/dotty/issues/16467
* https://github.com/lampepfl/dotty/issues/16437
* https://github.com/lampepfl/dotty/issues/16458
* https://github.com/lampepfl/dotty/issues/16443

The most notable changes: 
* `fastparse` is not yet published for Scala 3. It is using macros to implement parsers to all definitions of parsers were move to temporal `utils-parser` and `rudder-parsers` submodules. AFAIK most on the work in terms of fastparse for Scala3 is mostly done, so probably we should await a new release of it soon. At that point temporal modules can be removed
* `sealerate` library is Scala2 only - it's usages were replaced with redefinition of `sealed traits` with `enums`. This change can be potentially replaced with manual enumeration of possible cases or using a  macro/typeclass derivation. 
* There seemed to be problems with `specs2` - AFAIK Specs 4.x are having some problems with Scala 3 due to the changes in implicits resolution and syntax changes. I would recommend migration to Specs2 5.x whenever possible - it is only published for Scala 3 
* `Chimney` is not yet published for Scala 3. As a replacement `Ducktape` was used. 
* `silencer` plugin is not published for Scala 3. Since Scala 2.13 it can be safely removed with `scala.annotation.nowarn` All usages of nowarn were replaced with this annotation, however probably most of its usages can probably be removed
* Due to the Dotty issue #16443 usages of Serialization for NoTypeHints were replaced with `DefaultFormater` it should have the same semantics.
* Over 500 usages of `final` modifier used on objects were removed. All objects are by definition final. Each `final object` was generating a warning and though made compilation logs unreadable. This change can be safely cherry picked to Scala 2.13 build
* Most of the safe warnings were resolves: these were typically missing parentheses for the typed parameters of lambdas. 
* There is still quite large number of high priority warnings - especially in area of NonLocalReturns and usages of scala.reflect.Manifest. These issues should be resolved in the future as they might impact performance. 

Becouse of the large amount of changes I recommend reviewing changes based on the commits and not total amount of changes. It would be easier that way. I've tried to group most of the closely related changes together.


fanf:
Backported scala 2 source compatible changes in https://github.com/Normation/rudder/pull/4789

etorreborre:
> I would recommend migration to Specs2 5.x whenever possible - it is only published for Scala 3

@WojciechMazur @fanf is this something I can help with? I tried to checkout this branch and make it compile, I had several compilation errors like
```
No given instance of type zio.Trace was found for parameter trace
```
(while running `webapp/sources/ldap-inventory> mvn test` for example). Is that something that is supposed to work on that branch and I have something wrong in my environment or commands?

fanf:
oh wow @etorreborre - sorry, I totally missed your almost-one-year old comment. I think it's OK, but @mbaechler might ask you questions at some points if he discovers problems.

etorreborre:
@fanf no worries. @mbaechler don't hesitate to come back to me.

PR #4614, opened 2022-12-23
Fixes #22210: Do not trigger a group reload if last check is younged than 100 ms

fanf:
https://issues.rudder.io/issues/22210


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/DynGroupService.scala
```diff
@@ -153,7 +153,7 @@ class DynGroupServiceImpl(
   override def changesSince(lastTime: DateTime):   Box[Boolean]                              = {
     val n0 = System.currentTimeMillis
     if (n0 - lastTime.getMillis < 100) {
-      Full(true)
+      Full(false)
```
fanf, line 156: here, if I understand correctly the code and how it's used in `processUpdate` (`val need = dynGroupService.changesSince(lastUpdateTime).getOrElse(true)`), `true` means that we DO need to compute groups. But if last change is not far away, we likely want to avoid that.

PR #4616, opened 2022-12-23
Fixes #7526: Housekeeping of /var/rudder/inventories/received|failed|etc

fanf:
https://issues.rudder.io/issues/7526

This is my Christmas present for author of that very old ticket :)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
@@ -1,0 +1,121 @@
+/*
+ *************************************************************************************
+ * Copyright 2022 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.batch
+
+import better.files.File
+import com.normation.errors.IOResult
+import com.normation.inventory.domain.InventoryProcessingLogger
+import com.normation.utils.CronParser._
+import com.normation.zio._
+import cron4s.CronExpr
+import java.time.Instant
+import zio._
+import zio.clock.Clock
+import zio.duration._
+
+/**
+ * A scheduler which delete old inventory files under /var/rudder/inventories/{received, failed}
+ * It use a cron-like config parsed by cron4s.
+ */
+class PurgeOldInventoryFiles(
+    optCron:          Option[CronExpr],
+    maxAge:           Duration,
+    cleanDirectories: List[File]
+) {
+
+  val logger = InventoryProcessingLogger
+
+  // check is a file is older than max age (true) or not. Only true for regular file.
+  // maxAge in seconds.
+  def isOlderThanMaxAge(f: File, maxAge: Long, now: Instant): IOResult[Boolean] = {
+    IOResult.effect(f.isRegularFile && f.attributes().creationTime().toInstant.isBefore(now.minusSeconds(maxAge)))
+  }
+
+  // must not fail, will be in a cron
+  val cleanOldFiles: UIO[Unit] = for {
+    t0 <- currentTimeMillis
+    now = Instant.ofEpochMilli(t0)
+    _  <- ZIO
+            .foreach_(cleanDirectories) { dir =>
+              ZIO.foreach_(dir.list.to(Iterable)) { f =>
+                isOlderThanMaxAge(f, maxAge.toSeconds, now).flatMap { older =>
+                  if (older) {
+                    InventoryProcessingLogger.info(
+                      s"Deleting inventory file '${f.pathAsString}' (older than ${maxAge.toString})"
+                    ) *>
+                    IOResult
+                      .effect(f.delete())
+                      .catchAll(err =>
+                        InventoryProcessingLogger.error(s"Error when trying to clean old inventory file '${f.pathAsString}'")
+                      )
+                  } else {
+                    InventoryProcessingLogger.trace(
+                      s"Keeping inventory file '${f.pathAsString}' (younger than ${maxAge.toString})"
+                    )
+                  }
+                }
+              }
+            }
+            .catchAll(err => logger.error(s"Error when cleaning-up old inventory files: ${err.fullMsg}"))
+
+    t1 <- currentTimeMillis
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -452,6 +452,19 @@ history.inventories.rootdir=/var/rudder/inventories/historical
...
+# File in these two directories are purged independently from files under {incoming, accepted-nodes-updates}
+# which are purged by the new inventory processor.  There is cron to configure when to delete them, and
+# a max age in duration format.
+#
+# You can disable it totally with:
+#   @never
+# in place of the cron usual format.
+#
+rudder.inventories.cleanup.old.files.cron=0 32 3 * * ?
+rudder.inventories.cleanup.old.files.age=30 days
```
fanf, line 466: Check if we want to keep received & failed inventories more than 30 days by default. It sound like it's enought.

VinceMacBuche (approved):

VinceMacBuche (approved):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
```diff
@@ -0,0 +1,121 @@
...
+              ZIO.foreach_(dir.list.to(Iterable)) { f =>
+                isOlderThanMaxAge(f, maxAge.toSeconds, now).flatMap { older =>
+                  if (older) {
+                    InventoryProcessingLogger.info(
+                      s"Deleting inventory file '${f.pathAsString}' (older than ${maxAge.toString})"
+                    ) *>
+                    IOResult
+                      .effect(f.delete())
+                      .catchAll(err =>
+                        InventoryProcessingLogger.error(s"Error when trying to clean old inventory file '${f.pathAsString}'")
```
mbaechler, line 84: don't you want to log the stacktrace, at least in debug level?

mbaechler, line 84:
```suggestion
                        InventoryProcessingLogger.error(s"Error while trying to clean old inventory file '${f.pathAsString}'")
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
```diff
@@ -0,0 +1,121 @@
...
+                      )
+                  } else {
+                    InventoryProcessingLogger.trace(
+                      s"Keeping inventory file '${f.pathAsString}' (younger than ${maxAge.toString})"
+                    )
+                  }
+                }
+              }
+            }
+            .catchAll(err => logger.error(s"Error when cleaning-up old inventory files: ${err.fullMsg}"))
```
mbaechler, line 94: catching here means: if a directory is not readable (for example), others will never been tried, right?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
```diff
@@ -0,0 +1,121 @@
...
+
+    t1 <- currentTimeMillis
+    _  <- logger.debug(s"Cleaned-up old inventory files in ${Duration.fromMillis(t1 - t0).toString}")
+  } yield ()
+
+  // create the schedule cron or nothing if disabled.
+  // Must not fail.
+  val prog: URIO[Any with Clock, Unit] = optCron match {
```
mbaechler, line 102: Why do you need any env? Clock is now provided by the runtime, isn't it?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
```diff
@@ -0,0 +1,121 @@
...
+      logger.info(
+        s"Automatic cleaning of file older than ${maxAge.toString} starts for directories: '${cleanDirectories
+            .map(_.pathAsString)
+            .mkString("', '")}' (schedule 'sec min h dayMonth month DayWeek': '${cron.toString}')"
+      ) *>
+      cleanOldFiles.schedule(schedule).unit
+  }
+
+  // start cron
+  def start() = {
```
mbaechler, line 118: what about returning the Fiber (or something closable) to let the caller handle the lifecycle (for example in case of tests)?

mbaechler (approved):
I was curious to read one of your PRs, so I wrote a few comments along the way.

They are most likely useless as the PR is now merged but as I took time to wrote them, maybe someone wants to read them anyway?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
```diff
@@ -0,0 +1,121 @@
...
+
+  // check is a file is older than max age (true) or not. Only true for regular file.
+  // maxAge in seconds.
+  def isOlderThanMaxAge(f: File, maxAge: Long, now: Instant): IOResult[Boolean] = {
+    IOResult.effect(f.isRegularFile && f.attributes().creationTime().toInstant.isBefore(now.minusSeconds(maxAge)))
+  }
+
+  // must not fail, will be in a cron
+  val cleanOldFiles: UIO[Unit] = for {
+    t0 <- currentTimeMillis
```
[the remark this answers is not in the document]
fanf: for compat with previous version of ZIO in previous branches of scala. Perhaps we will be now at the point were we can migrate that (that API changed a lot in ZIO along the version, we needed a consistant way to access it)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
```diff
@@ -0,0 +1,121 @@
...
+ */
+class PurgeOldInventoryFiles(
+    optCron:          Option[CronExpr],
+    maxAge:           Duration,
+    cleanDirectories: List[File]
+) {
+
+  val logger = InventoryProcessingLogger
+
+  // check is a file is older than max age (true) or not. Only true for regular file.
```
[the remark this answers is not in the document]
fanf: right, the name is not the best, but in the context of purging inventory files, it's not so bad - but I agree, it needs conventional knowledge, so rise the threshold for contrib etc. 
Plus, we need to filter out directories in any cases, because Java nio API does not provide a consistent creation time for directories and leads to surprising results

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
fanf: Our users are not developpers. As a rule of thumb, that does not make sens for them, a stacktrace is a bug in the soft. 
So the general approach is to try to make the logs human(ops)-readable.

In that case, I forgot the the `err.fullMsg` though :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
fanf: yes, the the catch-all of last ressort. Given the soft convention, it should not be reached appart when things are very broken. I'm not sure that a per-directory handling in addition would be better, perhaps (ie: even after the fact and trying to answer your question, I'm not sure it matters since we will need the outer one in all case).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
fanf: Same as before: we use ZIO 1.0 on previous branch and try to minimize merge difficulties. It is not always the best choice (we tend to keep things a bit too long), but as a general rule of thumbs, it minimize our time pass on these things.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
fanf: It could be interesting. In general, I prefer to have the caller do the least possible things in the first implementations and open up API when needs arise. 
For that case, I think I just didn't think to the that possibility / way of doing it.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeOldInventoryFiles.scala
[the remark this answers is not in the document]
fanf: (the 7.2 branch on which that PR was done is on ZIO 1.0)

fanf:
Some feedback addressed in https://github.com/Normation/rudder/pull/4621

PR #4617, opened 2022-12-24
Fixes #22224: Replace usage of Int + comments as seconds to an actual Duration time

mbaechler:
That's a small contrib as a kind of a christmas present, nothing fancy but as I was browsing the codebase, I tried to actually do a change in the boyscout rule style.


fanf:
❤️
I'm on Holliday bared from my laptop but I will come back to you asap

fanf (commented):
Thank you! LGTM for, I updated the tittle so that our bots are able to close https://issues.rudder.io/issues/22224 (needed for our automatic changelog generation & internal review process). 

Can you just run `mvn spotless:apply`, and rebase into one commit before I validate? (we use the convention of "one commit for a fix/feature if possible")

VinceMacBuche (approved):
Thanks for your pull request @mbaechler I'm merging it!

PR #4618, opened 2022-12-24
As ZIO ZManaged as been replaced by Scope, let's change the type alias

mbaechler:
By the way, there's no longer value in IOManaged methods, it's
  as simple with ZIO methods now

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
@@ -247,58 +247,15 @@ final class AsyncDeploymentActor(
         case v @ (Right(PolicyGenerationTrigger.AllGeneration) | Left(_)) =>
           logTriggerError(v)
           currentDeployerState match {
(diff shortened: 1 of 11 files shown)


fanf:
Thanks for the PR!

When I did the port to ZIO, I kept `IOManaged` to ease merge from earlier branches (we need to support previous version - so with ZIO 1.x- for an extended period of time, so that alias allows to break less things if we need to touch to code related to  `IOManaged`). 
On the other hand, we will have to transition at one point (or keep `IOManaged` until the end of times), so it can be now and perhaps it's even the best time to do so, and there is little nice shortening in the PR too. 

@VinceMacBuche what do you think?

@mbaechler for the test error, you need to format code with `mvn spotless:apply` or use  `rudder/webapp/sources/.scalafmt.conf` file in intellij/whatever code editor you use if it supports it

fanf:
(this is not dead nor forgotten, we just got 2 very full months)

fanf:
I forgot to come back to that: 
- we will change the alias, 
- but only when we will have all the maintained code in branches in ZIO 2 (ie 7.3 up)
- and we will do the change in 7.3 and up to avoid merge breakage. 
I'm closing that PR, we will need it only in a couple more months (7.2 EOL is in Q3)

PR #4627, opened 2023-01-10
Fixes #22139: Running deleted campaign don't get their directive cleaned-up 

VinceMacBuche:
https://issues.rudder.io/issues/22139

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
@@ -54,6 +54,7 @@ import zio.syntax._
 
 trait CampaignHandler {
   def handle(mainCampaignService: MainCampaignService, event: CampaignEvent): PartialFunction[Campaign, IOResult[CampaignEvent]]
+  def delete(mainCampaignService: MainCampaignService, event: CampaignEvent): PartialFunction[Campaign, IOResult[CampaignEvent]]
 }
 
 object MainCampaignService {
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -75,6 +76,17 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
...
   private[this] var inner: Option[CampaignScheduler] = None
 
+  def deleteCampaign(c: CampaignId) = {
+
+    inner match {
+      case Some(s) => for {
+        _ <- s.deleteCampaign(c)
+        _ <- campaignRepo.delete(c)
+      } yield {  () }
+      case None    => Inconsistency("not initialized yet").fail
```
fanf, line 86: I think in that case we want to be idempotent and just say that everything is OK (perhaps in log debug say that the event was not found)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -92,6 +104,16 @@ class MainCampaignService(repo: CampaignEventRepository, campaignRepo: CampaignR
 
   case class CampaignScheduler(main: MainCampaignService, queue: Queue[CampaignEventId], zclock: Clock) {
 
+    def deleteCampaign(c : CampaignId) = {
+      for {
+        campaign <- campaignRepo.get(c)
+        events <- repo.getWithCriteria(campaignId = Some(c))
+        _ <- ZIO.foreach(events) { event => ZIO.foreach(services)(s => s.delete(main, event)(campaign)).catchAll(_ => CampaignLogger.warn(s"An error occured while cleaning campaign event ${event.id.value} during deletion of campaign ${c.value}") ) }
```
fanf, line 111: I'm not sure what are the consequences here. So we are in the case where we try to delete campaign event, the campaign is already deleted. Is there a way the user can correct the situation?

VinceMacBuche: The campaign is not deleted is there is an error here, you can still delete it again

the workflow is 

User decides to delete a campaign
main campaign service delegates to it inner services handler (l81-83)
inner services handler,
=> get the campaign (L 109
=> Get events linked to the campaign (L 110
=> Clean events if needed (line 111) (like  if they were running state )
=> delete events (L 112)
going back to parents, which deletes the campaign (L 84)


So if anything goes wrong you can run it again, until all events are cleaned, then deleted, then campaign is deleted

VinceMacBuche:
Commit modified

VinceMacBuche:
I do think we should merge this pr

VinceMacBuche:
Commit modified

VinceMacBuche:
OK, merging this PR

PR #4634, opened 2023-01-18
Fixes #22274: Add includeSubdomains to HSTS header with correct property name

amousset:
https://issues.rudder.io/issues/22274

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
@@ -73,7 +73,7 @@ rudder.server.hsts=false
 # You MUST NOT enable it if you rely on plain HTTP for other sites or applications
 # on any (present or future) subdomain of you Rudder server's domain.
 #
-rudder.server.hsts.includeSubDomains=false
+rudder.server.hstsIncludeSubDomains=false
 
 ##################
 # LDAP properties ###################################################################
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -73,7 +73,7 @@ rudder.server.hsts=false
 # You MUST NOT enable it if you rely on plain HTTP for other sites or applications
 # on any (present or future) subdomain of you Rudder server's domain.
 #
-rudder.server.hsts.includeSubDomains=false
+rudder.server.hstsIncludeSubDomains=false
```
amousset, line 76: I would have preferred 

```
rudder.server.hsts.enable = ...
rudder.server.hsts.includeSubDomains = ...
```

but I don't think it is worth the renaming/migration effort.

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
fanf: Ah yes, it breaks because we have `rudder.server.hsts` too, I missed that.

amousset:
OK, squash merging this PR

amousset:
OK, merging this PR

amousset:
OK, merging this PR

PR #4647, opened 2023-01-27
Fixes #22318: Refactor case insensitivity property for users

fanf:
https://issues.rudder.io/issues/22318

This PR refactor code around case insensitivity property to make is simpler to maintain in the future. 
The main goals are:
- remove code duplication and xml parsing duplication
- remove dependcies on global parameters
- make responsibility of function more self-contained (avoid scattering around the need to know how the code works to use it and make testing easier)

Also rename the currend test file to describe that it is about testing crypto algo, and add a test file for testing case sensitivity.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
@@ -1032,19 +1032,19 @@ object RudderConfig extends Loggable {
   val ldapInventoryMapper = inventoryMapper
 
   val pluginSettingsService = new FilePluginSettingsService(root / "opt" / "rudder" / "etc" / "rudder-pkg" / "rudder-pkg.conf")
-  ////////////////////////////////////////////////
-  ////////// plugable service providers //////////
-  ////////////////////////////////////////////////
+  /////////////////////////////////////////////////
+  ////////// pluggable service providers //////////
+  /////////////////////////////////////////////////
 
   /*
-   * Plugable service:
+   * Pluggable service:
    * - Rudder Agent (agent type, agent os)
    * - API ACL
    * - Change Validation workflow
    * - User authentication backends
    * - User authorization capabilities
    */
-  // Plugable agent register
+  // Pluggable agent register
   lazy val agentRegister = new AgentRegister()
 
   // Plugin input interface to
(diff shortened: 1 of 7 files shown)


fanf:
PR updated with a new commit

fanf:
PR rebased

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -1065,24 +1065,10 @@ object RudderConfig extends Loggable {
   // Plugin input interface for Authorization for API
   lazy val authorizationApiMapping = new ExtensibleAuthorizationApiMapping(AuthorizationApiMapping.Core :: Nil)
 
-  ////////// end plugable service providers //////////
+  ////////// end pluggable service providers //////////
 
   lazy val roleApiMapping = new RoleApiMapping(authorizationApiMapping)
 
-  lazy val rudderUsernameCaseSensitive: Boolean = {
-    (for {
```
fanf, line 1073: It is dangerous to define that as a global parameter while the underlying file providing it can be reloaded. In the best case, we end up with a desynchonisation between that property and the one in the file making debugging harder, in the worst case we actually don't do what the user asked in the config file and make a complicated WTF user experience where things only work when restarting rudder.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -146,16 +149,16 @@ object UserDetailList {
    *   - users with same username (according to the case sensitivity)
    *     will be removed from the returned list
    */
-  def sanityLoginFilter(userDetails: List[RudderUserDetail]): Map[String, RudderUserDetail] = {
+  def filterByCaseSensitivity(userDetails: List[RudderUserDetail], isCaseSensitive: Boolean): Map[String, RudderUserDetail] = {
```
fanf, line 152: a function must never ever rely on a global configuration. That's a blinking red flag. We know it will cause stange errors, hard to refactor code, and other unsustainable problem. Dont' do that, ever. 

On some hypothesis, we can use global values at class level (ie directly referenced in class, not in there parameter). 
And there might be exception for default, "the world is set up like that" values, but the general rule of thumb is that we have bad intuition on them, so just add a parameter to the function and define it at class level.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -166,17 +169,19 @@ object UserDetailList {
             Nil
         }
 
-        // Warn that some users  with name different only by their case are defined, and that disabling case sensitivity would break those users
+        // Notice that some users  with name different only by their case are defined, and that disabling case sensitivity would break those users
+        // (not a warning, since we are in the case sensitive case, it's ok)
         ApplicationLogger
-          .warn(s"Users with potential username collision if case sensitivity is disabled: ${res.keys.mkString(", ")}")
+          .info(s"Users with potential username collision if case sensitivity is disabled: ${res.keys.mkString(", ")}")
```
fanf, line 175: we don't warn for things configured by the user to work like that. A warning is something requiring a corrective action by the user (even if not urgent). Warning ends up as ticket created by the log analysis tools of ops team.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -360,50 +376,6 @@ object UserFileProcessing {
     }
   }
 
-  def parseCaseSensitivityOpt(resource: UserFile): Either[UserConfigFileError, Boolean] = {
```
fanf, line 363: This code was an almost direct duplication of the code above it. We don't do that for function that are that long (ie longer than 2 or 3 lines), especially when most of the code is just general purpose xml parsing code. This must be a red flag, the moment we know we are doing something that should not be done that way, and the time to think how to refactor a bit to avoid that consequence.

fanf:
PR updated with a new commit

VinceMacBuche (approved):

amousset:
OK, squash merging this PR

PR #4651, opened 2023-01-31
Fixes #22333: Avoid sync group when importing a rule archive

fanf:
https://issues.rudder.io/issues/22333

### DIFF
--- a/webapp/sources/api-doc/paths/archives/import.yml
+++ b/webapp/sources/api-doc/paths/archives/import.yml
@@ -14,6 +14,12 @@ post:
               type: string
               format: binary
               description: The ZIP archive file containing policies in a conventional layout and serialization format
+            merge:
+              type: string
+              description: Optional merge algo of the import. Default `override-all` means what is in the archive is the new reality. `keep-rule-groups` will keep existing target definition for existing rules (ignore archive value).
+              enum:
+                - override-all
+                - keep-rule-groups
   responses:
     "200":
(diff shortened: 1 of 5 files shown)


webapp/sources/api-doc/paths/archives/import.yml
```diff
@@ -14,6 +14,12 @@ post:
               type: string
               format: binary
               description: The ZIP archive file containing policies in a conventional layout and serialization format
+            merge:
+              type: string
+              description: Optionnal merge option of the import. Default `override-all`. `keep-rule-groups` will keep existing target definition for existing rules.
```
ncharles, line 19: does that means that "keep-rule-group" won't synchronize the group, or that t will merge (resulting in targets being existing group + group from the import )?

VinceMacBuche: If you import a rule that already exists, it will not change the current target of the rule (ignore the target of the imported rules)

fanf:
PR updated with a new commit

webapp/sources/api-doc/paths/archives/import.yml
fanf: updated the doc to specify that archive values are ignored in the seconde case.

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #4655, opened 2023-02-05
Fixes #22206: Allow user to define custom roles in rudder-user.xml

fanf:
https://issues.rudder.io/issues/22206

Introduce the possibility to have user-defined roles in the `rudder-users.xml` file. 

The main changes are:
- the user file parsing steps are not more cleanly separated, with the parsing first step that just check for the XML structure, extract data but doesn't do any consistancy check ; then we have a different step for custom role check and an other for user roles,
- for the parsing of users, both `role` (as before 7.3) and `roles` (7.3 normalization) tags are accepted,
- parsing authorisation (ie `node_read` etc) is moved to the `AuthorizationType` class,
- authorisation type (`read`, `write`, `edit`) are defined in a sealed class so that we can check that a custom role doesn't overlap an authorisation
- `RoleToAuthz` object is renamed `RudderRoles`. It's a *stateful* object and so should be a class+impl, but b/c of the existing object status everywhere in rudder/plugin, I wasn't sure it will be possible to change. With implementation insight, it seems that it can be done (perhaps during a latter phase).
- `RudderRoles` holds all known roles and is able to parse custom roles and register new ones. 
- `custom-roles` are case insensitive (because it was already the case for built-in roles, and we don't want to have two roles providing vastly different authorisations only diffing by case sensisitivity). No other restriction is put on role name legal characters, it may be a point to decide.
- a `custom-role` can't refere a non-existing role (the whole custom-role declaration will be ignored)
- order of custom-role definition does not matter
- the main difficulty is the cycle detection in custom role and so the `resolveCustomRoles` method. That method contains all the logic and relies on several sub-function for each step. The basic idea of the algo is the input/pending/done queues with callback for pending when a new role reaches done. Implemented with ZIO  `Ref`, it's ok-ish in clarity 
- some unit tests are added to check the different assumptions. In particular one try to use all the different edge cases (role with empty list of roles, overriding name, case insensitivity, several cycles detections)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ApplicationLogger.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ApplicationLogger.scala
@@ -60,6 +60,9 @@ object ApplicationLoggerPure extends NamedZioLogger {
     def loggerName = parent.loggerName + ".plugin"
   }
 
+  object Authz extends NamedZioLogger {
+    def loggerName = parent.loggerName + ".authorization"
+  }
 }
 
 object ApiLogger extends Logger {
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
@@ -36,6 +36,15 @@
  */
 package com.normation.rudder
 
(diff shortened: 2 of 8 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4657, opened 2023-02-06
Fixes #22346: New filters on campaign queries to get several campaign types in one request

VinceMacBuche:
https://issues.rudder.io/issues/22346

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
@@ -58,6 +58,17 @@ trait Campaign {
   def version: Int
 }
 
+object Campaign {
+  def filter(campaigns: List[Campaign], typeFilter: List[CampaignType], statusFilter: List[CampaignStatusValue]) = {
+    (typeFilter, statusFilter) match {
+      case (Nil, Nil) => campaigns
+      case (t, Nil)   => campaigns.filter(c => t.contains(c.campaignType))
+      case (Nil, s)   => campaigns.filter(c => s.contains(c.info.status.value))
+      case (t, s)     => campaigns.filter(c => t.contains(c.campaignType) && s.contains(c.info.status.value))
+    }
+  }
+}
+
 case class CampaignInfo(
     id:          CampaignId,
     name:        String,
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
```diff
@@ -84,14 +84,43 @@ object CampaignId                                                            {
...
+  case object Enabled  extends CampaignStatusValue {
+    val value = "enabled"
+  }
+  case object Disabled extends CampaignStatusValue {
+    val value = "disabled"
+  }
+  case object Archived extends CampaignStatusValue {
+    val value = "archived"
+  }
+  val allValues = Enabled :: Disabled :: Archived :: Nil
```
fanf, line 105: it can be replaced by a sealerate call to avoid unsync if value are added later on

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
```diff
@@ -84,14 +84,43 @@ object CampaignId                                                            {
...
+sealed trait CampaignStatusValue {
+  def value: String
+}
 @jsonDiscriminator("value")
-sealed trait CampaignStatus
+sealed trait CampaignStatus      {
+  def value: CampaignStatusValue
+}
+
+object CampaignStatus {
```
fanf, line 95: Can you call the companion object like the trait (ie `CampaignStatusValue`) else it's really confusing and against convention. Plus I'm pretty sure it will cause problems with the actual `CampaignStatus` trait & objects

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/CampaignApi.scala
```diff
@@ -57,8 +58,11 @@ class CampaignApi(
     val schema = API.GetCampaigns
 
     def process0(version: ApiVersion, path: ApiPath, req: Req, params: DefaultParams, authzToken: AuthzToken): LiftResponse = {
-      val res = (for {
-        campaigns  <- campaignRepository.getAll()
+      val campaignType  = req.params.get("campaignType").getOrElse(Nil).map(campaignSerializer.campaignType)
+      val (err, status) = req.params.get("status").getOrElse(Nil).map(v => CampaignStatus.getValue(v)).partitionMap(identity)
```
fanf, line 62: too much go :)
Still, you need to deal with `err` (because we are not really go :) )

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/MockServices.scala
```diff
@@ -3031,10 +3032,23 @@ class MockCampaign() {
...
+    override def getAll(typeFilter: List[CampaignType], statusFilter: List[CampaignStatusValue]): IOResult[List[Campaign]] = {
+      for {
+        campaigns <- items.get.map(_.valuesIterator.toList)
+      } yield {
+        (typeFilter, statusFilter) match {
+          case (Nil, Nil) => campaigns
+          case (t, Nil)   => campaigns.filter(c => t.contains(c.campaignType))
+          case (Nil, s)   => campaigns.filter(c => s.contains(c.info.status.value))
+          case (t, s)     => campaigns.filter(c => t.contains(c.campaignType) && s.contains(c.info.status.value))
+        }
```
fanf, line 3044: you already have that exact code in `CampaignRepository`, it likely means that that bit of logic needs to be refactored out in an utility object (`object CampaignRepository` likely) and reused in both places.

fanf (commented):
Some change to do but after that, I think it can be merged.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):

PR #4661, opened 2023-02-09
Fixes #22357: Reloading user must discared previously registered custom-roles

fanf:
https://issues.rudder.io/issues/22357


fanf:
OK, merging this PR

PR #4663, opened 2023-02-10
Fixes #22363: User should be able to accept change request even with divergences

fanf:
https://issues.rudder.io/issues/22363


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -125,18 +125,17 @@ class CommitAndDeployChangeRequestServiceImpl(
...
-                                    for {
-                                      (modId, triggerDeployment) <- saveConfigurationChangeRequest(config)
-                                      updatedCr                   = ChangeRequest.setModId(config, modId)
-                                    } yield {
-                                      (updatedCr, modId, triggerDeployment)
-                                    }
-                                  } else {
-                                    Failure(
-                                      "The change request can not be merge because current item state diverged since its creation"
+                                  if (!isMergeableConfigurationChangeRequest(config)) {
```
fanf, line 128: we are now always able to merge, and it's a legit user action, but it's not the nominal path so let an info message in that case

VinceMacBuche (approved):

PR #4669, opened 2023-02-16
Fixes #22375: Nodes not answering are seen in "missing" rather than in "no reports"

VinceMacBuche:
https://issues.rudder.io/issues/22375

So, the reason was that in the case were we had a valid old run, with a config id, but the expected report for that config id were missing (because they were old and so they got deleted since then), THEN we were going into a `UnexpectedUnknownVersion` in place of a `NoReportInInterval`. 
So now, we first check the date of the old run, and if the run is expired, we return `NoReturnInInterval`, else we return `UnexpectedUnknownVersion` (because in that case, we should really have the config id in base)


VinceMacBuche:
Commit modified

fanf (commented):
It's good, need to merge https://github.com/VinceMacBuche/rudder/pull/24 to have the test

fanf (approved):

PR #4671, opened 2023-02-17
Fixes #22383: Password tag should be optionnal

fanf:
https://issues.rudder.io/issues/22383

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
@@ -87,6 +87,12 @@ object PasswordEncoder {
 
   val random = new SecureRandom()
 
+  def randomHexa(byteLength: Int) = {
+    val token = new Array[Byte](byteLength)
+    random.nextBytes(token)
+    new java.math.BigInteger(1, token).toString(16)
+  }
+
   class DigestEncoder(digestName: String) extends PasswordEncoder {
     override def encode(rawPassword: CharSequence):                           String  = {
       val digest = MessageDigest.getInstance(digestName)
@@ -477,7 +483,7 @@ object UserFileProcessing {
     if (root.size != 1) {
       Inconsistency("Authentication file is malformed, the root tag '<authentication>' was not found").fail
     } else {
-      val hash = PasswordEncoder.parse((root(0) \ "@hash").text).getOrElse(PasswordEncoder.PlainText)
+      val hash = PasswordEncoder.parse((root(0) \ "@hash").text).getOrElse(PasswordEncoder.BCRYPT)
 
       val isCaseSensitive: IOResult[Boolean] = (root(0) \ "@case-sensitivity").text.toLowerCase match {
         case "true"  => true.succeed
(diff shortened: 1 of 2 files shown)


fanf:
PR updated with a new commit

VinceMacBuche (dismissed):

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -554,8 +554,14 @@ object UserFileProcessing {
...
               userRoles(node.attribute("role")) ++ userRoles(node.attribute("roles"))
             ) match {
-              case (Some(name :: Nil), Some(pwd :: Nil), roles) if (name.size > 0) =>
-                Some(ParsedUser(name, pwd, roles)).succeed
+              case (Some(name :: Nil), pwd, roles) if (name.size > 0) =>
+                // password can be optional when an other authentication backend is used.
+                // If the attribute is defined several time, use the first occurrence.
+                val password = pwd match {
+                  case Some(p :: _) => p
+                  case _            => ""
```
amousset, line 562: in "plain text" case, down't this make the account accessible without password?

VinceMacBuche: I thought that was what you wanted, i was wondering if this would make an account fallback to nopassword if other authentication fails

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
fanf: note that we weren't checking the lengh either before the change. 
I think that we can just set "password" to some random (long enough) string when not defined. In the case of external auth, it will be ignored, and in the case of file based, it will disable login in practice.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
```diff
@@ -554,8 +554,13 @@ object UserFileProcessing {
...
               userRoles(node.attribute("role")) ++ userRoles(node.attribute("roles"))
             ) match {
-              case (Some(name :: Nil), Some(pwd :: Nil), roles) if (name.size > 0) =>
-                Some(ParsedUser(name, pwd, roles)).succeed
+              case (Some(name :: Nil), pwd, roles) if (name.size > 0) =>
+                // password can be optional when an other authentication backend is used.
+                // If the attribute is defined several time, use the first occurrence.
+                pwd match {
+                  case Some(p :: _) => Some(ParsedUser(name, p, roles)).succeed
+                  case _            => Random.nextString(20).map(s => Some(ParsedUser(name, s, roles)))
```
amousset, line 562: Random is too predictable, we should do something like https://github.com/Normation/rudder/blob/2aba894711bda3be34660005363eaf1ff316ea15/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/TokenGenerator.scala#L51

fanf: it's not java random, but let me check what actual implementation it is using

amousset: that's already what we used for API tokens before, it uses Java Random https://github.com/scala/scala/blob/v2.12.4/src/library/scala/util/Random.scala#L29

fanf: but in doubt and for homogenity, let's go for TokenGeneratorImpl

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderUserDetailsFile.scala
fanf: (well, its logic at leat)

amousset: :+1:

VinceMacBuche: Can we log that this user is disabled somehow ?

VinceMacBuche: Maybe we should not log about that

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #4720, opened 2023-03-28
Fixes #22552: General improvements on Directive compliance API

VinceMacBuche:
https://issues.rudder.io/issues/22552

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ExpectedReportsRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ExpectedReportsRepository.scala
@@ -39,6 +39,7 @@ package com.normation.rudder.repository
 import com.normation.box._
 import com.normation.errors._
 import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.policies.DirectiveId
 import com.normation.rudder.domain.policies.RuleId
 import com.normation.rudder.domain.reports._
 import net.liftweb.common.Box
@@ -108,6 +109,13 @@ trait FindExpectedReportRepository {
    * only limited on the nodeIds in parameter (used when cache is incomplete)
    */
   def findCurrentNodeIdsForRule(ruleId: RuleId, nodeIds: Set[NodeId]): IOResult[Set[NodeId]]
+
+  /**
+   * Return node ids associated to the rule (based on expectedreports (the one still pending)) for this Rule,
+   * only limited on the nodeIds in parameter (used when cache is incomplete)
+   */
+  def findCurrentNodeIdsForDirective(ruleId: DirectiveId, nodeIds: Set[NodeId]): IOResult[Set[NodeId]]
+  def findCurrentNodeIdsForDirective(ruleId: DirectiveId): IOResult[Set[NodeId]]
 
   /*
    * Retrieve the expected reports by config version of the nodes.
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
@@ -206,10 +206,30 @@ class FindExpectedReportsJdbcRepository(
     }
   }
 
+  /**
+   * Return node ids associated to the directive (based on expectedreports (the one still pending)) for this Directive,
+   * only limited on the nodeIds in parameter (used when cache is incomplete)
+   */
+  override def findCurrentNodeIdsForDirective(directiveId: DirectiveId, nodeIds: Set[NodeId]): IOResult[Set[NodeId]]                              = {
+    if (nodeIds.isEmpty) Set.empty[NodeId].succeed
+    else {
+      transactIOResult(s"Error when getting nodes for directive '${directiveId.serialize}' from expected reports")(xa => sql"""
+        select distinct nodeid from nodeconfigurations
+        where enddate is null and configuration like ${"%" + directiveId.serialize + "%"}
+        and nodeid in (${nodeIds.map(id => s"'${id}'").mkString(",")})
+      """.query[NodeId].to[Set].transact(xa))
+    }
+  }
+  override def findCurrentNodeIdsForDirective(directiveId: DirectiveId):                       IOResult[Set[NodeId]]                              = {
+    transactIOResult(s"Error when getting nodes for directive '${directiveId.serialize}' from expected reports")(xa => sql"""
+      select distinct nodeid from nodeconfigurations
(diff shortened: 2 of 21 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -206,6 +206,26 @@ class FindExpectedReportsJdbcRepository(
     }
   }
 
+  /**
+   * Return node ids associated to the directive (based on expectedreports (the one still pending)) for this Directive,
+   * only limited on the nodeIds in parameter (used when cache is incomplete)
+   */
+  override def findCurrentNodeIdsForDirective(directiveId: DirectiveId, nodeIds: Set[NodeId]): IOResult[Set[NodeId]] = {
```
ncharles, line 213: why do you query the database when you have a cache of current expected reports?
in the NodeConfigurationService, you could do the same as findNodesApplyingRule but with a directive
it would be much more efficient as data is already there

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -186,6 +187,24 @@ trait RuleOrNodeReportingServiceImpl extends ReportingService {
     }
   }
 
+  override def findStatusReportsForDirective(directiveId: DirectiveId): IOResult[Map[NodeId, NodeStatusReport]] = {
+    // here, the logic is ONLY to get the node for which that rule applies and then step back
+    // on the other method
+    for {
+      time_0  <- currentTimeMillis
+      nodeIds <- nodeConfigService.findNodesApplyingDirective(directiveId)
+      rules <- rulesRepo.getAll()
```
ncharles, line 196: you don't use the rules

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ExpectedReportsRepository.scala
```diff
@@ -109,6 +110,13 @@ trait FindExpectedReportRepository {
    */
   def findCurrentNodeIdsForRule(ruleId: RuleId, nodeIds: Set[NodeId]): IOResult[Set[NodeId]]
 
+  /**
+   * Return node ids associated to the rule (based on expectedreports (the one still pending)) for this Rule,
+   * only limited on the nodeIds in parameter (used when cache is incomplete)
+   */
+  def findCurrentNodeIdsForDirective(ruleId: DirectiveId, nodeIds: Set[NodeId]): IOResult[Set[NodeId]]
+  def findCurrentNodeIdsForDirective(ruleId: DirectiveId): IOResult[Set[NodeId]]
```
fanf, line 118: you should not use the same function name for different signatures, it creates problems latter on (inference, refactoring, search not working as expected, etc). 
At least, each and every time I thought it was a good idea, I got bitten down the path latter on

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -206,10 +206,30 @@ class FindExpectedReportsJdbcRepository(
...
+        and nodeid in (${nodeIds.map(id => s"'${id}'").mkString(",")})
+      """.query[NodeId].to[Set].transact(xa))
+    }
+  }
+  override def findCurrentNodeIdsForDirective(directiveId: DirectiveId):                       IOResult[Set[NodeId]]                              = {
+    transactIOResult(s"Error when getting nodes for directive '${directiveId.serialize}' from expected reports")(xa => sql"""
+      select distinct nodeid from nodeconfigurations
+      where enddate is null and configuration like ${"%" + directiveId.serialize + "%"}
+    """.query[NodeId].to[Set].transact(xa))
+  }
```
fanf, line 228: can you factor out the common bit? I think I saw you use something for and part somewhere

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ExpectedReportsRepository.scala
fanf: Actually, given the code it could be that `nodeIds` is an option (or have the semantic "if set is empty, ignore id") and just add the `and` fragment when it's not empty

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -389,6 +388,134 @@ class ComplianceAPIService(
...
+
+    val (groupsComponents, uniqueComponents) = nodeComponents.partitionMap {
+      case (a, b: BlockStatusReport) => Left((a, b))
+      case (a, b: ValueStatusReport) => Right((a, b))
+    }
+
+    (if (groupsComponents.isEmpty) {
+       Nil
+     } else {
+       val bidule = groupsComponents.flatMap { case (nodeId, c) => c.subComponents.map(sub => (nodeId, sub)) }
```
fanf, line 403: :)

fanf (commented):
I would love some unit tests for `findDirectiveNodeStatusReports` and there is a bit of code to factor out, but it's small compared to the PR. My tests on dev data seems ok. 
So you can merge it like that so that we have a shippable rudder next week, and add unit test along the way.

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ExpectedReportsRepository.scala
VinceMacBuche: What is complex is that to keep the current behaviour:
If some nodes ids => Make the request with a where condition on nodeIds
If none => Make the request without where on node ids
If Some empty set => Don't make request 

I'm ok that None and some(Empty) makes sense to not have the and  So i need to look the semantic of the empty Set here

And yes i would totally love rewriting this query using more doobie fragments and tools 

So i may do another PR to improve this

PR #4727, opened 2023-03-30
Fixes #22534: List Managed Nodes API with any of the fields listed in the "full" never answer

fanf:
https://issues.rudder.io/issues/22534

Add a methode to be able to filter out entries in a subtree research. 
Then use it in the LDAP "get inventories" to limit the subtrees to the one of the looked for nodes. 
We need a second request to get relevant machine (which use the same id of only requesting machines actually linked to nodes by filtering out the tree)

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/services/core/FullInventoryRepository.scala
+++ b/webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/services/core/FullInventoryRepository.scala
@@ -109,6 +109,9 @@ trait ReadOnlyFullInventoryRepository {
 
   def getAllInventories(inventoryStatus: InventoryStatus): IOResult[Map[NodeId, FullInventory]]
 
+  // get inventories for a set of nodes
+  def getInventories(inventoryStatus: InventoryStatus, nodeIds: Set[NodeId]): IOResult[Map[NodeId, FullInventory]]
+
   def getAllNodeInventories(inventoryStatus: InventoryStatus): IOResult[Map[NodeId, NodeInventory]]
 }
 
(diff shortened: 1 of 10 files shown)


webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/services/core/FullInventoryRepository.scala
```diff
@@ -39,6 +39,7 @@ package com.normation.inventory.services.core
 
 import com.normation.errors._
 import com.normation.inventory.domain._
+import zio.Chunk
```
ncharles, line 42: this don't seem to be used

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/FullInventoryRepositoryImpl.scala
```diff
@@ -337,6 +338,79 @@ class FullInventoryRepositoryImpl(
...
+        .sorted
+        .map(a => Filter.create(s"entryDN:dnSubtreeMatch:=${a}"))
+      BuildFilter.OR(dnFilters: _*)
+    }
+
+    for {
+      con      <- ldap
+      dit       = inventoryDitService.getDit(inventoryStatus)
+      // Get base tree, we will go into each subtree after
+      nodeTree <- con
```
ncharles, line 358: there are several nodes, so naming it nodesTree would help maintenance

ncharles (changes requested):
this is a neat change. Two small cosmetic change, you can self merge after

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/FullInventoryRepositoryImpl.scala
fanf: It was to keep the same structure than above method, but the smart move would have been to factor out that actually

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
OK, merging this PR

PR #4731, opened 2023-03-31
Fixes #22579: Rudder can't boot when custom role uses cve_read without cve plugin

fanf:
https://issues.rudder.io/issues/22579

We were a bit lenient on the check for authorization: if the authorization had a valid syntaxt BUT no object of the authorization name existed, we still tried to create corresponding right (with an empty set, which failed). 
The solution is to:
- forbid for role the usage of `_`. I could have restricted it to existing `_all`, `read` etc pattern, but it's much much easier to be broad today, and make the restriction smaller latter, than the contrary. On the plus side, it gives the general rule "only built-in permissions (authorizations or role) can have an `_`
- for role name validity, check the syntax, and not the existence of a corresponding authorization, 
- adapt tests

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
@@ -36,6 +36,7 @@
  */
 package com.normation.rudder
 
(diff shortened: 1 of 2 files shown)


VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #4744, opened 2023-04-06
Fixes #22614: Error log at boot: Can not find active technique with id 'test_import_export_archive}'

fanf:
https://issues.rudder.io/issues/22614

- the change is just that we need to look for the parent breadcrump only in the case where the active technique exists
- need to be view with "?w=1" to see the change
- also remove a duplicated `}` in the log


fanf:
OK, merging this PR

PR #4746, opened 2023-04-06
Fixes #20396: It's hard to understand what authentication backend failed when a fallback happens

fanf:
https://issues.rudder.io/issues/20396

Tested with file, ldap and OIDC backends. That's a lot of different case from springsecurity, whose API is just plan horrible. 
So nothing facny: 
- we need a new class `RudderAuthenticationProvider` that implements `AuthenticationProvider` else we don't have access to our provider identifier at login
- then, the hard part was trying to understand where in Spring we can put the log - it's in a finally, because control flow by exception is a thing in java
- then trying to understand all the things that can get null, and well anything can be null. 

image


fanf:
Arg. There's still case where spring seems to cache a login attempt and does not show the log info.

fanf:
PR updated with a new commit

amousset (approved):

fanf:
OK, squash merging this PR

PR #4751, opened 2023-04-07
Fixes #22617: 

fanf:
https://issues.rudder.io/issues/22617

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
@@ -40,7 +40,6 @@ package com.normation.rudder.rest.data
 import cats.data._
 import cats.implicits._
 import com.normation.NamedZioLogger
-import com.normation.errors._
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
```diff
@@ -101,6 +99,31 @@ object Rest {
...
+
+  object JsonCodecNodeDetails {
+
+    import com.typesafe.config.ConfigRenderOptions
+    import zio.json._
+    import zio.json.ast._
+    import zio.json.internal.Write
+
+    implicit val codecConfigValue: JsonCodec[ConfigValue] = JsonCodec(
+      new JsonEncoder[ConfigValue] {
```
fanf, line 111: strictly speaking, we don't need codec but we are going more and more toward `zio-json`, so it makes sense to have codec for that directly I think

ElaadF:
I have tested, it seems to be ok

fanf:
OK, merging this PR

fanf:
OK, merging this PR

PR #4760, opened 2023-04-16
Fixes #22645: Deadlock in rudder 7.3 (and perhaps 7.2)

fanf:
https://issues.rudder.io/issues/22645

(you should look at that PR without spaces change: https://github.com/Normation/rudder/pull/4760/files?w=1)

I found a way to keep compat and implement something that both make locks on init object a non existing problem and make rudder starts much faster (30~50%, I sometime get to "webapp started" log under 10s and it is now really CPU bounded, ie switching to "low energy" CPU policy increase boot time, max speed decrease it, etc)

So, the strategy is: 
- add a `RudderConfigInit`, with nothing in its constructor so that it returns immediately and a massive `init()` method that does all what RudderConfig used to be
- everything is a `lazy val` in that `init()` method. It allows JVM to sort instanciations and ensure there is no deadlock (scala compiler has some guards if everything is `lazy val`) 
- to have access to parsed properties, create a `RudderParsedProperties` that contains all parsed variable definitions and import it in `RudderConfigInit`. It would have been better to define a set of case classes for the parsed, typed properties and pass it around(also betterfor unit tests etc), but it would break a lot of APIs.
- for the rare properties that are referenced directly in classes and not through constructors, add proxies definition in `RudderConfig`. That's not a deadlock problem, since properties were resolved in the `RudderParserProperties` that is fully instanciated at that point. 
- group "init" code of service toward the end, so that it doesn't go in the way of effecient class instanciation/sorting 
- have a big case class (`RudderServiceApi`) that is returned at end of init method with all the services that were in RudderConfig and used elsewhere. This also ensure that `lazy val` are referenced that their RHS is actually executed. 
- RudderConfig is now just `val rci = RudderConfigInit.init()` + a big list of `val myService: ServiceInterface = rci.myService` 

Important notice: all terminal services, especially batches that aren't refered by anything strict (`val`, `start()` methods, etc), must be at least reference one time, either in `RudderServiceApi` or by calling a `start` method, else they won't be run at all.

It looks like it solves the case I found, and it should solve all cases of early init since `RudderConfigInit` object instanciation can't fail (everything is in init method), and then `RudderConfig` call init (which can fail but don't have lock to other parts). 

I also tried to minimize the change so that:
- futur upmerge are still doable without (too much) pain
- existing plugins and rudder API is not impacted. I hope I didn't miss any reference to `RudderConfig` but if it's the case, we just need to add the corresponding `val missingService = rci.theService` declaration in `RudderConfig` (and add more things in `RudderServiceApi` if needed)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ApplicationLogger.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ApplicationLogger.scala
@@ -39,6 +39,7 @@ package com.normation.rudder.domain.logger
 
 import com.normation.NamedZioLogger
 import net.liftweb.common.Logger
+import org.slf4j
 import org.slf4j.LoggerFactory
 
 /**
@@ -46,6 +47,10 @@ import org.slf4j.LoggerFactory
  */
 object ApplicationLogger extends Logger {
   override protected def _logger = LoggerFactory.getLogger("application")
+
+  object Properties extends Logger {
+    override protected def _logger: slf4j.Logger = LoggerFactory.getLogger("application.properties")
+  }
 }
 
 object ApplicationLoggerPure extends NamedZioLogger {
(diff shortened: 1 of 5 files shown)


fanf:
This is still a draft because I need to:
- minimize API changes. Renaming must be avoided for what is exposed by `RudderConfid`
- check that all batches are correctly referenced in RudderConfig init. They are terminal leafs (not ref anywhere else) and if they are not called, they will never be instantiated nor started

amousset (approved):

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #4766, opened 2023-04-20
Fixes #22659: Remove heartbeat configuration from the settings

RaphaelGauthier:
https://issues.rudder.io/issues/22659


RaphaelGauthier:
There are still references to the Hearbeat configuration in the back-end code. I don't know if we should get rid of them too

amousset:
We should but in 8.0.

VinceMacBuche (approved):

VinceMacBuche:
Did you try to save an heartbeat mode without value defined ?

VinceMacBuche (approved):

RaphaelGauthier:
We can no longer change the heartbeat mode, saving works fine without a heartbeat value

VinceMacBuche:
Yes sorrty i meant "changes-only" but the api still needs a value, but you set it in fact (to 1) so it should be ok

VinceMacBuche:
(and i accepted the pr when i realised it but it does not merge :( )

PR #4771, opened 2023-04-21
Fixes #22672: Remove node information and inventories from LDAP

fanf:
https://issues.rudder.io/issues/22672


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -131,25 +131,28 @@ class NodeFactQueryProcessor(
...
 
-  def processNodeFactMatcher(m: NodeFactMatcher): IOResult[Chunk[NodeFact]] = {
+  def processNodeFactMatcher(m: NodeFactMatcher): IOResult[Chunk[NodeId]] = {
     nodeFactRepo
       .getAllOn(status)
-      .filterZIO(node => FactQueryProcessorPure.debug(m.debugString) *> processOne(m, node))
-      .run(ZSink.collectAll)
+      .mapConcatChunkZIO { case node =>
+        for {
+          b   <- FactQueryProcessorPure.debug(m.debugString) *> processOne(m, node)
```
ncharles, line 153: the logger here adds a massive overhead, i'm unsure as to why, but we should either remove it, or unerstand why it is so costly

ncharles: this syntax allows to get back performance for group computation of Rudder 7.3 on load test platform

`nodeFactRepo
                   .getAllOn(status)
                   .rechunk(100) // one node fact = 700kB, so 70Mo alloc
                   .mapChunksZIO {
                     case nodes =>
                       nodes.filterZIO(node => processOne(m, node)).map(_.map(_.id))
                   }
                   .run(ZSink.collectAll)`

ncharles: and the processOne should not have logger also for perfs

ncharles: Note: with the code i posted, it takes 25s to compute dynamic groups on a 10 000 nodes environment with 464 groups. In Rudder 7.3 it was 22s
In 7.3, computing a group was between 3ms and 3s, based on its complexity. Here, it's between 35ms and 110ms - there is something that looks suboptimal when fetching from cache (dropping totally the filtering part don't change the time, so it's really somewhere within: transforming the map to stream, chunking the stream, collecting all.
Note bis: adding more thread DOES improve perf: 24s with 1 thread, 8,2s with 4 threads,  4,6s with 6 threads
Note third: with rechunk(200), perf is a bit less, but barely noticable

fanf:
This PR is superseeded by #5167

PR #4775, opened 2023-04-24
Fixes #22646: We should initialize campaign service after plugin were added or else current events will be skipped (before being added again)

VinceMacBuche:
https://issues.rudder.io/issues/22646


amousset (approved):

VinceMacBuche:
Commit modified

PR #4783, opened 2023-05-03
Fixes #22723: Backport scala 2.13 change for scala 3.0 migration

fanf:
https://issues.rudder.io/issues/22723

Backport all changes compatible with scala 2.13. Commit [31d64d9](https://github.com/Normation/rudder/pull/4783/commits/31d64d9ea4cbeb1efdb2ec59317342fcc37bc85e) compensate for the drift since the PR was opened + for some changes that would require scala 3 or became unnecessary for now (moving `PropertyParser` in its own module for example).

The changes are mainly:
- a lot of added type ascription, added `()` and `.apply`
- removing all `final` in front of `case object` and `object`
- changing use of `silencer` by now builtin `nowarn` (and so `com.github.ghik.silencer.silent` by `scala.annotation.nowarn`)
- changing the scala 3 deprecated syntax for renaming self into a method call, e.g: 

```
object RULES extends OU("Rules", BASE_DN) {
    rules =>
    private[this] def rules = this
...
```
- some xml magic concatenation syntax now illicit
- changing lift json `implicit val formats  = Serialization.formats(NoTypeHints)` into `implicit val formats  = DefaultFormats`
- avoiding having two different method having the same name (for ex: `get` / `getIO` in `ConfigService`)
- adding `case` in `for` loop when we extract a case class, typically `case JString(ruleId) <- childs`

Nothing seems to change code semantic anywhere (but welcomong a second pair of eyes). 
If it's really the case, we should make the change in the oldest branch we support to avoid merge horror stories after that.


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

PR #4787, opened 2023-05-04
Fixes #19519: 

fanf:
https://issues.rudder.io/issues/19519

Funny, we have an `xml.Utility.escape(...)` and an `xml.Utility.unescape()`, but they don't do the same thing at all. Really funny. 

I checked that restoring a node group archive is identity


VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4791, opened 2023-05-05
Fixes #22733: Build webapp on java 17

VinceMacBuche:
https://issues.rudder.io/issues/22733


webapp/sources/pom.xml
```diff
@@ -163,7 +163,7 @@ limitations under the License.
             </compilerPlugin>
           </compilerPlugins>
           <args>
-            <arg>-target:11</arg>
+            <arg>-release:17</arg>
```
amousset, line 166: https://github.com/scala/scala/pull/9982 for reference

PR #4794, opened 2023-05-11
Fixes #21236: When we arrive on the node search page with a query pre-filed, the query is done twice

fanf:
https://issues.rudder.io/issues/21236


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #4797, opened 2023-05-15
Fixes #22764: Correct warning due to nowarn or release in 7.2 branch

fanf:
https://issues.rudder.io/issues/22764


fanf:
I will need to merge it by hand to avoid removing necessary nowarn in upper branches

PR #4805, opened 2023-05-24
Fixes #22800: Update webapp dependencies 

VinceMacBuche:
https://issues.rudder.io/issues/22800


VinceMacBuche:
Commit modified

fanf (approved):
Lgtm.

I think we are starting to have a maintenance problem with doobie, and we will need to look for alternative in the future

PR #4829, opened 2023-06-14
Fixes #22702: When we clone a group in Rudder 7.2, newly added properties are not cloned

ElaadF:
https://issues.rudder.io/issues/22702

The idea is to get the NodeGroup info refreshed, because when we clone we doesn't get back modifications that can be made without "Save" button up right
Property management is done by an angular app shared between node property and node group


webapp/sources/rudder/rudder-web/src/main/webapp/javascript/rudder/rudder.js
```diff
@@ -418,6 +418,11 @@ function makeDiff(beforeId,afterId,resultId) {
     fragment.appendChild(node);
   }
 
+  result.css({
+    "white-space": "pre-line",
+    "word-break": "break-word",
+    "overflow": "auto"
+  });
```
ElaadF, line 425: Should remove this changes, it's a commit from another branch

ElaadF:
PR updated with a new commit

ElaadF:
PR rebased

fanf (approved):
LGTM

PR #4847, opened 2023-06-22
Fixes #22933: diff on node properties in event log (at least) is broken

fanf:
https://issues.rudder.io/issues/22933


ncharles:
is it related with https://github.com/Normation/rudder/pull/4835/files ?

fanf:
Not directly, but I don't thing the correction in linked pr can work by itself since the js in xhr response won't be executed without the correction in that pr

VinceMacBuche (approved):

PR #4856, opened 2023-06-28
Fixes #22975: Correct scala tags encoder

VinceMacBuche:
https://issues.rudder.io/issues/22975

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/yaml/YamlTechnique.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/yaml/YamlTechnique.scala
@@ -64,7 +64,7 @@ case class Technique(
     items:         List[MethodItem],
     documentation: Option[String],
     category:      Option[String],
-    tags:          Option[List[(String, String)]],
+    tags:          Option[Map[String, String]],
     params:        Option[List[TechniqueParameter]]
 )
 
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/yaml/YamlTechnique.scala
```diff
@@ -108,17 +108,18 @@ class YamlTechniqueSerializer(parameterTypeService: ParameterTypeService, resour
     JsonEncoder[String].contramap(s => parameterTypeService.value(s).getOrElse(s.toString))
   implicit val encoderTechniqueParameter:     JsonEncoder[TechniqueParameter] = DeriveJsonEncoder.gen
   implicit val encoderTechnique:              JsonEncoder[Technique]          = DeriveJsonEncoder.gen
```
fanf, line 110: I think we must sort tags when encoding so that the yaml is always written with tags in the same order (easier for git diff and an user looking at the file). 
I'm not sure how it's done  thought - temporaly map to a `SortedMap` and encode that ? Directly use a `SortedMap` for tags ?

fanf (commented):
I think we need to take care of the order of tags in the serialized file so that they are always sorted

fanf (approved):
We will address the sorting of `tags` in a latter PR

PR #4858, opened 2023-06-28
Fixes #22978: Cache most current Version number to save lots of memory

fanf:
https://issues.rudder.io/issues/22978

We have a parser for `Version` object which create instance of `Numeric`, `Chars` etc part for each segment of a version (each part separated by `.`, `~`, etc). 
So typically for version `1.2.3`, we get 3 new instances of `Numeric(1)`, `Numeric(2)`, `Numeric(3)`. When we have thousands of versions processed, that create a lot of churn. We also know that the low numeric values are over represented in version, so we can easily create a cache for values under 100 and have only one instance of these `Numeric` in RAM at any point. 100 `Numeric` consume almost nothing, it doesn't change memory requirements for rudder. 

The cache is implemented by making the `Numeric` constructor private and always using its factory, which was already the usage case since it's a case class. Then, we just look if the value to construct is lower than the max value in our cache and either use the cached value or create a new instance.

### DIFF
--- a/webapp/sources/utils/src/main/scala/com/normation/utils/Version.scala
+++ b/webapp/sources/utils/src/main/scala/com/normation/utils/Version.scala
@@ -38,6 +38,7 @@
 package com.normation.utils
 
 import java.nio.charset.Charset
+import zio.Chunk
 
 /**
  * When comparing two version numbers, first the epoch of each are compared, then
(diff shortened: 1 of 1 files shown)


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4869, opened 2023-07-02
Fixes #22976: Use new NodeFact datastructure to save inventories in fact-reposiroty

fanf:
https://issues.rudder.io/issues/22976

Backport from https://github.com/Normation/rudder/pull/4869 `NodeFact` data structure and changes that are low risk (minor refactoring needed for the parent PR and which are good by themselves). 
Also, remove the use of directory `/var/rudder/inventories/historical` for inventory data linked to acceptation of pending nodes and replace it with a new `nodefacts` table in postgresql. 

There is a lot of changes, but more are minors (renaming, comments, typo). 
Important changes:

- well, `NodeFact` exists. This is a super structure that represent what a node is TODAY for rudder: not only inventory, but also rudder configuration, properties, etc. That structure should be the only one used everywhere in rudder (one day). For now, it have mapping from and to full or partial inventories, node info, node, etc. with the logic for the missing bit. It is also full serialisable / deserialisable to JSON, which make it amendable to storage, API, etc.
- `NodeFactRepository` is update to use `NodeFact`. Be careful: compared to referenced PR, it's only the `NodeFactStorage` so I change the name accordingly.
- there is a new table `nodefacts` (node id, date of acceptation, fact when accepted in json)
- `InventoryProcessor` is updated to be more independant fomr `LDIF` and be able to use JSON in post-commit action

Now for node acceptation/refusal:

- I moved file related to node into `webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/` (they were in `webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers` which made no sense)
- `NewNodeManager` implementation is splitted so that LDAP things are not mixed with other. In reference PR, that allows to use `factrepos` in place of LDIF. Some of its method gain their own trait like `ListNewNode` for the same reason. 
- `PurgeDeleteNodeImpl` get its own trait/implementation/file, because it's something independent from `RemoveNodeServe` only usefull for batch maintenance
-  `RemoveNodeService` has a lot of changes to make it more similar in structure to `NewNodeManager`, esp regarding post-hooks. That will help with the maintenance of all the things to clean-up when a node is removed. 

Now for inventory history:

- `ReadOnlyHistoryLogRepository` is cleaned-up (remove unused methods)
- `InventoryHistoryJdbcRepository` is created for the same feature, for storing `NodeFact` in place of inventory, and in base in place of files. 
     - in old services, we just use `NodeFact` compatibility methods (`.toFullInventory`)
 - `MigrateNodeAcceptationInventories` is the bootcheck that create the table if missing and migrate historical inventory data


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
```diff
@@ -0,0 +1,1028 @@
...
+      .using(_.sortBy(_.name))
+      .modify(_.videos)
+      .using(_.sortBy(_.name))
+      .modify(_.vms)
+      .using(_.sortBy(_.name))
+  }
+
+  def toMachineId(nodeId: NodeId) = MachineUuid("machine-" + nodeId.value)
+
+  implicit class IterableToChunk[A](it: Iterable[A]) {
```
VinceMacBuche, line 288: Should not this be in a more generic file in utils ? where we have all things related to implicit conversion of basic datastrcutures ?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateNodeAcceptationInventories.scala
```diff
@@ -0,0 +1,166 @@
...
+
+  val msg = "old inventory accept/refuse facts to 'NodeFacts' database table"
+
+  override def description: String =
+    "Check if table 'NodeFacts' exists and if data from /var/rudder/inventories/historical are migrated"
+
+  def createTableStatement: IOResult[Unit] = {
+    val sql = sql"""CREATE TABLE IF NOT EXISTS NodeFacts (
+        nodeId           text PRIMARY KEY
+      , acceptRefuseDate timestamp with time zone
```
VinceMacBuche, line 92: Why not simply date ? and fact ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
fanf: Perhaps, I wasn't sure why it was not in the base encoder (same for either). I wonder if it doesn't create implicit resolution divergence in the general case. I can try to move it there

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateNodeAcceptationInventories.scala
fanf: because I want to be able to have latter on `date` and `fact` for the current state of a node, and so to be able to clearly differentiate them appart (actually, it will likely be `modifyDate` and `currentFact` or something like that)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
fanf: I propose to look at that in a subsequent PR, because perhaps there is other such case already existing, we should merge them all in `utils` (I'm thinking to datetime, etc)

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4886, opened 2023-07-10
Fixes #23050: Switch quicksearch to elm

VinceMacBuche:
https://issues.rudder.io/issues/23050


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf:
Test are ok, but it's hard to get the general arch. Can you describe a bit what is done where, the general behavior? Can be in file comments or description here, something like: 
"The general approch is to build an UI with a text field that is observe for inputs and then make an API request. 
We use bebounce to avoid making a request at each keystroke, but only after some time (using default params).
The API is unchanged. 
The UI don't interpret at all the input. 
Etc"

webapp/sources/rudder/rudder-web/src/main/elm/sources/QuickSearch/View.elm
```diff
@@ -0,0 +1,157 @@
...
+                            , div [ hidden ((String.length model.search ) < 3 )] [ --ng-if="searching || (!searching &amp;&amp; (!results || results.length == 0));" class="ng-scope">
+                                div [ class "angucomplete-searching", hidden (model.state /= Searching) ] [
+                                  text "Searching..."
+                                ]--ng-show="searching &amp;&amp; searchStr.length > 2; " ng-bind="textSearching">Searching...</div>
+                              , div [ class "angucomplete-searching text-danger", hidden ((model.state /= Opened) || (List.length model.results > 0))  ] [-- ng-show="!searching &amp;&amp; (!results || results.length == 0) &amp;&amp; (searchStr.length > 2);">
+                                  span [ class "glyphicon glyphicon-exclamation-sign text-danger" ] []
+                                , text "No results found"
+                                ]
+                              ]
+                              --TODO : look for no result here <!-- ngIf: searching || (!searching && (!results || results.length == 0)); -->
```
fanf, line 140: this one seems to be a real todo

fanf: but the behavior in test is OK, I have the case well handled: 
image

PR #4897, opened 2023-07-17
Fixes #20071: Compliance incorrectly computed when two components have the same name and cfe vars are involved

fanf:
https://issues.rudder.io/issues/20071

Add tests to validate the current assumption on reporting:
- we check that when reports have report ID, everything is simpler: we can correctly split similar reports for different component. Merge works as expected
    - I added a `reportId` in `ComponentValueStatusReport` because it was extremelly hard to follow tests without it, and it was almost free to add, and actually it's quite strange it was not already there. Also added a method to find `ComponentValueStatusReport` in a block based on `reportId`
 - corrected the description of a test that was saying the opposite of what it was testing: we DO check that a component name matches the expected component name pattern when checking for reports.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
@@ -1232,10 +1232,10 @@ class ExecutionBatchTest extends Specification {
       withBad.componentValues.size === 4
     }
     "with bad reports return a component with the key values foo with one unexpected and one repaired " in {
-      withBad.componentValues("foo").flatMap(_.messages.map(_.reportType)) === Unexpected :: EnforceRepaired :: Nil
+      withBad.componentValues("foo").flatMap(_.messages.map(_.reportType)) === EnforceRepaired :: Unexpected :: Nil
     }
     "with bad reports return a component with the key values bar which is a success " in {
-      withBad.componentValues("bar").flatMap(_.messages.map(_.reportType)) === Missing :: Unexpected :: Nil
+      withBad.componentValues("bar").flatMap(_.messages.map(_.reportType)) === Unexpected :: Missing :: Nil
     }
   }
 
@@ -1592,9 +1592,9 @@ class ExecutionBatchTest extends Specification {
         "component",
         "keyvalue",
         executionTimestamp,
-        "message"
+        "message c1"
       ),
-      new ResultRepairedReport(
+      new ResultSuccessReport(
         executionTimestamp,
         "cr",
         "policy",
@@ -1603,8 +1603,8 @@ class ExecutionBatchTest extends Specification {
         "component",
         "keyvalue",
         executionTimestamp,
-        "message"
-      ),
+        "message c2"
+      ), // following are 3 iterations for b2c1
       new ResultSuccessReport(
         executionTimestamp,
         "cr",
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -1232,10 +1232,10 @@ class ExecutionBatchTest extends Specification {
       withBad.componentValues.size === 4
     }
     "with bad reports return a component with the key values foo with one unexpected and one repaired " in {
-      withBad.componentValues("foo").flatMap(_.messages.map(_.reportType)) === Unexpected :: EnforceRepaired :: Nil
+      withBad.componentValues("foo").flatMap(_.messages.map(_.reportType)) === EnforceRepaired :: Unexpected :: Nil
```
fanf, line 1235: not sure why the order changed, but order seems meaningless in that test

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -1232,10 +1232,10 @@ class ExecutionBatchTest extends Specification {
...
     }
     "with bad reports return a component with the key values bar which is a success " in {
-      withBad.componentValues("bar").flatMap(_.messages.map(_.reportType)) === Missing :: Unexpected :: Nil
+      withBad.componentValues("bar").flatMap(_.messages.map(_.reportType)) === Unexpected :: Missing :: Nil
```
fanf, line 1238: same

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -1444,14 +1444,15 @@ class ExecutionBatchTest extends Specification {
...
+  // see correction with report id in next test
   "Sub block with looping component and same component names are not correctly reported, see https://issues.rudder.io/issues/20071" should {
     val reportsWithLoop = Seq[ResultReports](
       new ResultRepairedReport(
         executionTimestamp,
         "cr",
         "policy",
         "nodeId",
-        "report_id12",
+        "report_id_b1c1",
```
fanf, line 1455: use different reportId just to prove that when we have expected reports not supporting them, they are not used. In the real life, it will be "0" everywhere in that case.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -2426,7 +2602,7 @@ class ExecutionBatchTest extends Specification {
     )
 
     /*
-     * we see that for now, we are quite lenient on the check: we don't check for the shape of returned component messages and names
+     * we check the pattern of returned component name
```
fanf, line 2605: this test was lying about what it was testing: it's the opposite

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
```diff
@@ -2491,7 +2667,7 @@ class ExecutionBatchTest extends Specification {
         List("""Success:"alice rights are correct"""", """Repaired:"bob rights are correct"""")
       ))
     }
-    "mallory can get green compliance with other things executed that what we though" in {
+    "we check that component name matches the variable regex from expected" in {
       (withBad.compliance === ComplianceLevel(unexpected = 2, missing = 2)) and
```
fanf, line 2671: that's why we get `unexpected` + `missing` here, not success

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/ExecutionBatchTest.scala
VinceMacBuche: I think it changed because the list of component values is built using two group by instead of one previously and the order might be different (but it's not important)

VinceMacBuche (approved):

fanf:
Finally, we will do the reportId addition change in 8.0 with https://issues.rudder.io/issues/23084, and tests will be updated with https://issues.rudder.io/issues/22388

PR #4899, opened 2023-07-19
Fixes #22388: Bad report maching when reportid are present

fanf:
https://issues.rudder.io/issues/22388

- backport the tests change from https://github.com/Normation/rudder/pull/4897
- make the pattern check for component name and variable with `$` more lenient: now, we only compare that what is before the `$` is identical (ie whole string when no variable, nothing when the variable is at the begining, etc)


fanf:
PR updated with a new commit

VinceMacBuche:
We may remove function replaceCFEngineVars or is it used somewhere else ? probably in another PR

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4901, opened 2023-07-19
Fixes #23084: Remove constraint on component name pattern for matching reports

fanf:
https://issues.rudder.io/issues/23084

Next step of https://github.com/Normation/rudder/pull/4897 and https://github.com/Normation/rudder/pull/4899

- we completely remove the pattern matching between expected component name & value and actual one for reports with a `reportId`
- we add a `reportId` field (default to `0` for case without a valid `reportId`) for `ComponentValueStatusReport` and addapt grouping in consequence, so that now, two report with same value but different report Id are splitted apart. 
    - the grouping is done `componentName` then `reportId` then  `keyValue` to minimize change (reportId first look like we should have something on top of `ComponentStatusReport` with Id, which doesn't match well the current reports structure where each report brings reportid)
- adpat tests to check all that

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -280,6 +280,10 @@ final case class DirectiveStatusReport(
                                |    ${components.sortBy(_.componentName).mkString("\n    ")}
                                |]"""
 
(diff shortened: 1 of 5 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #4908, opened 2023-07-24
Fixes #7008: After deleting a node, the user should be redirected to nodes list page

fanf:
https://issues.rudder.io/issues/7008

The redirection part is trivial: it's just two `RedirectTo` for the `onSuccess/onFailure` case. Can't be simpler. 
Safe, we want a notification that the node was deleted (or that there was an error), and we can't do that in JS since we are doing a redirect. 
So I created a simple snippet with a session var (`UnpublishedToasts`) that display the list of toast when the comon layout is loaded (ie at each actual load of a rudder page). 
You can register toasts with `RegisterToasts .register` anywhere in Rudder where the HTML session is defined (so mostly lift snippet). Toast are cleaned when displayed. 
The are session-bound, so linked to an user, and cleaned-up on loggout in any case.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
@@ -44,6 +44,7 @@ import com.normation.cfclerk.domain.HashAlgoConstraint.SHA1
 import com.normation.eventlog.ModificationId
 import com.normation.inventory.domain._
 import com.normation.rudder.batch.AutomaticStartDeployment
+import com.normation.rudder.domain.logger.NodeLoggerPure
 import com.normation.rudder.domain.nodes.NodeInfo
 import com.normation.rudder.domain.nodes.NodeKind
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -1127,89 +1130,19 @@ object DisplayNode extends Loggable {
...
-                </button>
-            </div>
-        </div>
-    </div>;
-    JsRaw("""$('#errorPopupHtmlId').bsModal('hide');""") &
-    SetHtml(errorPopupHtmlId, popupHtml) &
-    JsRaw(s""" callPopupWithTimeout(200,"${errorPopupHtmlId}")""")
+    RegisterToasts.register(
+      ToastNotification.Error(
+        s"An error happened when trying to delete node '${nodeId.value}'. " +
```
ncharles, line 1135: I would be awesome to have the hostname as well

fanf:
PR updated with a new commit

ncharles (approved):

fanf:
OK, squash merging this PR

PR #4914, opened 2023-07-26
Fixes #23136: Rudderc fails to render block conditions

amousset:
https://issues.rudder.io/issues/23136


amousset:
OK, merging this PR

PR #4918, opened 2023-07-27
Fixes #23148: Temporary workaround for tag creation in rudder 8.0 alpha

fanf:
https://issues.rudder.io/issues/23148

Workaround the tag limitation by providing a hard coded value that is clearly not something that we want to keep. 
Note: it will put that value in directive tags, so it must not outlive alpha cycle.


amousset (approved):

amousset:
OK, merging this PR

PR #4920, opened 2023-07-27
Fixes #23150: Sort bundles in CFEngine output

amousset:
Bundle should be in the order they run, not alphanumeric order.


Fdall (approved):

amousset:
OK, merging this PR

PR #4921, opened 2023-07-27
Fixes #23151: Allow booleans for condition fields

amousset:
https://issues.rudder.io/issues/23151

`true` and `false` are special value, let's accept them as boolean too.

### DIFF
--- a/policies/rudderc/tests/cases/general/ntp/technique.ps1
+++ b/policies/rudderc/tests/cases/general/ntp/technique.ps1
@@ -65,7 +65,7 @@
         TechniqueName = $techniqueName
     }
     
-    $class = ""linux.fedora""
+    $class = ""false""
     if ($localContext.Evaluate($class)) {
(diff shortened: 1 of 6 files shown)


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

policies/rudderc/tests/cases/general/ntp/technique.ps1
```diff
@@ -35,7 +35,7 @@
         TechniqueName = $techniqueName
     }
     
-    $class = ""debian""
+    $class = ""false""
```
amousset, line 38: these double quotes look dubious to me

Fdall: Double double quotes does not exist in powershell

Fdall (approved):

PR #4925, opened 2023-07-28
Fixes #23155: Regenerate yaml technique and filter generated files in rule archives

fanf:
https://issues.rudder.io/issues/23155

Archive adaptation to YAML technique. 

Missing: compilation of YAML techniques. This is harder than it seems (because we have a dozen different "Technique" and we need to rationnalize a bit access to things). 

What is done: 
- we filter out generated files when exporting YAML
- when importing a JSON technique, we migrate it

There is also some changes:

- I tried to remove string for path/file names/etc. For example, I added `TechniqueFiles` that keeps all common technique file names, like `TechniqueFiles.yaml` is `technique.yml`. 
- `MigrateOldTechniquesService` is now in indep object that is used in both bootstrap migration and archive import
- I refactored `YamlTechniqueSerialization` so that:
    - encode/decoder are pures (no I/O access) 
    - they are in an object and can be use anywhere

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
@@ -4,7 +4,6 @@ import better.files._
 import com.normation.errors._
 import com.normation.errors.Inconsistency
 import com.normation.errors.IOResult
-import com.normation.errors.PureChainError
 import com.normation.eventlog.ModificationId
 import com.normation.inventory.domain.AgentType
 import com.normation.rudder.domain.eventlog.RudderEventActor
@@ -18,6 +17,7 @@ import com.normation.rudder.ncf.ParameterType.ParameterTypeService
 import com.normation.rudder.ncf.yaml.YamlTechniqueSerializer
 import com.normation.rudder.repository.GitModificationRepository
 import com.normation.rudder.repository.xml.RudderPrettyPrinter
+import com.normation.rudder.repository.xml.TechniqueFiles
 import com.normation.rudder.repository.xml.XmlArchiverUtils
 import com.normation.rudder.services.user.PersonIdentService
 import com.normation.utils.StringUuidGenerator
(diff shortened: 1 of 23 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
```diff
@@ -67,29 +67,26 @@ class TechniqueReader(
       result
     }
   }
-  def readTechniquesMetadataFile:              IOResult[(List[EditorTechnique], Map[BundleName, GenericMethod], List[RudderError])] = {
-    import yamlTechniqueSerializer._
 
-    import zio.json.yaml._
```
VinceMacBuche, line 73: Here inmport should have been our yamlOps so zio.yaml, this would have allow to keep using fromYaml (but even i'm not sure there is a bug in decoding json, i think it just on encoding side)

VinceMacBuche: ok understood why you not use it anymore

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/yaml/YamlTechnique.scala
```diff
@@ -222,42 +263,23 @@ class YamlTechniqueSerializer(resourceFileService: ResourceFileService) {
...
-                item.reporting == Some("disabled")
-              )
-            )
-          case None         => Left(Consistancy("error"))
+class YamlTechniqueSerializer(resourceFileService: ResourceFileService) {
+  import YamlTechniqueSerializer._
 
-        }
+  // shortcut to avoid having to import yamlOps / toIO, but can be avoided
+  def toYml(technique: Technique): IOResult[String] = technique.toYaml().toIO
```
VinceMacBuche, line 272: Yep, I'd like to remove them as i think they may be useless. maybe they are useful in case of classes and services that mix yaml and json and you need to import nearer to the use case and in this case we could keep a global json import and use the service function

VinceMacBuche (approved):

PR #4927, opened 2023-07-30
Fixes #23159: Use the secrecy crate for wrapping passwords

amousset:
https://issues.rudder.io/issues/23159

Improves secret security:

* Zero-ed memory on Drop
* Immutable secrets (avoid accidental copies)
* No access possible outside of `expose_secret()`, even for only a comparison
* `expose_secret()` is more explicit than `value()`


amousset:
PR updated with a new commit

Fdall (approved):

amousset:
OK, squash merging this PR

PR #4944, opened 2023-08-02
Fixes #23208: Add a warning in the logs when using local auth with non-bcrypt passwords

amousset:
https://issues.rudder.io/issues/23208

* Remove the ability to use the fallback admin account with a clear text password in config
* Add a warn log when local users use an unsafe storage


amousset:
PR updated with a new commit

fanf (commented):
Perhaps we could already have the warning even if the health check is not satisfaying ?

PR #4949, opened 2023-08-03
Fixes #23218: Windows agent cannot get immediatly its policies right after being accepted

ncharles:
https://issues.rudder.io/issues/23218

Windows agent fetchs their policies with apache2, based on their certificate
This changes runs the part that is made to update the access rules, and reload apache2 if necessary, so that we can do a rudder agent update on window node once it's been accepted

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/50-reload-policy-file-server
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/50-reload-policy-file-server
@@ -12,6 +12,14 @@
 # Signal to cf-serverd that it shall look-up the new promises
 
 export PATH="/opt/rudder/bin:$PATH"
+
+# For windows agents
+# we need to reload the configuration on apache2, based on the policies
+# it must be done even if the hash didn't change, as cf-serverd doesn't have the lists of
+# windows agent in it, and it doesn't change on windows node addition/deletion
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/50-reload-policy-file-server
```diff
@@ -12,6 +12,14 @@
 # Signal to cf-serverd that it shall look-up the new promises
 
 export PATH="/opt/rudder/bin:$PATH"
+
+# For windows agents
+# we need to reload the configuration on apache2, based on the policies
+# it must be done even if the hash didn't change, as cf-serverd doesn't have the lists of
+# windows agent in it, and it doesn't change on windows node addition/deletion
+rudder agent run -b system_rudder_apache_configuration,system_reload_rudder_services
```
amousset, line 20: I think I'd prefer to copy the file and reload the service directly.

ncharles: there are 2 parts that need to be done:
* the certificate copy (which is a copy)
* the rudder-networks-24.conf (which needs to be built)
the second point simply cannot be do without duplicating code, the first one has a bit of plumbing and checks embedeed in the agent

i'd rather not duplicate code, and reuse what already exists, to ease maintenance

PR #4957, opened 2023-08-06
Fixes #23227: 

fanf:
https://issues.rudder.io/issues/23227

Correct right checking for old APIs that don't take benefits of the framework with shema definition that manages that automatically. 
The idea is just to encapsulate each API endpoint logic with a `with[ActionType][ActionObject]` check that will do the correct check or else return an error.
For example for quicksearch:

```
....
    case Get("secure" :: "api" :: "quicksearch" :: Nil, req) => {
      OldInternalApiAuthz.withReadUser {
        ... the same code than before ...
     }
...
```

The error is logged two times: by hand at warn level for logger `api`, and automaticaly by the rest logger of the old API because it's an API error and they are always logger. Perhaps it's useless, but it seems strange that no logs are traced for `api`. The warn log can be removed by removing the line `ApiLogger.warn(fail.messageChain)`

Nothing else fancy, just be particulary careful when checking the PR that:
- only `GET` api has a `read` check, other have a `write` one, 
- the role of the check seems OK

Oh, and it shows that we must port all these old API to the new ones, checking by hand rights for each one is far too fragile.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
@@ -40,6 +40,7 @@ import com.normation.box._
 import com.normation.eventlog._
 import com.normation.rudder.repository.EventLogRepository
 import com.normation.rudder.repository.json.DataExtractor.CompleteJson
+import com.normation.rudder.rest.OldInternalApiAuthz
 import com.normation.rudder.rest.RestExtractorService
 import com.normation.rudder.rest.RestUtils._
 import com.normation.rudder.services.user.PersonIdentService
@@ -130,7 +131,14 @@ class EventLogAPI(
         }
       }
 
-      (for {
+      implicit val prettify = restExtractor
+        .extractBoolean("prettify")(req)(identity)
+        .getOrElse(Some(false))
+        .getOrElse(
+          false
+        )
+      implicit val action: String = "eventFilterDetails"
+      OldInternalApiAuthz.withWriteAdmin((for {
         json   <- req.json
         draw   <- CompleteJson.extractJsonInt(json, "draw")
         start  <- CompleteJson.extractJsonInt(json, "start")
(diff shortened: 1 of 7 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -201,12 +209,12 @@ class EventLogAPI(
...
           val fail = eb ?~! "Error when fetching event logs"
           JsonResponse(errorFormatter(fail.messageChain), 500)
-      }
+      })
 
     case Get(id :: "details" :: Nil, req) =>
       implicit val prettify = restExtractor.extractBoolean("prettify")(req)(identity).getOrElse(Some(false)).getOrElse(false)
       implicit val action: String = "eventDetails"
-      (for {
+      OldInternalApiAuthz.withReadAdmin((for {
```
amousset, line 217: This could give access to tokens with full write access. Until we hash them I think it should be write here too.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
```diff
@@ -36,24 +36,40 @@ class RestApiAccounts(
...
-      readApi.getAllStandardAccounts.either.runNow match {
+      implicit val prettify = restExtractor
+        .extractBoolean("prettify")(req)(identity)
+        .getOrElse(Some(false))
+        .getOrElse(
+          false
+        )
+      implicit val action: String = "getAllAccounts"
+
+      OldInternalApiAuthz.withReadAdmin(readApi.getAllStandardAccounts.either.runNow match {
```
amousset, line 47: This could give access to tokens with full write access. Until we hash them I think it should be write here too.

fanf: ok

fanf:
PR updated with a new commit

amousset (approved):

amousset:
OK, squash merging this PR

PR #4960, opened 2023-08-07
Fixes #23228: 

amousset:
https://issues.rudder.io/issues/23228


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -65,6 +65,9 @@ object ApiToken {
     case tokenRegex(v) => Some(ApiToken(v))
     case _             => None
   }
+
+  // Avoid printing the value in logs
+  override def toString: String = s"[REDACTED ApiToken]"
```
fanf, line 70: It should be in the case class I think, line 58

amousset: indeed

amousset:
PR updated with a new commit

PR #4971, opened 2023-08-09
Fixes #23234: Hash API tokens

amousset:
https://issues.rudder.io/issues/23234

See https://www.notion.so/rudderio/Hash-des-tokens-d-API-ddb9ec15820e4a81875b2a235d3f32e5 for details.

Technically, in the webapp:

* Nothing changes in the LDAP schema, we only change the content of the token field
* Centralize token management (hash, generation) in `ApiToken`. This will ease future modifications and ensure consistency.
* Changes in the API:
  * On get request, return an empty token field if it is hashed.
  * On put/post that modify the token, return the clear-text password in the token field to pass it to the user

The frontend changes are straightforward:

* New window based on modals, allowing to copy the clear-text token, opened after account creation or regeneration of a token.
* Changes in the token column:
  * Don't display it if there are not v1 tokens (and use the earned space for displaying creation date)
  * Move the regeneration button to the actions column


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

fanf (approved):
Tested:
- using old token
- creating new token
- migrating token
- user token
- new token with API ACLS

Everything works great, message are :+1: , and all code related topic were previously addressed.

amousset:
OK, squash merging this PR

PR #4972, opened 2023-08-09
Fixes #23252: Improve serde deserialization error messages

amousset:
image

A lot of line numbers are incorrect, but that was already the case, I'll treat it in a separate PR.


amousset:
OK, squash merging this PR

PR #4985, opened 2023-08-17
Fixes #23303: When drag'n drop, the selected method and the targeted drop zone should be more highlighted

RaphaelGauthier:
https://issues.rudder.io/issues/23303

To improve feedback during drag'n drop, I've added an animation of the border colour and box-shadow of the targeted drop zone :
highlighted-drop-zone

However, I couldn't do anything to highlight the selected method - no css can be applied to the moving element during drag'n drop.

I took the opportunity to fix a bug with empty blocks: 2 drop zones would appear when a method was dragged over them :
bug-block-no-method


fanf (approved):
Works great !

PR #4995, opened 2023-08-22
Fixes #23324: Don't accept inventories from agents without a certificate

amousset:
https://issues.rudder.io/issues/23324


fanf (commented):
I believe that you need to also clean-up `com.normation.inventory.domain.SecurityToken` and only keep certificate. It may be a bit involving (IIRC, the chosen encoding was a pain). And it could go in a second PR if prefered.

fanf:
We will do the clean-up in an other PR, at least here we are sure no new inventories without certificate are used.

fanf (approved):

amousset:
OK, squash merging this PR

amousset:
OK, squash merging this PR

PR #4997, opened 2023-08-22
Fixes #23326: Compile yaml technique on archive import

fanf:
https://issues.rudder.io/issues/23326

This PR needs https://github.com/Normation/rudder/pull/4925 to be merged. :heavy_check_mark: 

The main added changes are: 

- `TechniqueCompiler` earn more compilation methods to ease the different cases we know: `compile an EditorTechnique` (always needed for the fallback case) or `we just know the base path of technique` (archive, migration...). 
- `TechniqueCompiler` also get a "migrate json and regenerate yaml if needed", which is basically "do what need to be done to have an up-to-date technique
- On the other hand, I removed methods parameter, since it's only needed for the fallback case, which should not be the  most used (and so it's ok to have on demand query of these info). This change is on a separated commit for now to ease the review.
- the `GitTechniqueArchive#saveTechnique` always check if migration/regen is needed before saving (since it reads `metadata.xml`). It seems to be a good place to have a "catch all" for doing that. 

The import yaml does not have unit test, I'm not sure how to set-up everything for that (note: it will be done in technique migration test and it need quite a lot of work)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
@@ -29,7 +29,15 @@ import zio.ZIO._
 import zio.json._
 import zio.syntax._
 
-class EditorTechniqueReader(
+trait EditorTechniqueReader {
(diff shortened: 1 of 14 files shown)


fanf:
PR rebased

VinceMacBuche (approved):

PR #5004, opened 2023-08-24
Fixes #23098: Plugin cannot add custom roles or it will be overwritten by boot custom roles 

fanf:
https://issues.rudder.io/issues/23098

This PR add the possibility to have new/overridden built-in roles. 
For that, builtin roles are no more object but a case class `Builtin` and are identified by their name (lower case). 
We still keep two special roles, `Administrator` and `NoRights` since they have special semantic. 
Role trait also get a new `addAuthz` method to facilitate the extension of roles - since we don't wan't to remove authz to roles, and that semantic would not be very clear. (note: it could be still possible by making a new role with an existing name and less rights, but it's likely that rudder won't work well then for user with that role).

Then, we add a new `RudderRoles.registerBuiltin` method that will change builtin role set, either by overriding an existing one or adding a new one. Contrary to custom roles, these role can't be unloaded by user reload, and that method should be reserved for plugin use with tested behavior. 

With that, plugin can extends roles and define their own ones.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
@@ -40,6 +40,8 @@ import cats.implicits._
 import com.normation.errors.Inconsistency
 import com.normation.errors.IOResult
 import com.normation.errors.PureResult
+import com.normation.rudder.Role.Builtin
+import com.normation.rudder.Role.BuiltinName
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.zio._
 import scala.collection.immutable.SortedMap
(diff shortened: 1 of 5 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
```diff
@@ -330,62 +330,78 @@ sealed trait Role {
...
+
+    def rights = Rights(Set(A.NoRights))
+
+    // no-op
+    override def addAuthz(authz: Set[A]): Role = this
+  }
+
+  // standard predefined built-in roles
+  def standardBuiltIn: Map[String, Role] = List(
+    Builtin("user", Rights(ua ++ A.nodeKind ++ A.configurationKind)),
```
VinceMacBuche, line 378: I think all built in should have been objects that extends BuiltIn not instances of BuiltIn so that we can find their key more easily in code 

ie:

object ReadOnly extends BuiltIn("read_only", Rights(ua ++ A.allKind.collect { case x: ActionType.Read => x }))


and in plugin:
instead of redifining a builtIn

 Builtin("read_only", Rights(Set(right1, right2)))

you can do

ReadOnly.addRights(Set(right1,right2))

VinceMacBuche: addAuthz)

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/RudderUserDetailsTest.scala
```diff
@@ -145,11 +145,11 @@ class RudderUserDetailsTest extends Specification {
     import AuthorizationType._
     val userDetailList = getUserDetailList(userXML_3, "userXML_3")
 
-    val roleA0 = NamedCustom("role-a0", List(Role.forRight(Node.Read), Role.forRight(Node.Write), Role.Configuration))
+    val roleA0 = NamedCustom("role-a0", List(Role.forRight(Node.Read), Role.forRight(Node.Write), Role.values("configuration")))
     val roleA1 = NamedCustom("role-a1", List(roleA0))
-    val roleB0 = NamedCustom("ROLE-B0", List(Role.Inventory))
-    val roleC0 = NamedCustom("role-c0", List(Role.RuleOnly))
-    val roleC1 = NamedCustom("role-c1", List(Role.RuleOnly))
+    val roleB0 = NamedCustom("ROLE-B0", List(Role.values("inventory")))
```
VinceMacBuche, line 150: and here keep using an object

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
```diff
@@ -433,6 +459,30 @@ object RudderRoles {
...
 
+  /*
+   * Register built-in roles. These roles are typically provided by plugins and can't be unloaded or
+   * changed at run-time, contrary to user-provided custom roles.
+   * If the registered role is a new name, then it is just added to the list of built-in roles.
+   * If the registered role has an existing name, then it *replaces* the existing built-in roles. This is typically
+   * done by a plugin to extends an existing core built-in role with new permissions.
+   * Administrator and NoRights roles are special and can't be overridden.
+   */
+  def registerBuiltin(roles: List[Role]): IOResult[Unit] = {
```
VinceMacBuche, line 470: I think this function should extend the existing instead of replacing. this would allow getting the existing roles in plugins, you just define your plugin roles, with authorization provided by the plugin (and only them, you don't need too get existing ones)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/Authorizations.scala
fanf: I thought to do it, but I have a problem with that: if you add several time rights, with the object case you (at least me) would be ver very tempted to do `ReadOnly.addRights(Set(right1,right2))`. But the `read_only` was already modified, so you must have done a look-up on `read_only` (but since the object is there, it is not what is on least resistance path). 

So the alternative I see are:
- base built-in roles are stateful object, their rights are mutable in a `Ref` (to avoid parrallel pb). I almost did that but then base built in roles are more special than plugins, it breaks simitry, and there was a lot of boilerplate added (`getRights` is `ZIO` etc). But perhaps it just need more polishing
- what is stable is just the name/a typed version of the name (ie `val READ_ONLY = "read_only"` or `object ReadOnly extends BuiltinRoleName("read_only")` for ex) and we keep the remaning parts identical, just lookup identifier are from stable values in place of a string that can have typo.

VinceMacBuche: In fact as stated here: https://github.com/Normation/rudder/pull/5004/files#r1306393613 i think registerBuiltIn should extend existing Role, not replace them. so you don't have to fetch them beforehand.

Maybe it's just a role identifier and list of authz granted (added) to that role to pass as parameter to registerBuiltIn. 

I think i really want at least an object with builtIn identifiers : 

object BuildInRole {

val READ_ONLY = "read_only" 

}

(whatever the case, i would prefere lowercase, but i think you would prefer uppercase

VinceMacBuche: In fact as stated here: https://github.com/Normation/rudder/pull/5004/files#r1306393613 i think registerBuiltIn should extend existing Role, not replace them. so you don't have to fetch them beforehand.

Maybe it's just a role identifier and list of authz granted (added) to that role to pass as parameter to registerBuiltIn. 

I think i really want at least an object with builtIn identifiers : 

object BuildInRole {

val READ_ONLY = "read_only" 

}

(whatever the case, i would prefere lowercase, but i think you would prefer uppercase

fanf: OK, I think I understand what you meant, coding an update.

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

fanf:
OK, merging this PR

PR #5008, opened 2023-08-26
Fixes #23352: Import/Export yaml with technique editor

VinceMacBuche:
https://issues.rudder.io/issues/23352

Adds a new API to check whether a technique is valid or not (accepts a yaml, returns a json) to use to import a yaml technique

Modify getTechnique api to add a format parameter to return a yaml instead of a json to use to export a technique as yaml

modify the technique editor to use both new api/parameters

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
@@ -52,12 +52,14 @@ import com.normation.rudder.domain.logger.ApiLoggerPure
 import com.normation.rudder.domain.policies.Directive
 import com.normation.rudder.ncf._
 import com.normation.rudder.ncf.BundleName
+import com.normation.rudder.ncf.yaml.YamlTechniqueSerializer
 import com.normation.rudder.repository.RoDirectiveRepository
 import com.normation.rudder.repository.xml.TechniqueRevisionRepository
 import com.normation.rudder.rest.{TechniqueApi => API, _}
 import com.normation.rudder.rest.RestUtils.ActionType
 import com.normation.rudder.rest.RestUtils.response
 import com.normation.rudder.rest.implicits._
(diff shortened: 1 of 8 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -780,7 +816,7 @@ class TechniqueAPIService14(
     techniqueRevisions.getTechniqueRevision(name, version).map(_.map(JRRevisionInfo.fromRevisionInfo))
   }
 
-  def getTechniqueWithData(techniqueName: TechniqueName, version: Option[TechniqueVersion]) = {
+  def getTechniqueWithData(techniqueName: TechniqueName, version: Option[TechniqueVersion], format: String) = {
```
fanf, line 819: format should not be a string there

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor.elm
```diff
@@ -448,11 +450,8 @@ update msg model =
       let
         action = case model.mode of
                    TechniqueDetails t _ _ ->
-                     let
-                       data =  encodeExportTechnique t
-                       content = Json.Encode.encode 2 data
-                     in
-                       File.Download.string (t.id.value ++ ".json") "application/json" content
+                       getTechniqueYaml model t
```
fanf, line 453: this does not seems to work: I see the remote call being done, if I do it by hand I get the yaml, but with the button it does not open a download windows in the browser. The remote call result is just ignored.

fanf (commented):
Import seems to work as expected, but export does not. Once corrected and with the format change in a sealed trait/enum, you can self merge.

fanf:
OK, actually export works too, elm just needed to be recompiled/reloaded. So just the format thing, and self merge!

VinceMacBuche:
Commit modified

PR #5012, opened 2023-08-29
Fixes #23357: Display compilation output  in technique details

VinceMacBuche:
https://issues.rudder.io/issues/23357

The PR adds compilation output to technique api, output entry is defined if there is a valid compilation-output.yml file

it also add the display of it in technique editor, tab is displayed only if there is an output  

data are displayed within the tab if this the valus is not empty (no stdout in the screen, but stderr is displayed

Maybe we could add interpretation of Result code in a next PR (with a check if code is equal to 0, a cross for error return codes)

Screenshot from 2023-08-30 01-27-13

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompiler.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompiler.scala
@@ -135,6 +135,10 @@ trait TechniqueCompiler {
            }
     } yield x
   }
+
+  def getCompilationOutputFile(technique: EditorTechnique): File
+
+  def getCompilationConfigFile(technique: EditorTechnique): File
 }
 
 /*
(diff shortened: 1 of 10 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
I forgot to add output when saving so that could remove quite some mislead you got

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf:
I still don't think it's what we want: now, when we save and there is a compilation error, we get the error message, the technique becomes a draft, but the compilation output tab is not displayed so we don't have any way to see the compilation pb after the fact: 

image

In that case, I would like to have something like:
- error message on save "your technique was saved, but something went wront during compilation. Please see compilation output for details" and in tab, instruction about what is actionnable for the user.

fanf:
The case when I the fallback happens is OK. 
So we need to better handle the case where rudderc has returned an error code for the user to handle.

fanf (approved):
LGTM for now, we need to polish the errors cases management

PR #5021, opened 2023-08-31
Fixes #23335: Tables in Rudder UI are sorted alphabetically but should also follow a numerical sort

RaphaelGauthier:
https://issues.rudder.io/issues/23335
Whenever we want to compare the name or value of an object in string form, we want to use a natural sort so that decimals in string form are correctly sorted, like :
`["1.2", "1.1", "1.10"] => ["1.1", "1.2", "1.10"]`

To do that we use the compare function from [mcordova47/elm-natural-ordering package](https://package.elm-lang.org/packages/mcordova47/elm-natural-ordering/latest/).


fanf:
Missing a description with something like: "we want to use natural ordering in place of alpha-num. For that, we add a dependency to the relevant elm module and use it where appropriate.

fanf (approved):

RaphaelGauthier:
OK, merging this PR

PR #5024, opened 2023-09-01
Fixes #23379: DEFAULT constraints are now rendered in metadata.xml

Fdall:
https://issues.rudder.io/issues/23379


Fdall:
Closing as we want to keep the default value of a parameter as a param field in the yaml file and not as a constraint.

PR #5026, opened 2023-09-03
Fixes #23381: Allow to edit yaml in editor

VinceMacBuche:
https://issues.rudder.io/issues/23381

This pr adds a way to display and edit a Technique yaml directly instead of using the editor UI

This feature is based on the check api that was added when adding yaml import, this api allow to check wether a technique is valid or not, it used to only accept yaml as input and responded a json.

I added a parameter to determine input and output types, (yaml or json) that allow to send a yaml and get back a json (will be used when we make some changes in our yaml and modify the technique in UI by getting back a valid json) 
But also to send a json and get a yaml back, so that we can send a json that represents what is currently displayed withtin the inventory and get back our  yaml representing what is displaed in the UI

I added a TechniqueEditInfo, which contains data related to yaml edition (is it displayed, what is the value, is it valid (not displayed yet))

Added a CheckMode type, that is used to handle how we will make a request to check API, how to handle the response and what to with it

There is three cases to handle:
  Importing a technique (make a call yaml -> Json, create a new technique in UI
 Get Json from yaml (Make a call yaml -> Json, update current technique)
 Get Yaml From Json (Make a Json-> Yaml, getting current technique yaml (not what is saved, but what is actually displayed) 

[Screencast from 2023-09-04 17-12-13.webm](https://github.com/Normation/rudder/assets/1238978/90833711-d7e7-4542-980d-4d538e3248cc)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
@@ -59,7 +59,7 @@ import com.normation.rudder.rest.{TechniqueApi => API, _}
 import com.normation.rudder.rest.RestUtils.ActionType
 import com.normation.rudder.rest.RestUtils.response
 import com.normation.rudder.rest.implicits._
-import com.normation.rudder.rest.lift.TechniqueApi.OutputFormat
+import com.normation.rudder.rest.lift.TechniqueApi.QueryFormat
 import com.normation.utils.ParseVersion
 import com.normation.utils.StringUuidGenerator
 import com.normation.utils.Version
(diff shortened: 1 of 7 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -481,25 +481,37 @@ class TechniqueApi(
     val restExtractor = restExtractorService
     implicit val dataName:                                                                                     String       = "techniques"
     def process0(version: ApiVersion, path: ApiPath, req: Req, params: DefaultParams, authzToken: AuthzToken): LiftResponse = {
-      case class Content(content: String)
-      def charset:          String               = RestUtils.getCharset(req)
-      implicit val decoder: JsonDecoder[Content] = DeriveJsonDecoder.gen[Content]
+      def charset: String = RestUtils.getCharset(req)
 
+      val input  = req.params.get("input").flatMap(_.map(OutputFormat.parse).headOption).getOrElse(OutputFormat.Json)
```
fanf, line 486: It's confusing to use `OutputFormat` here, since it is also used as input. It should be `FileFormat` (or `SerializationFormat`)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -481,25 +481,37 @@ class TechniqueApi(
...
+            input match {
+              case OutputFormat.Yaml =>
+                import YamlTechniqueSerializer._
+                content.fromYaml[EditorTechnique].toIO
+              case OutputFormat.Json =>
+                import techniqueSerializer._
+                content.fromJson[EditorTechnique].toIO
+            }
           }
           json      <- {
```
fanf, line 505: it's not `json` anymore, it's `response` (likely)

PR #5027, opened 2023-09-04
Fixes #23372: Add a way to set a message if "change audit logs" setting is enabled

RaphaelGauthier:
https://issues.rudder.io/issues/23372

Putting back the change audit message / change request popup after it was removed when the rules interface was rewritten in version 7.0.

We want to control this popup in Elm, so we're rewriting an Elm module to handle change requests and change audit messages, which we're integrating into the Rules application.

The interface is almost identical to what existed before, the aim here is not to improve the user experience but to restore this functionality which had disappeared.


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche (approved):

PR #5034, opened 2023-09-06
Fixes #23406: Clean-up acceptation inventory

fanf:
https://issues.rudder.io/issues/23406

- create a json codec for `DateTime` in `DateFormaterService` so that it's available everywhere (`zio.json` is available in `utils`)
- we had a problem with log handler for sql statements, it has also a dedicated bug: https://github.com/Normation/rudder/pull/5037
- Until now, we were relying on `eventlog` to know who/when accepted/refused/deleted a node, which can be quite costly for deletion. We want to store these information in the `NodeFact` table so that it is self contained for clean-up. The data will be cleaned-up with the node inventory expiration time. There is duplication, but for immutable log data, and it's rather small (date, actor). 
-  we want to use the same structure for deletion and for accept / refuse => this first commit is just to change format for accept/refuse and handle sql table migration at boot.
- there is two configurable duration for clean-up: 
    - `rudder.inventories.pendingChoiceHistoryCleanupLatency.deletedNode` is the inteval we keep the inventory log for a refused/deleted node (default to 30 days ; 0 means immediately)
    - `rudder.inventories.pendingChoiceHistoryCleanupLatency.acceptedNode` is the interval we keep the inventory log for an accepted (non deleted) node (default 0 mean "as long as node is not deleted")
- to avoid yet another daily cron for cleaning things, I added the clean-up logic to `PurgeOldInventoryFiles` which a new name: `PurgeOldInventoryData`
- I have the SQL for the new table in three places, it really doesn't seems optimal, but I don't know how to reduce that. We will need to carefully check they are the same.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
@@ -423,7 +423,8 @@ CREATE INDEX event_state_index ON CampaignEvents ((state->>'value'));
  */
 
 CREATE TABLE NodeFacts (
-  nodeId           text PRIMARY KEY
-, acceptRefuseDate timestamp with time zone
-, acceptRefuseFact jsonb
+  nodeId            text PRIMARY KEY
+, acceptRefuseEvent jsonb  -- { 'date': 'rfc3339 timestamp', 'actor': 'actor name', 'statue'; 'accepted or refused' }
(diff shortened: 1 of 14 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

VinceMacBuche:
OK, merging this PR

PR #5039, opened 2023-09-13
Fixes #23424: Make writing node state to fact repos optionnal

fanf:
https://issues.rudder.io/issues/23424

Add a new rudder config property, `rudder.facts.repo.writeNodeState`, that when true totally disable node fact-repos (replaced by a no-op implementation). The property is false by default (in new install and migration, too).


fanf:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #5049, opened 2023-09-20
Fixes #23466: Focus reporting is not working 

VinceMacBuche:
https://issues.rudder.io/issues/23466

Focus Reporting logic, which is created to focus a "reportId", matches on component name in later parts of compliance check and not on Id (that we don't have yet, but in 8.0 yes) to fix this in 7.3 i need to translate ids when defining expected reports to a valid component to match.

I also added an id on each block in metadata.xml for techniques in technique editor (which was already defined, but disappeared in metadatra.xl) so that we can match on block children and not just on methods

This adds an id to some expectedBlocks, but since it is optionnal it won't break our serialisation


VinceMacBuche:
Commit modified

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
PR updated with a new commit

fanf (approved):
LGTM !

fanf:
OK, squash merging this PR

PR #5050, opened 2023-09-21
Fixes #23469: Reporting by node is not correct on directives and Rules

VinceMacBuche:
https://issues.rudder.io/issues/23469

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -328,7 +328,8 @@ final case class BlockStatusReport(
     subComponents.find(_.componentName == componentName).toList :::
     subComponents.collect { case g: BlockStatusReport => g }.flatMap(_.findChildren(componentName))
   }
-  def compliance:                          ComplianceLevel             = {
+
+  override lazy val compliance: ComplianceLevel = {
     import ReportingLogic._
     reportingLogic match {
       // simple weighted compliance, as usual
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -343,23 +344,25 @@ final case class BlockStatusReport(
...
+
+  def withFilteredElement(predicate: ComponentValueStatusReport => Boolean): Option[ComponentStatusReport] = {
     subComponents.flatMap(_.withFilteredElement(predicate)) match {
       case Nil => None
       case l   => Some(this.copy(subComponents = l))
     }
   }
 
-  def status: ReportType = {
+  val status: ReportType = {
```
fanf, line 365: I think we are oscillating between def and val because val (or lazy val) makes the number of allocation/memory consumption exploses. Did you look in history to see if we did the opposite change at some point?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -131,27 +137,143 @@ final case class ByRuleDirectiveCompliance(
...
+
+  def getValues(predicate: ByRuleNodeCompliance => Boolean): Seq[ByRuleNodeCompliance]
+
+  def withFilteredElement(predicate: ByRuleNodeCompliance => Boolean): Option[ByRuleComponentCompliance]
+
+  def componentValues: List[ByRuleNodeCompliance]
+
+  def componentValues(v: String): List[ByRuleNodeCompliance] = componentValues.filter(_.values.exists(_.componentValue == v))
+
+  def status: ReportType
```
fanf, line 149: the meaning of status here is "the worst report type in sub component, based on target rules?"

VinceMacBuche: In fact i don't think it's used by component but i may be wrong, but for value it's is the worst in all cases. Probably in a block it should depends on reporting logic, But I'm not sure it is really used for a block
.

fanf (commented):
Can you add a unit test that is failing before the change and green after ?

PR #5051, opened 2023-09-21
Fixes #23474: Ensure yaml technique id and technique directory path agree

fanf:
https://issues.rudder.io/issues/23474

That was a bit more complicated than wished, because given the scenario (rudder loading, ncf techniquer regen, etc) we have several entry point where the code can read/load a technique and notice the user - and then, its on cache / not reparsed by technique lib, so the warning may not happen anymore. 

So the different entry point:
- `GitTechniqueReader`: for the case were user is compiling its yaml technique by hand, and then copying all the file into a bad path. We can only detect when the technique lib is reloaded that something is wrong
- `TechniqueCompiler`: that's the case where we are seing outdated generated files compared to `technique.yml` (at different point, notably on technique write) and call the compiler based on path
- `TechniqueReader`: that's the case where a recompilation of ncf technique is asked at boot for example. 

image

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
@@ -48,11 +48,13 @@ import com.normation.rudder.git.ExactFileTreeFilter
 import com.normation.rudder.git.GitFindUtils
 import com.normation.rudder.git.GitRepositoryProvider
 import com.normation.rudder.git.GitRevisionProvider
+import com.normation.rudder.repository.xml.TechniqueFiles
 import com.normation.zio._
 import java.io.File
 import java.io.FileNotFoundException
 import java.io.InputStream
 import java.io.IOException
+import java.nio.charset.StandardCharsets
 import net.liftweb.common._
 import org.eclipse.jgit.diff.DiffEntry.ChangeType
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/GitTechniqueReader.scala
```diff
@@ -822,46 +824,93 @@ class GitTechniqueReader(
...
+  def checkTechniqueIdMatchesYaml(techniqueRelativePath: File, techniqueId: TechniqueId): Task[Unit] = {
+    import zio._
+    import zio.json.yaml._
+    import com.normation.rudder.ncf.yaml.{Technique => YamlTechnique}
+    import com.normation.rudder.ncf.yaml.YamlTechniqueSerializer._
+
+    // looking at descriptor on file path, because it's what TechniqueCompiler will look at
+    val descriptor = repo.rootDirectory / (relativePathToGitRepos.getOrElse(
+      ""
+    ) + "/" + techniqueRelativePath.getPath + "/" + TechniqueFiles.yaml)
```
VinceMacBuche, line 878: why not using better files "/" here ? 

repo.rootDirectory / (relativePathToGitRepos.getOrElse("" ) / techniqueRelativePath.getPath  / TechniqueFiles.yaml ?

VinceMacBuche:
PR updated with a new commit

amousset:
OK, squash merging this PR

PR #5053, opened 2023-09-22
Fixes #23454: Cannot override global property in Rudder 8.0 beta 3

RaphaelGauthier:
https://issues.rudder.io/issues/23454

We need to check whether the properties are inherited or not. If they are, we can then replace their value.


RaphaelGauthier:
Commit modified

fanf (approved):

PR #5057, opened 2023-09-25
Fixes #23483: import technique doesn't show yaml files

amousset:
https://issues.rudder.io/issues/23483

I had to resort to file extension, there does not seem to be usable media type for yaml files...


amousset:
PR updated with a new commit

PR #5076, opened 2023-10-02
Fixes #23524: rudder package should not totally fail when parsing index if one entry

VinceMacBuche:
https://issues.rudder.io/issues/23524


amousset (approved):

VinceMacBuche:
Commit modified

PR #5081, opened 2023-10-06
Fixes #23541: Add new Windows OS conditions to the technique editor

amousset:
https://issues.rudder.io/issues/23541

Still a draft while we discuss compatiblity https://rudderio.slack.com/archives/C3X0V8ZL6/p1696609924041829

image

I considered doing something "smarter" than an big enum, but it seems nothing makes sense if we want to keep it easy to use. We could put the enum on a second level but it would require more clicks and the amound os supported OS versions should stay faily low. Plus it is after Linux in the list so most users won't need to scroll anyway.


amousset:
replaced by https://github.com/Normation/rudder/pull/5082

PR #5090, opened 2023-10-11
Fixes #23555: When migrating json technique to yaml, canonify parameter name

fanf:
https://issues.rudder.io/issues/23555

Principal changes: 

- add a `variableName` in `EditorTechnique -> parameters`. This is the canonified version of the parameter, the one used in .cf and .ps1 files, 
- remove `canonify` from `NcfId` since we now have access to it in technique param directly
- when we read old technique, do the canonification of `name` for params, and put the result in `variableName`
- map yaml field `name` to editor `variableName`, and `description` to editor `name` 

I didn't touch `JResponseObject`, they will likely need to be updated to add `variableName` field etc


VinceMacBuche:
I don't think that what we want. As discussed in our notion with @amousset We'd like to make "name" to be the canonified name of the parameter and "description" be the current name so that we don't create new fields anymore and the mapping to yaml is direct.

So migration from 7.3 should: take the name, use it as description, and canonify it to use it as name

VinceMacBuche:
i made a PR : https://github.com/fanf/rudder/pull/15/files

PR #5092, opened 2023-10-11
Fixes #23551: Broken handling of non-printable characters in techniques

amousset:
https://issues.rudder.io/issues/23551

This happens only when using double-quoted strings in YAML. In this case some special chars are escaped using a backslash. We end up with parameters containing unwanted chars (like backspace from `\b`). In addition `quick_xml` does not check its input for forbidden chars, and happily generates invalid XML.

This PR makes a first line of check at the end of the pipeline, to prevent writing broken files. We should find a better way to handle this (but it is tricky as using these escaped chars it is often not what users want...), and maybe make quick_xml properly escape these chars.

### DIFF
--- a/policies/rudderc/docs/src/test.md
+++ b/policies/rudderc/docs/src/test.md
@@ -17,6 +17,9 @@ The format of the test case file is:
 params:
   param1: value1
   param2: true
+conditions:
+  - condition1
+  - condition2
 setup:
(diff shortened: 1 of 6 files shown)


Fdall (approved):

amousset:
OK, squash merging this PR

PR #5109, opened 2023-10-18
Fixes #15019: Tag are not displayed with a nice diff in eventLog 

clarktsiory:
https://issues.rudder.io/issues/15019

* Adding a `defaultValue` constant in `PolicyMode` that is used across many files to specify that the policy mode is inferred from the global context

* Add display of diffs for tags using 'simple lines diff' (the same that is used when displaying diffs on properties) :
 Screenshot from 2023-10-18 09-41-18

* Add display of diffs also for the policy mode of a directive :  
from-default-to-enforce

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/DirectiveDiff.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/DirectiveDiff.scala
@@ -75,19 +75,44 @@ final case class ModifyDirectiveDiff(
     id:            DirectiveId,
     name:          String, // keep the name around to be able to display it as it was at that time
 
-    modName:             Option[SimpleDiff[String]] = None,
-    modTechniqueVersion: Option[SimpleDiff[TechniqueVersion]] = None,
-    modParameters:       Option[SimpleDiff[SectionVal]] = None,
-    modShortDescription: Option[SimpleDiff[String]] = None,
-    modLongDescription:  Option[SimpleDiff[String]] = None,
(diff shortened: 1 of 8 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/DirectiveDiff.scala
```diff
@@ -75,22 +75,47 @@ final case class ModifyDirectiveDiff(
...
+    modName:             Option[SimpleDiff[String]],
+    modTechniqueVersion: Option[SimpleDiff[TechniqueVersion]],
+    modParameters:       Option[SimpleDiff[SectionVal]],
+    modShortDescription: Option[SimpleDiff[String]],
+    modLongDescription:  Option[SimpleDiff[String]],
+    modPriority:         Option[SimpleDiff[Int]],
+    modIsActivated:      Option[SimpleDiff[Boolean]],
+    modIsSystem:         Option[SimpleDiff[Boolean]],
+    modPolicyMode:       Option[SimpleDiff[Option[PolicyMode]]],
+    modTags:             Option[SimpleDiff[Tags]]
```
fanf, line 87: For information: when you do that kind of change, you need to ensure that it does not impact a plugin (here, the one that could be impacted is `change-validation`). 
(and it's ok, it is not, it's just so that you won't be surprise the first time a change in rudder breaks a plugin, and so you will understand why our plugin build system is so complex)

fanf:
OK, squash merging this PR

PR #5112, opened 2023-10-18
Fixes #23320: Quotes in group name generate JS errors when using GroupId criteria

VinceMacBuche:
https://issues.rudder.io/issues/23320

form was added using a raw js script, doing a jquery.replace(id, 'content') where content was a the string containing the new form to insert. if this forms include a ' , the string is then broken.

Using List replace jscommand instead will sanitize your content correctly and escape the single quote


VinceMacBuche:
Commit modified

PR #5118, opened 2023-10-19
Fixes #23607: When a mandatory field in a technique is not defined, we can save a technique and we have a nasty error

amousset:
https://issues.rudder.io/issues/23607

In 8.0, rudderc has (inadvertently) changed the output format from a list of plain strings to a list of

```json
{
  "name": "optional",
  "value": "previous value"
}
```

And regex from a string to:

```json
{
  "error_message": "optional",
  "value": "previous value"
}
```

This change was not expected in the webapp, but the parser for this file was rewritten at the same time, with a typo on the constraint key name (`constraint` instead of `constraints`), which lead to totally ignoring constraints (and hide the error).

As the change in rudderc was made to improve consistency with select parameter type in techniques, this PR changes the expected format on the webapp side. We can do this as rudderc and the webapp come in the same package and are always in sync.

This PR fixes the checks on parameters in the editor:
image

Also adding an annotation in rudderc to avoid the null `error_message` in regex:

```json
"regex": {
    "value": "^[A-z0-9._-]+$",
    "error_message": null
}
```

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
@@ -219,12 +219,12 @@ final case class JsonGenericMethodParameterRename(old: String, `new`: String)
 final case class JsonGenericMethodParameter(
     name:        String,
     description: String,
-    constraint:  Option[JsonParameterConstraint],
+    constraints: Option[JsonParameterConstraint],
     `type`:      String
 ) {
   def toMethodParameter(parameterTypeService: ParameterTypeService): PureResult[MethodParameter] = {
     parameterTypeService.create(`type`).map { t =>
-      val c = constraint match {
+      val c = constraints match {
         case None    => Nil
         case Some(x) => x.toConstraint
       }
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechniqueReader.scala
```diff
@@ -251,13 +254,14 @@ final case class JsonParameterConstraint(
...
     not_regex.map(NotMatchRegex.apply).toList :::
-    select.map(FromList.apply).toList
+    select.map(_.map(_.value)).map(FromList.apply).toList
   }
 }
 
 object GenericMethodSerialization {
 
   implicit val decodeJsonParameterConstraint:          JsonDecoder[JsonParameterConstraint]          = DeriveJsonDecoder.gen
+  implicit val decodeJsonParameterConstraintSelect:    JsonDecoder[JsonParameterConstraintSelect]    = DeriveJsonDecoder.gen
```
fanf, line 264: Should be defined before `JsonParameterConstraint` since it's used as a field in it (or put `implicit lazy val` to get at-nead initialization)

amousset: thanks!

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

VinceMacBuche (approved):

PR #5121, opened 2023-10-20
Fixes #23440: Add users table to better track user and sessions

fanf:
https://issues.rudder.io/issues/23440

The change is quiet substantive but it touches peripherical parts of Rudder, so it should be rather low risk. 

# General information about the feature

## User information in PostgreSQL

The main change is that now, user information is managed in `postgresql` in a new table `users` with different provisionning sources, like (default) the `rudder-user.xml` file, or (futur work) `oidc`, `ldap` etc when the central authentication succeeds. 

That table also store an `user status` (active, deleted, etc) ; an history of status changed (when and what provisionning source initiated the change); and general information (name, etc). 
User are managed by one provisionning source once they are created. Only that source or system/direct user managed can disable/delete users. For example, if an user comes from `rudder-users.xml`, then if latter on the user is not in `rudder-users.xml`, it will be considered deleted. 

`deleted` is only a flag, information related to a deleted user are kept in base and only removed when the user is purged. 
That allows to keep user information for compliance (for ex) and post mortem analysis (getting user status history), or restore user that should not have been deleted without loosing status history or user related information. 
A purge delete all information related to an user, again to be able to comply to other compliance framework like RGPD. 

Note that user permissions are not managed in these tables, user roles still come from the provisionning sources. 

### Summary of `users` table:

- `id` : login id, must be non empty and unique (same as current user name in `rudder-users.xml`)
- `creationdate`: date/time of user creation (or migration to DB for migration)
- `status`: status of user (`active`, `disabled`, `deleted`). For now, `disabled` is ignored and behave like `active`.
- `managedby`:  what backend "owns" the user (provenance)
- `name`, `email`, `otherinfo`: user related things, not used for now 
- `lastlogin`: last user login date/time
- `statushistory`: trace of all change of the user status, management backend, etc.

## User session in PostgreSQL

In addition to managing user session are stored in `postgresql` too. 
A session store when the user logged in, from which authentication back end and with what role (maybe dynamically related by the authentication provider), and when the session ended. 

### Summary of `usersessions` table:

- `userid` : user id
- `sessionid` : session id as reported by jetty
- `creationdate`: date of creation of the session (after successful user login)
- `authmethod`: what backend accepted user login (file, oidc, ldap, etc)
- `permissions`: list of permission of the user for that session, as actually computed, including runtime permissions provided by an IdP
- `enddate`: when the session was terminated
- `endcause`: how the session was terminated (mainly user logout or rudder restart)

## Cleaning

Different cleaning batches allow to clean outdated information in these tables. 
There is 3 properties directing user information cleaning: 
- `rudder.users.cleanup.account.disableAfterLastLogin`: auto disable users after a given duration since last login (or creation date), 
- `rudder.users.cleanup.account.deleteAfterLastLogin`: auto delete an user after a given duration since last login (or creation date),
- `rudder.users.cleanup.purgeDeletedAfter`: auto purge user after a duration since they get the `deleted` status. 

There is one property directing session cleaning: 
- `rudder.users.cleanup.sessions.purgeAfter`: clean user session older than the duration. 

# Main architecture changes

## High level 

The main architecture change is that the authentication general scheme evolved : 
- before, an user must have been declared in `rudder-users.xml` (or the special root user) before any login attempt was done
- now, we look at the authentication backend ; and given the sucess of it or not, we check if the user is present in `users` table ; then, if it is not, we check if the backend is allowed to auto-provision users or not and given that accept or not login. 
- `rudder-user.xml` get a special treatment: now, at each reload of that file (either at boot, or when the user management plugin is present), we provision all user from it and change user status accordingly (creating new ones, deleting missing ones). This allow to delete users not present in the file, and so keep the exact same user experience than before the change. 

The other main change is the session management. It is implemented so that authentication backend don't have to manage it when authentication succeed or on logout.

## Lower level

- some user related class were moved from `rudder-rest` to `rudder-core`: `User.scala`, current user, etc. That implies a lot of changes in import in lots of files (import renaming).
- `UserRepository.scala` (in `rudder-core`) is the main class that manage users. It has a default implementation in `postgresql`, and an in memory one, used for test. It manage both user and user sessions, since some time, information are need from both side (ex: when an user is logged, we need to update both session table and user last login date). 
- `UserRepositoryTest.scala` tests user lifecycle logic, which is non trivial. The same tests work with both repository implementation. 
- `CleanupUsers.scala` is the new batch that manage everything related to user/sessions clean-up
- user session can be not correctly closed when rudder is shut downed. At boot, we close all opened session with a special reason in base. 
- `UserDetails` are now always retrieved from base. Thus the `RudderInMemoryUserDetailsService` now depends upon `UserRepository`


fanf:
PR updated with a new commit

fanf:
Given the time frame, we will first merge in 8.1 (master) and once we are OK, we will backport in 7.3.

fanf:
Before rebase to 8.1 branch stored here: https://github.com/fanf/rudder/tree/archive/7_3/enh_23440/add_users_table_to_better_track_user_and_sessions

fanf:
@clarktsiory said "LGTM"

clarktsiory (commented):
Overall LGTM, I think it can be merged

PR #5138, opened 2023-10-26
Fixes #23010: When we are logged out, the logout button doesn't work

clarktsiory:
https://issues.rudder.io/issues/23010

The ajax button no longer works after logout because the session is already terminated and all ajax variables mapping are cleaned in Lift...
We need to know if user is already logged out, in that case we do simply redirect instead of making the ajax call.


clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #5151, opened 2023-11-02
Fixes #23655: Import archive does not support technique upgrade

fanf:
https://issues.rudder.io/issues/23655

So, there were two main problems: 

- we were throwing the `NoSuchFileException` when deleting a file that was not here while it's convergent and we should not, 
- but just correcting would have hidden the real problem: in `zip`, directories are not stored, so we never have them, so in the diff they are always marked as deleted. We need to filter out directories in the `Files.walk`

Clean-up:
- also factored-out the code responsible for making the diff so that I was able to write a unit test just on that part (the new test was failing before the fix). 
- make `trace` log a little better for the deletion (display `none` if nothing is deleted)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
@@ -99,6 +99,7 @@ import java.io.ByteArrayInputStream
 import java.io.InputStream
 import java.io.OutputStream
 import java.nio.charset.StandardCharsets
+import java.nio.file.NoSuchFileException
 import java.text.Normalizer
 import java.util.zip.ZipEntry
 import net.liftweb.http.FileParamHolder
@@ -1076,6 +1077,48 @@ trait SaveArchiveService {
   def save(archive: PolicyArchive, mergePolicy: MergePolicy, actor: EventActor): IOResult[Unit]
 }
 
+object SaveArchiveServicebyRepo {
+
+  /*
+   * Create the resource diff between the archive technique and the corresponding path
+   * (assumed to be the technique base directory, ie ../category/1.0/techniqueId).
+   * If file does not exists, everything is marked as "new", including the base directory.
+   */
+  def buildDiff(t: TechniqueArchive, techniqueBaseDirectory: File): IOResult[Chunk[ResourceFile]] = {
+    for {
+      // path ot 't' are relative to technique. Files we need to add as new in the end
+      addedRef <- Ref.make(Chunk.fromIterable(t.files.map(_._1)))
+      diffRef  <- Ref.make(Chunk[ResourceFile]())
+      // get all file path (relative to technique dir)
+      existing <- IOResult.attempt {
+                    if (techniqueBaseDirectory.exists) {
+                      techniqueBaseDirectory
+                        .collectChildren(!_.isDirectory) // we don't have directories in the archive, only files
+                        .to(Chunk)
+                        .map(_.pathAsString.replaceFirst("\\Q" + techniqueBaseDirectory.pathAsString + "/\\E", ""))
+                    } else { // technique or technique version does not exists
+                      Chunk.empty
+                    }
+                  }
+      _        <- ZIO.foreachDiscard(existing) { e =>
+                    for {
+                      keep <- addedRef.modify(a => {
+                                if (a.contains(e)) {
+                                  (true, a.filterNot(_ == e))
+                                } else {
+                                  (false, a)
+                                }
+                              })
+                      _    <- diffRef.update(_ :+ ResourceFile(e, if (keep) ResourceFileState.Modified else ResourceFileState.Deleted))
+                    } yield ()
+                  }
+      added    <- addedRef.get.map(_.map(ResourceFile(_, ResourceFileState.New)))
+      updated  <- diffRef.get
+    } yield added ++ updated
+  }
+
+}
+
 /*
  * This implementation does not check for possible conflicts and just
  * overwrite existing policies if updates are available in the archive.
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -1054,6 +1055,50 @@ trait SaveArchiveService {
...
+      // path ot 't' are relative to technique. Files we need to add as new in the end
+      addedRef <- Ref.make(Chunk.fromIterable(t.files.map(_._1)))
+      diffRef  <- Ref.make(Chunk[ResourceFile]())
+      // get all file path (relative to technique dir)
+      existing <- IOResult.attempt {
+                    if (techniqueBaseDirectory.exists) {
+                      techniqueBaseDirectory
+                        .collectChildren(!_.isDirectory) // we don't have directories in the archive, only files
+                        .toList
+                        .map(_.pathAsString.replaceFirst("\\Q" + techniqueBaseDirectory.pathAsString + "/\\E", ""))
```
fanf, line 1078: we must quote the `replaceFirst` because path can contains `.` and `+` and other chars that would be interpreted as regex directives while we just want to cut the fisrt few path segments

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -1054,6 +1055,50 @@ trait SaveArchiveService {
   def save(archive: PolicyArchive, mergePolicy: MergePolicy, actor: EventActor): IOResult[Unit]
 }
 
+final case class TechniqueArchiveMod(added: Chunk[ResourceFile], updated: Chunk[ResourceFile])
```
clarktsiory, line 1058: This case class does not seem to be used :thinking:

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -1054,6 +1055,50 @@ trait SaveArchiveService {
...
+      existing <- IOResult.attempt {
+                    if (techniqueBaseDirectory.exists) {
+                      techniqueBaseDirectory
+                        .collectChildren(!_.isDirectory) // we don't have directories in the archive, only files
+                        .toList
+                        .map(_.pathAsString.replaceFirst("\\Q" + techniqueBaseDirectory.pathAsString + "/\\E", ""))
+                    } else { // technique or technique version does not exists
+                      Chunk.empty
+                    }
+                  }
```
clarktsiory, line 1082: The inferred type is a bit weird because in the first conditional branch the type is `List[String]` and in the second one a `zio.Chunk`, so maybe an empty List should be there ? In the end it is used as an iterable so even keeping an iterable would do...

clarktsiory:
```suggestion
      existing <- IOResult.attempt {
                    if (techniqueBaseDirectory.exists) {
                      techniqueBaseDirectory
                        .collectChildren(!_.isDirectory) // we don't have directories in the archive, only files
                        .map(_.pathAsString.replaceFirst("\\Q" + techniqueBaseDirectory.pathAsString + "/\\E", ""))
                        .to(Iterable)
                    } else { // technique or technique version does not exists
                      Iterable.empty[String]
                    }
                  }
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -1086,69 +1131,50 @@ class SaveArchiveServicebyRepo(
...
+                  case f if (f.state == ResourceFileState.Deleted) => f.path
+                }.toList match {
+                  case Nil => "none"
+                  case l   => l.mkString(", ")
+                }
+                s"Deleting technique files for technique '${t.technique.id.serialize}': ${deleted}"
+              }
+      _    <- ZIO.foreachDiscard(diff) { u =>
+                if (u.state == ResourceFileState.Deleted) {
+                  IOResult.attempt(File(techniqueDir.pathAsString).delete()).catchSome {
```
clarktsiory, line 1150: Have I missed something there because the whole technique directory is deleted ? Shouldn't it be the delete file `u` ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
clarktsiory: I realized that it is code that has been moved, but I think it should be changed because it slows the IDE because of the huge type being inferred `Seq[String] with StrictOptimizedSeqOps[String,[_]Seq[_] with immutable.StrictOptimizedSeqOps[_,[_]Seq[_] with immutable.StrictOptimizedSeqOps[_,[_]Seq[_] with Serializable,Seq[_] with Serializable] with Serializable,Seq[_] with immutable.StrictOptimizedSeqOps[_,[_]Seq[_] with Serializable,Seq[_] with Serializable] with Serializable] with Serializable{def iterableFactory: scala.collection.SeqFactory[[A]Seq[A] with Serializable]},Seq[String] with StrictOptimizedSeqOps[String,[_]Seq[_] with immutable.StrictOptimizedSeqOps[_,[_]Seq[_] with Serializable,Seq[_] with Serializable] with Serializable,Seq[String] with StrictOptimizedSeqOps[String,[_]Seq[_] with Serializable,Seq[String] with Serializable] with Serializable] with Serializable{def iterableFactory: scala.collection.SeqFactory[[A]Seq[A] with java.io.Serializable]}] with Serializable{def iterableFactory: scala.collection.SeqFactory[[A]Seq[A] with scala.collection.immutable.StrictOptimizedSeqOps[A,[_]Seq[_] with java.io.Serializable,Seq[A] with java.io.Serializable] with java.io.Serializable]}` :sweat_smile:

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
clarktsiory: This may also be the reason why the `NoSuchFileException` was raised when more that 1 file was deleted from the archive

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
fanf: you're right, forgot to delete it (it was before a simplification)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
fanf: yes, right!

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/ArchiveApiTest.scala
```diff
@@ -110,8 +109,8 @@ class ArchiveApiTest extends Specification with AfterAll with Loggable {
   def children(f: File) = f.children.toList.map(_.name)
 
   // format: off
-//  org.slf4j.LoggerFactory.getLogger("application.archive").asInstanceOf[ch.qos.logback.classic.Logger].setLevel(ch.qos.logback.classic.Level.TRACE)
-//  org.slf4j.LoggerFactory.getLogger("configuration").asInstanceOf[ch.qos.logback.classic.Logger].setLevel(ch.qos.logback.classic.Level.TRACE)
+  org.slf4j.LoggerFactory.getLogger("application.archive").asInstanceOf[ch.qos.logback.classic.Logger].setLevel(ch.qos.logback.classic.Level.TRACE)
```
clarktsiory, line 112: The changes are ok !
But here maybe the trace logs should be kept as comments... :innocent:

fanf:
OK, merging this PR

PR #5153, opened 2023-11-02
Fixes #23688: Import archive must refuse yaml technique with mismatch directory and id

fanf:
https://issues.rudder.io/issues/23688

Add a check that Technique ID from path pattern matches technique from descriptor for YAML and JSON case after parsing the descriptor. We already have everything, just need to test at the correct place. 

The most part of the change is an unit test.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
@@ -766,14 +766,30 @@ class ZipArchiveReaderImpl(
      * Both technique.json and technique.yaml have priority above metadata.xml
      */
     def checkParseTechniqueDescriptor(
-        id:      TechniqueId,
-        optTech: Ref[Option[TechniqueInfo]],
-        name:    String,
-        content: Array[Byte]
+        basePath: String,
+        id:       TechniqueId,
+        optTech:  Ref[Option[TechniqueInfo]],
+        name:     String,
+        content:  Array[Byte]
     ): IOResult[(String, Array[Byte])] = {
       import YamlTechniqueSerializer._
       import com.normation.rudder.ncf.yaml.{Technique => YTechnique}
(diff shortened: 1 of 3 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -1003,25 +1020,30 @@ class CheckArchiveServiceImpl(
...
+              s"technique with that name already exists in category '${existing}': it must be imported in the " +
+              s"same category, please update your archive."
+            ).fail
+          }
+      }
     }
+
+    checkTechniqueExistSameCat()
   }
+
```
clarktsiory, line 1046: Instead of wrapping the method body with a more descriptive name, shouldn't it be better renaming `checkTechnique` to `checkTechniqueExistsSameCat` then ?
 Or simply comment on the method ? There already is a line comment but inside the method...

clarktsiory:
LGTM ! (apart from my comment on the method name)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
fanf: I change that b/c I though I would be able to just chain check here, so the shape would be: 
```
def checkTechnique(techInfo: TechniquesInfo, techArchive: TechniqueArchive): IOResult[Unit] = { 
  def checkTechniqueExistSameCat() = { ... }
  def otherCheck() = {... }

  checkTechniqueExistSameCat() *> otherCheck()
```

I was not able to do it b/c that chek need to be done before hand, and there's no simple way to factor out all checks at the same place, but I prefer to let it that way, it will be more simple to add other checks latter on (and the cost in perf/compilation/whatever is negligible)

fanf:
OK, squash merging this PR

PR #5156, opened 2023-11-03
Fixes #23606: Creating files with the file explorer fails when using invalid character

clarktsiory:
https://issues.rudder.io/issues/23606

We don't want the user to rename a file with invalid characters : we display a notification, and we can also avoid making an API request.

Also, when a API request fails, a generic error message will be displayed as notification


clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5160, opened 2023-11-06
Fixes #23699: Event log migration takes too long

VinceMacBuche:
https://issues.rudder.io/issues/23699

COALESCE on the whole table takes a looooong time. An update to remove all null before modifying the column should be enough


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):
LGTM. We have an indicator of end of migration in log, so it's ok to fordaemon without more log FMPOV.

PR #5163, opened 2023-11-07
Fixes #23707: Implement common zio-json utils for plugins in rudder 

clarktsiory:
https://issues.rudder.io/issues/23707

It is about moving code from [a plugin](https://github.com/Normation/rudder-plugins-private/pull/468/files#diff-53797078b2054da1a7e7aee8e315cd630043045f9e1ff19916e01a88a39c32e1L26), because it needs to be shared by many plugins and also probably in the API endpoints definitions when migrating lift-json to zio-json...

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RudderJsonRequest.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RudderJsonRequest.scala
@@ -1,0 +1,59 @@
+/*
+ *************************************************************************************
+ * Copyright 2023 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RudderJsonRequest.scala
```diff
@@ -0,0 +1,40 @@
+package com.normation.rudder.rest
```
fanf, line 1: You need to add the copyright header on new files

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RudderJsonRequest.scala
```diff
@@ -0,0 +1,40 @@
...
+      // end copy
+
+      req.body match {
+        case eb: EmptyBox => Left(Unexpected((eb ?~! "error when accessing request body").messageChain))
+        case Full(bytes) => decoder.decodeJson(new String(bytes, charset)).left.map(Unexpected(_))
+      }
+    } else {
+      Left(Unexpected("Cannot parse non-JSON request as JSON; please check content-type."))
+    }
+  }
```
fanf, line 31: This code exactly is already in `ZioJsonExtractor#parseJson`

We don't want to duplicate such code, maintenance would be harder. Is there an intent with the copy ?

fanf:
You need to add a description explaining the architecture/implem choice and - if needed - impact.

Question: why a new file in

fanf (changes requested):
Please:
- complete description (for ex "similar to RudderJsonResponse for the request side, nothing special")
- add missing headers
- avoid copy/pasting

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RudderJsonRequest.scala
clarktsiory: It's just that I initially moved that code from the plugins repo, and I didn't realize it was the same :smile:

I reused `ZioJsonExtractor#parseJson` then, it is also more suited that only methods about requests are exposed in `RudderJsonRequest`

clarktsiory:
PR updated with a new commit

fanf (approved):

fanf:
OK, squash merging this PR

PR #5166, opened 2023-11-09
Fixes #23724: 

fanf:
https://issues.rudder.io/issues/23724

So, most of the PR is about changing string to `doobie.Fragment` which leads to correctly escaped prepared statement (ie `?` + parameter). 
I also cut down information going back to UI, and kept the SQL errors for people able to see webapp logs.

Some subtilities:
- we need to parse `asc` `desc` because they are SQL keyword, so can't be passed with fragment (it seems)
- we need to use correct types for fragments, so for example we need `java.sql.Timestamp` in place of strings
- `Fragment.const` is necessary only for case where we want to avoid to interpolate (in the SQL prepared statement way) a variable but interpolate it as a string. I chose to use it for anything without SQL variable, included constant strings

There does not seems to have a lot of tests on event logs appart `EventLogJdbcRepositoryTest` which is ok, so I tested by hand:
- the fact that last generation time is correctly found, 
- in event log admin page, `filter` field with sql chars,
- dates (no start, no end, none (it's long), both), changing number of events by page

No plugin seems to be impacted, even `change-validation` (while it uses eventLog, but not the problematic  methods)


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #5170, opened 2023-11-15
Fixes #23736: Implement common utils for compatibility of zio-json API response

clarktsiory:
https://issues.rudder.io/issues/23736

When migrating from lift-json to zio-json, we do not currently have a way in our zio-json custom definitions to provide an `id` in the json response in case of error. We will need that in order to be compatible and have the same output when migrating.


clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM

fanf:
OK, squash merging this PR

PR #5171, opened 2023-11-15
Fixes #23738: Error in upmerge (empty TestMigrateSystemTechnique7_0.scala)

fanf:
https://issues.rudder.io/issues/23738


fanf:
Merging by hand since it breaks the build (and @VinceMacBuche knows about it)

PR #5174, opened 2023-11-16
Fixes #23741: Testing json from API definitions in YAML should have a nice display

clarktsiory:
https://issues.rudder.io/issues/23741

Now when testing for json equality (for instance in this test class [here](https://github.com/Normation/rudder/blob/e62443c97bbcd4f08516479fd974b658b5383438/webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestFromFileDef.scala#L93)), we can call `doTest(Nil, semanticJson = true)` that will check for semantic equality only (key-values in json object have no order).

We could as well do :

```scala
doTest(List("api_x.yaml")) // false by default
doTest(List("api_x.yaml", "api_y.yaml"), semanticJson = true)
```

!


clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5178, opened 2023-11-16
Fixes #23735: Exception when bad input for inventory duration config

clarktsiory:
https://issues.rudder.io/issues/23735

The only message that an `ExceptionInInitializerError` provides is `null`, so we do not want to show the message but the cause instead...

Also for config value errors, adopting a standard error format and catching errors could always be beneficial so I used that to improve the errors logs during boot when config errors should prevent the server from running...   

I also took into account the suggestions to add comments to the sample configuration file, and use the main value `rudder.batch.reportsCleaner.deleteLogReport.TTL` in the configuration in favor of `rudder.batch.reportsCleaner.deleteReportLog.TTL` (which we do not use in the code)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
@@ -218,6 +218,7 @@ import org.bouncycastle.jce.provider.BouncyCastleProvider
 import org.joda.time.DateTimeZone
 import scala.collection.mutable.Buffer
 import scala.concurrent.duration.FiniteDuration
+import scala.util.Try
 import zio.{Scheduler => _, System => _, _}
 import zio.syntax._
 
@@ -387,6 +388,7 @@ object RudderProperties {
 }
 
 object RudderParsedProperties {
+  import RudderConfigInit._
   import RudderProperties.config
 
   val logger = ApplicationLogger.Properties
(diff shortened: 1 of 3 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -644,11 +646,16 @@ object RudderParsedProperties {
   val RUDDER_BATCH_REPORTSCLEANER_COMPLIANCE_DELETE_TTL =
     config.getInt("rudder.batch.reportscleaner.compliancelevels.delete.TTL") // AutomaticReportsCleaning.defaultDeleteTTL
   val RUDDER_BATCH_REPORTSCLEANER_LOG_DELETE_TTL = {
-    try {
+    (Try {
+      config.getString("rudder.batch.reportscleaner.deleteLogReport.TTL")
```
fanf, line 650: it should be `rudder.batch.reportsCleaner.deleteLogReport.TTL` (with 'C' for cleaner)

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM

Nice output on error: 
image
:+1:

fanf:
OK, squash merging this PR

PR #5181, opened 2023-11-20
Fixes #23756: Comparison json should allow to display json diff

clarktsiory:
https://issues.rudder.io/issues/23756

I should not have used `^^` because it returns a generic `Matcher[T]` instead of an `EqualityMatcher[T]`, this suppressed the _Click to see difference_ button in IntelliJ...


fanf (approved):
Looks like it works!

PR #5199, opened 2023-11-24
Fixes #23786: Inventory accept test sometimes fails on callback check

fanf:
https://issues.rudder.io/issues/23786

Add a retry on the check for the callback part since it is async and can fail. 200 ms should be enough for all case, as Microsoft says.


clarktsiory:
LGTM !

clarktsiory (approved):

PR #5157, opened 2023-11-03
Fixes #23671: Locale issue: date picker outputs date in wrong order 

ElaadF:
https://issues.rudder.io/issues/23671
:warning: We should validate that this change doesn't break anything and need more attention. We should not merge this without validating that is not breaking any other part of Rudder

PR #5201, opened 2023-11-24
Bug #23671 : Locale issue: date picker outputs date in wrong order 

ElaadF:
https://issues.rudder.io/issues/23671

This is a simple fix, it only affects the JavaScript date picker module in group criteria that was not sync with the expected [defined format](https://github.com/Normation/rudder/pull/5201/files#diff-5192ead8ee6e8bbdfc1c1de061965986ff9d79c1e6216b2dc3ac06ba68e61582L567) in the backend to `dd/MM/yyyy`.

In this case, this bug affects the criteria `Last inventory date` in group criteria page. It appears that the format of inventory date is like `2023-08-17 14:31:49+0200` so to the format `yyyy/mm/dd`.

image

I think the date picker should use the same format. Despite the fact that the date picker and the inventory date doesn't use the same format, this PR fix the problem and the filter seems to work

datepicker

This is not an ideal solution since we use the format `yyyy/mm/dd` in general in the interface, but changing the format in the backend, means higher risk of breaking anything else, and we should consider refactoring this part and think about what standard we want to apply everywhere to be consistent.
Here there are some ideas about what can we do to change the format in backend: https://github.com/Normation/rudder/pull/5157


VinceMacBuche (dismissed):

VinceMacBuche:
Commits are not well formed !!

VinceMacBuche (approved):

PR #5204, opened 2023-11-24
Fixes #23791: Add enable/disable commands to rudder-package

Fdall:
https://issues.rudder.io/issues/23791


relay/sources/rudder-package/src/lib.rs
```diff
@@ -193,4 +195,67 @@ pub mod action {
         }
         Ok(())
     }
+
+    pub fn enable(
+        mut w: Webapp,
```
amousset, line 200:
```suggestion
        w: &mut Webapp,
```

PR #5206, opened 2023-11-28
Fixes #23798: rudderc method parsing breaks when method contains several bundles

amousset:
https://issues.rudder.io/issues/23798

Modify the regex to prevent capturing previous bundle calls in the match, so we're matching each bundle, and get back on the previous "first bundle is the main bundle" convention.

Before 
image

After
image


amousset:
Doesn't work, don't merge yet.

amousset:
PR updated with a new commit

Fdall (approved):

amousset:
OK, squash merging this PR

PR #5210, opened 2023-11-28
Fixes #23802: rudderc should export a technique with ids

amousset:
https://issues.rudder.io/issues/23802

* Remove the `export` subcommand as we now need to be able to build the techniques, hence identical parameters to `build`, so it becomes an option of the `build` command
* Fix condition serialization. For some reason the `false` condition was serialized to `null`.
* Run export test on all test techniques, and also check validity of the `technique.ids.yml` file by re-building it.

### DIFF
--- a/policies/.gitignore
+++ b/policies/.gitignore
@@ -7,3 +7,4 @@
 /rudderc/docs/examples/*.cf
 /rudderc/docs/examples/*.ps1
 rudderc/src/methods.json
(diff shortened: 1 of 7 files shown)


amousset:
PR updated with a new commit

Fdall (approved):

amousset:
OK, squash merging this PR

PR #5214, opened 2023-11-29
Fixes #23806: Add Amazon Linux in the technique editor conditions

amousset:
https://issues.rudder.io/issues/23806


amousset:
Not finished

PR #5224, opened 2023-11-30
Fixes #23818:  Import archive for an existing technique delete the previous technique

fanf:
https://issues.rudder.io/issues/23818

We didn't update the status of file in [GitArchivers.scala](https://github.com/Normation/rudder/pull/5224/files#diff-bb5734a3493073add0df1291f147ac6cd515b8a211f543610446796ed2374441) after the compilation was done, and we were basing our cleaning based on the "before compilation" status. 

So the main change is to track what files are produce by `rudderc` and weave that info up to the `GitArchiver` where we need it to compute the updated status. 

Appart that, add logs to better understand what is happening for a next time.


clarktsiory (approved):
LGTM !

PR #5225, opened 2023-12-01
Fixes #23789: system status tab show normal directives rather than system directives

VinceMacBuche:
https://issues.rudder.io/issues/23789

Add a parameter to node compliance api to get only system rules

Pass this parameter to api calls of nodecompliance

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -102,6 +102,7 @@ class ComplianceApi(
           case API.GetRulesCompliance       => GetRules
           case API.GetRulesComplianceId     => GetRuleId
           case API.GetNodesCompliance       => GetNodes
+          case API.GetNodeSystemCompliance  => GetNodeSystemCompliance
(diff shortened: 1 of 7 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -297,7 +298,8 @@ class ComplianceApi(
       (for {
         level     <- restExtractor.extractComplianceLevel(req.params)
         precision <- restExtractor.extractPercentPrecision(req.params)
-        nodes     <- complianceService.getNodesCompliance()
+        onlySystem = req.params.get("system").flatMap(_.headOption.map(_ == "true")).getOrElse(false)
```
fanf, line 301: shouldn't that change be removed to not modify public API?

fanf (commented):
I think there is a change remaining from early iteration that should not be here. Else, LGTM

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
OK, squash merging this PR

PR #5226, opened 2023-12-01
Fixes #23332: Serde errors do not point to the correct line number

amousset:
https://issues.rudder.io/issues/23332

image

Solves probably 75% of the problem with a low effort.

The idea is to avoid parsing an untagged enum with serde (as it totally breaks error reporting), and use a plain struct instead, and then convert manually to the different kinds.
The higher level errors still do not point to a line in the source (but it was never the case, and would require heavy lifting, like replacing serde with a lower level YAML parser), but all reporting error line are now correct.


amousset:
OK, squash merging this PR

PR #5229, opened 2023-12-03
Fixes #23797: Backend evolution for node RBAC: add tenants domain concept

fanf:
https://issues.rudder.io/issues/23797

This PR adds the foundation to have RBAC for nodes and demonstrates the usage on node list/node details pages. 
It brings: 
- a new attribute in node, `SecurityTag`, which is a list of `Tenant`  (just an ID container)
- a new user attribute, `tenants`, for user access to tenants
- the filtering of nodes based on tenants in `CoreNodeFactRepository`
- a way to weave user context from UI to `CoreNodeFactRepository` via `QueryContext`

# Easing and limiting migration

I didn't want to alter existing API by adding `QueryContext` to for ex `FullNodeInventory`, so I prefered to replace these services by direct use of `CoreNodeFactRepository` where possible. 

For the same reason, I change the signature of `CoreNodeFactRepository.getAll` to just return the underlying map of `CoreNodeInfo` and avoid stream. It then (almost) mirror existing `NodeInfo` / `NodeIventory` methods, making further changes in plugins etc easier. 
It should also bring an interesting performance improvement by avoiding a lots of transform in that method. 

# Tenant /securityTag spec

The tenant ID is chosen to be only alpha-num non empty strings (with _-). It doesn't seems necessary to have more for the ID, and we will be able to add a "tenants" structure in the future, with more info (description, etc). 

The security tag is added in `CoreNodeFact.rudderSettings` zone. 
It's persisted in LDAP with its json serialisation, which should use the futur addition/change of what is containted in a security tag (and we now know how to search and do things on these thing in LDAP if needed, but since it's in `CoreNodeFact`, we will have access at the scala object most of the time). 

For the user side, it's standard parsing or `rudder-users.xml`, similar but simplier to roles (since we don't have security tag depending upon others). 
I just added two special value, `*` (any nodes can be accessed) and `-` (no node can be access). 

## Migration logic

Since we want to compatible with previsous version of rudder, we need to have:
- no `tenants` means `tenants="*"` 
- `*` have access to nodes without a `SecurityTag`

Then, it felt better to have the rule:
- `tenants=""` means `tenants="-"` and NOT `*` (which would be better aligned with the previous rule) because I forsee case where tenant list will be managed automatically, with action like "remove tenant XXX to all users", and we don't want that user who had only that tenant gain the right to see all nodes. 

The consistent behaviour with/without the futur plugin will be implemented once the code is plugin-ized. 

# Known missing bits. 

This is just a PoC of the foundation. It does not provide:
- a way to set the securitytag for nodes. An LDAP borwser is needed
- the API dont' have tenants for now. I will need a specific dev (same migration consistency than user). For internal API, we can rely on `CurrentUser` value for authentication (even in API processing)
- only the minimum part to be able to demonstrate the feature to node details was done. In particular, there's a billion place that use `NodeInfoService` and that will need to use `CoreNodeFactRepository` to provide the `QueryContext` (and I can't modify `NodeInfoService` without massive impact on plugins, and non iterable work).

# About `CurrentUser` and  `SessionVar`

Spring now have a security fonction that changes the session ID at each connection. This is good from a security point of view, since it makes session fixation attacks hard. 
But it seems to bode ill with regard to the init of our `CurrentUser` value. 
It looks like `spring-security` is rewriting the value of session and of its `SecurityContextHolder` at different moment, which are not compatible with lift session management of session vars. 
This is not exposed in our previous usage of `CurrentUser` because it happened "late enough". Now, with the `QueryContext` calling on it very early, it breaks. And since we clear everything session-related if the `CurrentUser` is not well defined, it brings havoc.

Since actually, the value of the current user is *already* in a session variable, I just removed the use of lift `SessionVar`.

We have another `SessionVar` for toasts, but since there is no interaction between lift and an other framework to set/get the value, I believe it will be OK for it. 

# PR commits 

I reorganized the PR by commit to help understand what is going on at each step. I will rebase before merge.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
```diff
@@ -1289,17 +1302,139 @@ object NodeFactChangeEvent {
...
+ * latter need, by-item solution need to be used (see for ex: SelectFacts for nodes)
+ */
+final case class QueryContext(
+    actor:     EventActor,
+    nodePerms: NodeSecurityContext
+)
+
+object QueryContext {
+  // for test
+  implicit val testQC: QueryContext = QueryContext(eventlog.RudderEventActor, NodeSecurityContext.All)
```
VinceMacBuche, line 1347: I think it should be created in test classes. to prevent we use it in code by mistake

VinceMacBuche (approved):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
VinceMacBuche: but that could be done in another time i think

PR #5232, opened 2023-12-05
Fixes #23836: Event log migration should not be done every time

clarktsiory:
https://issues.rudder.io/issues/23836

The current migration have some cost with queries such as `UPDATE col SET col=${defaultValue} WHERE col is NULL`, we want to avoid doing that when we know the migration was already applied : it is a simple check on whether the column is nullable or not, from the PG table `information_schemas.columns`.

The should be executed if the result of the check on the column returns a row : 
```sql
SELECT DISTINCT column_name, table_name, is_nullable
FROM INFORMATION_SCHEMA.COLUMNS 
WHERE table_name = tablename AND column_name = migrationcolumn and is_nullable = 'YES'
```

We add a debug log to inform when the migration does not need to be made, and we improve the error log whenever the migration fails (also added database tests)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateEventLogEnforceSchema.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateEventLogEnforceSchema.scala
@@ -39,11 +39,15 @@ package bootstrap.liftweb.checks.migration
 
 import bootstrap.liftweb.BootstrapChecks
 import bootstrap.liftweb.BootstrapLogger
-import com.normation.errors.IOResult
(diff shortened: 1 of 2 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf:
Good explanation, good work with @VinceMacBuche
will need to be thoroughly tested on nightly on non migrated machines

fanf (approved):
Gg

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #5233, opened 2023-12-05
Fixes #23837: Enforce change request and workflow table schema

clarktsiory:
https://issues.rudder.io/issues/23837

We need to add `NOT NULL` constraint to the change-validation plugin tables, in order to enforce the table schema  (until now there does not seem to be a `null` value that has been written in the table, in that case the migration would fail but the user would have a nice error log). I added database tests.

Note that this PR is very similar to https://github.com/Normation/rudder/pull/5232

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateChangeValidationEnforceSchema.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateChangeValidationEnforceSchema.scala
@@ -1,0 +1,121 @@
+/*
+ *************************************************************************************
+ * Copyright 2023 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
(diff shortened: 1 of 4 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateChangeValidationEnforceSchema.scala
```diff
@@ -0,0 +1,128 @@
...
+  private def alterTableStatement(table: Fragment, column: Fragment): Fragment = {
+    sql"alter table $table alter column $column set not null"
+  }
+
+  private def shouldMigrateColumn(tableName: String, columnName: String): ConnectionIO[Boolean] = {
+    sql"""
+      select count(*)
+      from information_schema.columns
+      where table_name = $tableName
+      and column_name = $columnName
```
fanf, line 82: ]we have a convention of always using `. ${Var}` for interpolation

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/MigrateChangeValidationEnforceSchema.scala
```diff
@@ -0,0 +1,128 @@
...
+      and is_nullable = 'YES'
+    """.query[Int].unique.map(_ > 0)
```
fanf, line 84: This method has been written several times.
Shouldn't it be time to put it in a `DbCommonMigration` or something?

fanf (commented):
Looks OK once feedback are addressed, I still need to test it on real data

fanf (approved):
LGTM

fanf:
OK, squash merging this PR

PR #5241, opened 2023-12-08
Fixes #23826: Reorganize the header of the node page

RaphaelGauthier:
https://issues.rudder.io/issues/23826

As part of the integration of the compliance score in the 'Summary' nodes tab , we have moved some information to the page header, so that it is constantly accessible and frees up space in this tab.

These changes not only affect the nodes page, they are also visible in the pending nodes table

Here is the result : 
node-header-8 1
node-header-pending-8 1


RaphaelGauthier:
Commit modified

RaphaelGauthier:
Commit modified

RaphaelGauthier:
PR rebased

VinceMacBuche (approved):

PR #5246, opened 2023-12-13
Fixes #23895: We can access system rules detail page from node system status tab

VinceMacBuche:
https://issues.rudder.io/issues/23895


fanf:
It's ok for the rule part, but we still have the pen appearing for system directives: 
image

VinceMacBuche:
Commit modified

fanf (approved):

PR #5254, opened 2023-12-14
Fixes #23446: Change request on special:all_nodes_without_role lead to error

fanf:
https://issues.rudder.io/issues/23446

I think we don't want to modify event logs (as a general rule). We should perhaps be able to store old event logs in a "cold storage" (other table, or crypted file, or whatever) and wait for old format to disappear. 

In the meantime, we will just parse the old and new format (it's extermelly limited in that case, so almost without impact)


VinceMacBuche (approved):

VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #5255, opened 2023-12-14
Fixes #22170: A new group display no nodes while all linux are selected

fanf:
https://issues.rudder.io/issues/22170

Nothing special: just remove the init of group at creation, only let `None` - it doesn't seems to break anything (generation, test, coming back to the group page, update dyn groups, api - at least)


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5261, opened 2023-12-14
Fixes #23915: Remove BadPolicyMode case from computation

VinceMacBuche:
https://issues.rudder.io/issues/23915

We have no way to decide wether a policy is sending report for correct mode or not since it can now be embedded within the technique details. and can produce report for a different type


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ReportType.scala
```diff
@@ -81,8 +81,6 @@ object ReportType {
   def apply(report: Reports, policyMode: PolicyMode): ReportType = {
     import PolicyMode._
```
amousset, line 82:
```suggestion
```

PR #5267, opened 2023-12-15
Fixes #23920: Lift Async system is not able to find spring SecurityContextHolder

fanf:
https://issues.rudder.io/issues/23920

Explanation are in comment: basically, things are complicated and several tools trying to store things in thread local vars lead to problems.


fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #5271, opened 2023-12-18
Fixes #23932: Plugin license implementation is not coherent with interface

VinceMacBuche:
https://issues.rudder.io/issues/23932

Some explanation within issue description. The goal here is to be coherent between the trait and all implementation (All to def), this breaks release of some plugins


VinceMacBuche:
Commit modified

fanf (approved):

PR #5272, opened 2023-12-19
Fixes #23935: API for node group compliance

clarktsiory:
https://issues.rudder.io/issues/23935

This adds two endpoints that return compliance computation by node group : 
- one `/compliance/groups/{id}` for global compliance by taking into account any rule that apply to nodes within the group
- another `/compliance/groups/{id}/target` for targeted compliance by taking into account only rules that target the group (using include/exclude targeting)

In the yaml tests I used concrete examples (a simple one and a more complex one) and the result correspond to what is expected with a some caveats : 
- lists are not sorted, and this could cause some issues because sometimes we use the `Set` collection. Maybe we should sort by 'compliance', from the "worst" ones to best ones ? Or by the name of the _node_/_rule_ ? I left them as is so there is no specific order
- when there are 'skipped' reports, with overrides, we remap them to "No report" with the `NoAnswer` report type, which could be confusing (but may be better than omitting it completely). Later we could handle that case specifically to display which rule/directive have been overriden

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -43,10 +43,14 @@ import com.normation.inventory.domain.NodeId
 import com.normation.rudder.api.ApiVersion
 import com.normation.rudder.domain.logger.TimingDebugLogger
 import com.normation.rudder.domain.logger.TimingDebugLoggerPure
+import com.normation.rudder.domain.nodes.NodeGroup
+import com.normation.rudder.domain.nodes.NodeGroupId
 import com.normation.rudder.domain.policies.Directive
 import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.GroupTarget
 import com.normation.rudder.domain.policies.Rule
 import com.normation.rudder.domain.policies.RuleId
+import com.normation.rudder.domain.policies.RuleTarget
 import com.normation.rudder.domain.reports.BlockStatusReport
 import com.normation.rudder.domain.reports.ComplianceLevel
 import com.normation.rudder.domain.reports.CompliancePrecision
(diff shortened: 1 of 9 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -679,6 +750,179 @@ class ComplianceAPIService(
...
+      currentGroupNodeIds = allGroups.getOrElse(nodeGroup.id, Chunk.empty)
+
+      targetedNodeIdsByRuleId =
+        rules.map(rule => (rule.id, RoNodeGroupRepository.getNodeIdsChunk(allGroups, rule.targets, nodeInfos))).toMap
+
+      filteredRules = rules.filter(rule => {
+                        val targetedNodeIds      = targetedNodeIdsByRuleId(rule.id)
+                        val isRuleTargetingGroup =
+                          RuleTarget.merge(rule.targets).includes(GroupTarget(nodeGroup.id)) // targeted compliance
+                        (isGlobalCompliance || isRuleTargetingGroup) && targetedNodeIds.exists(currentGroupNodeIds.contains(_))
```
clarktsiory, line 793: (:exploding_head:)

This adds the check for `isRuleTargetingGroup` only when compliance is not global (i.e. targeted compliance)

There are multiple of occurences of : 
```
(condA || condB) && condC 
``` 
in this PR and the way to read them is : "I always want to check `condC`", and "I only want to check `condB` when `condA`"

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

VinceMacBuche:
@clarktsiory Tests are not ok !

I wondering about the NoAnswer case. I need to look if this is the same behavior on nodes/rules/directives

clarktsiory:
PR updated with a new commit

clarktsiory:
I rebased on master (some changes from the use of `NodeFactRepository` instead of `NodeInfoService`)

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #5274, opened 2023-12-19
Fixes #23941: Improve rudderc test error message when no library is passed in the parameters

Fdall:
https://issues.rudder.io/issues/23941

### DIFF
--- a/policies/rudderc/src/lib.rs
+++ b/policies/rudderc/src/lib.rs
@@ -339,6 +339,9 @@ pub mod action {
             if libraries.len() > 1 {
                 bail!("Tests only support one library path containing a full 'ncf' library");
             }
+            if libraries.is_empty() {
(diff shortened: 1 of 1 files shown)


policies/rudderc/src/lib.rs
```diff
@@ -339,6 +339,9 @@ pub mod action {
             if libraries.len() > 1 {
                 bail!("Tests only support one library path containing a full 'ncf' library");
             }
+            if libraries.is_empty() {
+                bail!("At least one library path must be passed using the '--library' option");
```
amousset, line 343:
```suggestion
                bail!("One library path must be passed using the '--library' option");
```

amousset: For the test command it must be exactly one

PR #5276, opened 2023-12-19
Control boxSequence/sequence are using early exist, which is not supp…

mbaechler:
…orted in Scala 3

	By the way, sequence has been renamed to traverse and boxSequence to sequence
	because it's how these concepts are called in the community.

	Early exit has been eliminated by ... plug-in cats


fanf:
> By the way, sequence has been renamed to traverse and boxSequence to sequence
because it's how these concepts are called in the community.

Ah yes, a time posterior to their inception IIRC :) 
I'm not sure it worth it, the goal is to delete box everywhere (but then, there's not that many so why not :shrug: )

mbaechler:
> > By the way, sequence has been renamed to traverse and boxSequence to sequence
> > because it's how these concepts are called in the community.
> 
> Ah yes, a time posterior to their inception IIRC :) I'm not sure it worth it, the goal is to delete box everywhere (but then, there's not that many so why not 🤷 )

As the implementation is already there and getting rid of early return is required to compile with scala 3, this solution seems the most straightforward.

Of course, dropping `Box` sounds nice, but we can do it at a later, right?

fanf:
Closing, superseeded by #5280

PR #5281, opened 2023-12-21
Fixes #23930: lost audit/enforce on the node compliance page

RaphaelGauthier:
https://issues.rudder.io/issues/23930


webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeCompliance/ApiCalls.elm
```diff
@@ -31,6 +32,38 @@ getPolicyMode model =
...
+        , headers = []
+        , url     = getUrl model [ "directives" ] []
+        , body    = emptyBody
+        , expect  = expectJson GetDirectivesInfo decodeGetDirectives
+        , timeout = Nothing
+        , tracker = Nothing
+        }
+  in
+    req
+
```
fanf, line 66: I don't think we want to do that. It will be extremely costly when the number of rules/directives growths, especially since we get all rules/directives, and not the ones for the node. 

A better approach would be to extends `getNodeComplianceId` api to provide the compliance mode - and moreover, I'm pretty sure it's something that would be interesting to have in the public API too.

fanf (changes requested):
We can't go that way, it will be too costly. A better fashion would be to extend `getNodeComplianceId` to directly have the wanted information.

webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeCompliance/ApiCalls.elm
clarktsiory: The API part is done here : https://github.com/Normation/rudder/pull/5308/files
You can change the decoder at the level of the `RuleCompliance` and `DirectiveCompliance` data types : just adding a `policyMode` field will do, it simplifies also a lot of things in this PR I guess :smile:

RaphaelGauthier:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
I did the change :
image

clarktsiory:
Closing this PR, the fix is done in this other one : https://github.com/Normation/rudder/pull/5351

PR #5282, opened 2023-12-21
Fixes #23951: Fix some small things in rudder-package

Fdall:
https://issues.rudder.io/issues/23951

### DIFF
--- a/relay/sources/rudder-package/src/repository.rs
+++ b/relay/sources/rudder-package/src/repository.rs
@@ -20,7 +20,7 @@ use crate::{
     config::{Configuration, Credentials},
     signature::{SignatureVerifier, VerificationSuccess},
     webapp::Webapp,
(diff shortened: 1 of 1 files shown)


relay/sources/rudder-package/src/repository.rs
```diff
@@ -79,11 +79,24 @@ impl Repository {
         let res = req.send()?;
 
         // Special error messages for common errors
+        let error_conf_msg = format!("Please check your configuration in {}", CONFIG_PATH);
```
amousset, line 82: It might not be `CONFIG_PATH` depending on the CLI parameter.

Fdall:
Setting as draft as the fix is not as trivial as initially thought.
The effective config path is not exposed to the error catching message and so, printing which path the user need to modify requires some extra changes.

PR #5284, opened 2023-12-21
Fixes #22197: We can accept a node with an existing hostname even if node_accept_duplicated_hostname is false

fanf:
https://issues.rudder.io/issues/22197

Not sure what happened in the last refactoring, but here, the failure was yield (as a `Full(Failure(..))` in place of being flatMapped. 

Also: I remove the possibility to write several time the hostname, we know they are all the same, since that's what we are checking. 

Finaly, move that check in first place, because it's really useless to move into accepted, then rollback into pending if the hostname is duplicated: just start by checking if it's ok or not to move.


VinceMacBuche:
I tried to make a suggested change through github UI but it made a commit directly (i also tried to make a PR but it failed ....) 

I may not pass spotless ...

VinceMacBuche:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5286, opened 2023-12-21
Fixes #18952: Static group which depends on dynamic subgroup must be forbiden

fanf:
https://issues.rudder.io/issues/18952

Manage the cases where subgroup changes cause chaos on other subgroup, either by being deleted, or becoming static (with dyn dep) or decoming dynamic (used in subgroup of static group). 

There is a lot of lines changed, but most are error handling. 

The logic is similar for both "look if the node becoming dynamic is a dependency of a static group" or "look if the group becoming static as a dynamic group as a subgroup": each time, unwind the the query of the interesting group, look for the nature of dependencies. 

The error is not returned in the IOResult error channel, since it's user-actionnable: it's an option (in that case, the `Some` contains a string for the list of bad groups). 

The configirmation pop-up on deletion is handle by cssSel magic, I was happy to not have to change its HTML at all. 

- the warning for deletion when used on other group: 
 
image

- the error when a dyn group depending on dyn group becomes static:

image

- the error when a static group dependency becomes dynamic: 

image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
@@ -39,13 +39,17 @@ package com.normation.rudder.web.components
 
 import bootstrap.liftweb.RudderConfig
 import com.normation.box._
+import com.normation.errors.IOResult
 import com.normation.plugins.DefaultExtendableSnippet
 import com.normation.rudder.AuthorizationType
 import com.normation.rudder.domain.nodes._
 import com.normation.rudder.domain.policies._
+import com.normation.rudder.domain.queries.CriterionLine
 import com.normation.rudder.domain.queries.NewQuery
 import com.normation.rudder.domain.queries.Query
+import com.normation.rudder.domain.queries.QueryTrait
 import com.normation.rudder.domain.queries.ResultTransformation
+import com.normation.rudder.domain.queries.SubGroupComparator
 import com.normation.rudder.domain.workflows.ChangeRequestId
 import com.normation.rudder.repository.FullNodeGroupCategory
 import com.normation.rudder.services.workflows.DGModAction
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
```diff
@@ -528,12 +644,22 @@ class NodeGroupForm(
...
-          case Some(_) =>
-            SetHtml("confirmUpdateActionDialog", popup.popupContent()) &
-            JsRaw("""createPopup("confirmUpdateActionDialog")""")
+        if (popup.popupWarningMessages.isEmpty && dependingSubgroups.isEmpty) {
+          popup.onSubmit()
+        } else {
+          val html: NodeSeq = dependingSubgroups match {
+            case None      => popup.popupContent()
+            case Some(msg) =>
+              val cssSel: CssSel = "#explanationMessageZone *+" #>
```
fanf, line 653: that cssSel means: "add after the last child in content"

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
clarktsiory: woaaah it's quite inventive

clarktsiory (approved):
Overall this works fine, except for one little edge case : 
* I set a static parent group to have a sub-group A which is static, it works fine
* Then I change the sub-groups of the parent to another sub-group B which is also static
* Then I change the sub-group A to be dynamic (which is fine, it has no parent)
* Finally I change the parent group to have A as sub-group... And it's still saving without any error ! But there should be, because the parent is static and A is now dynamic, it breaks the rule !

Maybe there is some state left in memory here, but it seems hard to debug. Ideally we should later move that whole logic to another part (`NodeFactRepo`) so it's fine for now !

fanf:
OK, squash merging this PR

PR #5289, opened 2023-12-22
Fixes #23957: Testing the API with a custom user role in yaml is not possible 

clarktsiory:
https://issues.rudder.io/issues/23957

Somehow, when `CurrentUser.set` is run outside of the specs2 `Fragment` which executes the tests, it simply does not set anything... We actually have to `set` in individual tests, in order to set the current user so that they are visible when calling `CurrentUser.get` from the services...


fanf:
I don't think you wan to do that like that. 
I think you need to have a `test` user service, that let you set whoever you want as an user. 
So I think you should change: 
- the fact that there is a `TestActor` in mock service ,
- the implementation of `userService` so that it accepts a setter
- possibly, the service will need to go into `MocService` to get the actor from there

Then, you have a real integration system, that really exercizes what the API subsystem is doing regarind authentication. 
Then, you can change on a test by test or file by file case what user you want.

clarktsiory:
Indeed, the [UserService](https://github.com/Normation/rudder/blob/6281da8a4c4cd0bd36587a5d4deb362e30c4ccc2/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/UserService.scala#L80) should be the one used for mocking the user...

But I learned from in this PR that when testing something that need to access `CurrentUser` directly from the session, in a global context (like lift snippets), we need to execute tests in a `S.initIfUninitted(session) { /* tests here */ }` or else the `set` will not have effect...

PR #5169, opened 2023-11-13
Fixes #23593: Upgrading Bootstrap to version 5

RaphaelGauthier:
https://issues.rudder.io/issues/23593

First PR on the upgrade to Bootstrap 5. Here the biggest change is the renaming of all the html attributes "data-xxx" to "data-bs-xxx", as well as the renaming of CSS classes that changed between version 3 and 5.

PR #5290, opened 2023-12-26
Fixes #23926: Display group compliance

clarktsiory:
https://issues.rudder.io/issues/23926

Creates an Elm app for display of compliance by group (I copied most of the app structure from directives).
It also changes a bit of Scala frontend code in the summary page of a group, which therefore uses javascript to display compliance (this page has not been rewritten in Elm yet).

There is also some change in the Scala backend to add the policy mode of a rule and a node to the API response.

The 'global' and 'targeted' compliance is specific to groups, and it can be seen as filters for the rules and nodes displayed in the compliance tree : 
image
When the filter value is changed (e.g. from 'global' to 'targeted'), I made the compliance table display in "reload" state until the new compliance data is fetched, as it may take some time for the new data to be shown.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupCompliance/ViewNodesCompliance.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupCompliance/ViewNodesCompliance.elm
@@ -1,0 +1,90 @@
+module GroupCompliance.ViewNodesCompliance exposing (..)
+
+import Html exposing (..)
+import Html.Attributes exposing (..)
+import Html.Events exposing (onClick, onInput)
+import List
+import List.Extra
+import String
+import Tuple3
+import Dict
+
+import GroupCompliance.ApiCalls exposing (..)
+import GroupCompliance.DataTypes exposing (..)
+import GroupCompliance.ViewUtils exposing (..)
+import Compliance.Utils exposing (displayComplianceFilters, filterDetailsByCompliance)
+
+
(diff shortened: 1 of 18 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupCompliance/ViewNodesCompliance.elm
```diff
@@ -0,0 +1,90 @@
...
+  in
+    ( if model.ui.loading then
+      generateLoadingTable
+      else
+      div[]
+      [ div [class "table-header extra-filters"]
+        [ div [class "d-inline-flex align-items-baseline pb-3 w-25"]
+          [
+            div [class "btn-group yesno"]
+            [ label [class ("btn btn-default" ++ if isGlobalMode then " active" else ""), style "box-shadow" (if isGlobalMode then "inset 0 3px 5px rgba(0,0,0,.125)" else "none"), onClick (LoadCompliance GlobalCompliance)]
```
clarktsiory, line 53: The styling is done "by-hand" here for testing, it is currently broken since the [bootstrap 5 migration](https://github.com/Normation/rudder/pull/5169), the shadow is no longer present when the button is active...

(This should be fixed globally in another PR)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

PR #5295, opened 2023-12-28
Fixes #23402: In quicksearch, query error are not displayed which leads to infinite loading result

clarktsiory:
https://issues.rudder.io/issues/23402

This formats errors as a nicer message for the user. There is a caveat when doing error transformation from `PureResult` (`Either`s) to `Box`, which is still the "effect type" in this API : if `.toBox` is not used at the end, we will end up with a exception cause message being serialized with both `err1; cause was : err2` and `err2 <- err3` formats...

Screenshot from 2023-12-28 18-32-30


clarktsiory:
Commit modified

fanf (approved):
LGTM, just need a rebase on latest HEAD

clarktsiory:
PR rebased

clarktsiory:
PR rebased successfully

fanf (approved):

fanf:
OK, merging this PR

PR #5296, opened 2023-12-29
Fixes #23967: Rudder score backend and API

VinceMacBuche:
https://issues.rudder.io/issues/23967

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
@@ -453,3 +453,20 @@ CREATE TABLE NodeFacts (
 , acceptRefuseFact  jsonb  -- the big node fact data structure
 , deleteEvent       jsonb  -- { 'date': 'rfc3339 timestamp', 'actor': 'actor name' }
 );
+
+
+Create table GlobalScore (
+  nodeId  text primary key
+, score   text
+, message text
+, details jsonb
+);
+
+Create table scoreDetails (
+  nodeId  text
+, scoreId    text
+, score   text
+, message text
+, details jsonb
+, PRIMARY KEY (nodeId, scoreId)
+);
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
@@ -93,6 +93,44 @@ final case class CompliancePercent(
   val compliance = success + repaired + notApplicable + compliant + auditNotApplicable
 }
 
+final case class ComplianceSerializable(
+    applying:                   Option[Double],
+    successNotApplicable:       Option[Double],
+    successAlreadyOK:           Option[Double],
+    successRepaired:            Option[Double],
+    error:                      Option[Double],
+    auditCompliant:             Option[Double],
+    auditNonCompliant:          Option[Double],
+    auditError:                 Option[Double],
+    auditNotApplicable:         Option[Double],
+    unexpectedUnknownComponent: Option[Double],
+    unexpectedMissingComponent: Option[Double],
+    noReport:                   Option[Double],
+    reportsDisabled:            Option[Double],
+    badPolicyMode:              Option[Double]
+)
+
+object ComplianceSerializable {
+  def fromPercent(compliancePercent: CompliancePercent) = {
+    ComplianceSerializable(
+      if (compliancePercent.pending == 0) None else Some(compliancePercent.pending),
+      if (compliancePercent.notApplicable == 0) None else Some(compliancePercent.notApplicable),
+      if (compliancePercent.success == 0) None else Some(compliancePercent.success),
+      if (compliancePercent.repaired == 0) None else Some(compliancePercent.repaired),
+      if (compliancePercent.error == 0) None else Some(compliancePercent.error),
+      if (compliancePercent.compliant == 0) None else Some(compliancePercent.compliant),
+      if (compliancePercent.nonCompliant == 0) None else Some(compliancePercent.nonCompliant),
+      if (compliancePercent.auditError == 0) None else Some(compliancePercent.auditError),
+      if (compliancePercent.auditNotApplicable == 0) None else Some(compliancePercent.auditNotApplicable),
+      if (compliancePercent.unexpected == 0) None else Some(compliancePercent.unexpected),
+      if (compliancePercent.missing == 0) None else Some(compliancePercent.missing),
+      if (compliancePercent.noAnswer == 0) None else Some(compliancePercent.noAnswer),
+      if (compliancePercent.reportsDisabled == 0) None else Some(compliancePercent.reportsDisabled),
+      if (compliancePercent.badPolicyMode == 0) None else Some(compliancePercent.badPolicyMode)
+    )
+  }
+}
+
 object CompliancePercent {
 
   // a correspondance array between worse order in `ReportType` and the order of fields in `ComplianceLevel`
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
@@ -52,6 +52,8 @@ import com.normation.rudder.facts.nodes.QueryContext
 import com.normation.rudder.hooks.HookEnvPairs
 import com.normation.rudder.hooks.PureHooksLogger
 import com.normation.rudder.hooks.RunHooks
+import com.normation.rudder.score.InventoryScoreEvent
+import com.normation.rudder.score.ScoreServiceManager
 import com.normation.utils.StringUuidGenerator
 import com.normation.zio.currentTimeMillis
 import zio._
@@ -173,5 +175,12 @@ class TriggerPolicyGenerationPostCommit[A](
     (if (inventory.node.main.status == AcceptedInventory) {
        IOResult.attempt(asyncGenerationActor ! AutomaticStartDeployment(ModificationId(uuidGen.newUuid), RudderEventActor))
      } else ZIO.unit) *> records.succeed
+  }
+}
+
+class TriggerInventoryScorePostCommit[A](scoreServiceManager: ScoreServiceManager) extends PostCommit[A] {
+  override def name:                                    String      = "trigger Score computation on inventory update"
+  override def apply(inventory: Inventory, records: A): IOResult[A] = {
+    scoreServiceManager.handleEvent(InventoryScoreEvent(inventory.node.main.id, inventory)) *> records.succeed
   }
 }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
@@ -1,0 +1,83 @@
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.score
+
+import com.normation.errors.PureResult
+import com.normation.inventory.domain.InventoryError.Inconsistency
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.reports.ComplianceSerializable
+import com.normation.rudder.score.ScoreValue.A
+import com.normation.rudder.score.ScoreValue.B
+import com.normation.rudder.score.ScoreValue.C
+import com.normation.rudder.score.ScoreValue.D
+import com.normation.rudder.score.ScoreValue.E
+import zio.json.DeriveJsonEncoder
+import zio.json.EncoderOps
+import zio.json.JsonEncoder
+
+object ComplianceScoreEventHandler extends ScoreEventHandler {
+  implicit val compliancePercentEncoder: JsonEncoder[ComplianceSerializable]     = DeriveJsonEncoder.gen
+  def handle(event: ScoreEvent):         PureResult[List[(NodeId, List[Score])]] = {
+
+    event match {
+      case ComplianceScoreEvent(n, percent) =>
+        (for {
+          p <- ComplianceSerializable.fromPercent(percent).toJsonAST
+        } yield {
+          val scoreId = "compliance"
+          val score   = if (percent.compliance >= 100) {
+            Score(scoreId, A, "Node is compliant at 100%", p)
+          } else if (percent.compliance >= 75) {
+            Score(scoreId, B, "Node is compliant at least at 75%", p)
+          } else if (percent.compliance >= 50) {
+            ComplianceSerializable.fromPercent(percent)
+            Score(scoreId, C, "Node is compliant at least at 50%", p)
+          } else if (percent.compliance >= 25) {
+            Score(scoreId, D, "Node is compliant at least at 25%", p)
+          } else {
+            Score(scoreId, E, "Node is compliant at less then 25%", p)
+          }
+          ((n, score :: Nil) :: Nil)
+        }) match {
+          case Left(err) => Left(Inconsistency(err))
+          case Right(r)  => Right(r)
+        }
+      case _                                => Right(Nil)
+    }
+  }
+}
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
@@ -1,0 +1,119 @@
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.score
+
+import com.normation.NamedZioLogger
+import com.normation.errors.PureResult
+import com.normation.inventory.domain.Inventory
+import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.reports.CompliancePercent
+import zio.json.DeriveJsonDecoder
+import zio.json.DeriveJsonEncoder
+import zio.json.JsonDecoder
+import zio.json.JsonEncoder
+import zio.json.ast.Json
+
+sealed trait ScoreValue {
+  def value: String
+}
+
+object ScoreValue {
+  case object A extends ScoreValue { val value = "A" }
+  case object B extends ScoreValue { val value = "B" }
+  case object C extends ScoreValue { val value = "C" }
+  case object D extends ScoreValue { val value = "D" }
+  case object E extends ScoreValue { val value = "E" }
+
+  val allValues: Set[ScoreValue] = ca.mrvisser.sealerate.values
+
+  def fromString(s: String) = allValues.find(_.value == s.toUpperCase()) match {
+    case None    => Left(s"${s} is not valid status value, accepted values are ${allValues.map(_.value).mkString(", ")}")
+    case Some(v) => Right(v)
+  }
+}
+
+case class NoDetailsScore(scoreId: String, value: ScoreValue, message: String)
+case class Score(scoreId: String, value: ScoreValue, message: String, details: Json)
+
+case class GlobalScore(value: ScoreValue, message: String, details: List[NoDetailsScore])
+
+object GlobalScoreService {
+  def computeGlobalScore(oldScore: List[NoDetailsScore], scores: List[Score]): GlobalScore = {
+
+    val correctScores = scores.foldRight(oldScore) {
+      case (newScore, acc) =>
+        NoDetailsScore(newScore.scoreId, newScore.value, newScore.message) :: acc.filterNot(_.scoreId == newScore.scoreId)
+    }
+    import ScoreValue._
+    val score         = if (correctScores.exists(_.value == E)) { E }
+    else if (correctScores.exists(_.value == D)) { D }
+    else if (correctScores.exists(_.value == C)) {
+      C
+    } else if (correctScores.exists(_.value == B)) {
+      B
+    } else A
+    GlobalScore(score, s"There is at least a Score with ${score.value}", correctScores)
+  }
+}
+
+trait ScoreEvent
+
+case class InventoryScoreEvent(nodeId: NodeId, inventory: Inventory)                  extends ScoreEvent
+case class ComplianceScoreEvent(nodeId: NodeId, compliancePercent: CompliancePercent) extends ScoreEvent
+
+trait ScoreEventHandler {
+  def handle(event: ScoreEvent): PureResult[List[(NodeId, List[Score])]]
+}
+
+object ScoreSerializer {
+  implicit val scoreValueEncoder: JsonEncoder[ScoreValue] = JsonEncoder[String].contramap(_.value)
+  implicit val scoreValueDecoder: JsonDecoder[ScoreValue] = JsonDecoder[String].mapOrFail(ScoreValue.fromString)
+
+  implicit val noDetailsScoreEncoder: JsonEncoder[NoDetailsScore] = DeriveJsonEncoder.gen
+  implicit val noDetailsScoreDecoder: JsonDecoder[NoDetailsScore] = DeriveJsonDecoder.gen
+
+  implicit val globalScoreEncoder: JsonEncoder[GlobalScore] = DeriveJsonEncoder.gen
+  implicit val globalScoreDecoder: JsonDecoder[GlobalScore] = DeriveJsonDecoder.gen
+
+  implicit val jsonScoreEncoder: JsonEncoder[Score] = DeriveJsonEncoder.gen
+  implicit val jsonScoreDecoder: JsonDecoder[Score] = DeriveJsonDecoder.gen
+}
+
+object ScoreLoggerPure extends NamedZioLogger {
+  override def loggerName: String = "score"
+}
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
@@ -1,0 +1,130 @@
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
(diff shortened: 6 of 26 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -453,3 +453,20 @@ CREATE TABLE NodeFacts (
...
+  nodeId  text primary key
+, score   text
+, message text
+, details jsonb
+);
+
+Create table scoreDetails (
+  nodeId  text
+, score   text
+, name    text
```
fanf, line 468: `name` should be `scoreId` I think

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -93,6 +93,44 @@ final case class CompliancePercent(
...
+    unexpectedMissingComponent: Option[Double],
+    noReport:                   Option[Double],
+    reportsDisabled:            Option[Double],
+    badPolicyMode:              Option[Double]
+)
+
+object ComplianceSerializable {
+  def fromPercent(compliancePercent: CompliancePercent) = {
+    ComplianceSerializable(
+      if (compliancePercent.pending == 0) None else Some(compliancePercent.pending),
```
fanf, line 116: you should have a private def to avoid duplicating that logic and avoid risking micopying the name in both place

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/ComplianceLevel.scala
```diff
@@ -93,6 +93,44 @@ final case class CompliancePercent(
...
+    error:                      Option[Double],
+    auditCompliant:             Option[Double],
+    auditNonCompliant:          Option[Double],
+    auditError:                 Option[Double],
+    auditNotApplicable:         Option[Double],
+    unexpectedUnknownComponent: Option[Double],
+    unexpectedMissingComponent: Option[Double],
+    noReport:                   Option[Double],
+    reportsDisabled:            Option[Double],
+    badPolicyMode:              Option[Double]
```
fanf, line 110: the elements are not in the same order than `CompliancePercent`, that seems to be an extremely easy way to have errors. 
For the naming, it seems better to use the event name like you did, but I wonder why we don't use them in CompliancePercent too.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactServiceProxies.scala
```diff
@@ -74,12 +76,14 @@ import zio.syntax._
...
-    val basePostPipeline:  Seq[PostCommit[Unit]]
+    val basePostPipeline:  Seq[PostCommit[Unit]],
+    scoreServiceManager:   ScoreServiceManager
 ) extends PipelinedInventorySaver[Unit] {
 
   override def commitChange(inventory: Inventory): IOResult[Unit] = {
     implicit val cc = ChangeContext.newForRudder()
-    backend.updateInventory(FullInventory(inventory.node, Some(inventory.machine)), Some(inventory.applications)).unit
+    backend.updateInventory(FullInventory(inventory.node, Some(inventory.machine)), Some(inventory.applications)).unit *>
+    scoreServiceManager.handleEvent(InventoryScoreEvent(inventory.node.main.id, inventory)).unit
```
fanf, line 86: Why puting it here and not in a `PostCommit` handler? They should be able to match that use case, and I don't see the need for hard implementation coupling here. And I don't think it gives any additional consistency benefits.

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -453,3 +453,20 @@ CREATE TABLE NodeFacts (
 , acceptRefuseFact  jsonb  -- the big node fact data structure
 , deleteEvent       jsonb  -- { 'date': 'rfc3339 timestamp', 'actor': 'actor name' }
 );
+
+
+Create table GlobalScore (
+  nodeId  text primary key
+, score   text
+, message text
```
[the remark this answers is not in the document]
VinceMacBuche: It's the score explanation message. I think it is really a message. It is not a configuraiton structure and i do think it should not follow the same rules than other items

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
VinceMacBuche: if you want

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -453,3 +453,20 @@ CREATE TABLE NodeFacts (
...
+, details jsonb
+);
+
+Create table scoreDetails (
+  nodeId  text
+, score   text
```
fanf, line 467: I think it could be an enumerated type, it seems to be the perfect case for that, and it would likely help regarding perf and size. But I don't know what is the migration story if for ex we want to add new values.

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -453,3 +453,20 @@ CREATE TABLE NodeFacts (
...
+, name    text
+, message text
+, details jsonb
+, PRIMARY KEY (nodeId, name)
+);
```
fanf, line 472: you can add not null not empty constraints for nodeId, score, name

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
```diff
@@ -0,0 +1,251 @@
...
+
+  val allValues: Set[ScoreValue] = ca.mrvisser.sealerate.values
+
+  def fromString(s: String) = allValues.find(_.value == s.toUpperCase()) match {
+    case None    => Left(s"${s} is not valid status value, accepted values are ${allValues.map(_.value).mkString(", ")}")
+    case Some(v) => Right(v)
+  }
+}
+
+trait Score[T] {
```
fanf, line 42: I don't think we want to have that parametrization. We never use a generic access to details, and `List[Score[_]]` is a signal that we maybe are not doing things like we should. 

Plus, that parametrization risk to create some unpleasant situation since `Score[ComplianceSerializable]` is not the same as `ComplianceScore` (at type level), but in our code, we really really want to have that equality.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  implicit val globalScoreEncoder: JsonEncoder[GlobalScore] = DeriveJsonEncoder.gen
+  implicit val globalScoreDecoder: JsonDecoder[GlobalScore] = DeriveJsonDecoder.gen
+
+  implicit val jsonScoreEncoder: JsonEncoder[JsonScore] = DeriveJsonEncoder.gen
+  implicit val jsonScoreDecoder: JsonDecoder[JsonScore] = DeriveJsonDecoder.gen
+}
+
+trait ScoreSerializerService {
+  def parse(score:  Score[Json]): IOResult[Option[Score[_]]]
```
fanf, line 29: I think it should be a pure result, I don't see why we should have IO here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+  def toJson(score: Score[_]):    IOResult[Option[Score[Json]]]
```
fanf, line 30: I don't see how it can fail. I understand to potential `None` case if there is no serializer for that score, but if there is one, it should not ever fail (and so just be a `Option[Score[Json]]`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+  override def toJson(score: Score[_]): IOResult[Option[Score[Json]]] = {
+    score.details match {
+      case details: SystemUpdateStats =>
+        details.toJsonAST.toIO.map(JsonScore(score.value, score.name, score.message, _)).map(Some(_))
+      case _ => None.succeed
+    }
+  }
+}
+
+class ScoreSerializer {
```
fanf, line 77: about naming: I think that this one is the `ScoreSerializerService`, and the it has a `val serializers: Ref[List[ScoreSerializer]]`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
fanf: I wonder if that can't be simplier. 
It's used in two place: 
- in DB to read/write from a JSON column (so a string), and using the intermediary `Json` means that we do `db -> get String -> parse to Json -> serialize to String -> parse to T` - we could certainly directly use the string, even if that means having a `StringScore` to simplify doobie interaction, 
- in node API, to serialize these info. But here, we should really just need the `JsonEncoder` which we already have (and again avoid the `T -> to String -> parse Json -> to String`)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
```diff
@@ -0,0 +1,114 @@
...
+
+  val serializers: Ref[List[ScoreSerializerService]] =
+    Ref.make(ComplianceScoreSerializerService :: SystemUpdateScoreSerializerService :: List.empty[ScoreSerializerService]).runNow
+
+  def registerSerializer(handler: ScoreSerializerService) = {
+    serializers.update(handler :: _)
+  }
+
+  def parse(score: Score[Json]): IOResult[Score[_]] = {
```
fanf, line 86: I looks like the things are done so that several serializer can work on a Score (because of the logic of "try all serializer and get the first" vs "find a serializer that can handle that case and make it do its work"). Is there a case where it would be interesting to have several serializer succeeding or is this just no specific reasons ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
fanf: Something like the first commit in that PR: https://github.com/VinceMacBuche/rudder/pull/29

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
fanf: General idea draft in the second commit of that PR: https://github.com/VinceMacBuche/rudder/pull/29 (not sure it floats till the end, but we can in the worst case keep `JsonScore` for `NodeAPI`)

VinceMacBuche: I prefer to remove invalid score (that are not json) at the beginning. and it will end up parsed as json (either in api, could be also used to init score if we wanted to recompute a score ) I don't think there is a usage in using a String and not a Json, which is the underlying type

VinceMacBuche: We already have everything to manage json from database. and here you have to translate into json at the end which can be already done

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
VinceMacBuche: maybe i need some services that may not be pure

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
VinceMacBuche: In fact we could remove all ComplianceScore, and juste avec a Score that has a Json as details. Just everything translated into a jsonScore directly from the score event

fanf: I like the idea to have different score implementation for the different cases, it makes the score computation very clean and really feels like each module can do whatever they want from a structured data type, and not Json. But perhaps I misunderstood ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
fanf: I didn't see any and parsing should really always be pure (and perhaps you need some impure one before geting the data, but having IO involved in the parsing feels really like not what we want to have to deal with)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
fanf: You can still check for invalid data structure at the begining, it doesn't change that. You certainly can use directly the `JsonEncoder` and `JsonDecoder` in the score repository, I just didn't take the time to implem it. 
Here, each time you use `as` or `toJsonAST`, you really go through a string parsing in the middle. 

And I really believe we should not have one unique JsonScore, it will make the score calculation part less clear.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
fanf: The third commit in https://github.com/VinceMacBuche/rudder/pull/29 there is an example of using `collectFirst` with view to stop at the first matching one

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
```diff
@@ -175,3 +177,10 @@ class TriggerPolicyGenerationPostCommit[A](
      } else ZIO.unit) *> records.succeed
   }
 }
+
+class TriggerInventoryScorePostCommit[A](scoreServiceManager: ScoreServiceManager) extends PostCommit[A] {
```
fanf, line 181: there is no need to add that one if there is already an `CoreNodeFact` event handler. 
Here, it's really only for the low level things related to inventory translation

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
```diff
@@ -0,0 +1,251 @@
...
+    } else if (correctScores.exists(_.value == B)) {
+      B
+    } else A
+    GlobalScore(score, s"There is at least a Score with ${score.value}", correctScores)
+  }
+}
+
+trait ScoreEvent
+
+case class InventoryScoreEvent(nodeId: NodeId, inventory: Inventory)                  extends ScoreEvent
```
fanf, line 91: It should not be Inventory here, but `CoreNodeFact` or `NodeFact` if the data is not in the first. Inventory is now an internal datastructure that will be removed (hopefully) in a later release

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
```diff
@@ -0,0 +1,251 @@
...
+
+    event match {
+      case InventoryScoreEvent(n, inventory) =>
+        val sum         = inventory.node.softwareUpdates.size
+        val security    = inventory.node.softwareUpdates.filter(_.kind == SoftwareUpdateKind.Security).size
+        val patch       = inventory.node.softwareUpdates.filter(_.kind == SoftwareUpdateKind.None).size
+        val defect      = inventory.node.softwareUpdates.filter(_.kind == SoftwareUpdateKind.Defect).size
+        val enhancement = inventory.node.softwareUpdates.filter(_.kind == SoftwareUpdateKind.Enhancement).size
+        val other       = inventory.node.softwareUpdates.filter { s =>
+          s.kind match {
```
fanf, line 133: it looks like we will want to add these info into `CoreNodeFact` at some point in the future

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
```diff
@@ -0,0 +1,104 @@
...
+  }
+
+  override def getScore(nodeId: NodeId, name: Option[String]): IOResult[List[Score[_]]] = {
+
+    val whereNode = Some(fr"nodeId = ${nodeId.value}")
+    val whereName = name.map(n => fr"name = ${n}")
+    val where     = Fragments.whereAndOpt(whereNode, whereName)
+    val q         = sql"select score, name, message, details from scoreDetails " ++ where
+    transactIOResult(s"error when getting scores for node")(xa => q.query[Score[Json]].to[List].transact(xa))
+      .flatMap(c => ZIO.foreach(c)(v => scoreSerializer.parse(v)))
```
fanf, line 66: I don't know enough of doobie nor looked at it, but it seems that here, you should be able to directly get a `JsonEncoder` from `scoreSerializer` for the correct type of score. 
Or just get a string from DB and translate it directly to the final type (I really don't think there is any optimisation in doobie for the Json type, and that it is retrieved as a full utf string before being decoded to json, ie you don't loose anything by telling you get a string and then decode it vs have doing both)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
fanf: (it's in `NodeFact`)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
fanf: Hum, the fact that the `InventoryScoreEvent` is triggered on inventory update (here or in nodefactrepository) means that we have to thing about the migration story to have score correctly init after a rudder upgrade. 
I doesn't seems trivial either, if we don't want for it to be too costly at boot, but something like "look if table score exist but is empty" is not enough, since it can fail in the middle. Right now I don't see how to do that but with a flag "migration finish" that can be quick checked at boot, and have a long migration where we look for each node without a system update score and compute it.

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -453,3 +453,20 @@ CREATE TABLE NodeFacts (
 , acceptRefuseFact  jsonb  -- the big node fact data structure
 , deleteEvent       jsonb  -- { 'date': 'rfc3339 timestamp', 'actor': 'actor name' }
 );
+
+
+Create table GlobalScore (
+  nodeId  text primary key
```
fanf, line 459: that will need a migration

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
```diff
@@ -0,0 +1,251 @@
...
+          case Some(g) => g.succeed
+          case None    => Inconsistency("No global score for node").fail
+        }
+    } yield {
+      res
+    }
+  }
+
+  def getScoreDetails(nodeId: NodeId): IOResult[List[Score[_]]] = {
+    for {
```
fanf, line 182: I think we should always return the list of score for ALL registered modules, and if the node is missing some score, have a default policy (perhaps configurable by score module). 
Else, if the node doesn't have a score (for ex because it was not computed yet), then it appears as not have it all all (and not like a missing score) ; plus the final score value won't be correct (we should have a way to tell "the value of that score is not computable yet because some data are missing", and not "C because it's the only data I have")

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
```diff
@@ -0,0 +1,251 @@
...
+    Ref.make(ComplianceScoreEventHandler :: InventoryEventHandler :: List.empty[ScoreEventHandler]).runNow
+
+  def registerHandler(handler: ScoreEventHandler) = {
+    handlers.update(handler :: _)
+  }
+
+  def handleEvent(scoreEvent: ScoreEvent) = {
+    (for {
+      h       <- handlers.get
+      _       <- ReportLoggerPure.info(s"new event ${scoreEvent}")
```
fanf, line 244: the need to be trace level (or debug perhaps?).
Perhaps it need its own `ScoreLoggerPure` too, we will likely need it to investigate strange computation results or things like that

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -339,46 +339,70 @@ object DirectiveApi       extends ApiModuleProvider[DirectiveApi]
...
+
+  final case object NodeInheritedProperties extends NodeApi with GeneralApi with OneParam with StartsAtVersion11 with SortIndex {
     val z              = implicitly[Line].value
-    val description    = "Get all proporeties for that node, included inherited ones"
+    val description    = "Get all propreties for that node, included inherited ones"
     val (action, path) = GET / "nodes" / "{id}" / "inheritedProperties"
   }
-  final case object ApplyPolicyAllNodes     extends NodeApi with GeneralApi with ZeroParam with StartsAtVersion8 with SortIndex  {
+
+  final case object NodeGlobalScore extends NodeApi with GeneralApi with OneParam with StartsAtVersion11 with SortIndex {
```
fanf, line 378: We need documentation for these two new API

fanf:
In the whole, this is working as expected, and the general architecture seems well done :clap: 
There is a lot of things missing but I don't think any of them are blocking: the interface points (database, API level, interplugin communication) seems OK. They will need to be better defined and documented with unit/integration tests, especially at elm level to demonstrate how a plugin will work (even if CVE is ready and does that too). 

Still, I would be more at ease with some changes on `Score[_]` structure now, it seems too complicated for the need, which makes everything harder to understand and evolve.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
VinceMacBuche: yes ScoreEvent will be different witth diffrent structure.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
VinceMacBuche: Also it's uised mostly in ZIO context and i have to transform my PureResult into IORestult to i guess it does not gain much

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
VinceMacBuche: I think we can wait for an inventory

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
VinceMacBuche: debug

VinceMacBuche: There is a lot of debug output that still needs to be removed, and tests not passing. told you this pr was not ready yet and UI PR needs to be merged before ;)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactServiceProxies.scala
VinceMacBuche: done

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
VinceMacBuche: I thought we wanted them private for now, don't we ? 

Also version is not ok yet

VinceMacBuche:
A lot simpler with latest commit (No more generic, details is a json that was serialized from data, we never translate anymore once it was writtten. It's just goes directly from DB to API, no translation in between)

VinceMacBuche:
Hence Serializer disappear. We may want to add it back one day if we want to translate from json to real classes (maybe to recompute a score ?), but that's no use for now

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreSerializer.scala
fanf: it gains in intent conveyed, and in unit test easyness. `IOResult` is a bit like global variables, they are easy and powerfull, but we should always try to restraint as much as possible to use them because they are *too* powerful.

fanf: (not a subject anymore since that part of the code disapeared, it's for info)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
fanf: discussed online, it will be better to have the score init as soon as possible to make user WOUAH, but clearly not blocking for that PR.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
fanf: discussed irl: 
- we need - in the future - way to decide for each node what score are enabled for it and only display these ones (and use these one for computation). Plus, some module could be optionnal, or only enabled on a given subset of nodes. 
- at some point in the future, we will have a "computation rule for that score" parameter, and it will perhaps need to be a per-node param, including the previous point ("enable/disable XXX score for that node")
- we still need a way to display a "not computed" score, ie a greyed/disabled version of score, which is a better feedback for the user than "nothing"
- we don't want for now to include disabled module as part of the computation

For now, only the last two points are relevant, but they are not blockers for that PR.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
fanf: ah yes, private seems more secure for now.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
```diff
@@ -0,0 +1,84 @@
...
+import com.normation.rudder.score.ScoreValue.C
+import com.normation.rudder.score.ScoreValue.D
+import com.normation.rudder.score.ScoreValue.E
+import zio.json.DeriveJsonEncoder
+import zio.json.EncoderOps
+import zio.json.JsonEncoder
+
+object ComplianceScoreEventHandler extends ScoreEventHandler {
+  implicit val compliancePercentEncoder: JsonEncoder[ComplianceSerializable]     = DeriveJsonEncoder.gen
+  def handle(event: ScoreEvent):         PureResult[List[(NodeId, List[Score])]] = {
```
fanf, line 56: it looks like we always have that structure: 
- a `Score` computation from a subtype of `ScoreEvent`
- a partial function from a `ScoreEvent` to a `Score` (ie a "can handle that type" + "if yes, use the computation function with the correct subtype of score event"

I think you need to split appart these two bits so that the `score from score event computation` is unit-testable by itself on two aspects:
- the correctness of the computation (easy here, but it will be a kind of spec so users will question it a lot, we need to be able to accumulate all the strange corner case they will find)
- the JSON format, because it becomes a storage format that will need to never change without a migration

Plus, once you have that separation, the implementation of the `handle` will be almost automatic, so more easy for other people needing to do so to understand what's going on.

fanf (commented):
LGTM for me, once spotless/tests pass. 
We tracked all the non blocking change elsewhere for latter PR.

webapp/sources/rudder/rudder-web/src/main/elm/sources/SystemUpdateScore.elm
```diff
@@ -0,0 +1,84 @@
...
+          [ toBadge "security"    "warning" details.security
+          , toBadge "bugfix"      "bug"     details.defect
+          , toBadge "enhancement" "plus"    details.enhancement
+          , toBadge "update"      "box"     details.patch
+          ]
+
+main =
+  Browser.element
+    { init = init
+    , view = always (Html.text "")
```
clarktsiory, line 59: Is it intended to not display anything ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -393,6 +397,7 @@ trait CachedFindRuleNodeStatusReports
       // * no report from the node (compliance expires): recompute compliance
       {
         (for {
+          _ <- ReportLoggerPure.warn(s"Hello ! ${actions} ")
```
clarktsiory, line 400: Please do not forget to remove this !

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+      (NodeId(d._1), Score(d._3, d._2, d._4, d._5))
+    }
+  }
+
+  import doobie._
+  override def getAll(): IOResult[Map[NodeId, List[Score]]] = {
+    val q = sql"select nodeId, score, name, message, details from scoreDetails "
+    transactIOResult(s"error when getting scores for node")(xa => q.query[(NodeId, Score)].to[List].transact(xa))
+      .map(_.groupMap(_._1)(_._2))
+  }
```
clarktsiory, line 92: Those tuple indices look really confusing between both implicits (but they look correct)
Normally doobie can generate the `Read[(NodeId, Score)]` by re-using the `Read[Score]` so we can remove the noise from `scoreWithIdRead` here

webapp/sources/rudder/rudder-web/src/main/elm/sources/SystemUpdateScore.elm
VinceMacBuche: Yep! these apps only exist to provide details for generic node details api that does not know how to handle them (in fact it could handle system-updates and compliance by itself but i wanted them to be like the plugins so that we know the recipe to make a plugin from Rudder

VinceMacBuche:
Commit modified

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
```diff
@@ -0,0 +1,132 @@
...
+  override def saveScore(nodeId: NodeId, score: Score): IOResult[Unit] = {
+
+    val query = {
+      sql"""insert into scoreDetails (nodeId, scoreId, score, message, details) values (${(nodeId, score)})
+           |  ON CONFLICT (nodeId, scoreId) DO UPDATE
+           |  SET score = ${score.value}, message = ${score.message}, details = ${score.details} ; """.stripMargin
+    }
+
+    transactIOResult(s"error when inserting global score for node '${nodeId.value}''")(xa =>
+      query.updateWithLabel("test").run.transact(xa)
```
clarktsiory, line 120: test !

VinceMacBuche:
PR updated with a new commit

clarktsiory (commented):
LGTM, it only misses the database migration and the not null constraint ! (And of course to fix the compilation errors)

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #5297, opened 2023-12-29
Fixes #23961: Display of related rules for groups should avoid detailing compliance

clarktsiory:
https://issues.rudder.io/issues/23961

* We display a rules tree in place of the previous table (most code is copied from the sidebar tree in the rules page).
* We have a badge like the one we have for policy mode, but with the `Included` or `Excluded` label instead.
* For that we create a new Elm app `Grouprelatedrules`. We already have another Elm app for groups compliance, which could be merged with this one later when we migrate the groups page to Elm

image

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
@@ -862,7 +862,7 @@ class RuleApiService14(
 ) {
 
   // this Id is special and use client side to identify missing rules
-  private val MISSING_RULE_CAT_ID = RuleCategoryId("ui-missing-rule-category")
+  val MISSING_RULE_CAT_ID = RuleCategoryId("ui-missing-rule-category")
 
   private def createChangeRequest(
       diff:      ChangeRequestRuleDiff,
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupRelatedRules/ApiCalls.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupRelatedRules/ApiCalls.elm
@@ -1,0 +1,27 @@
+module GroupRelatedRules.ApiCalls exposing (..)
+
+import Http exposing (..)
+import Url.Builder exposing (QueryParameter, string)
+
+import GroupRelatedRules.DataTypes exposing (..)
+import GroupRelatedRules.JsonDecoder exposing (..)
+
+getUrl: Model -> List String -> List QueryParameter -> String
+getUrl m url p=
+  Url.Builder.relative (m.contextPath :: "secure" :: "api"  :: url) p
+
+getRulesTree : Model -> RelatedRules -> Cmd Msg
+getRulesTree model relatedRules =
+  let
+    req =
+      request
+        { method  = "GET"
+        , headers = []
+        , url     = getUrl model [ "rulesinternal", "relatedtree" ] [string "rules" (String.join "," (List.map (.value) relatedRules.value))]
+        , body    = emptyBody
+        , expect  = expectJson (GetRulesResult relatedRules) decodeGetRulesTree
+        , timeout = Nothing
+        , tracker = Nothing
+        }
+  in
+    req
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
@@ -517,6 +517,7 @@ object JsonResponseObjects {
   // when returning a root category, we have a "data":{"ruleCategories":{.... }}. Seems like a bug, though
   final case class JRCategoriesRootEntryFull(ruleCategories: JRFullRuleCategory)
   final case class JRCategoriesRootEntrySimple(ruleCategories: JRSimpleRuleCategory)
+  final case class JRCategoriesRootEntryInfo(ruleCategories: JRRuleCategoryInfo)
 
   final case class JRSimpleRuleCategory(
       id:          String,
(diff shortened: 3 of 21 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupRelatedRules/ApiCalls.elm
```diff
@@ -0,0 +1,27 @@
...
+  Url.Builder.relative (m.contextPath :: "secure" :: "api"  :: url) p
+
+getRulesTree : Model -> RelatedRules -> Cmd Msg
+getRulesTree model relatedRules =
+  let
+    req =
+      request
+        { method  = "GET"
+        , headers = []
+        , url     = getUrl model [ "rules", "tree" ] []
```
fanf, line 20: I don't think we want to do that request: it will be huge when there is a lot of rules/directives, when we can have groups with just a few of them. 
And sending big json on the wire is costly. 
We should use a server-side filtered version of that request that only present rules and directives relevant for that group.

fanf:
When I click on the `Related rules` tab and there is no related rules, the workflow seems broken because we just have a list of nodes, but no information that there is no related rules. We should have the same general layout, just without content: 
image

webapp/sources/rudder/rudder-web/src/main/elm/sources/GroupRelatedRules/ApiCalls.elm
clarktsiory: I thought about adding a list of `ruleIds` to filter the tree on the API side, but I realized we also need to have the info : "Is the current group excluded or included by the each rule" and that blocked me. We [already have that info](https://github.com/Normation/rudder/pull/5297/commits/b602c955a7d570d4d5773e610fd30d7e947c55be#diff-b244376771e5d6dd4b7fa139fc009fa57b16707de2606cfdb1cabd35c817fc3cR222-R244)

So YES !  I guess we could just add the filter on that endpoint now... :smile:

clarktsiory:
PR updated with a new commit

clarktsiory:
@fanf  The issue about related rules screen being empty was just a "bug", I had it handled but in the wrong place, so the fix is there (just removing the "if tree is empty" condition) : 
https://github.com/Normation/rudder/pull/5297/commits/559672878e19b43fa0fe064f05a7f82969051799

image

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -454,7 +454,8 @@ class RuleApi(
 
     def process0(version: ApiVersion, path: ApiPath, req: Req, params: DefaultParams, authzToken: AuthzToken): LiftResponse = {
       implicit val qc: QueryContext = authzToken.qc
-      serviceV14.getCategoryTree().toLiftResponseOne(params, schema, s => Some(s.ruleCategories.id))
+      val ruleIdsFilter = req.params.get("ruleId").map(_.flatMap(RuleId.parse(_).toOption))
```
fanf, line 457: Can you add a comment explaining what that param does, the expected format ? And update the corresponding API doc?

fanf: It's likely that we need to increment the API version number too, since the bevavior won't be the same

fanf: (I think we said it's ok in that case, because we don't add a public api endpoint and we don't change the behavior of existing calls)

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
clarktsiory: Sure ! Here is the commit : https://github.com/Normation/rudder/pull/5297/commits/24d8d1059456a8cbe4b8bad622655b5af054cb8e

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
clarktsiory: + another one, I copied from an existing API doc but the format should be `?ruleId=id1&ruleId=id2...`, `exploded` should be `true`

fanf: we don't use that format anywhere, you should use `ruleIds=id1,id2...`

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
clarktsiory: Okay, I also found that `rules` is more consistent than `ruleIds` from other endpoints, so I changed the key : https://github.com/Normation/rudder/pull/5297/commits/b9e129d3eda2081bc6c3521b83666fda2fab98a3

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
clarktsiory: As we discussed, I created another private endpoint instead of changing this public one. It should return only needed info for rules related to a given group : https://github.com/Normation/rudder/pull/5297/commits

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
I added display also for "system" groups, it was not working before : https://github.com/Normation/rudder/pull/5297/commits/679242881a510c5c431de3d3fed4dd5186120f77

VinceMacBuche:
OK, squash merging this PR

PR #5304, opened 2024-01-04
Fixes #23956: Rudder agent scheduled runs are not triggered at regular interval

fanf:
https://issues.rudder.io/issues/23956

We want to present a new value in `rudder.json`. For that, we need to create a new  `SystemVariable` and its spec. 

The content of the variable is the real start time + a splay time derived from node ID.

The splay time computation was the easy part, and we just assume that the input is random (it's supposed to be an uuid) and get its byte as a BitInt, then just `modulo` so that it fits in the expected interval given by agent run period (in seconds).

I chose to use `java.time` structure since it seems that we will need to switch to it in a near futur, at least we can start to use them on new, isolated content. 

The hardest part was to rediscover that the `start hour` and `start minutes` in the `AgentRunInterval` data structure are not really the start time of the agent. They are more like a shift from midnight modulo agent run interval, which means that if the agent run is 5 minutes, whatever `start hour` contains, the actual start hour will always be `0`.

Finaly, that new data does not impact any of the existing system vars, and in particular it DOES NOT change the schedule string used by CFEngine agent - which is by design, since the agent will do localy its splaytime computation.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
@@ -60,6 +60,10 @@ import com.normation.rudder.reports._
 import com.normation.rudder.services.servers.PolicyServerManagementService
 import com.normation.rudder.services.servers.RelaySynchronizationMethod
 import com.normation.zio._
+import java.nio.charset.StandardCharsets
+import java.time.Duration
+import java.time.LocalTime
+import java.time.format.DateTimeFormatter
 import net.liftweb.common.Box
 import net.liftweb.common.EmptyBox
 import net.liftweb.common.Failure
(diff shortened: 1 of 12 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/SystemVariableService.scala
```diff
@@ -676,16 +690,49 @@ object ComputeSchedule {
...
+  /*
+   * Convert an agent schedule into the starttime string which include the splaytime
+   */
+  def getSplayedStartTime(nodeId: String, startTime: LocalTime, interval: Duration): LocalTime = {
+    startTime.plus(computeSplayTime(nodeId, interval))
+  }
 
+  def formatStartTime(startTime: LocalTime): String = {
+    // the withNano(0) is to only print "hh:mm:ss", not "hh:mm:ss.sss"
+    startTime.withNano(0).format(DateTimeFormatter.ISO_TIME)
```
clarktsiory, line 735: Since we use a `LocalTime` we will have no offset, but `DateTimeFormatter.ISO_TIME` could theoretically add one... `DateTimeFormatter.ISO_LOCAL_TIME` would have been more suited but it's also fine as is !

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestComputeSchedule.scala
```diff
@@ -83,16 +100,73 @@ class TestComputeSchedule extends Specification {
...
+    "be 0 when nodeId is empty" in {
+      ComputeSchedule.computeSplayTime("", fiveMinutes) must beEqualTo(Duration.ofSeconds(0))
+    }
+
+    "be stable for a given uuid" in {
+      val d1 = ComputeSchedule.computeSplayTime(uuid1, fiveMinutes)
+      val d2 = ComputeSchedule.computeSplayTime(uuid1, fiveMinutes)
+
+      (d1 === uuid1Splay) and (d2 === uuid1Splay)
+    }
```
clarktsiory, line 154: :+1:

clarktsiory (approved):
Nice change ! LGTM :+1:

fanf:
OK, merging this PR

PR #5308, opened 2024-01-05
Fixes #23987: Add policy mode info to node compliance API

clarktsiory:
https://issues.rudder.io/issues/23987

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -46,6 +46,8 @@ import com.normation.rudder.domain.logger.TimingDebugLoggerPure
 import com.normation.rudder.domain.nodes.NodeInfo
 import com.normation.rudder.domain.policies.Directive
 import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.GlobalPolicyMode
+import com.normation.rudder.domain.policies.PolicyMode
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -412,12 +416,13 @@ class ComplianceApi(
  * compliance for all rules/nodes/directives.
  */
 class ComplianceAPIService(
-    rulesRepo:                   RoRuleRepository,
-    nodeInfoService:             NodeInfoService,
-    nodeGroupRepo:               RoNodeGroupRepository,
-    reportingService:            ReportingService,
-    directiveRepo:               RoDirectiveRepository,
-    val getGlobalComplianceMode: () => Box[GlobalComplianceMode]
```
fanf, line 420: are you sure it's not used in a plugin or something? It's strange enough to have a val here to be cautious

fanf: ok, I didn't find anything :shrug:

fanf (approved):
LGTM

fanf:
OK, merging this PR

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
clarktsiory: I also didn't find anything :smile_cat:

PR #5309, opened 2024-01-10
Fixes #23998: Completly ignore source target of rules when importing an archive

fanf:
https://issues.rudder.io/issues/23998

Add the new merge policy which is just setting target to empty in addition to previous one when in use.

Also, I want to have a common naming convention, so use a set for names now to keep backward compat with the old name.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
@@ -175,23 +175,29 @@ object ArchiveScope       extends Enum[ArchiveScope] {
   }
 }
 
-sealed trait MergePolicy extends EnumEntry         { def value: String }
-object MergePolicy       extends Enum[MergePolicy] {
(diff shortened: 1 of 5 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -1574,11 +1576,17 @@ class SaveArchiveServicebyRepo(
       _ <- x match {
              case Some(value) =>
                // if merge policy is `keep-rule-groups`, update rule from archive with existing groups before saving
-               val ruleToSave = if (mergePolicy == MergePolicy.KeepRuleGroups) {
-                 r.copy(targets = value.targets)
-               } else r
+               val ruleToSave = {
+                 if (mergePolicy == MergePolicy.KeepRuleGroups || mergePolicy == MergePolicy.IgnoreSourceRuleGroups) {
```
clarktsiory, line 1580: the comment above also needs to be updated

PR #5315, opened 2024-01-12
Fixes #24016: Implement CSP strict headers with nonce and apply to healtcheck page

clarktsiory:
https://issues.rudder.io/issues/24016

Implementing CSP strict headers : 
```
Content-Security-Policy:
  script-src 'nonce-{RANDOM}' 'strict-dynamic';
  object-src 'none';
  base-uri 'none';
  ...
``` 
Given an html page to be rendered with an HTTP request which should be responded with these strict headers :
* We generate a unique random value during the rendering of request
* We include the nonce value in scripts in the page using `data-lift="with-nonce"` attribute, which will add `nonce="nonce-{RANDOM}"` as attribute to the script 
* For scripts defined in Scala code and generated in the page usually with `Script(OnLoad(JsRaw("...")))`, we also need a call to `WithNonce` to add the attribute to the scripts
* For `onclick` attributes containing javascript code, they are blocked by browser with those strict headers, so we need to create a separate script tag with a nonce attribute, and add it to that element

This PR implements it and applies it to the "Healtcheck" page which now has strict CSP headers :tada: 

> [!TIP]
> To migrate a page so that it works with those strict CSP headers, first step is to add the page regex : https://github.com/Normation/rudder/pull/5315/files#diff-7ad40f3003f8941c211ddaae2aa0bd0d82328a6c9cd8114c492b4a22f3abfe95R102-R105, next is to watch for errors and just follow those steps above to fix console errors due to blocked scripts.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/CustomPageJs.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/CustomPageJs.scala
@@ -1,0 +1,27 @@
+package com.normation.rudder.web.snippet
+
+import net.liftweb.http.DispatchSnippet
+import net.liftweb.http.LiftRules
+import net.liftweb.http.S
+
+/**
+  * We need to provide both lift.js and page js to the client since the scripts appended by lift is hard to intercept.
+  * TODO: find a way to remove these duplicates after merging, or parse the whole html and add nonces to the scripts after merging
+  * , see lift source code : 
+  * https://github.com/lift/framework/blob/master/web/webkit/src/main/scala/net/liftweb/http/LiftSession.scala#L995
+  * 
+  * For now both scripts will be duplicated in the page, it does not induce weird behavior
+  */
+class CustomPageJs extends DispatchSnippet {
+
+  def dispatch: DispatchIt = { case "pageScript" => _ => pageScript }
+
+  def pageScript = {
+    <script data-lift="with-nonce" src="/classpath/lift.js"></script>
+    <script type="text/javascript" data-lift="with-nonce" src={scriptUrl(s"page/${S.renderVersion}.js")}></script>
+  }
+
+  private def scriptUrl(scriptFile: String) = {
+    S.encodeURL(s"/${LiftRules.liftContextRelativePath}/$scriptFile")
+  }
+}
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/index.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/index.html
@@ -71,7 +71,7 @@
     </a>
   </div>
 </div>
-<script>
+<script data-lift="with-nonce">
   $(document).ready(function(){
     (function(){
       var urlVars = {};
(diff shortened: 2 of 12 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/CustomPageJs.scala
```diff
@@ -0,0 +1,27 @@
...
+  def dispatch: DispatchIt = { case "pageScript" => _ => pageScript }
+
+  def pageScript = {
+    <script data-lift="with-nonce" src="/classpath/lift.js"></script>
+    <script type="text/javascript" data-lift="with-nonce" src={scriptUrl(s"page/${S.renderVersion}.js")}></script>
+  }
+
+  private def scriptUrl(scriptFile: String) = {
+    S.encodeURL(s"/${LiftRules.liftContextRelativePath}/$scriptFile")
+  }
```
clarktsiory, line 26: Lifts adds a `<script src="/rudder/classpath/lift.js"/>` and another page script every time an html page is rendered, but it does not currently provide an easy way in the rendering lifecycle to replace/add the attributes to these elements. 

So for now we have console errors but at least each of the scripts are loaded once, and blocked once.

The only level at which it is possible to do it is when generating a `LiftResponse` i.e. with the whole response xml content already generated, and replace those scripts to add the nonce to them... 
I don't know which solution is "cleaner" though

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

amousset (approved):

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/CustomPageJs.scala
fanf: OK, we will need to correct that duing alpha (it means: "it's a new task in the milestone `ready for beta`)
Perhaps just have a hook in `LftRules.beforeSend` (or at the last moment before request is still xml, so that we can easily add nonce to script elements, see https://exploring.liftweb.net/master/index-9.html#sec:Request/Response-Lifecycle for the request lifecycle).

webapp/sources/rudder/rudder-web/src/main/webapp/index.html
```diff
@@ -71,7 +71,7 @@
     </a>
   </div>
 </div>
-<script>
+<script data-lift="with-nonce">
```
fanf, line 74: it feels scaringly error prone to have to add that attribute in *all* script tags. 
Since it's a completly automatic process, it looks like we should be able to totally abstract it away with a logic like "before sending response to client, add to ALL script tags the required nonce attribute". Less work to do, only one place to change in case something is wrong.  
(it can be done like that for now, it's a PoC. We will need to see if we can do better in alpha).

fanf (approved):
LGTM (modulo evolution to look for in alpha)

fanf:
OK, squash merging this PR

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/CustomPageJs.scala
clarktsiory: That's funny in the documentation they mean something different to the source code : 
```
  /**
   * Holds user functions that are executed before sending the response to client. The functions'
   * result will be ignored.
   */
  val beforeSend = RulesSeq[(BasicResponse, HTTPResponse, List[(String, String)], Box[Req]) => Any]
```
VS
```
Call any functions defined in LiftRules.beforeSend. This is the last place where you can modify the response before it’s sent to the user 
```
I'll try that, but since our `LiftResponse` is a `case class` I hope HTTPResponse can be modified easily !

PR #5316, opened 2024-01-12
Fixes #24018: AGENT_RUN_STARTTIME is not correctly computed with max delay

fanf:
https://issues.rudder.io/issues/24018


clarktsiory (approved):
LGTM !

PR #5326, opened 2024-01-17
Fixes #24034: Group compliance tab is loaded only with a double-click

clarktsiory:
https://issues.rudder.io/issues/24034

There is a `not` that was missing in the boolean condition reloading the tab with new data...

Also added accessibility in the html of groups tabs


VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #5329, opened 2024-01-18
Fixes #24024: case-sensitivity parameter of rudder-user.xml is not documented

ElaadF:
https://issues.rudder.io/issues/24024


fanf (approved):
LGTM, we will enhance a next time if still unclear.

PR #5331, opened 2024-01-18
Fixes #24036: If REST url ends with .com, then the .com is removed from path segments

fanf:
https://issues.rudder.io/issues/24036

The problem is that we rely on Lift URL parsing (because URL parsing is complicated and full of worms and we don't want to do by ourself), but lift, being helpfull, remove the last bit if it looks like an extension and put it in an `extension` parameter. 

Here, we just put it back when parsing looking for a REST URL. It won't impact the case where it would be a real site extension, since we don't authorize REST parsing to result in a root site path.


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #5334, opened 2024-01-22
Fixes #24047: Compliance has unoptimized usage of map view and policy mode computation  

clarktsiory:
https://issues.rudder.io/issues/24047

* The computation of policy mode of a rule requires knowing the node ids of the rule. Before, the node ids were re-computed every time even if it had already been... Now, they re-use existing values and save some iterations on collections
* The `nodeFacts` map being a `MapView[NodeId, CoreNodeFact]`, we better store the values we will need in an actual `Map` instead of recomputing from the view every time (e.g. a `Map[NodeId, RudderSettings]`, instead of storing the whole `CodeNodeFact` in memory)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -62,9 +62,9 @@ import com.normation.rudder.domain.reports.ValueStatusReport
 import com.normation.rudder.facts.nodes.CoreNodeFact
 import com.normation.rudder.facts.nodes.NodeFactRepository
 import com.normation.rudder.facts.nodes.QueryContext
+import com.normation.rudder.facts.nodes.RudderSettings
 import com.normation.rudder.reports.GlobalComplianceMode
 import com.normation.rudder.repository.FullActiveTechnique
-import com.normation.rudder.repository.FullNodeGroupCategory
 import com.normation.rudder.repository.RoDirectiveRepository
 import com.normation.rudder.repository.RoNodeGroupRepository
 import com.normation.rudder.repository.RoRuleRepository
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -771,61 +771,69 @@ class ComplianceAPIService(
...
 
-      // filter our rules that are applicable to any node in this group and get their compliances
-      currentGroupNodeIds = allGroups.getOrElse(nodeGroup.id, Chunk.empty)
+      nodeSettings = nodeFacts.mapValues(_.rudderSettings).toMap
 
-      targetedNodeIdsByRuleId = {
+      // A map for constant time access for the set of targeted nodes and policy mode:  a Map[RuleId, (Chunk[NodeId], Option[PolicyMode])]
+      // The set is reused to directly compute the policy mode of the rule
+      allRuleInfo = {
+        val allGroups = groupLib.allGroups.map { case (k, v) => (k, Chunk.from(v.nodeGroup.serverList)) }
```
ncharles, line 802: you certainly ought to time this part

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -1064,44 +1070,64 @@ class ComplianceAPIService(
                         )
                         .toIO
     } yield {
-
-      // get nodeIds by rules
-      val nodeByRules = rules.map(rule => (rule, RoNodeGroupRepository.getNodeIds(allGroups, rule.targets, nodeFacts)))
+      // A map to access the node fqdn and settings
+      val nodeInfos: Map[NodeId, (String, RudderSettings)] = nodeFacts.map {
```
fanf, line 1074: you can use `mapValues(..).toMap` here

fanf:
Since it's a PR to enhance perf, can you explicit the gain in either time/ram/io ? 
Ram is complicated to measure, and in first pass not our main concern. But time is very important, and should be checked with big dataset (reproducibility is hard, because your CPU is much faster than CI, but we need to find a way to check that)

fanf:
And, and so: can you explicit how much faster things are with that PR?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
clarktsiory: I'm not sure this would be interesting for this specific line of code. This operation is included with the 

```
_ <- TimingDebugLoggerPure.trace(s"getByNodeGroupCompliance - filterGlobalOrTargetedRules in ${t7 - t6} ms")
```
line for logging, which includes the timing for "all computations needed to filter the rules for the group", so this one included

clarktsiory: Do you mean to separate the timing in two different parts : "getRuleInfos" (policy mode and targeted nodes) and "filterRules" ?

clarktsiory:
PR updated with a new commit

clarktsiory:
* There is an unnecessary call to LDAP taking about 200ms which is removed per API call, replaced by an instantaneous call
* Apart from that there is only duplicate computations which could be removed, and I managed to compare timings and get x8 faster API calls with a set of 10000 nodes, and 10 rules targeting 1000 nodes, so the complexity is heavily reduced.

**Details :** the previous version was actually fetching all node facts, each time we wanted the policy mode of a rule (so when fetching the compliance of a node, it fetches all 10000 nodes facts 10 times and filters each time to take the 1000 nodes the rule apply to :upside_down_face:). Now we can avoid doing it per node so we do almost 10000 times less the same operation on the CPU. 
Instead in-memory we need to store a map of the node facts with the smallest size (rudder settings, fqdn, etc.), by fetching the node facts only once per REST API call. From the example I think the compromise CPU vs memory is worth when complexity increases

fanf (approved):
LGTM :+1:

fanf:
OK, squash merging this PR

PR #5337, opened 2024-01-22
Fixes #24051: Wrong agent type for dsc node in tests

VinceMacBuche:
https://issues.rudder.io/issues/24051

Wrong agent type for dsc node

also improve logging of policy build failing in test

PR #5339, opened 2024-01-23
Fixes #24051: Wrong agent type for dsc node in tests

fanf:
https://issues.rudder.io/issues/24051

Same as https://github.com/Normation/rudder/pull/5337 with indent spotlessed and an additional tests for interval check related to question in https://github.com/Normation/rudder-plugins-private/pull/497


webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestComputeSchedule.scala
```diff
@@ -183,5 +184,14 @@ class TestComputeSchedule extends Specification {
     "if the splay time is shorter than the interval, then the splay time length is taken" in {
       ComputeSchedule.computeSplayTime(uuid1, Duration.ofMinutes(10), fiveMinutes) === uuid1Splay
     }
+
+    "Check that the splaytime is strictly smaller that it's duration" in {
```
amousset, line 188:
```suggestion
    "Check that the splaytime is strictly smaller than its duration" in {
```

PR #5342, opened 2024-01-23
Fixes #24062: Implementing CSP headers without duplicating Lift scripts

clarktsiory:
https://issues.rudder.io/issues/24062

 ~One of a few solutions to remove the scripts appended to the HTML page by Lift : use the response byte array. The only straightforward way is to make use of mutability (array copying here) because Lift only provides us the final response (which is a case class).~

Lift has a mutable partial function `convertResponse` which we can re-assign to intercept the script if it contains duplicates, then remove the duplicates.

~There is also some fix in the 2nd commit for the healthcheck page, which cannot have 'onclick' event handlers as from CSP restrictions...~

Lift has a boolean attribute `extractInlineJavaScript` that, when the value is true, allows to include all inline Javascript into the lift page script (which already have a nonce). So we can actually leave `onclick` handlers now and Lift will do the rewrite !

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -85,6 +85,8 @@ import org.reflections.Reflections
 import org.springframework.security.core.context.SecurityContextHolder
 import scala.concurrent.duration.DAYS
 import scala.concurrent.duration.Duration
+import scala.xml.Elem
+import scala.xml.Node
 import scala.xml.NodeSeq
 
 /*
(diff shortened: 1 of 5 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):
No real idea about performances, but it looks like the Correct Way to do it in lift. 
Headers seems OK with no js error and no duplicate. 
LGTM :+1:

PR #5353, opened 2024-01-25
Fixes #24086: You can create a static group which depends on a dynamic one

fanf:
https://issues.rudder.io/issues/24086

We don't want to forbid only group changing from dynamic to static to have dynamic subgroup, but anything that is now static.


clarktsiory:
LGTM, but the previous message could be changed : 
> you can't change the nature of current group to static because it uses following dynamic groups as a subgroup criteria
It is ok when changing the field from "dynamic" to "static" but now we have the same exact error message when assigning an already "static" group to have a "dynamic" subgroup...

clarktsiory (approved):
Error message is now ok !

fanf:
OK, squash merging this PR

PR #5354, opened 2024-01-26
Fixes #24084: 404 not found in technique editor resource manager Ajax query

VinceMacBuche:
https://issues.rudder.io/issues/24084


fanf (approved):
Works as expected (but I found https://issues.rudder.io/issues/24089 in the way :) )

PR #5356, opened 2024-01-26
Fixes #24090: Score table migration

VinceMacBuche:
https://issues.rudder.io/issues/24090


VinceMacBuche:
Commit modified

fanf (approved):
works as expected. 

We still need to find a general pattern to not to write the same SQL both here and in the schema SQL file, because it's extremely fastidious and error prone.

PR #5357, opened 2024-01-26
Fixes #24085: We need a consistant presence of policyMode in compliance API outputs

clarktsiory:
https://issues.rudder.io/issues/24085

Basically just adding the _policyMode_ field at every level of the compliance tree.

For nodes and directives, _policyMode_ is already an attribute, but for a rule we need to compute it, so this is a bit less performant as it requires the set of node ids by group to be fetched.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -58,7 +58,6 @@ import com.normation.rudder.domain.reports.DirectiveStatusReport
 import com.normation.rudder.domain.reports.ValueStatusReport
 import com.normation.rudder.reports.GlobalComplianceMode
 import com.normation.rudder.repository.FullActiveTechnique
-import com.normation.rudder.repository.FullNodeGroupCategory
 import com.normation.rudder.repository.RoDirectiveRepository
 import com.normation.rudder.repository.RoNodeGroupRepository
 import com.normation.rudder.repository.RoRuleRepository
@@ -471,8 +470,9 @@ class ComplianceAPIService(
   }
 
   private[this] def getByDirectivesCompliance(
-      directives: Seq[Directive],
-      level:      Option[Int]
+      directives:    Seq[Directive],
+      allDirectives: Map[DirectiveId, (FullActiveTechnique, Directive)], // to compute policy mode for each rule
+      level:         Option[Int]
   ): IOResult[Seq[ByDirectiveCompliance]] = {
     val computedLevel = level.getOrElse(10)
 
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -741,11 +773,11 @@ class ComplianceAPIService(
...
-      allGroups    <- nodeGroupRepo.getAllNodeIds()
+      allGroups    <- nodeGroupRepo.getAllNodeIdsChunk()
       groupLib     <- nodeGroupRepo.getFullGroupLibrary()
       directiveLib <- directiveRepo.getFullDirectiveLibrary().map(_.allDirectives)
       allNodeInfos <- nodeInfoService.getAll()
-      nodeInfos    <- onlyNode match {
+      nodeTmpInfos <- onlyNode match {
                         case None     => allNodeInfos.succeed
                         case Some(id) =>
                           allNodeInfos
```
clarktsiory, line 783: Upmerge will be a little scary on this one because of the `NodeInfoService` to `NodeFactsRepo` migration, but the same implementation already exists in 8.1...

Apart from this block, there should be nothing that is non-complicated to change ^^

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -757,49 +793,68 @@ class ComplianceAPIService(
...
                         )
                         .toIO
     } yield {
 
-      // get nodeIds by rules
-      val nodeByRules = rules.map(rule => (rule, RoNodeGroupRepository.getNodeIds(allGroups, rule.targets, allNodeInfos)))
+      // A map to access the node fqdn and settings
+      val nodeInfos = nodeTmpInfos.view.mapValues(n => (n.name, n)).toMap
+
+      val nodeAndPolicyModeByRules = rules
```
fanf, line 805: I wonder if we shouldn't have a cache for that. 
The "node/rule policy mode" is something that is costly to compute, and vary only at specific point (new node, change on rule/directive, change on groups - that's all?). So it looks like a good candidate to keep these data. I think actually the changes are the same than when a generation is needed, and the same than what we would like to have in the expected reports. 

So, perhaps not something for that PR, but we need to keep in mind that when perf will need to get better

fanf (approved):
LGTM, tests are consistant and datastructures are homogeneous between rules and nodes prism. 
The performance impact is complicated to assess, it will need a specific task if needed.

fanf:
OK, squash merging this PR

PR #5358, opened 2024-01-26
Fixes #24091: Add manufacturer field in create node API

fanf:
https://issues.rudder.io/issues/24091

Trivial change, most of it is just adding test and API doc. 
We just add a new case for `machine` deserialisation and choose either it or the old `machineType`. The new `machine` is an object matching format from the GET which allows `manufacturer`.

### DIFF
--- a/webapp/sources/api-doc/components/schemas/timezone.yml
+++ b/webapp/sources/api-doc/components/schemas/timezone.yml
@@ -13,4 +13,4 @@ properties:
   offset:
     type: string
     description: Timezone offset compared to UTC, in +/-HHMM format
-    example: +0200
+    example: "+0200"
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
@@ -50,12 +50,9 @@ import com.normation.rudder.domain.properties.NodeProperty
 import com.normation.rudder.rest.data.Creation.CreationError
 import com.normation.rudder.rest.data.NodeTemplate.AcceptedNodeTemplate
 import com.normation.rudder.rest.data.NodeTemplate.PendingNodeTemplate
+import com.softwaremill.quicklens._
 import com.typesafe.config.ConfigValue
-import net.liftweb.json.JArray
-import net.liftweb.json.JField
-import net.liftweb.json.JObject
-import net.liftweb.json.JString
-import net.liftweb.json.JValue
+import net.liftweb.json._
 
 /**
  * Applicative log of interest for Rudder ops.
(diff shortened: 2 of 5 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/api-doc/components/schemas/timezone.yml
```diff
@@ -13,4 +13,4 @@ properties:
   offset:
     type: string
     description: Timezone offset compared to UTC, in +/-HHMM format
-    example: +0200
+    example: "+0200"
```
fanf, line 16: not related to the change, but currently the doc is incorrect, showing int `200` in place of string `+0200`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
```diff
@@ -85,13 +82,20 @@ object Rest {
...
+   */
   final case class NodeDetails(
       id:             String,
       hostname:       String,
       status:         String,
       os:             OS,
       policyServerId: Option[String],
-      machineType:    String,
+      machineType:    Option[String],
+      machine:        Option[MachineCreate],
```
fanf, line 98: it looks like we relax a bit things by making both optional, but in fact, we were already using `physical` as a default value.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
```diff
@@ -362,7 +378,10 @@ object Validation {
 
   def getMachine(id: NodeId, status: InventoryStatus, machineType: String): MachineInventory = {
     val machineId = MachineUuid(IdGenerator.md5Hash(id.value))
-    val machine   = machineType.toLowerCase
+    val machine   = machineType.toLowerCase match {
+      case "virtual" => "vm"
```
fanf, line 382: this is to keep compatibility with old format where `vm` used to mean `Virtual, unknown provider`

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_settings.yml
```diff
@@ -12,15 +12,13 @@ response:
...
+              "id":"node-dsc",
+              "allowed_networks":[]
             },
             {
-              "id":"node-dsc",
+              "id":"root",
               "allowed_networks":[
-
+                "192.168.2.0/32"
               ]
```
fanf, line 22: I'm not sure why that change is needed. I suspect it's not related to that commit but to the change with dsc/splaytime which, for some reason, changed the order of root/dsc node (it's just a flip between the two even if the diff doesn't make it clear)

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #5360, opened 2024-01-29
Fixes #24096: Add score on top of tabs

VinceMacBuche:
https://issues.rudder.io/issues/24096

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
@@ -50,6 +50,10 @@ import zio.json.DeriveJsonEncoder
 import zio.json.EncoderOps
 import zio.json.JsonEncoder
 
+object ComplianceScore {
+  val scoreId = "compliance"
+}
+
 object ComplianceScoreEventHandler extends ScoreEventHandler {
   implicit val compliancePercentEncoder: JsonEncoder[ComplianceSerializable]     = DeriveJsonEncoder.gen
   def handle(event: ScoreEvent):         PureResult[List[(NodeId, List[Score])]] = {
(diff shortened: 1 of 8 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/ShowNodeDetailsFromNode.scala
```diff
@@ -305,7 +310,11 @@ class ShowNodeDetailsFromNode(
     "#node_parameters -*" #> agentPolicyModeEditForm.cfagentPolicyModeConfiguration(Some(node.id)) &
     "#node_parameters -*" #> (if (node.rudderSettings.isPolicyServer) NodeSeq.Empty
                               else agentScheduleEditForm(node).cfagentScheduleConfiguration) &
-    "#node_tabs [id]" #> s"details_${id}").apply(serverDetailsTemplate)
+    "#node_tabs [id]" #> s"details_${id}" &
+    "href=#node_summary -*" #> <span class={s"badge-compliance-score ${globalScore.value.value} sm"}></span> &
+    "href=#node_reports -*" #> <span class={
+      s"badge-compliance-score ${globalScore.details.find(_.scoreId == "compliance").map(_.value).getOrElse(NoScore).value} sm"
```
fanf, line 316: Can you declare `compliance` as the ID of the compliance score, it's a special keyword for scores, we need to have a convention for them. 
Perhaps in the handler, it looks like the only place where we give an identity to a compliance score: 
```
object ComplianceScoreEventHandler extends ScoreEventHandler {
  implicit val compliancePercentEncoder: JsonEncoder[ComplianceSerializable]     = DeriveJsonEncoder.gen

  val id = "compliance"

  def handle(event: ScoreEvent):         PureResult[List[(NodeId, List[Score])]] = {
    event match {
      case ComplianceScoreEvent(n, percent) =>
```

Or have a `ComplianceScoreId` object somewhere ?

fanf (commented):
Appart for the normalization of ID of rudder score, LGTM.

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
```diff
@@ -59,8 +63,8 @@ object ComplianceScoreEventHandler extends ScoreEventHandler {
         (for {
           p <- ComplianceSerializable.fromPercent(percent).toJsonAST
         } yield {
-          val scoreId = "compliance"
-          val score   = if (percent.compliance >= 100) {
+          import ComplianceScore.scoreId
```
fanf, line 66: I think I never thought to do that, it's rather elegant :+1:

PR #5367, opened 2024-01-30
Fixes #24105: Machine is not correctly create with node create API

fanf:
https://issues.rudder.io/issues/24105

So the problem was only for node created in `Accepted` status. For them, we created the machine directly in `Accepted`, while the node part was created in `Pending`, then accepted. 
So at one point, we have a machine in `Accepted`, and the node is still in `Pending`. So we clean the machine because we chose at some point to enforce the fact that machine and node must have the same status (when we deeply linked machine to node with the machine id being a hash of the node id). And there is no machine with the correct ID in pending. 
Works as expected. We just need to also create the machine in pending.

Other changes are just better logs and log correction.


fanf:
OK, merging this PR

PR #5372, opened 2024-02-01
Fixes #24112: We must not generate policies for nodes without a well formed certificate

fanf:
https://issues.rudder.io/issues/24112

So, we want to not generate policies for nodes without a correct security token, for the value of correct == "is pasable as what it's supposed to be". 

We also want to skip these node from appearing into `nodesList.json`. 

=> check that in `system variable` part of node config handling:
- for policy servers, just avoid nodes with an empty `keyHashBase64Sha256` which does what need to be done to check parsability etc 
- for nodes, add a "check security token" check, which uses the same heuristic. The node will be avoided in that case. 

With the change, we get that kind of error message:
image


VinceMacBuche (approved):

fanf:
Merging by hand, since the problem is not related with that PR

PR #5374, opened 2024-02-01
Fixes #24129: For delete API for node, add a message for node already deleted

fanf:
https://issues.rudder.io/issues/24129

Adapt the response to delete so that user using the API can know they perhaps didn't do want they wanted. 
The method seems to use at exactly one place, so it's rather simple to change. 

image


amousset (approved):

fanf:
OK, merging this PR

PR #5378, opened 2024-02-05
Fixes #24146: Backport user table to Rudder 7.3

fanf:
https://issues.rudder.io/issues/24146

Backport of https://github.com/Normation/rudder/pull/5121 for branche 7.3.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -1,0 +1,834 @@
+package com.normation.rudder.users
+
+/*
+ *************************************************************************************
+ * Copyright 2023 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+import cats.data.NonEmptyList
+import com.normation.errors.IOResult
+import com.normation.rudder.db.Doobie
+import com.normation.rudder.domain.logger.ApplicationLoggerPure
+import com.normation.utils.DateFormaterService
+import com.softwaremill.quicklens._
+import doobie._
+import doobie.free.connection.{unit => connectionUnit}
+import doobie.implicits._
+import doobie.postgres.implicits._
+import org.joda.time.DateTime
+import zio._
+import zio.interop.catz._
+import zio.json.ast._
+
+/*
+ * Repository that deals with users and sessions
+ */
+
+trait UserRepository {
+
+  /*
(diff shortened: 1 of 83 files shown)


webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/users/UserRepositoryTest.scala
```diff
@@ -0,0 +1,275 @@
...
+  val AUTH_PLUGIN_NAME_REMOTE = "test-oidc"
+
+  lazy val repo = if (doJdbcTest && doobie != null) {
+    new JdbcUserRepository(doobie)
+  } else {
+    InMemoryUserRepository.make().runNow
+  }
+
+  val actor    = EventActor("test")
+  val dateInit = DateTime.parse("2023-09-01T01:01:01Z")
```
clarktsiory, line 68:
```suggestion
  val actor:    EventActor = EventActor("test")
  val dateInit: DateTime   = DateTime.parse("2023-09-01T01:01:01Z")
```

clarktsiory (commented):
Some types that should ideally be added right now (I think there are many more, I did only some `git diff` between the 8.1 branch and this one to find that types were deleted)

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableUsers.scala
```diff
@@ -0,0 +1,99 @@
...
+    , creationDate timestamp with time zone NOT NULL
+    , authMethod   text
+    , permissions  text[]
+    , endDate      timestamp with time zone
+    , endCause     text
+    );"""
+
+    transactIOResult(s"Error with 'Users' table creation")(xa => sql1.update.run.transact(xa)).unit *>
+    transactIOResult(s"Error with 'v' table creation")(xa => sql2.update.run.transact(xa)).unit
+  }
```
clarktsiory, line 85: Some database compatibility issues that may arise : 
* The users is version 7.3 will have the tables created, and when upgrading to 8.1 they will need to add a column `tenants` to the `UserSessions` table : we should create a migration for that  
* I imagine there will be no user rolling back from version 8.1 to 7.3, but for developement I have an error reading user sessions from the table when the webapp starts (see the fix I suggested in `UserRepository` file)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -0,0 +1,834 @@
...
+  override def closeAllOpenSession(endDate: DateTime, endCause: String): IOResult[Unit] = {
+    val sql = sql"""update usersessions set enddate = ${endDate}, endcause = ${endCause} where enddate is null returning userid"""
+    transactIOResult(s"Error when closing opened user session")(xa => sql.query[String].to[List].transact(xa)).flatMap(users =>
+      ApplicationLoggerPure.User.info(s"Close open sessions with reason '${endCause}' for users '${users.mkString("', '")}'")
+    )
+  }
+
+  override def getLastPreviousLogin(userId: String): IOResult[Option[UserSession]] = {
+    val sql =
+      sql"""select * from usersessions where userid = ${userId} and enddate is not null order by creationdate desc limit 1"""
```
clarktsiory, line 592: The "*" is not a good practice and now we know why : the `tenants` column should not be included in 7.3 but only since 8.1...

```suggestion
      sql"""select userid, sessionid, creationdate, authmethod, permissions, enddate, endcause from usersessions where userid = ${userId} and enddate is not null order by creationdate desc limit 1"""
```


One more reliable and maintainable way to do `select *` in multiple queries is to declare it as a static fragment value : 
```scala
// maps to `UserSession` case class
val all: Fragment = fr"userid, sessionid, creationdate, authmethod, permissions, enddate, endcause"
```
, which would be in 8.1 : 
```scala
// maps to `UserSession` case class
val all: Fragment = fr"userid, sessionid, creationdate, authmethod, permissions, tenants, enddate, endcause"
```

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableUsers.scala
fanf: yes, we will need an additionnal migration to add the tenants column

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -0,0 +1,834 @@
...
+ *   endDate      timestamp with time zone
+ *   endCause     text
+ */
+class JdbcUserRepository(doobie: Doobie) extends UserRepository {
+  import com.normation.rudder.db.Doobie.DateTimeMeta
+  import com.normation.rudder.db.json.implicits._
+  import com.normation.rudder.users.UserSerialization._
+  import doobie._
+
+  implicit val logger = Doobie.slf4jDoobieLogger
```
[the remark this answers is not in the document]
fanf: no, that one is not important: what we want to have is a nice upmerge experience. For that, we want to limit changes between branches, and so want want to have the same changes on all branches (for ex adding types). 
But here, since the code was removed in upper branch, adding the type would likely cause a merge conflict, while letting it unchanged is ok (the deletion was already merged).

fanf:
The failing tests have nothing to do with the code (they will be investigated appart). Merging by hand because it targets a branch unknown of our bot.

PR #5379, opened 2024-02-06
Fixes #24124: Links and buttons don't have the right colour since the bootstrap update

RaphaelGauthier:
https://issues.rudder.io/issues/24124

There's still a lot of optimisation to be done on this (like replacing all our colour codes with variables, making sure we override as few bootstrap rules as possible but instead using variable overrides, etc.) but to limit its scope I've just fixed and harmonised the appearance of Rudder's main links and buttons.

This first PR already brings a few changes:

- There is a new **`_rudder-variables.scss`** file
    - the underscore means in Sass that this is a "partial" file, i.e. a file that will only be imported into other scss files. It is not compiled in css
    - it will contain all our colour codes (text, buttons, compliance, policy mode, warning/error etc.). Only variables, no css rules!
    - It is completely independent of bootstrap
    - In future, the code in this file will be divided into several partial files, and all these files will be imported into this file, which will serve as the root, to make code maintenance easier.
    
- We no longer load the **`bootstrap.css`** file directly into common-layout.html, instead I've created a new **`rudder-bootstrap.scss`** file into which bootstrap is imported. This allows :
    - cleanly override bootstrap variables
    - have a file where you can find all the overridden bootstrap rules
    - in the future, not to import ALL bootstrap, but only the modules that interest us

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/style/rudder/_rudder-variables.scss
+++ b/webapp/sources/rudder/rudder-web/src/main/style/rudder/_rudder-variables.scss
@@ -1,0 +1,58 @@
+/*
+*************************************************************************************
+* Copyright 2024 Normation SAS
+*************************************************************************************
+*
+* This file is part of Rudder.
+*
+* Rudder is free software: you can redistribute it and/or modify
+* it under the terms of the GNU General Public License as published by
+* the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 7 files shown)


RaphaelGauthier:
Commit modified

RaphaelGauthier:
Commit modified

VinceMacBuche (approved):

PR #5382, opened 2024-02-07
Fixes #24152: Jstree links are broken

clarktsiory:
https://issues.rudder.io/issues/24152

This looks as a hack but it works as a fix : 
* we store the initial event handlers in the tree, rendered in Scala code by Lift, they always have an `id` prefixed with `lift-event-js-`,
* we create a jstree "plugin" which provides a function `redraw_node` defined by jstree, in which we re-bind the event handler

I think this is the only convenient solution... The plugin would simply delete all events for the node in the DOM if we add an event handler anywhere else

In the groups tree, there was no convenient selector for a node of the tree so I created one "groupTreeNode".

There is also a fix to add _hrefs_ to all nodes of the tree, so that they can be opened in a new tab (e.g. with a middle click)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/tree.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/tree.js
@@ -259,21 +259,55 @@ var buildGroupTree = function(id, appContext, initially_select, select_multiple_
    */
   var select_system_node_allowed = false;
 
+  // Save node event handlers, in order to customize node by reassigning the click event handler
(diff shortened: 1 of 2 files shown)


clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

RaphaelGauthier (approved):

clarktsiory:
OK, squash merging this PR

PR #5384, opened 2024-02-08
Fixes #24169: Synntax errors in schema script and init order changes

VinceMacBuche:
https://issues.rudder.io/issues/24169

error in sql script (missing ;)

bootstrap check was not working to create globalscore table because you can't check empty value on an enum. just null

Also change order of init so that bootstrap checks are run before initialisation of score Service (which has at class init time)


fanf (approved):
LGTM

PR #5386, opened 2024-02-08
Fixes #24172: Active buttons no longer have shadow

clarktsiory:
https://issues.rudder.io/issues/24172

image

* Added the shadow of active buttons in bootstrap 3 here with the new bootstrap 5 version.
* Some hardcoded style was already patched in Elm code previously, now it can be removed.

I don't know if the shadow color should go in the `rudder-variables` partial, it seems too specific to be reused anywhere else for now...


clarktsiory:
PR updated with a new commit

RaphaelGauthier (approved):

clarktsiory:
OK, squash merging this PR

PR #5387, opened 2024-02-08
Fixes #24177: Make spotless only check source of current project

fanf:
https://issues.rudder.io/issues/24177

The previous pattern was looking for all subdirectories with scala file. This new one only look at the `src` directory, thus limiting to current project scala sources.


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #5388, opened 2024-02-09
Fixes #24178: Initialize score on webapp start up

VinceMacBuche:
https://issues.rudder.io/issues/24178


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -162,10 +162,14 @@ class ScoreServiceImpl(globalScoreRepository: GlobalScoreRepository, scoreReposi
...
+    Ref.make(ComplianceScoreEventHandler :: List.empty[ScoreEventHandler]).runNow
 
-  def registerHandler(handler: ScoreEventHandler): UIO[Unit] = {
-    handlers.update(handler :: _)
+  def registerHandler(handler: ScoreEventHandler): IOResult[Unit] = {
+    handlers.update(handler :: _) *>
+    ZIO.ifZIO(readScore.getAll().map(_.exists(g => handler.initForScore(g._2))))(
+      handler.initEvents.flatMap(ZIO.foreach(_)(handleEvent(_))).unit,
+      ().succeed
+    )
```
fanf, line 172: I think handle event needs to be an `UIO[Unit]. Not catching an error in it leads to unroverable errors and a non working rudder - for ex, when the init event has a problem, it rudder starts but in a bad state .

fanf (approved):
Changed the init method to be always succeding so that an error during init does not break Rudder but is correctly logged.

fanf:
OK, merging this PR

PR #5390, opened 2024-02-09
Fixes #24183: Add an Alias type of Role to track role mapping

fanf:
https://issues.rudder.io/issues/24183

- add an `Role.Alias` type that is just an alias of an other Role. It use the other role `name` so that rudder internal resolution is transparent for it, but allows to have an additionnal `aliasName` (typically provided by an external source to avoid referencing Rudder internal name)
- add a facility to transform roles into meaningfull string usable in log/session log
- also add some facility to provide a `logout` redirect URL. This is a bit of work, since we don't use Spring security logout for historical reasons (it didn't work well with lift logout, session cleaning, etc)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -37,6 +37,7 @@
 
 package bootstrap.liftweb
 
+import com.normation.errors.IOResult
 import com.normation.eventlog.EventActor
 import com.normation.eventlog.EventLog
 import com.normation.eventlog.EventLogDetails
@@ -78,10 +79,13 @@ import net.liftweb.sitemap.Menu
 import net.liftweb.util.TimeHelpers._
 import org.joda.time.DateTime
 import org.reflections.Reflections
+import org.springframework.security.core.Authentication
 import org.springframework.security.core.context.SecurityContextHolder
 import scala.concurrent.duration.DAYS
 import scala.concurrent.duration.Duration
 import scala.xml.NodeSeq
+import zio.{System => _, _}
+import zio.syntax._
 
 /*
  * Utilities about rights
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -227,39 +231,71 @@ object FatalException {
  */
 object UserLogout {
 
-  def cleanUpSession(session: LiftSession, endCause: String) = {
-    SecurityContextHolder.getContext.getAuthentication match {
+  val logoutActions = Ref.make(Chunk[LogoutPostAction]()).runNow
```
fanf, line 234: a storage for action that plugins want to register post logout, like a redirect for SLO (single logout).
Action can return an URI, which is understood as the redirect. Only the first URI is used (because, well, can't do anything else).

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/UserInformation.scala
```diff
@@ -91,13 +91,16 @@ class UserInformation extends DispatchSnippet with DefaultExtendableSnippet[User
...
-            ApplicationLogger.debug("Logout called for a non existing session, nothing more done")
+        {
+          val redirect = (S.session match {
+            case Full(session) => // we have a session, try to know who is login out
+              UserLogout.cleanUpSession(session, "User asked for logout")
+            case e: EmptyBox => // no session ? Strange, but ok, nobody is login
+              ApplicationLogger.debug("Logout called for a non existing session, nothing more done")
+              None
+          })
+          JsCmds.RedirectTo(redirect.map(_.toString).getOrElse("/"))
```
fanf, line 102: if the logout action returned on URI, we use that for redirect in place of Rudder normal redirect to login page.

clarktsiory (approved):
LGTM ! It works great !

PR #5391, opened 2024-02-11
Fixes #24140: Have a source of known-tenants in rudder-core

fanf:
https://issues.rudder.io/issues/24140

That PR introduce a source of "tenants known in that Rudder server". Only tenants know will be allowed to be used for node access, so that we have a kind of prefiltering before the "user/node/tenant" check on "node tenants/available tenants" 

The filtering is done in a `TenantService` that knows about available tenants and wrap low level methods on node. 
These wrapping are then used  in `CoreNodeFactRepository`. 

Nothing in Rudder core manage that list of available tenants. It will be done in plugins. 

Also add tests to check that the available tenants are doing there job.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
@@ -58,6 +58,7 @@ import com.normation.rudder.domain.properties.NodeProperty
 import com.normation.rudder.domain.properties.PropertyProvider
 import com.normation.rudder.domain.servers.Srv
 import com.normation.rudder.reports._
+import com.normation.rudder.tenants.TenantId
 import com.normation.utils.ParseVersion
 import com.normation.utils.Version
 import com.normation.zio._
@@ -72,7 +73,6 @@ import net.liftweb.json.JsonAST
 import net.liftweb.json.JsonAST._
 import net.liftweb.json.JsonAST.JValue
 import org.joda.time.DateTime
-import scala.util.matching.Regex
 import zio.Chunk
 import zio.json._
 import zio.json.ast.Json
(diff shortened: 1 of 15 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
```diff
@@ -1446,28 +1446,6 @@ object QueryContext {
   implicit val systemQC: QueryContext = QueryContext(eventlog.RudderEventActor, NodeSecurityContext.All)
 }
 
-/*
- * A node can belong to one (or technically more, but we will limit that for now) `tenant`.
- * A `tenant` is a segregation limit defining an isolated zone.
- * Tenants should be \ascii\num_-
- */
-final case class Tenant(value: String) extends AnyVal {
```
fanf, line 1454: move into new package `com.normation.rudder.tenants`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactRepository.scala
```diff
@@ -82,14 +83,6 @@ import zio.syntax._
  *
  */
 trait NodeFactRepository {
-
-  /*
-   * Check if the node can be seen in the given query context. Return none if it can't.
-   */
-  def securityFilter[A <: MinimalNodeFactInterface](n: A)(implicit qc: QueryContext): Option[A] = {
-    if (qc.nodePerms.nsc.canSee(n)) Some(n) else None
-  }
```
fanf, line 91: this logic is move into `TenantService`

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5393, opened 2024-02-13
Branch 7.3 scala 3 migration part 2

mbaechler:
Same PR as https://github.com/mbaechler/rudder/pull/2 but targeting 7.3 branch


fanf:
Merging that PR by hand so that I can handle the upmerge to branche 8.0/8.1 correctly.

PR #5396, opened 2024-02-14
Fixes #24194: Add space in the login form between inputs

RaphaelGauthier:
https://issues.rudder.io/issues/24194

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/index.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/index.html
@@ -17,7 +17,7 @@
 </head>
 <body>
 <div class="flex-container">
-  <div class="col-lg-4 col-sm-6 col-xs-12">
+  <div class="col-xl-4 col-md-6 col-12">
 
     <div data-lift="Login.display">
 
@@ -32,7 +32,7 @@
 
         <div class="motd"></div>
 
-        <div class="form-group group-username">
+        <div class="form-group group-username mb-3">
           <label>
             Username
             <span class="warning-info"><i class="fa fa-warning"></i>This field is required</span>


webapp/sources/rudder/rudder-web/src/main/webapp/index.html
```diff
@@ -32,7 +32,7 @@
 
         <div class="motd"></div>
 
-        <div class="form-group group-username">
+        <div class="form-group group-username mb-3">
```
clarktsiory, line 35: :+1:

webapp/sources/rudder/rudder-web/src/main/webapp/index.html
```diff
@@ -17,7 +17,7 @@
 </head>
 <body>
 <div class="flex-container">
-  <div class="col-lg-4 col-sm-6 col-xs-12">
+  <div class="col-xl-4 col-md-6 col-12">
```
clarktsiory, line 20: The width on the smallest screen is 100% with `col-12`, which was not the case before. To make a smoother transition and to keep the same behavior : 
```suggestion
  <div class="col-xl-4 col-md-6 col-8">
```

RaphaelGauthier: According to Bootstrap 4 migration guide :

> - Added a new sm grid tier below 768px for more granular control. We now have xs, sm, md, lg, and xl. This also means every tier has been bumped up one level (so .col-md-6 in v3 is now .col-lg-6 in v4).
> - xs grid classes have been modified to not require the infix to more accurately represent that they start applying styles at min-width: 0 and not a set pixel value. Instead of .col-xs-6, it’s now .col-6. All other grid tiers require the infix (e.g., sm).

So I think we want to keep col-12 for the smallest screen size, as we want it to take 100% of the width (I just forgot to add extra padding for this case because new col-* classes don't provide any).

To smooth the transition, we can add a col-xs class, so that it looks like :
`<div class="col-xl-4 col-lg-6 col-sm-8 col-12">`

clarktsiory: My bad ! Indeed the behavior is the same as before so it can be left as is. I just didn't like seeing the `100%` without padding/margin and the rough transition... I will be fine with your suggestion : ` <div class="col-xl-4 col-lg-6 col-sm-8 col-12">` :+1:

RaphaelGauthier:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/webapp/index.html
clarktsiory: :fire:

clarktsiory (approved):

PR #5398, opened 2024-02-15
Fixes #24150: when we have conditions on blocks, the condition text is not correctly positionned when the block is opened

ElaadF:
https://issues.rudder.io/issues/24150

Please note that the offset label `Audit from command` that you can see in the gif, for the method after the nested block, is another bug, not introduced by this fix

condition_position


RaphaelGauthier (approved):

ElaadF:
OK, merging this PR

PR #5408, opened 2024-02-21
Fixes #24230: Authentication providers and role mapping settings should be exposed

clarktsiory:
https://issues.rudder.io/issues/24230

We need to know providers properties e.g. : Do they override or extend the user roles defined in the `users.xml` file ? Can they write password in the users file ?

These properties will be used in user-management plugin to handle some providers logic regarding roles and also for the logic of the UI : https://github.com/Normation/rudder-plugins/pull/668

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -625,12 +625,13 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
           fr")"
 
         def update(updated: Vector[UserInfo]): ConnectionIO[Int] = {
+          // never update the managedBy of an existing user which may have been provided by another origin
           val sql =
             """insert into users (id, creationdate, status, managedby, name, email, lastlogin, statushistory, otherinfo)
                   values (?,?,?,?,? , ?,?,?,?)
                on conflict (id) do update
-                set (creationdate, status, managedby, name, email, lastlogin, statushistory, otherinfo) =
-                  (EXCLUDED.creationdate, EXCLUDED.status, EXCLUDED.managedby, EXCLUDED.name, EXCLUDED.email, EXCLUDED.lastlogin, EXCLUDED.statushistory, EXCLUDED.otherinfo)"""
+                set (creationdate, status, name, email, lastlogin, statushistory, otherinfo) =
+                  (EXCLUDED.creationdate, EXCLUDED.status, EXCLUDED.name, EXCLUDED.email, EXCLUDED.lastlogin, EXCLUDED.statushistory, EXCLUDED.otherinfo)"""
 
           Update[UserInfo](sql).updateMany(updated)
         }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -801,7 +801,7 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
 
   // only disabled user can be set back to active
   override def setActive(userId: List[String], trace: EventTrace): IOResult[Unit] = {
-    changeStatus(userId, None, Nil, trace, UserStatus.Active, Some(fr"status = '${UserStatus.Disabled.value}'")).unit
+    changeStatus(userId, None, Nil, trace, UserStatus.Active, Some(fr"status = ${UserStatus.Disabled.value}")).unit
   }
 
   override def getAll(): IOResult[List[UserInfo]] = {
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -625,13 +625,13 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
           fr")"
 
         def update(updated: Vector[UserInfo]): ConnectionIO[Int] = {
-          // never update the managedBy of an existing user which may have been provided by another origin
+          // never update the user personal information and managedby of an existing user which may have been provided and modified by another origin
           val sql =
             """insert into users (id, creationdate, status, managedby, name, email, lastlogin, statushistory, otherinfo)
                   values (?,?,?,?,? , ?,?,?,?)
                on conflict (id) do update
-                set (creationdate, status, name, email, lastlogin, statushistory, otherinfo) =
-                  (EXCLUDED.creationdate, EXCLUDED.status, EXCLUDED.name, EXCLUDED.email, EXCLUDED.lastlogin, EXCLUDED.statushistory, EXCLUDED.otherinfo)"""
+                set (creationdate, status, lastlogin, statushistory) =
+                  (EXCLUDED.creationdate, EXCLUDED.status, EXCLUDED.lastlogin, EXCLUDED.statushistory)"""
 
           Update[UserInfo](sql).updateMany(updated)
         }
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -625,12 +625,13 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
           fr")"
 
         def update(updated: Vector[UserInfo]): ConnectionIO[Int] = {
+          // never update the managedBy of an existing user which may have been provided by another origin
           val sql =
             """insert into users (id, creationdate, status, managedby, name, email, lastlogin, statushistory, otherinfo)
                   values (?,?,?,?,? , ?,?,?,?)
                on conflict (id) do update
-                set (creationdate, status, managedby, name, email, lastlogin, statushistory, otherinfo) =
```
clarktsiory, line 632: We should not be allowed to replace existing users' `managedby` value because it is subject to very specific user-management logic e.g. a user can exist in the users static configuration file but also be managed by an external provider

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -625,12 +625,13 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
...
-                  (EXCLUDED.creationdate, EXCLUDED.status, EXCLUDED.managedby, EXCLUDED.name, EXCLUDED.email, EXCLUDED.lastlogin, EXCLUDED.statushistory, EXCLUDED.otherinfo)"""
+                set (creationdate, status, name, email, lastlogin, statushistory, otherinfo) =
+                  (EXCLUDED.creationdate, EXCLUDED.status, EXCLUDED.name, EXCLUDED.email, EXCLUDED.lastlogin, EXCLUDED.statushistory, EXCLUDED.otherinfo)"""
```
clarktsiory, line 634: The update of the `managedby` field can break available information about the provider of existing users.

We should not be allowed to replace existing users' `managedby` value because it is subject to very specific user-management logic e.g. a user can exist in the users static configuration file but also be managed by an external provider

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -418,6 +447,55 @@ class RudderXmlUserDetailsContextMapper(authConfigProvider: UserDetailListProvid
...
+/**
+  * A description of some properties of an authentication backend
+  * @param hasAdditionalRoles if the backend can provide additional roles than the ones of the default backend 
+  * @param overridesRoles if the backend is configured to override the roles of the default backend and cannot change them
+  * @param hasModifiablePassword if the backend is configured to allow password modification
+  */
+final case class AuthBackendProviderProperties private (
+    hasAdditionalRoles:    Boolean,
+    overridesRoles:        Boolean,
+    hasModifiablePassword: Boolean
```
fanf, line 459: I think the logic here is not clear (or I don't understand it). 
LDAP does not have updatable password by rudder: they are managed by the LDAP directory, out of Rudder domain of responsibility. 
But if we have a configuration with `provider = ldap, file`, ie first trying LDAP, then trying file, then we want to let the user password for the *file* case be updatable. 
But `file` is not defined as an `AuthBackendProviderProperties`. Why so ?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -418,6 +447,55 @@ class RudderXmlUserDetailsContextMapper(authConfigProvider: UserDetailListProvid
...
+      hasAdditionalRoles:    Boolean,
+      overridesRoles:        Boolean,
+      hasModifiablePassword: Boolean
+  ): AuthBackendProviderProperties = {
+    if (overridesRoles && !hasAdditionalRoles) {
+      // This is not consistent so we force overridesRoles to false
+      ApplicationLogger.error(
+        s"Backend provider properties for '${name}' are not consistent: backend does not enables providing roles but roles overrides is set to true. " +
+        s"Overriding roles will not be enabled."
+      )
```
fanf, line 476: If you need to do that kind of check, then it means that you are not enforcing "Make invalid states unrepresentable", which we really want to enforce as much as possible. 

You likely want an ADT like:
```
sealed trait RoleExtension
object RoleExtension {
  case object None extends RoleExtension
  case object NoOverride extends RoleExtension
  case object WithOverride extends RoleExtention
}
```

fanf: Plus with such an AST, it's easy to build an order relationship (for ex, add a priority index `sealed trait RoleExtension { def index: Int }` and compare on it for the priority rules, or take the max, etc)

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
clarktsiory: Maybe the `hasModifiablePassword` should be renamed `canProviderBeMigratedToFile` because that is what happens with LDAP...

I didn't really want to introduce this boolean, but its sole purpose is to allow the UI to "force" a password for an external provider (which LDAP is), as in this Elm code : https://github.com/clarktsiory/rudder-plugins/blob/bug_24219/display_user_detail_using_information_from_database/user-management/src/main/elm/sources/View.elm#L144.

Indeed with `provider = ldap, file` we want to let the user password for the file case be updatable, but we don't want it to be updatable with `provider = oidc, file` (unless I was misunderstanding the spec), so a boolean attribute for making distinction between both would be actually needed

clarktsiory: `file` is always handled very specifically even in the UI, so I didn't want to abstract all the properties that make the distinction between file and other providers, and I guess there are plenty...

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
clarktsiory: :+1: thanks for the suggestion, I would only need to move the existing definition from the user-management plugin to here in Rudder : https://github.com/Normation/rudder-plugins/pull/668/commits/b8cd1a9e225b694e0b57307fa494f904c5ab43cd#diff-50791aeada8e5190976dae763b438a4d990a8a3bbcf23b7e063a7c9ac11274c0R94-R101

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
fanf: you can completely have `provider = oidc, file` with updatable passwords *safe* oidc provisioning is set to true (for at least one provider, because I don't think can know from which oidc provider users come... Which is perhaps a problem in fact - for that case, and for the case where a new OIDC provider is added or an old one removed...). 
Ok, appart that new interesting subject: an oidc provider without auto-provisioning is almost exactly like LDAP conceptually: user are declared in `rudder-users.xml`, but the authentication is done by some other entity. 

And LDAP could, one day, with lots of incentives, get an auto-provisioning mode.

fanf:
Even with the feedbacks, I will merge that so that it's easier to iterate (it becomes complicated with severa PR here and on plugin). Opening an other ticket to track needed enhancements.

PR #5413, opened 2024-02-28
Fixes #24282: Display user roles on hover of user profile

clarktsiory:
https://issues.rudder.io/issues/24282

Hopefully the roles are custom ones and have a pretty display name, otherwise if there are custom roles there will be `anon[node_write], anon[...], ...` in the list but they would still be needed because we can't lie to the user...


fanf (approved):
LGTM

fanf:
Merging by hand for feature branch

PR #5414, opened 2024-02-28
Fixes #24284: Log on user api authorizations should be more concise

clarktsiory:
https://issues.rudder.io/issues/24284

This will have effect on debug logs that use string interpolations, the change is introduced to fix this one : https://github.com/Normation/rudder-plugins/pull/655/files#diff-8fe302fecad34e77c8d7fa30f3efdf0110c77048e1f14006e4855dc1c60e73d9R658-R659.

Now we should be able to read it more fluently :
```
[2024-02-28 16:18:00+0100] DEBUG auth-backends - Principal 'test-user' final roles: [inventory], and API authz: /settings/global_policy_mode:[GET];/settings/global_policy_mode_overridable:[GET];/user/api/token:[GET,PUT,POST,DELETE];/nodes/details/property/*:[POST];/nodes/details/software/*:[POST];/nodes/*/displayInheritedProperties:[GET];/nodes/*/inheritedProperties:[GET];/nodes/pending/*:[GET];/nodes/details:[POST];/nodes/pending:[GET];/nodes/status:[GET];/nodes/*:[GET];/nodes:[GET];/inventories/info:[GET]
```

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
@@ -94,6 +94,14 @@ final case object HttpAction {
 
   def values: Set[HttpAction] = ca.mrvisser.sealerate.values[HttpAction]
 
+  // sort the action by the importance of the action, please DON'T FORGET TO ADD NEW VALUES HERE
+  implicit val orderingHttpAction: Ordering[HttpAction] = new Ordering[HttpAction] {
+    val order = List(HEAD, GET, PUT, POST, DELETE)
+    override def compare(x: HttpAction, y: HttpAction): Int = {
+      order.indexOf(x) - order.indexOf(y)
+    }
+  }
+
   def parse(action: String): Either[String, HttpAction] = {
     val lower = action.toLowerCase()
     values.find(_.name == lower).toRight(s"Action '${action}' is not recognized as a supported HTTP action")
@@ -133,7 +141,8 @@ object AclPathSegment {
  */
 sealed trait AclPath extends Any {
   def value: String = parts.toList.map(_.value).mkString("/")
-  def parts: NonEmptyList[AclPathSegment]
+  def parts:   NonEmptyList[AclPathSegment]
+  def display: String
 }
 
 object AclPath {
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -246,7 +257,10 @@ object ApiAuthorization       {
   case object None                               extends ApiAuthorization { override val kind = ApiAuthorizationKind.None }
   case object RW                                 extends ApiAuthorization { override val kind = ApiAuthorizationKind.RW   }
   case object RO                                 extends ApiAuthorization { override val kind = ApiAuthorizationKind.RO   }
-  final case class ACL(acl: List[ApiAclElement]) extends ApiAuthorization { override def kind = ApiAuthorizationKind.ACL  }
+  final case class ACL(acl: List[ApiAclElement]) extends ApiAuthorization {
+    override def kind = ApiAuthorizationKind.ACL
+    override def toString: String = acl.map(_.display).mkString(";")
```
fanf, line 262: can you add an alias "debugString" and use it where appropriate ? The jvm platform force us to have `.toString`, but it makes for unbeliable hard to correct bugs (related to string concat when it's not what you want at all) and should be avoided. So as a good practice, we don't use "toString" but `.debugString` for which the intent is clear (and we wait for a linter rule that can scream at use when toString is used). 

Plus, it makes it clear that "debugString" is not something to try to parse again, what `toString` never does, make assumption quite unclear.

clarktsiory: Yes that is the cleaned solution, I'll have to also change the culprit log on auth-backends and call `.debugString` on it then :see_no_evil:

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
clarktsiory: renamed the `toString` and added it also in https://github.com/Normation/rudder-plugins/pull/675 :white_check_mark:

fanf (approved):
Will merge by hand since feature branch

PR #5428, opened 2024-03-01
Fixes #24304: Deleted users should change status on file reload 

clarktsiory:
https://issues.rudder.io/issues/24304

* The change of not updating user origin should not have been made [in this previous PR](https://github.com/Normation/rudder/pull/5408/files#diff-4a2aaf2f76c5556341804a45895b09d281d02773e3dc0ff723507cced5178652R633) : it breaks tests and users can no longer be set as managed by OIDC when logging-in the first time and already being a file user. 

* Users that are from the same origin should not be taken into account when changing user status (`setExistingUsers`)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -234,20 +234,22 @@ object UserRepository {
         )
     }.toMap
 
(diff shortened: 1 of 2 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased and squashed

fanf (approved):
LGTM, good work!

PR #5435, opened 2024-03-05
Fixes #24318: Existing deleted user managed by file cannot be reactivated 

clarktsiory:
https://issues.rudder.io/issues/24318


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -232,6 +232,16 @@ object UserRepository {
...
+      // deleted users that are in active users list should be active again if they are from the same origin
+      case k if zombies.get(k).map(u => u.status == UserStatus.Deleted && u.managedBy == origin).getOrElse(false) =>
+        (
+          k,
+          zombies(k)
+            .modify(_.status)
+            .setTo(UserStatus.Active)
+            .modify(_.statusHistory)
+            .using(StatusHistory(UserStatus.Active, trace) :: _)
+        )
```
clarktsiory, line 244: I don't know if there is a more elegant way, instead of having `zombies.get(k)` then `zombies(k)` later...

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
fanf: Put the condition below and pattern match on get(K)  in place of having a guard. Iirc it's similar in perf and alloc

clarktsiory: But the `case k if ... =>` is passed to a `collect` and should not be a function but a partial function.
There is always the option to change to a function and `flatMap` on `Option` but I don't know if that way it is worth the change...

fanf:
I will need to be rebased after retarget to branche 7.3

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
clarktsiory: I just realized that maybe you were exactly referring to that :sweat_smile: : https://github.com/Normation/rudder/pull/5435/commits/bf74ff154569ef9e5418e27e3e8963b270c3fc63

fanf (approved):
Works as expected !

clarktsiory:
OK, squash merging this PR

PR #5449, opened 2024-03-07
Fixes #24363: Display compliance for system groups

clarktsiory:
https://issues.rudder.io/issues/24363

We need to support either a `groupId` (e.g. an uuid) or a target (e.g. `special:...` or even `group:...`). 

Common fields that we need for the compliance result are `id, name, serverList`, the code is refactored accordingly.

**Screenshots** : 

Screenshot from 2024-03-08 14-37-36

Screenshot from 2024-03-08 14-34-00

This one needs some change in the UI (in another PR) :
image

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
@@ -38,7 +38,6 @@
 package com.normation.rudder.rest.data
 
 import com.normation.inventory.domain.NodeId
-import com.normation.rudder.domain.nodes.NodeGroupId
 import com.normation.rudder.domain.policies.DirectiveId
 import com.normation.rudder.domain.policies.PolicyMode
 import com.normation.rudder.domain.policies.RuleId
@@ -127,7 +126,7 @@ final case class ByDirectiveByNodeRuleCompliance(
 )
 
 final case class ByNodeGroupCompliance(
-    id:         NodeGroupId,
+    id:         String,
     name:       String,
     compliance: ComplianceLevel,
     mode:       ComplianceModeName,
@@ -838,7 +837,7 @@ object JsonCompliance {
   implicit class JsonByNodeGroupCompliance(val nodeGroup: ByNodeGroupCompliance) extends AnyVal {
 
     def toJson(level: Int, precision: CompliancePrecision): JObject = {
-      (("id"                 -> nodeGroup.id.serialize)
+      (("id"                 -> nodeGroup.id)
       ~ ("name"              -> nodeGroup.name)
       ~ ("compliance"        -> nodeGroup.compliance.complianceWithoutPending(precision))
       ~ ("mode"              -> nodeGroup.mode.name)
--- a/webapp/sources/api-doc/components/parameters/target-or-node-group-id.yml
+++ b/webapp/sources/api-doc/components/parameters/target-or-node-group-id.yml
@@ -1,0 +1,12 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2024 Normation SAS
+name: "targetOrNodeGroupId"
+in: path
+required: true
+description: Id of the node group or rule target
+schema:
+  oneOf:
+    - type: string
+      format: uuid or string
+      example: 47e3f2c0-0b1a-4b1a-9b0a-9e9e9e9e9e9e
+    - $ref: ../schemas/rule-target.yml
--- a/webapp/sources/api-doc/openapi.src.yml
+++ b/webapp/sources/api-doc/openapi.src.yml
@@ -162,9 +162,9 @@ paths:
     $ref: paths/compliance/nodes.yml
   "/compliance/nodes/{nodeId}":
     $ref: paths/compliance/node.yml
-  "/compliance/groups/{nodeGroupId}":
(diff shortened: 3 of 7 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -838,7 +837,7 @@ object JsonCompliance {
   implicit class JsonByNodeGroupCompliance(val nodeGroup: ByNodeGroupCompliance) extends AnyVal {
 
     def toJson(level: Int, precision: CompliancePrecision): JObject = {
-      (("id"                 -> nodeGroup.id.serialize)
```
fanf, line 841: why not using the serialisation of node group as a `SimpleTarget` ? That datastructure is here just for that (ie abstracting over the different kind of targets: groups, system ones, etc). 
You will get a string lile: `group:90994f11-1638-4945-add0-70dc543e80d0` or `special:all` or `policyServer:root` etc.

And then just always parse back a `SimpleTarget`, and fail - ie `parseSimpleTargetOrNodeGroupId` does not seems necessary and inconsistent with the general way of representing rule targets.

fanf:
There is a CSS problem when there is a lot of nodes: 
image

Please open an other ticket to correct it indepently from that one (ie it doesn't already exist).

fanf (commented):
Appart for the serialization subject, it seems to work as expected. Once corrected, you will be able to just add the "ready for merge" tag. Plus need to open an other ticket for the css problem when there is a lot of nodes.

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
clarktsiory: It is indeed clearer to directly use it within the API.
As a consequence, the Elm app will need to reason over both target and group ids, because it parses and uses the raw group uuid as returned in the group tree. The change is here : https://github.com/Normation/rudder/pull/5449/commits/5c521e48d27e5c73a78ae82821fd3da335ad1d0d

And I am not sure if `parseSimpleTargetOrNodeGroupId` should actually be removed, we have an API endpoint that looks like `api/compliance/groups/9362944f-781d-4d9c-b8af-bfcdee8e63bb` now, and we should assume this is the basic "Public API" interface, but also allow for the `api/compliance/groups/group:9362944f-781d-4d9c-b8af-bfcdee8e63bb`. 
I don't think we want to return "not found" with : `api/compliance/groups/9362944f-781d-4d9c-b8af-bfcdee8e63bb`.

fanf: Ah, yes. It's a public API used internally. 
OK, if it's not only a private API, `api/compliance/groups/group:9362944f-781d-4d9c-b8af-bfcdee8e63bb` doesn't make much sense for an user, esp since we don't have that anywhere in the public API (plus the fact it's "group" and not "target" in the url). 

Could it be `api/compliance/targets/group:...` - it would at least bring some consistency. Perhaps we need to ask other.

clarktsiory: Indeed that's exactly what the parse function does and always ends up with a target  : 
```scala
private[this] def parseSimpleTargetOrNodeGroupId(str: String): PureResult[SimpleTarget] = {
    // attempt to parse a "target" first because format is more specific
    RuleTarget.unserOne(str) match {
      case None        => NodeGroupId.parse(str).map(GroupTarget(_)).left.map(Inconsistency(_))
      case Some(value) => Right(value)
    }
  }
```

VinceMacBuche: I think you should answer with the id and not the target in json response

fanf: OK, so this kind of subject gives more whatever to the idea that elm application should only use internal, RPC like APIs, and that public API have a different set of goals and contracts. 

In that example, the correct, simple RPC like thing that the elm app need is in contradiction with what user expects : elm does care about writting  `.../group:9362944f-781d-4d9c-b8af-bfcdee8e63b`, but it's very weird for an human. 

Since that API is a public one, yes we need to support the scheme and convention we use elsewhere, so: 
- `.../groups/9362944f-781d-4d9c-b8af-bfcdee8e63b` needs to be supported for human convenience in the scala API,  
- `.../groups/special:all` needs to be supported because the model is like that. 

And so, even if the elm app always use the correct, full serialisation of target, the backend needs to have a parser like the one you had initially. 

Sorry to have missed the fact that the API was public and set you on a bad path.

clarktsiory: Okay, it is not an issue, I didn't emphasis on the API being public either...

So I am reverting the change, and for the details, Elm also doesn't always make requests with `group:...` but directly with the UUID because it receives the UUID from the "group tree"

VinceMacBuche (approved):

PR #5456, opened 2024-03-08
Fixes #24366: Group compliance stays empty

clarktsiory:
https://issues.rudder.io/issues/24366

The existing test did not catch the bug because it the mock was not doing what happens when getting reports : filter by node and rule ids. I added the filters, and also corrected a test which was also wrong (the expected JSON for targeted compliance on nodes should have `nodes: []` in the test case, because the rules are skipped hence there are no node reports)


clarktsiory:
PR updated with a new commit

fanf:
Can you add a unit test for that please ?

clarktsiory:
PR updated with a new commit

fanf (approved):

PR #5460, opened 2024-03-08
Fixes #24383: Show logs button in node compliance tab does nothing

RaphaelGauthier:
https://issues.rudder.io/issues/24383


amousset (approved):

clarktsiory (approved):
The issue of having nothing "on click" is solved, but the button is still not doing what is expected : it is a problem of event handlers which are not bound to the button because the DOM tree of the compliance tab does not exist yet. 

This will need to be fixed in another PR along with other tabs.

PR #5463, opened 2024-03-11
Fixes #24405: Node tabs have onclick event handlers assigned on unrendered elements

clarktsiory:
https://issues.rudder.io/issues/24405
We display the `logs` table in two different tabs `Compliance` and `System status`, but having the same `id` for different elements across tables makes it impossible to dispatch event handlers correctly.
Also, the event handlers are not set on the `Show logs` buttons because the tab is has not yet been rendered.  

We just append the tab id to every HTML element, and set the buttons HTML with the correct event handler when the tab is rendered.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
@@ -660,10 +660,10 @@ function callRemoteRun(nodeId, refreshCompliance) {
   });
 }
 
-function showHideRunLogs(scrollTarget, init, refresh) {
-  $("#AllLogButton").toggle()
-  $("#logRun").toggle()
-  if ( ! $.fn.DataTable.isDataTable( '#complianceLogsGrid' ) ) {
+function showHideRunLogs(scrollTarget, tabId, init, refresh) {
+  $("#allLogButton-" + tabId).toggle()
+  $("#logRun-" + tabId).toggle()
+  if ( ! $.fn.DataTable.isDataTable( '#complianceLogsGrid-' + tabId ) && init !== undefined) {
     init()
   }
   if (refresh !== undefined) {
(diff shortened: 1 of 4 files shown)


clarktsiory:
PR updated with a new commit

fanf:
I don't know if it's that PR or if it was like that before, but if the (server) local API is not available, that I get a JS console error in place of a correct error:

``̀` 
XML Parsing Error: syntax error
Location: http://localhost:8082/rudder-web/secure/api/nodes/0c6ea44e-db4a-474a-bfac-0546bfd20371/applyPolicy
Line Number 1, Column 1:
``` 

The API seems to return an error correctly: 
```
Error occurred when contacting internal remote-run API to apply classes on Node '0c6ea44e-db4a-474a-bfac-0546bfd20371': Can not connect to local remote run API (POST:https://localhost/rudder/relay-api/remote-run/nodes/0c6ea44e-db4a-474a-bfac-0546bfd20371) 
``` 
(and then a lot of \u0 which is maybe a problem)

fanf:
It works as expected for the "show log" in compliance, but it doesn't seems to do anything for the system status page. Can you check it's OK ?

clarktsiory:
It works fine in both tabs on my side, did you have any specific error log in the console for the "System status" tab ?

clarktsiory:
Also, running the previous version also return the ` XML Parsing Error: syntax error` on my side, but it still displays the content correctly. I found out this is a browser error : https://stackoverflow.com/a/51000139 (I guess none of us is using Chome ^^)

fanf:
OK, finally understood: I had a previous version of the PR :) 
So it's working. But now, each time I click on a tab, I have a very viewable loading effect (and it's normal, there's 4 AJAX queries for each tab click o_O). It looks like that PR just "revealed" the underlying inefficient pattern, but now it is very, very annoying. 
I'm opening an other ticket for that.

fanf:
OK, squash merging this PR

PR #5469, opened 2024-03-13
Fixes #24432: Broken pipe when piping rudder-package output

amousset:
https://issues.rudder.io/issues/24432

There were two big options:
* Catch panics and prevent failure with broken pipe
* Replace `println!` and others by a fallible alternative and handle errors everywhere

As there can be `println!` everywhere and properlt handling these errors early would be expensive, let's go for the ugly solution.

### DIFF
--- a/policies/rudder-cli/src/lib.rs
+++ b/policies/rudder-cli/src/lib.rs
@@ -2,3 +2,30 @@
 // SPDX-FileCopyrightText: 2022 Normation SAS
 
 pub mod logs;
+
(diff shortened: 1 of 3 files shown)


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

Fdall (approved):

amousset:
OK, squash merging this PR

PR #5472, opened 2024-03-14
Fixes #24456: Missing primary key statement for user sessions table creation when migrating

fanf:
https://issues.rudder.io/issues/24456

What is in the migration and in SQL is not synched. We need to find a way to make it synched, else we will have things like that all the time. 

So:
- correct the initial migration of the table
- add a bit of script to add primary relation if it doesn't exists


clarktsiory (approved):
LGTM !

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #5477, opened 2024-03-15
Fixes #24416: Elm properties app is loaded multiple times

clarktsiory:
https://issues.rudder.io/issues/24416

The fix consist of : 
* loading the script when the group page loads, or else it will be loaded multiple times
* mounting and initializing an Elm app for node properties when the "Properties" tab content is shown
So we just move everything related to the Elm app "earlier" in the HTML structure, from `NodeGroupForm` details to the main `Group` page.

Caveats of this is that a new Elm app will be initialized each time the "Properties" tab is shown, but changing the Elm model to re-render with every group change action is heavier (we would likely need to make Elm apps global variables, to call ports on every action from the Scala templates). We could do that in another PR.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
@@ -317,39 +317,7 @@ class NodeGroupForm(
       </div>)
 
     def tabProperties = ChooseTemplate(List("templates-hidden", "components", "ComponentNodeProperties"), "nodeproperties-tab")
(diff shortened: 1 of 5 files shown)


RaphaelGauthier:
The error no longer appears when I click on different groups, but if I click several times on the Properties tab I get a new error: 

On firefox : 
errorjsfirefox

On chrome :
errorjschrome

clarktsiory:
PR updated with a new commit

clarktsiory:
@RaphaelGauthier The error is cryptic but obviously the tab is already mounted by the Elm app so the selector `nodeproperties-app` is empty :grin:

RaphaelGauthier (approved):

PR #5486, opened 2024-03-18
Fixes #24520: Async tests are failing for node certificates

clarktsiory:
https://issues.rudder.io/issues/24520

This is an async test so we use `eventually` on the expression to match (content of the output stream byte array)


fanf (approved):
LGTM, nice catch

fanf:
Oups, need to be retarget to 7.3

fanf (approved):

PR #5489, opened 2024-03-18
Fixes #24475:  API account UI needs to provide a configuration for tenants

fanf:
https://issues.rudder.io/issues/24475

:warning: 
There is `a lot` of changes that are due to the use of `elm-format` which is extremely happy to give new lines everywhere. I think the result is a bit better, but hell, do they love spaces. You should really look at diff with `?w=1`.

That PR adds the infrastructure to be able to configure tenants for API account. 
The logic is similar to the `api-authorization` plugin: 
- on rudder core, we have check in place to see if the plugin is enable or not, 
- if the plugin is disable, we only diplay the current tenants for the account, in a non updatable way, 
- if the plugin is enable, the the user can change tenants. 

We have three kind of tenants: 
- `*` means "all tenants", and is the defaults for compat reason
- `-` meains "no tenants", and the poor accout won't be able to do much with nodes, 
- last option is a list of tenants. On that case, we display a list of selectable tenants with the details (description, doc, etc). 

There is the corresponding changes in backend for API parsing on post/put, diff event update, etc. Nothing fancy. 
In LDAP, we just have one attribute in API account, `apiTenants`, which is just the string for the chosen tenants (ie `*`, `-`, or `tenant1,tenant2,etc` - like for users in `rudder-users.xml`).

There is one thing that is really not cool: the ACL plugins is appending its content at the end of the modal, and so do that new plugin, because we need and existing anchor when the page is init to init the corresponding elm app. And the place where it would be logic to put the list of tenant (just below the select for tenant mode) does not always exist (and certainly not when the page is loaded). 

This is the UI when plugin is disabled: 
image

And when it's enabled: 

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/ApiAccountDiff.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/ApiAccountDiff.scala
@@ -59,5 +59,6 @@ final case class ModifyApiAccountDiff(
     modExpirationDate:      Option[SimpleDiff[Option[DateTime]]] = None,
     modAPIKind:             Option[SimpleDiff[String]] = None,
     modAccountKind:         Option[SimpleDiff[String]] = None,
-    modAccountAcl:          Option[SimpleDiff[List[ApiAclElement]]] = None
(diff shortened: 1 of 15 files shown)


fanf:
Bad branch  / ticket, closing: see https://github.com/Normation/rudder/pull/5490 for the correct one

PR #5508, opened 2024-03-21
Fixes #24361: Status point next the node hostname doesn't provide much info

RaphaelGauthier:
https://issues.rudder.io/issues/24361


webapp/sources/rudder/rudder-web/src/main/style/rudder/rudder-node.scss
```diff
@@ -103,33 +106,50 @@ $node-state-size: 14px;
...
+  }
+  &.ignored{
     background-color: #a4aebf;
+    &:after{
+      content : "IGNORED";
+    }
   }
-  &.empty-policies:after{
+  &.empty-policies{
     background-color: #7591c599;
```
ElaadF, line 127: Isn't a typo on the color here ? I think it's `#7591c5` ?

VinceMacBuche:
I think that's too big and it really catches the eye and you see to much. I don't see the node name anymore

VinceMacBuche:
I would : 

 * Keep a circle but the color needs to change with some status. 
 * When not enabled (and maybe others), the status needs to be displayed in a pretty label not too much invasive (no caps) and probably on the right (after RAM) 
 * probably a tooltip on the circle

webapp/sources/rudder/rudder-web/src/main/style/rudder/rudder-node.scss
RaphaelGauthier: It's not a typo, Hexa he last two digits represent the alpha channel. So in this case, 99 corresponds to 60% opacity!

ElaadF:
- It is almost a standard to have an indication of the status next to the name, and doesn't make sens to me to move it else where, since this is important information when it is not `enabled`. Two locations for the same information is not the best choice IMO
- If the meaning of the status circle color is not obvious (the circle green is ok) we should use a label, like the previous screen with the name in it, tooltips are nightmare to maintain and it is not the best way to provide information, I think other status than `enabled` are not very common, making them usefull when they are written and visible, and keep the green circle when it is the most frequent state to not attract too much attention because, I assume, it will be the most frequent scenario

RaphaelGauthier:
PR rebased

RaphaelGauthier:
PR updated with a new commit

RaphaelGauthier:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
At least it covers the needs to display the status so i merge it

RaphaelGauthier:
OK, squash merging this PR

fanf:
OK, merging this PR

PR #5509, opened 2024-03-21
Fixes #24562: When moving an inventory to "received", check if other file for same node uuid exists and delete it

fanf:
https://issues.rudder.io/issues/24562

Normalize the file name before moving to received folder. 
We never fail on the normalization, it seems less risky to have several time similar files in received than an unmatch error for an unforseen case.


fanf:
PR updated with a new commit

amousset (approved):

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

PR #5520, opened 2024-03-22
Fixes #24583: We should not be able to save a technique with empty enum parameters 

ElaadF:
https://issues.rudder.io/issues/24583
image
image
image
image
image
image

Edit from https://github.com/Normation/rudder/pull/5520#issuecomment-2017581636

required_enum_trigger

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
@@ -88,6 +88,20 @@ techniqueParameter model technique param =
               TechniqueParameterModified param.id {param | constraints = newConstraint }
 
     constraint = param.constraints
+    (invalidEnumTypeClass, invalidEnumTypeElem) =
+        if ("Enum" == selectTypeValue) then
+          case param.constraints.select of
+            Nothing ->
+              ("error",  ul [class "row"] [ li [ class "text-danger col-md-8" ] [ text "Enum type should contain at least one element" ] ] )
+            Just opts ->
+              if(List.isEmpty opts) then
+                ("error",  ul [class "row"] [ li [ class "text-danger col-md-8" ] [ text "Enum type should contain at least one element" ] ] )
+              else
+                ("", text "")
+        else
+          ("", text "")
+
+
   in
     li [] [
       span [ class "border" ] []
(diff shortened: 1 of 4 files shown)


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
PR rebased

ElaadF:
Commit modified

VinceMacBuche:
An enum value may be empty, if the value is not "required" 

If required is checked, then you have an error

VinceMacBuche:
but <3

ElaadF:
cc @RaphaelGauthier I was not able to put the error message below the input of Enum value and the emtpy Enum, it seems to affect only the last input

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
```diff
@@ -141,46 +155,63 @@ techniqueParameter model technique param =
...
                                     let
-                                      newConstraint = { constraint | select = Just (List.Extra.updateAt index (\val -> {val | value = s}) l  ) }
+                                      newConstraint = { constraint | select = Just (List.Extra.updateAt index (\val -> {val | name = if String.isEmpty s then Nothing else Just s}) l  ) }
                                     in
                                       TechniqueParameterModified param.id {param | constraints = newConstraint }), required True] []
+                        , invalidEnumName
+                        ]
+                      , div [ class " form-group" ] [
+                          label [ ] [ text "Value"]
+                        , div [class "input-group" ] [
```
clarktsiory, line 196: the "input-group" class is not used elsewhere in the form, it should be removed because it doesn't look the same as the "Display name"

ElaadF:
Commit modified

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
```diff
@@ -141,46 +155,63 @@ techniqueParameter model technique param =
             , attribute "data-html" "true"
             ] [ i [ class "text-info fa fa-question-circle" ] []]
           ]
+          , invalidEnumTypeElem
```
clarktsiory, line 158: I think this should be moved upwards by 1 container : in the current "input-group" div, this looks like : 

image

ElaadF: In fact, we should move it downwards :D
image

clarktsiory (approved):
GG ! LGTM !

PR #5521, opened 2024-03-22
Fixes #24295: API export of groups doesn't export the categories as dependencies

fanf:
https://issues.rudder.io/issues/24295

The chosen new mapping is one that match the LDAP/UI representation of group categories and groups:

```
archive/
 `- groups
      |- category_with_name
      |     |- category.json // the category info: uuid, name, etc
      |     `- some_group.json // the group as before
      `- some_group_under_root_category.json
```

This mapping is so-so, since the group serialisation also contains the group category id, so we duplicate the information, and if we move a group, it doesn't actually change its category, which is surprising. But I wanted to have the same json serialisation than the one from API.

I hesitated with a mappring like:

```
archive/
 |- group-categories.json // one big json will the categories 
 `- groups
      |- some_group.json // the group as before
      `- some_group_under_root_category.json
``` 

But it was strang to have a mixe of "one item for each groups" and "one big json with lots of categories".
Still, maybe it would be easier to manipulate.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
@@ -64,10 +64,14 @@ import com.normation.rudder.domain.appconfig.FeatureSwitch
 import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.rudder.domain.nodes.NodeGroup
+import com.normation.rudder.domain.nodes.NodeGroupCategory
 import com.normation.rudder.domain.nodes.NodeGroupCategoryId
 import com.normation.rudder.domain.nodes.NodeGroupId
 import com.normation.rudder.domain.policies.Directive
 import com.normation.rudder.domain.policies.DirectiveId
+import com.normation.rudder.domain.policies.FullGroupTarget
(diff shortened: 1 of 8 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -435,12 +464,18 @@ class ZipArchiveBuilderService(
     val name = fileArchiveNameService.toFileName(origName) + extension
 
     // find a free name, avoiding overwriting a previous similar one
-    usedNames.modify(m => {
-      val realName = if (m(category).contains(name)) {
```
fanf, line 439: never do that, it breaks when the thing doesn't exist

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
```diff
@@ -499,13 +538,124 @@ class ZipArchiveBuilderService(
...
+     * Here: base
+     */
+    def recMapToZippable(
+        basePath:    String,
+        cat:         FullNodeGroupCategory,
+        catFileName: String,
+        usedNames:   Ref[Map[String, Set[String]]]
+    ): IOResult[Chunk[Zippable]] = {
+      for {
+        ref   <- Ref.make(Chunk.empty[Zippable])
```
fanf, line 613: we use a ref to avoid building list of list and then flatmapping in the end

fanf:
OK, merging this PR

PR #5528, opened 2024-03-25
Fixes #24592: Allow technique id starting with a number

amousset:
https://issues.rudder.io/issues/24592

* Create a separate type for technique id as the constraints are different
* Modify the schema accordingly
* Add a `Technique-` prefix to all Windows techniques (for consistency, I prefer to avoid a "surprise" prefix in corner cases).


Fdall (approved):

amousset:
OK, merging this PR

PR #5529, opened 2024-03-25
Fixes #24417: Promote a node to relay command failed with parsing message error

VinceMacBuche:
https://issues.rudder.io/issues/24417


fanf:
We would prefer to remove these methods and correct the callee. The `*Proxies` classes are supposed to be removed as soon as possible, so each time we have an opportunity to disengage from them, we should take it.

VinceMacBuche:
Replaced by https://github.com/Normation/rudder-plugins/pull/698 and https://github.com/Normation/rudder/pull/5530

PR #5530, opened 2024-03-25
Fixes #24593: Remove unused node function used by scale out relay plugin

VinceMacBuche:
https://issues.rudder.io/issues/24593


fanf (approved):
Good, let's remove that !

PR #5531, opened 2024-03-25
Fixes #24594: We can't save a technique with an enum value empty while required option is unchecked

ElaadF:
https://issues.rudder.io/issues/24594


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
PR rebased

clarktsiory:
I don't understand why we would want to allow that, unless there is a meaningful difference difference between "empty text" and "empty enum" :
* it's not consistent with the UX, why would a user change from `Text` to `Enum` but still provide an empty value ?
* when one or many parameters don't have any `Value` but have a `Display name`, the directive parameters form is misleading as the actual value is empty (this could be improved here by making the difference more explicit)

clarktsiory (approved):
LGTM ! It is for the use case of specifically having more than 1 enum possibilities but one of them is empty : 
```
param1 -> ""
param2 -> "a value"
param3 -> "another value"
```

PR #5533, opened 2024-03-26
Fixes #24493: Node detail page does not show the "Accepted since" information

ElaadF:
https://issues.rudder.io/issues/24493


amousset:
We want to keep it in the Rudder information as it is not linked to the node health but only a indicative information.

PR #5535, opened 2024-03-26
Fixes #24436: Tooltip on compliance in rule page is not rendered

RaphaelGauthier:
https://issues.rudder.io/issues/24436

Nothing magical here, I've just added more tooltip initialisation events until I find a definitive solution to this problem.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
@@ -817,7 +817,7 @@ function logout(cb){
 function initBsTooltips(){
   var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
   return tooltipTriggerList.map(function (tooltipTriggerEl) {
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
```diff
@@ -817,7 +817,7 @@ function logout(cb){
 function initBsTooltips(){
   var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
   return tooltipTriggerList.map(function (tooltipTriggerEl) {
-    return new bootstrap.Tooltip(tooltipTriggerEl,{container : "body", html : true});
+    return new bootstrap.Tooltip(tooltipTriggerEl,{container : "body", html : true, trigger : 'hover'});
```
RaphaelGauthier, line 820: By default, the trigger list is : `["hover", "focus"]`.
In some cases, the "focus" trigger can prevent certain tooltips from disappearing, so it is recommended not to use this trigger if this problem occurs.
However, this does not fix the problem of tooltips remaining displayed when the page is loading.

VinceMacBuche (approved):

PR #5541, opened 2024-03-27
Fixes #24590: Directive errors section is not consistent and always redirects to first tab

clarktsiory:
https://issues.rudder.io/issues/24590

Peek 2024-03-27 12-44

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
@@ -334,7 +334,7 @@ class DirectiveEditForm(
                         }) &
       "#directiveRulesTab *" #> ruleDisplayer &
       "#save" #> { SHtml.ajaxSubmit("Save", onSubmitSave _) % ("id" -> htmlId_save) % ("class" -> "btn btn-success") } &
-      "#notifications" #> updateAndDisplayNotifications() &
+      "#notifications" #> displayNotifications() &
       "#showTechnical *" #> <button type="button" class="btn btn-technical-details btn-default" onclick="$('#technicalDetails').toggle(400);$(this).toggleClass('opened');">Technical details</button> &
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
```diff
@@ -840,15 +855,13 @@ class DirectiveEditForm(
     }
   }
 
-  private[this] def updateAndDisplayNotifications(): NodeSeq = {
-
+  private[this] def displayNotifications(): NodeSeq = {
     val notifications = formTracker.formErrors
-    formTracker.cleanErrors
```
clarktsiory, line 846: This is the faulty line, it cleans errors too early. The form is re-rendered with errors near each field, so we need to clean only after re-rendering

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/DirectiveEditForm.scala
```diff
@@ -391,16 +391,27 @@ class DirectiveEditForm(
...
 
   private[this] def onNothingToDo(): JsCmd = {
     formTracker.addFormError(error("There are no modifications to save."))
     showErrorNotifications()
   }
 
+  private[this] def showVariablesErrorNotifications(): JsCmd = {
+    // only replace notification container to avoid resetting the tab state
+    onFailureCallback() & Replace("notification", displayNotifications())
+  }
```
clarktsiory, line 413: This allows to avoid going back to the first tab when we know there are errors in the parameters tab.

If there are errors though in the first tab, they will not be reported (the parameters errors need to be fixed first)

clarktsiory:
PR updated with a new commit

fanf (approved):
Works as expected !

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #5543, opened 2024-03-27
Fixes #24591: Tenant should be displayed in Node UI detailed

fanf:
https://issues.rudder.io/issues/24591

Nothing special in the code, this looks like that on a node with a tenant: 

image


VinceMacBuche:
I believe this should be a label in header, I do think it's as important a node state  (more important in fact)  maybe we can get rid of RAM if that's too much data

VinceMacBuche:
Anyway this is already a welcomed change and i will accept it. other changes can come in anoter PR

VinceMacBuche (approved):

PR #5547, opened 2024-03-27
Fixes #12791: Have a reset-keys button on the node screen

fanf:
https://issues.rudder.io/issues/12791

Add a button to reset the key status if:
- current user has write admin rights, 
- current key status is "certified". 

On backend, it's just a call to the relevant `woNodeRepository` function, plus some html/ajax to make it works. 
I needed to extract the div with security info to make it displayable again after change. 
I prefered to rebuild it all than just changing the icon because the key/key status logic is complex and I wanted to use the same logic for displaying (so that nothing change after reload). 

[change-key-status.webm](https://github.com/Normation/rudder/assets/44649/3d406506-21d7-45d1-984e-46bf86ca5b56)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
@@ -43,6 +43,7 @@ import com.normation.box.*
 import com.normation.cfclerk.domain.HashAlgoConstraint.SHA1
 import com.normation.eventlog.ModificationId
 import com.normation.inventory.domain.*
+import com.normation.rudder.AuthorizationType
 import com.normation.rudder.batch.AutomaticStartDeployment
 import com.normation.rudder.domain.logger.NodeLoggerPure
 import com.normation.rudder.domain.nodes.NodeInfo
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -630,6 +563,124 @@ object DisplayNode extends Loggable {
       """)))
   }
 
+  /*
+   * Display the div with ID `security-info` with cfenfine and curl key/certificate hash,
+   * the key/certificate detail, and the button to untrust the key
+   */
+  private def displaySecutiryInfo(node: NodeInventory): NodeSeq = {
```
VinceMacBuche, line 570: Secutiry ;)

amousset:
I'd prefer the button to be visible and not require to unfold the key (and we'll need to do it for 8.1 anyway)

VinceMacBuche:
Why Administration rights ? Administration is to manage rudder server and here to manage a Node parameter. I know that's a sensitive operation and can understand the will to put it behind walls, but I would have set "Node write" 

The api to update a Node, and reset it's key status is behind Node Edit right, (maybe not a good choice, like all edit) but i would have put the same right

fanf:
PR updated with a new commit

fanf:
[change-key-status-2.webm](https://github.com/Normation/rudder/assets/44649/d0623e73-89b8-4fea-aec9-9cc8dc66dbf1)

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5548, opened 2024-03-28
Fix new warnings introducted by scala 2.13.13 and -Xsource:3-cross option

mbaechler:
* boolean arguments need to be passed by name
* variable shadowing is now a warning
* case class constructor reference requires a `.apply`


fanf:
I'm closing this one since it was merged with https://github.com/Normation/rudder/pull/5639 - it remains only the first commit about pom change, that will need to go elsewhere.

PR #5558, opened 2024-03-29
Fixes #24637: Improvements on score: F score, rework tooltip, improve messages

VinceMacBuche:
https://issues.rudder.io/issues/24637


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/Score.scala
```diff
@@ -57,6 +57,7 @@ object ScoreValue extends Enum[ScoreValue] {
   case object C       extends ScoreValue { val value = "C" }
   case object D       extends ScoreValue { val value = "D" }
   case object E       extends ScoreValue { val value = "E" }
+  case object F extends ScoreValue { val value = "F" }
```
ElaadF, line 60: indent

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -456,7 +456,7 @@ CREATE TABLE NodeFacts (
 );
 
 
-CREATE TYPE score AS enum ('A', 'B', 'C', 'D', 'E');
+CREATE TYPE score AS enum ('A', 'B', 'C', 'D', 'E', 'F', 'X');
```
fanf, line 459: This leads to SQL errors for people with the beta2 since there is no migration. Pondering if it's important, not sure, but we should at least add in release notes: 
"if you upgrade from a beta version, you need to do: 
```
rudder=> ALTER TYPE score ADD VALUE 'F';
ALTER TYPE
rudder=> ALTER TYPE score ADD VALUE 'X';
ALTER TYPE
rudder=>
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ComplianceScore.scala
```diff
@@ -64,17 +65,19 @@ object ComplianceScoreEventHandler extends ScoreEventHandler {
...
-            Score(scoreId, C, "Node is compliant at least at 50%", p)
+            Score(scoreId, C, "Compliance is between 50% and 80%", p)
+          } else if (percent.compliance >= 20) {
+            Score(scoreId, D, "Compliance is between 20% and 50%", p)
           } else if (percent.compliance >= 25) {
-            Score(scoreId, D, "Node is compliant at least at 25%", p)
+            Score(scoreId, E, "Compliance is between 5% and 20%", p)
           } else {
-            Score(scoreId, E, "Node is compliant at less than 25%", p)
+            Score(scoreId, F, "Compliance is lower then 5%", p)
```
clarktsiory, line 80:
```suggestion
            Score(scoreId, F, "Compliance is lower than 5%", p)
```

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
VinceMacBuche: We don't support update between beta, we never did nor we want to support it

PR #5566, opened 2024-04-02
Fixes #24629: Score details columns in node tables cannot be saved 

clarktsiory:
https://issues.rudder.io/issues/24629

Peek 2024-04-02 16-00

We also need to await the the ajax for the score list have finished, before parsing the columns data from the cache. So, we just change the `createNodeTable` function to an async one (which is widely [browser-compatible](https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Operators/async_function))

PR #5568, opened 2024-04-02
Fixes #24627: We can add the same score details column many times in nodes tables

clarktsiory:
https://issues.rudder.io/issues/24627
**:warning: This PR is based on https://github.com/Normation/rudder/pull/5566 which should be merged before :warning:**

The second commit is the meaningful one.

Peek 2024-04-02 18-10

Now we cannot add a column when the score details is already in the table. 
We always keep the "Score details" selected, even if the value is empty (all score options have been added).


VinceMacBuche:
Same as https://github.com/Normation/rudder/pull/5566 i need some time to review it. i have not forgotten it

PR #5572, opened 2024-04-03
Fixes #24665: Number of nodes in rule badge is not correct with tenants

fanf:
https://issues.rudder.io/issues/24665

It's likely an older bug than that, I need to check it better. 
EDIT: it's from 7.3. 
But the only usage of that method in all rudder seems to be for badge, so not too important.


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #5575, opened 2024-04-04
Fixes #24234: Add warning for properties not case sensitive on Windows

ElaadF:
https://issues.rudder.io/issues/24234
image


ElaadF:
I'm closing this one, we don't want to display a waning message on every node, but only nodes that are affected, which is Windows node.
I will put a warn in the documentation

PR #5578, opened 2024-04-04
Fixes #24677: Improve apt/rpm package dependency detection

amousset:
https://issues.rudder.io/issues/24677


amousset:
PR updated with a new commit

PR #5594, opened 2024-04-10
Fixes #24701: Add daily schedule for campaign

VinceMacBuche:
https://issues.rudder.io/issues/24701

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
@@ -421,7 +421,7 @@ object MainCampaignScheduler {
 
   def nextCampaignDate(schedule: CampaignSchedule, date: DateTime): IOResult[Option[(DateTime, DateTime)]] = {
     schedule match {
-      case OneShot(start, end)        =>
+      case OneShot(start, end) =>
         if (start.isBefore(end)) {
           if (end.isAfter(date)) {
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -433,6 +433,31 @@ object MainCampaignScheduler {
...
+
+        val endDate = {
+          (if (end.realHour < start.realHour || (end.realHour == start.realHour && end.realMinute < start.realMinute)) {
+             startDate.plusDays(1)
+           } else {
+             startDate
+           }).withHourOfDay(end.realHour).withMinuteOfHour(end.realMinute).withSecondOfMinute(0).withMillisOfSecond(0)
+        }
+
+        Some((startDate, endDate)).succeed
```
fanf, line 459: I think we want to limit the max span to 20h or so to avoid all the borderline cases when two campaign override (ie, we don't even want to have to wonder what would happen). 

At least, it needs to be less than 24h

fanf:
I think the doc for system update API needs to be updated to explain the new data type alternative

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

VinceMacBuche:
api doc added

fanf (approved):

PR #5600, opened 2024-04-11
Fixes #22525: Directives applied twice don't show in rule details (they should be skipped)

clarktsiory:
https://issues.rudder.io/issues/22525

We need to add all the details of a skipped directive in the compliance API (here only for the `Rules` compliance API)
Here we are reusing [the logic](https://github.com/Normation/rudder/blob/branches/rudder/7.3/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala#L438) used in the server-rendered compliance which is now added to the API in a new JSON field : 
```
"skippedDetails": { "overridingRuleId": ..., "overridingRuleName": ... }
```

**Result :**
 
Screenshot from 2024-04-11 18-14-59

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
@@ -38,12 +38,15 @@
 package com.normation.rudder.rest.data
 
 import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.policies.Directive
 import com.normation.rudder.domain.policies.DirectiveId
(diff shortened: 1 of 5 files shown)


clarktsiory:
PR updated with a new commit

fanf:
OK, it's now working in rules: 
image

But it's still not present in the compliance part of node: 

image

fanf:
LGTM. I'm not sure if we want to use the "skipped" tag when a rule has all its directives skipped. 

image

I think it's a rare enought case to not matter a lot, and that version is clear. 

I'm merging that one, and if we want to do the skipped tag for rule too, we will open an other ticker.

fanf (approved):

fanf:
OK, squash merging this PR

PR #5601, opened 2024-04-11
Fixes #24713: Dynamic group are slow to compute in Rudder 8.1

fanf:
https://issues.rudder.io/issues/24713


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/LdapQueryProcessor.scala
```diff
@@ -440,13 +440,16 @@ class InternalLDAPQueryProcessor(
...
    */
   def rawInternalQueryProcessor(query: Query, debugId: Long = 0L): IOResult[Seq[NodeId]] = {
-    for {
-      nq  <- normalize(query).toIO.chainError("Error when normalizing LDAP query")
-      ids <- getMapDn(nq, debugId).flatMap {
-               case None      => Seq().succeed
-               case Some(dns) => executeLdapQueries(dns, nq, Seq("1.1"), debugId)
-             }
-    } yield ids
+    if (query.criteria.isEmpty) Seq().succeed
```
fanf, line 443: That method seems to be called only in the `NodeFactQueryProcessor` so we could not have guarded for empty (since it's also added their) ; but the old semantic seems strange: if there is no criteria line, then we shouldn't get any nodes, not all nodes. 

I'm not changing it elsewhere for fear of unsuspected side effects on other implementation relying on that semantic.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -99,7 +100,7 @@ object GroupAnd extends Group {
   }
   def inverse(a: NodeFactMatcher):                     NodeFactMatcher =
     NodeFactMatcher(s"!(${a.debugString})", (n: CoreNodeFact) => a.matches(n).map(!_))
-  val zero:                                            NodeFactMatcher = NodeFactMatcher("true", _ => true.succeed)
+  val zero:                                            NodeFactMatcher = NodeFactMatcher("and0_true", _ => true.succeed)
```
fanf, line 103: changing the debug string just to make it clearer that that `true` (or the corresponding `false` in `orGroup` case is really the zero case of the group, not a constant coming from a matcher.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -112,7 +113,22 @@ object GroupOr extends Group {
...
+ * A case class to sort a list of criterion lines into the different kind we know about:
+ * - core node fact
+ * - sub group
+ * - ldap (historical)
+ */
+final case class CriterionLines(
+    coreNodeFact: Chunk[CriterionLine],
+    subGroup:     Chunk[CriterionLine],
+    ldap:         Chunk[CriterionLine]
+)
```
fanf, line 129: That case class and the following lines correct one the big problem, ie the fact that we weren't really splitting between matchers kind, but just collecting several times, creating duplicates along the way. 
Now we cleanly check each line in turn and put in one (and only one) of the matcher kind and then create the matcher for its lines.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -146,9 +162,11 @@ class NodeFactQueryProcessor(
         _   <- FactQueryProcessorLoggerPure.Metrics.debug(s"Analyse query in ${t1 - t0} ms")
         res <- nodeFactRepo
                  .getAll()(QueryContext.todoQC, s)
-                 .flatMap(all =>
-                   ZIO.filter(all.values)(node => FactQueryProcessorLoggerPure.debug(m.debugString) *> processOne(m, node))
-                 )
+                 .flatMap(all => {
+                   ZIO
+                     .filterPar(all.values)(node => FactQueryProcessorLoggerPure.debug(m.debugString) *> processOne(m, node))
+                     .withParallelism(4)
```
fanf, line 168: a little optimisation that can reclam ~1/3 of the (small) process time. Here, we are just checking for all nodes in turn, so its highly parallelisable. But on the other hand, it's also extremelly cpu-cache efficient to just match over and over the same things. So 4 is arbitrary, but it should not take too much of a toil on small CPU machine, and still get a boost on higher count ones, without splitting unecessarly a Chunk of at most 10k nodes on too many small bits.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -176,28 +194,36 @@ class NodeFactQueryProcessor(
...
+          case _ if l.attribute.cType.isInstanceOf[LDAPCriterionType]               =>
+            lines.modify(_.ldap).using(_.appended(l)).succeed
+          case _                                                                    =>
+            // ??? Should not happen, perhaps log ?
+            FactQueryProcessorLoggerPure.warn(
+              s"The criterion line with attribute '${l.attribute.name}' of type ${l.attribute.cType} can't be handled. Please report that error to maintainers"
+            ) *>
+            lines.succeed
+        }
+    }
```
fanf, line 215: so, this the place where we attribute only one matcher kind for each line, with higher priority for NodeFact matcher than for LDAP matcher.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -216,15 +242,20 @@ class NodeFactQueryProcessor(
...
-    val ldapQuery = q.copy(transform = ResultTransformation.Identity, criteria = lines.toList)
-    Seq(for {
-      nodeIds <- ldapQueryProc.rawInternalQueryProcessor(ldapQuery)
-    } yield {
-      NodeFactMatcher(
-        s"[sub-ldap query ${ldapQuery.toString()}]",
-        (n: CoreNodeFact) => nodeIds.contains(n.id).succeed
-      )
-    })
+    if (lines.nonEmpty) {
```
fanf, line 245: The second big problem is solved here: when there is no LDAP line, we just don't create any LDAP matcher at all, period.

fanf:
PR updated with a new commit

VinceMacBuche:
Sorry forgot to mment, LGTM, I think you checked it again and it's ok ?

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #5604, opened 2024-04-16
Fixes #24729: Allow using a different password hash algorithm for each local user

amousset:
https://issues.rudder.io/issues/24729

* Makes a new `RudderPasswordEncoder` class that acts as a redirection to the correct encoder based on encoded password format
* Try to encapsulate better the internals of password management inside `RudderPasswordEncoder`
* Add a simple `securityLevel` lever to only allow legacy hashes when necessary
* Add the migration logic on the users file, run at startup and on reload, by using a boolean attribute `unsafe-hashes` to still be able to use legacy hashes

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
@@ -78,20 +78,18 @@ final case class UserFile(
 
 // Password encoder definition of parsing values
 sealed abstract class PasswordEncoderType(override val entryName: String) extends EnumEntry {
-  def name:        String = entryName
-  def displayName: String = name
+  def name: String = entryName
 }
 
 object PasswordEncoderType extends Enum[PasswordEncoderType] {
+  final case object MD5    extends PasswordEncoderType("MD5")
+  final case object SHA1   extends PasswordEncoderType("SHA-1")
+  final case object SHA256 extends PasswordEncoderType("SHA-256")
+  final case object SHA512 extends PasswordEncoderType("SHA-512")
+  final case object BCRYPT extends PasswordEncoderType("BCRYPT")
 
-  final case object PlainText extends PasswordEncoderType("plain") {
-    override def displayName: String = "plain text"
-  }
-  final case object MD5       extends PasswordEncoderType("MD5")
-  final case object SHA1      extends PasswordEncoderType("SHA-1")
-  final case object SHA256    extends PasswordEncoderType("SHA-256")
-  final case object SHA512    extends PasswordEncoderType("SHA-512")
-  final case object BCRYPT    extends PasswordEncoderType("BCRYPT")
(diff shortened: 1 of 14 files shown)


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
PR rebased since the merge of user-management. Since the algorithm is now by user, the API of user-management always return the default digest (BCRYPT), which is displayed as a hint when filling the password in the UI. We will need a PR to replace that UI field, and maybe infer and apply validations now that there are always known constraints.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckUsersFile.scala
```diff
@@ -0,0 +1,82 @@
...
+
+class CheckUsersFile(migration: UserFileSecurityLevelMigration) extends BootstrapChecks {
+
+  override def description: String = "Check if hash algorithm for user password is a modern one, if not enable unsafe_hashes"
+
+  def allChecks(currentSecurityLevel: SecurityLevel): IOResult[Unit] = {
+    for {
+      _ <- currentSecurityLevel match {
+             case SecurityLevel.Modern => ZIO.unit
+             case SecurityLevel.Legacy => migration.migrateToModern(UserManagementIO.getUserFilePath(migration.file))
```
amousset, line 58: I think `migrateToModern` is not a good name as it does the exact opposite, maybe "allowLegacy" or something like this.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -377,46 +493,49 @@ object UserFileProcessing {
...
+          ) *> Unexpected(s"rudder-users configuration file not found at path: '${config.path}'").fail
         }
     }
   }
 
+  def readUserFile(resource: UserFile): IOResult[Elem] = {
+    IOResult.attempt {
+      scala.xml.XML.load(resource.inputStream())
+    }.mapError {
+      // map a SAXParseException to a technical butmore user-friendly but error message
```
fanf, line 515: too many but in that comment

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckUsersFile.scala
fanf: `allowLegacy` seems good

fanf (commented):
Seems good. 

Can you had a unit test of : 
- a user file with unsafe algo getting chechekd and migrated ?
- a user file with bcrypt getting checked and not migrated ?

clarktsiory:
PR updated

amousset:
There is now a conflict, rebase required @clarktsiory

clarktsiory:
PR rebased

PR #5606, opened 2024-04-17
Fixes #24722: Update gulp related js dependencies

RaphaelGauthier:
https://issues.rudder.io/issues/24722

Only two js dependencies have been updated in this PR :
- del : from 6.1.1 to 7.1.0
- gulp : from 4.0.2 to 5.0.0

The gulpfile.js file (now **gulpfile.mjs**) has been converted into an Ecmascript module , so that the new version of del package can be imported as a module (that package has become an esm-only package in this new version). `gulp-cli` supports ESM so this change has no side effect.

This implies a few changes to the file :
- Most of them are listed in this guide : https://gist.github.com/noraj/007a943dc781dc8dd3198a29205bae04
- clean() function has been rewritten because of the change in the behaviour of the del package functions
- and in addition, merge-stream has been replaced by merge2 as it did not work with the new changes, and gulp-sourcemaps has been downgraded to the 2.6.5 version to avoid vulnerabilities present in the 3.0.0 version.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/gulpfile.mjs
+++ b/webapp/sources/rudder/rudder-web/src/main/gulpfile.mjs
@@ -1,21 +1,24 @@
-const fs = require('fs');
-const path = require('path');
-const { watch, series, parallel, src, dest } = require('gulp');
-const rename = require('gulp-rename');
-const mode = require('gulp-mode');
+import gulp from 'gulp';
+const { task, watch, series, parallel, src, dest } = gulp;
+import fs from 'fs';
+import path from 'path';
+import rename from 'gulp-rename';
+import mode from 'gulp-mode';
 const profile = mode();
-const terser = require('gulp-terser');
-const elm_p = require('gulp-elm');
-const merge = require('merge-stream');
-const del = require('del');
-const through = require('through2');
-const sass = require('gulp-sass')(require('sass'));
-const sourcemaps = require('gulp-sourcemaps');
+import terser from 'gulp-terser';
+import elm_p from 'gulp-elm';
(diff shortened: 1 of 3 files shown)


RaphaelGauthier:
Do not merge yet, I'm having issues with fontawesome and OTS errors :

font-error

RaphaelGauthier:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/gulpfile.mjs
```diff
@@ -133,7 +149,7 @@ function vendor_js(cb) {
 };
 
 function css(cb) {
-    src(paths.css.src)
+    src(paths.css.src, { encoding: false })
```
RaphaelGauthier, line 152: I had to add this option to make ionicons and fontawesome work.

Gulp 5 stream encoding is now set to UTF-8 by default, which is not compatible with these icon libs :

> Our streams now default to UTF-8 encoding. Previously, our streams took whatever data was emitted without considering any encoding. However, in this release, we resolved a 10 year old [issue](https://github.com/gulpjs/vinyl-fs/issues/23) to support custom encodings and we default these to UTF-8. Most usage of gulp won’t need to change anything, but certain plugins may produce non-UTF-8 output and you’ll need to set `{ encoding: false }` on your gulp streams.

RaphaelGauthier:
> Do not merge yet, I'm having issues with fontawesome and OTS errors :
> 
> font-error

It's now working, merge can be done !

RaphaelGauthier:
PR rebased

VinceMacBuche (approved):

PR #5607, opened 2024-04-17
Fixes #24712: ExpiredCompliance events are pilling up

fanf:
https://issues.rudder.io/issues/24712

Correct three loops (one in pending, two in unexpected handled in one case). Tested in the load server, which is now doing nothing in place of continously expiring / computing compliance.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
@@ -709,9 +709,10 @@ object ExecutionBatch extends Loggable {
                         )
                       } else {
                         // standard case: we changed version and are waiting for a run with the new one.
+                        // Pending status expires in expirationDate.
                         runType(
                           s"last run at ${t} was for expired configId ${rv.value} and no report received for current configId ${currentConfig.nodeConfigId.value}, but ${now} is before expiration time ${configurationExpirationTime}, Pending",
-                          Pending(currentConfig, Some((t, runConfig)), eolExpiration)
+                          Pending(currentConfig, Some((t, runConfig)), configurationExpirationTime)
                         )
                       }
                     }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
@@ -117,8 +117,6 @@ object CacheComplianceQueueAction       {
   final case class ExpectedReportAction(action: CacheExpectedReportAction)            extends CacheComplianceQueueAction {
     def nodeId = action.nodeId
   }
-  final case class InitializeCompliance(nodeId: NodeId, nodeCompliance: Option[NodeStatusReport])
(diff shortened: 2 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -709,9 +709,10 @@ object ExecutionBatch extends Loggable {
                         )
                       } else {
                         // standard case: we changed version and are waiting for a run with the new one.
+                        // Pending status expires in expirationDate.
                         runType(
                           s"last run at ${t} was for expired configId ${rv.value} and no report received for current configId ${currentConfig.nodeConfigId.value}, but ${now} is before expiration time ${configurationExpirationTime}, Pending",
-                          Pending(currentConfig, Some((t, runConfig)), eolExpiration)
+                          Pending(currentConfig, Some((t, runConfig)), configurationExpirationTime)
```
fanf, line 715: this is for avoiding the loop when we have a node without recent run and a which is pending

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -517,7 +515,12 @@ trait CachedFindRuleNodeStatusReports
     val nodeWithOutdatedCompliance = cache.filter {
       case (id, compliance) =>
         compliance.runInfo match {
-          case t: ExpiringStatus => t.expirationDateTime.isBefore(now)
+          // here, we have a special case for unexpected version: it is useless to recompute compliance until we don't have a new run,
+          // ie the node config was changed elsewhere. It means that "unexpected version" wins above "No report in interval",
+          // ie that that error is bigger.
+          case _: UnexpectedVersion => false
```
fanf, line 521: this is for avoiding the loops for when a node is un "unexpected version". We don't have to compute again and again it's compliance, since it won't change until we get a new run.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
```diff
@@ -117,8 +117,6 @@ object CacheComplianceQueueAction       {
   final case class ExpectedReportAction(action: CacheExpectedReportAction)            extends CacheComplianceQueueAction {
     def nodeId = action.nodeId
   }
-  final case class InitializeCompliance(nodeId: NodeId, nodeCompliance: Option[NodeStatusReport])
-      extends CacheComplianceQueueAction // do we need this?
```
fanf, line 121: this case is never used anywhere in the code

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ReportingServiceImpl.scala
ncharles: just to be sure, when policy are regenerated, then the compliance is invalidated and it's no longer unexpected but pending, right ?

ncharles:
The impact is impressive (you can see where the change happened)
image

ncharles (approved):

PR #5615, opened 2024-04-18
Fixes #24749:  Disabled LDAP users can still login and use Rudder

clarktsiory:
https://issues.rudder.io/issues/24749

Authentication validation should be the same as with "file" or "OIDC" : call `loadUserByUsername`, which takes into account the user status from the database to refuse authentication of disabled users.

There is some slight change in behavior : for LDAP users which were not found in the "users.xml" file we proceeded with a successful login but with no rights. Now, we fail at the login step. It is more consistent with the behavior for disabled users from "file" or "OIDC" backends.


clarktsiory:
PR updated with a new commit

fanf (approved):
Works as expected :+1:

clarktsiory:
OK, squash merging this PR

PR #5617, opened 2024-04-18
Fixes #24755: In case a NULL was introduced earlier in sessionid, login is impossible

fanf:
https://issues.rudder.io/issues/24755

Rationnal: we prefer to have an incorrect previous `sessionId` (I'm not even sure it's displayed) than makin login impossible.


clarktsiory (changes requested):
I think we need to close this : the error log may have been from the `tenants` column and not `sessionid` and because of dev schema shenanigans between 8.1 and 7.3 versions

PR #5618, opened 2024-04-19
Fixes #24757: Unable to log in - Server error

clarktsiory:
https://issues.rudder.io/issues/24757

The `tenants` column in the database is nullable (it is null especially when server is upgraded from 8.0 to 8.1).
Changing the case class changes the doobie Meta and now the null values before migration will not throw an error


fanf (approved):
LGTM.

PR #5621, opened 2024-04-22
Fixes #24723: Tree group is slow to load up because it contains the list of nodes in the tree

clarktsiory:
https://issues.rudder.io/issues/24723

Adding an internal groups API and using it instead of the public one.

All the JSON datatype is made with zio-json and chimney, and it is almost the same model as a full group category, because all the information is needed in the Elm app except the node ids.

Also added API tests for that (FYI the tests for the public groups API does not have one for the tree endpoint)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
@@ -67,6 +67,7 @@ import com.normation.rudder.ncf.ResourceFile
 import com.normation.rudder.ncf.TechniqueParameter
 import com.normation.rudder.repository.FullActiveTechnique
 import com.normation.rudder.repository.FullActiveTechniqueCategory
+import com.normation.rudder.repository.FullNodeGroupCategory
 import com.normation.rudder.rule.category.RuleCategory
 import com.normation.rudder.rule.category.RuleCategoryId
 import com.normation.rudder.services.queries.CmdbQueryParser
@@ -76,6 +77,7 @@ import com.normation.utils.DateFormaterService
 import com.softwaremill.quicklens.*
 import com.typesafe.config.ConfigRenderOptions
 import com.typesafe.config.ConfigValue
+import io.scalaland.chimney.Transformer
 import io.scalaland.chimney.dsl.*
 import zio.*
 import zio.Tag as _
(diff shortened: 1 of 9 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -870,6 +887,66 @@ object JsonResponseObjects {
...
+            cat.ownGroups.values.toList.map(t => {
+              implicit val categoryId: Option[NodeGroupCategoryId] = cat.categoryByGroupId.get(t.nodeGroup.id)
+              t.nodeGroup.transformInto[JRGroupInfo]
+            })
+          }
+        )
+        .withFieldComputed(
+          _.targetInfos,
+          _.targetInfos.collect {
+            case t @ FullRuleTargetInfo(_: FullOtherTarget, _, _, _, _) => t.toTargetInfo.transformInto[JRRuleTargetInfo]
```
clarktsiory, line 943: `FullOtherTarget` because group targets are already in `groups` fields

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -870,6 +887,66 @@ object JsonResponseObjects {
...
+      description:                            String,
+      @jsonField("categories") subCategories: List[JRGroupCategoryInfo],
+      groups:                                 List[JRGroupCategoryInfo.JRGroupInfo],
+      @jsonField("targets") targetInfos:      List[JRRuleTargetInfo]
+  )
+
+  object JRGroupCategoryInfo {
+    final case class JRGroupInfo(
+        id:                              NodeGroupId,
+        @jsonField("displayName") name:  String,
```
fanf, line 905: for this one and other: why not using the json name directly ? Is it an choice given other constraints ?

fanf: (I imagine with chimney, but if so: is there a reason to chose that way ?

clarktsiory: It avoid a boilerplate for the `Transformer` definition in chimney : we can derive the field transformation if the name is the same as the source datatype.  If we have the `displayName` field, we would have to write `withFieldRenamed(_.displayName, _.name)` (it is even more verbose if the transformer could be `derive`-d directly)

fanf: but then, you wrote `@jsonField("displayName")̀` :) OK, there's less chars in that latter option. And perhaps it's a bit clearer.

fanf (approved):
LGTM !

PR #5625, opened 2024-04-23
Fixes #24772: RuleTarget compliance is very slow to compute

fanf:
https://issues.rudder.io/issues/24772

So, the main idea is to have as little computing done as possible in the recurring part. Especially, we want to avoid any `contains` or `set.intersec` in that part. 

Since the parent change was introduced because we need to always know what the real set of node Ids is, and not assume that the content of group's `nodeIds` is the reality, we just always intersect the result of the recurring function with the keySet of the map of "allNodes". 
Also remove reference to map view here, to avoid unecessary cast from SetView to Set, even if that change was tested to account for almost no perf change at all.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
@@ -292,20 +292,23 @@ object RuleTarget extends Loggable {
    * allNodes pair is: (nodeid, isPolicyServer)
    */
   def getNodeIdsChunk(
-      targets:          Set[RuleTarget],
-      allNodes:         Map[NodeId, Boolean /* isPolicyServer */ ],
-      groups:           Map[NodeGroupId, Chunk[NodeId]],
-      allNodesAreThere: Boolean = true // if we are working on a subset of node, set to false
+      targets:  Set[RuleTarget],
+      allNodes: Map[NodeId, Boolean /* isPolicyServer */ ],
+      groups:   Map[NodeGroupId, Chunk[NodeId]]
   ): Chunk[NodeId] = {
-    val result = getNodeIdsChunkRec(Chunk.fromIterable(targets), allNodes, groups, allNodesAreThere)
-    result.distinct
+    val result = getNodeIdsChunkRec(Chunk.fromIterable(targets), allNodes.toMap, groups)
+
+    // we always must intersect with the keySet of nodes, because sometime node ids in groups is a super set
+    // of the given allNodes, for ex!
+    // - for static group where some of the nodes were deleted,
+    // - when we have tenants (Rudder 8.1+=
+    Chunk.fromIterable(result.toSet.intersect(allNodes.keySet))
   }
 
   def getNodeIdsChunkRec(
-      targets:          Chunk[RuleTarget],
-      allNodes:         Map[NodeId, Boolean /* isPolicyServer */ ],
-      groups:           Map[NodeGroupId, Chunk[NodeId]],
-      allNodesAreThere: Boolean = true // if we are working on a subset of node, set to false
+      targets:  Chunk[RuleTarget],
+      allNodes: Map[NodeId, Boolean /* isPolicyServer */ ],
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
```diff
@@ -315,49 +318,36 @@ object RuleTarget extends Loggable {
...
 
           // here, if we don't find the group, we consider it's an error in the
           // target recording, but don't fail, just log it.
           case GroupTarget(groupId)         =>
             val groupNodes = groups.getOrElse(groupId, Chunk.empty)
             val filtered   = {
-              if (allNodesAreThere) groupNodes
-              else {
-                val keys = Chunk.fromIterable(allNodes.keys)
-                groupNodes.filter(keys.contains(_))
```
clarktsiory, line 337: I think this is the culprit : a `contains` on a `Chunk` is `O(n)`, it should have been `allNodes.contains(_)` on the Map which is `O(1)`

clarktsiory (approved):
LGTM, I'm fine with the idea of applying the intersect at the end. But maybe later we should change the resulting type of the `getNodeIdsX` methods to `Set[NodeId]` instead of `Chunk[NodeId]` as we have an additional `Chunk.fromIterable(...)` here

clarktsiory:
It needs `spotless:apply` though

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
```diff
@@ -295,17 +295,20 @@ object RuleTarget extends Loggable {
       targets:          Set[RuleTarget],
       allNodes:         Map[NodeId, Boolean /* isPolicyServer */ ],
       groups:           Map[NodeGroupId, Chunk[NodeId]],
-      allNodesAreThere: Boolean = true // if we are working on a subset of node, set to false
   ): Chunk[NodeId] = {
-    val result = getNodeIdsChunkRec(Chunk.fromIterable(targets), allNodes, groups, allNodesAreThere)
-    result.distinct
+    val result = getNodeIdsChunkRec(Chunk.fromIterable(targets), allNodes.toMap, groups)
```
ncharles, line 299: i don't understand the toMap here; allNodes is already a map

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
```diff
@@ -295,17 +295,20 @@ object RuleTarget extends Loggable {
...
+
+    // we always must intersect with the keySet of nodes, because sometime node ids in groups is a super set
+    // of the given allNodes, for ex!
+    // - for static group where some of the nodes were deleted,
+    // - when we have tenants (Rudder 8.1+=
+    Chunk.fromIterable(result.toSet.intersect(allNodes.keySet))
```
ncharles, line 305: if we are converting to a set at the end, wouldn't it be cheaper to deal with a set all along ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
fanf: it's a mapview (or is in latter release) and it will be a noop if the underlying instance is a map without pending operation

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/RuleTarget.scala
fanf: it's not the case in latter branches, so if we change here, we will have to change in upmerge. 7.3 
I can do it for clarity on that PR, but it's not what we will get in 8.0/8.1

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #5629, opened 2024-04-25
Fixes #24788: Empty space in global compliance bar

RaphaelGauthier:
https://issues.rudder.io/issues/24788

Previously, the width of the compliance bars was calculated as a percentage, which involved many calculations to determine the correct width of each of the bars according to their value.

Thanks to the flexbox CSS property, we no longer need to calculate the width of the bars, and the minimum size is defined in CSS.

This avoids calculation errors and incorrect total bar widths.

### Before:
compliance-1

### After:
compliance-2


VinceMacBuche (approved):

RaphaelGauthier:
OK, merging this PR

PR #5630, opened 2024-04-25
Fixes #24097: Skipped directive does not show in node compliance tree

clarktsiory:
https://issues.rudder.io/issues/24097

Essentially the same as https://github.com/Normation/rudder/pull/5600, and with some refactoring : 
* move the Elm model for `SkippedDirective` to the Elm `Compliance.Datatypes` module
* compute the skipped directives in a method which takes the conversion `DirectiveComplianceOverride => T` as a param, either `ByRuleDirectiveCompliance` or `ByRuleDirectiveCompliance` (this PR) 

Screenshot from 2024-04-25 17-48-22


fanf:
So, the code looks good and we now have skipped in node, but I still think there is something broken: in the case where the rule has only one directive and that directive is skipped, then the rule doesn't appear at all in the node. 

Step to reproduce: in your screenshot example, remove the directive "File download (HTTP)" from rule "Install package skip"

fanf:
I'm going to merge that one as it is, it's already a nice enhancement, and then I'm opening a new issue for the remaining problem for rule with all directive skipped.

fanf (approved):

clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #5632, opened 2024-04-25
Fixes #24784: Nodes page has console error from the tree script

clarktsiory:
https://issues.rudder.io/issues/24784

Within the node details page, the group tree was still having disappearing event handlers (even without the console error) : 
* it seems like the previous selectors for tree elements with `lift-event-js-` id was not applied to the _pencils_ (in the `.treeActions` container) 
* the rendering of the tree only populates event handlers when the `ready.jstree` event is fired

N.B. : the use case here and the fix here are specific to the node details page, the groups page renders the tree using Elm and does not use the modified code


clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

clarktsiory:
PR rebased

PR #5635, opened 2024-04-26
Fixes #24779: Groups compliance summary need API pagination

clarktsiory:
https://issues.rudder.io/issues/24779


clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

fanf (commented):
Is this still relevant ? If so, we should make it go in Rudder 8.3. It looks like people can live without that in 8.1/8.2.

clarktsiory:
Yes it is still relevant, and it is a UX bug for users with more than 50 groups...
It likely depends on when these users will want to upgrade to 8.3. Until now the issue has only been reported once in 8.1

PR #5638, opened 2024-04-30
Fixes #24799: Move user-management plugin to rudder

clarktsiory:
https://issues.rudder.io/issues/24799

Moving all the code from the plugin involves : 
* moving HTML and snippets to the "Administration" menu
* moving Elm code into a new Elm package similar to existing ones, and reusing our notifications library
* moving Scala code (including tests) to _rudder-rest_ and move back some definitions in _rudder-web_ to _rudder-rest_ when possible


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -337,50 +340,12 @@ final class FileUserDetailListProvider(roleApiMapping: RoleApiMapping, authorisa
...
-          ) *>
-          UserFile(config.getAbsolutePath, () => new FileInputStream(config)).succeed
-        } else {
-          ApplicationLoggerPure.Authz.error(
-            s"Can not find configuration file specified by JVM property ${JVM_AUTH_FILE_KEY}: ${config.getPath}; aborting"
-          ) *> Unexpected(s"rudder-users configuration file not found at path: '${config.getPath}'").fail
-        }
-    }
-  }
+class UserFileProcessing(bcryptCost: Int) {
```
clarktsiory, line 343: this class is needed to pass the config value because we don't have access to `RudderConfig.RUDDER_BCRYPT_COST` in rudder-rest

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -76,34 +75,18 @@ final case class UserFile(
     inputStream: () => InputStream
 )
 
+sealed abstract class PasswordEncoder(override val entryName: String) extends EnumEntry {
+  def name:        String = entryName
+  def displayName: String = name
+  def encoder:     PasswordEncoder.Rudder
+}
```
clarktsiory, line 82: We need this ADT because BCRYPT specifically has 1 additional parameter (`cost: Int`) which is a configuration value used to implement the `encoder`. It also allows easier parsing since it is an enum from _enumeratum_

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -117,61 +100,100 @@ object PasswordEncoder {
...
+  type Rudder = SpringSecurityPasswordEncoder
+
+  override def values: IndexedSeq[PasswordEncoder] = findValues
+
+  override def extraNamesToValuesMap: Map[String, PasswordEncoder] = Map(
+    "sha"    -> SHA1,
+    "sha1"   -> SHA1,
+    "sha256" -> SHA256,
+    "sha512" -> SHA512
+  )
```
clarktsiory, line 161: a very useful trick in enumeratum

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/MockServices.scala
```diff
@@ -677,3 +697,164 @@ class MockCompliance(mockDirectives: MockDirectives) {
...
+  val providerRoleExtension: Map[String, ProviderRoleExtension] = Map("file" -> ProviderRoleExtension.WithOverride)
+  val authBackendProviders:  Set[String]                        = Set("file")
+}
+
+object MockUserManagement {
+  // Default mock with fake values and returning the temporary directory for cleanup
+  def apply(): (File, MockUserManagement) = {
+    val tmpDir = File.newTemporaryDirectory("rudder-users")
+    (tmpDir, new MockUserManagement(fakeUsers, fakeUserSessions, fakeUserFile(tmpDir)))
+  }
```
clarktsiory, line 799: Each time we instantiate a `MockUserManagement` we will need to cleanup the temp directory

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
fanf: yep

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -76,34 +75,18 @@ final case class UserFile(
...
 
+sealed abstract class PasswordEncoder(override val entryName: String) extends EnumEntry {
+  def name:        String = entryName
+  def displayName: String = name
+  def encoder:     PasswordEncoder.Rudder
+}
+
 //Password encoder type definition. Done like that to avoid
 //a deprecation warning on `PasswordEncoder` thanks to @silent annotation
-object PasswordEncoder {
```
fanf, line 81: For all that, get a look to what @amousset is doing in https://github.com/Normation/rudder/pull/5604

I think it's compatible with it, but I don't how hard the other subject will become.

fanf (commented):
Looks good and work as intended, just can you take a look for the refactoring around passwork encoder and the other PR where we want to be able in 8.2 to use a password hash algo by user (with a global default one) to make change more iterative ?

clarktsiory:
@fanf :+1: Okay I'll have a look at the other PR

clarktsiory:
PR updated : 
* fixed Elm `List.FlatMap` dependency removal
* renaming `SpringPasswordEncoder` to initial import `PasswordEncoder` and our custom password encoder to `RudderPasswordEncoder` instead 
* renaming the enum from `PasswordEncoder` to `PasswordEncoderType` and use the simple enum whenever possible 
* using a new class `PasswordEncoderDispatcher` to map the parsed enum type to the actual encoder : we need to use it to create a `PasswordEncoder` instance. In https://github.com/Normation/rudder/pull/5604, the encoder type (sha256, bcrypt, ...) is only inferred from the actual password value, so we will need to use the dispatcher when matching the password. It needs to be a separate class which depends on the bcrypt cost global configuration value

fanf (approved):
Works as expected!

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

PR #5644, opened 2024-05-02
Fixes #24504: The pending nodes history is not working anymore

clarktsiory:
https://issues.rudder.io/issues/24504
:warning: **We need to add event logs for deletion, acceptation, refusal events and consider migrating back** 

We need to use both sources for compatibility : 
* the event logs
* the nodefacts

I added a `getAll` method for nodefacts, and did not eventually filter by date. There is a dilemna : if we filter nodefacts since the most recent event log, we are sure to not get duplicates and have a consistent history. But normally we completely switch to nodefacts when we apply the migration, so there is no reason that it would be inconsistent if we don't filter...

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogDetailsService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogDetailsService.scala
@@ -560,7 +560,6 @@ class EventLogDetailsServiceImpl(
       actorIp      <-
         (details \ "actorIp").headOption.map(_.text) ?~! ("Missing attribute 'actorIp' in entry type node : " + entry.toString())
     } yield {
-      println("OK: found inventory log details")
       InventoryLogDetails(
         nodeId = NodeId(nodeId),
         inventoryVersion = ISODateTimeFormat.dateTimeParser.parseDateTime(version),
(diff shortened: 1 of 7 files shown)


clarktsiory:
PR updated with a new commit

fanf:
I don't get the case for a getAll with a date. 
What you seems to want is avoid duplicate, and that's not done by limiting by date. Just deduplicate based on nodeId if it's the goal. 
But actually, it seems to be ok to even have several event for one node (not what we are storing in base, we only keep last). So what you really want (likely) is dedup for a given pair (nodeId, eventDate) to avoid duplicate in the case where the migration didn't do what it was supposed to ? 

Even so, I'm not sure it's worth it: we would have a failed migration without user knowing about it. It's generaly better to have a clear problem than an obfuscated one. I would just remove all the code related to the filter.

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf:
I get an error when I try to access a node history without a corresponding fact: 

```2024-05-03 15:09:44+0200 DEBUG sql - Failed Resultset Processing [9 ms total (1 ms exec + 8 ms processing)]
  select nodeId, acceptRefuseEvent, acceptRefuseFact from nodefacts where nodeId = ? and acceptRefuseEvent ->>'date' = ?
 arguments = [0c6ea44e-db4a-474a-bfac-0546bfd20371, 2016-12-18T00:01:03+01:00]
   failure = ResultSet exhausted; more rows expected.``` 

![image](https://github.com/Normation/rudder/assets/44649/9f2a433e-8d9c-461b-ba25-260a0ce30c68)


This node didn't have an history anymore since it was cleaned since then. 
You can reproduce by: 
- accept or refuse a node, 
- delete the corresponding entry from postgresql nodefacts table. 

The error is the expected behavior, but it looks like in 8.0 it is now very user unfriendly. 
I wonder if ```  def get(id: ID, version: V): IOResult[HLog] ```  shouldn't return an option so that we can customize the message for the user with "details for that event are not available" on none, and "an error occured, please contact your admin to see what is going on" on failure with logs for that case. 

Else it's ok and seems to work as expected.

fanf:
Resolved by https://github.com/Normation/rudder/pull/5668

PR #5645, opened 2024-05-02
Fixes #24627: We can add the same score details column many times in nodes tables

VinceMacBuche:
https://issues.rudder.io/issues/24627


webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -1473,14 +1478,15 @@ function createNodeTable(gridId, refresh) {
...
-  });
-
+  if (scoreList.length === 0) {
+    $.ajax({
+      type: 'GET',
+      url: "/rudder/secure/api/scores/list",
+      success: function (data) {
+        scoreList = data.data
+      },
+    }).done( function() {refresh(gridId)} )
```
clarktsiory, line 1488: This does a 2nd API request for node details every time we create the table, is there a way without the `refresh` ?

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -43,6 +43,15 @@ var inventories = {};
...
 var scoreList = [];
+
+$.ajax({
+  type: 'GET',
+  url: "/rudder/secure/api/scores/list",
+  success: function (data) {
+    scoreList = data.data
+  },
+})
+
```
clarktsiory, line 54: We still have a race condition with this block, that executes asynchronously (`$.ajax` is fire-and-forget).
When it finished BEFORE the execution of the code that looks into localStorage for score columns, it works normally. But it often finishes AFTER, and the list is `[]`, so we are still displaying `system-updates Score` instead of `System updates Score` (because there is a "guard" to just display the `id` when we don't know the `name` of the score...)

VinceMacBuche:
Commit modified

VinceMacBuche:
PR rebased

VinceMacBuche:
Commit modified

PR #5646, opened 2024-05-02
Fixes #24817: When there are more nodes than supported by licenses, no logs in webapp states it

fanf:
https://issues.rudder.io/issues/24817

Successfully where to add the check to be useful. 

Now, the log looks like that at start: 

image

At at run time, when a limited feature is triggered: 

image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PluginStatus.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PluginStatus.scala
@@ -37,6 +37,7 @@
 
 package com.normation.plugins
 
+import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import org.joda.time.DateTime
 
 /**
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PluginStatus.scala
```diff
@@ -77,9 +78,17 @@ trait PluginStatus {
...
-    case _                                                                          => false
+  def isEnabled(): Boolean = {
+    current match {
+      case PluginStatusInfo.EnabledNoLicense | _: PluginStatusInfo.EnabledWithLicense => true
+      case PluginStatusInfo.Disabled(reason, optInfo)                                 =>
+        val name = optInfo match {
+          case Some(i) => s"'${i.softwareId}' "
+          case None    => ""
+        }
+        ApplicationLoggerPure.Plugin.logEffect.warn(s"Plugin ${name}is disabled: ${reason}")
```
fanf, line 89: this will be outputed each time a limited feature is accessed. It may spam logs, but it seems more important to have the message if number of nodes grow out of bound after start

VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #5648, opened 2024-05-02
Fixes #24829: Replace compliance chart with Score chart and and new details score charts

VinceMacBuche:
https://issues.rudder.io/issues/24829

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/homePage.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/homePage.js
@@ -114,8 +114,8 @@ function homePage (
     globalCompliance
   , globalGauge
   , nodeCompliance
-  , nodeComplianceColors
   , nodeCount
+  , scoreDetails
 ) {
   var opts = {
     lines: 12, // The number of lines to draw
@@ -173,6 +173,7 @@ function homePage (
       }
       stats += "<br>There "+ verb +" also " + pendingNodes + " for which we are still waiting for data (" + nodeCount.pending.percent + "%)."
     }
+
     $("#globalComplianceStats").html(stats);
 
     gauge.set(function() {
@@ -183,8 +184,21 @@ function homePage (
     $("#gauge-value").text(globalGauge+"%");
 
     var complianceHColors = nodeCompliance.colors.map(x => complianceHoverColors[x]);
-    doughnutChart('nodeCompliance', nodeCompliance, allNodes, nodeCompliance.colors, complianceHColors);
+    doughnutChart('nodeCompliance', nodeCompliance, nodeCompliance.colors, complianceHColors);
   }
+
+
+  scoreDetails.forEach(function(score) {
+    $("#scoreBreakdown .node-charts").append(
+              `<div class="node-chart">
+                <h4 class="text-center">${score.name}</h4>
+                <canvas id="score-${score.scoreId}" > </canvas>
+                <div  id="score-${score.scoreId}-legend"></div>
+              </div>`)
+    var complianceHColors = score.data.colors.map(x => complianceHoverColors[x]);
+    doughnutChart('score-'+ score.scoreId, score.data, score.data.colors, complianceHColors);
+  })
+
   initBsTooltips();
 }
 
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

fanf:
It seems that there is a problem with the percent computing: 
image

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/HomePage.scala
```diff
@@ -69,10 +70,21 @@ import net.liftweb.http.js.JsCmds.*
...
+
+    import com.normation.rudder.score.ScoreValue.*
+    scoreValue match {
+      case A       => "#13beb7"
+      case B       => "#68c96a"
+      case C       => "#b3d337"
+      case D       => "#fedc04"
+      case E       => "#f0940e"
+      case F       => "#da291c"
+      case NoScore => ""
```
fanf, line 84: Why not grey ?

VinceMacBuche:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/HomePage.scala
VinceMacBuche: In fact i think we want to filter Nodes without a score, but i don't know really. Yes in this case we should display them in grey

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/HomePage.scala
```diff
@@ -336,6 +314,7 @@ class HomePage extends StatefulSnippet {
           , ${data.toJsCmd}
           , ${diagramColor.toJsCmd}
```
fanf, line 315: I think this is not used anymore and should be removed

fanf:
The problem with percent is because the total number of node is not the same in the array than in the total :
- call js function `doughnutChart('nodeCompliance', nodeCompliance, allNodes, nodeCompliance.colors, complianceHColors);`
With: 

``` 
nodeCompliance:
        {
            "labels": ["E", "F"],
            "values": [1, 148],
            "colors": ["#f0940e", "#da291c"]
        }

allNodes: 130
```

I have disabled and pending nodes, which seems to be excluded from `allNodes`, but not from the data given with label/values. It must be consistent, likely by being all computed server-side and used everywhere

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/homePage.js
```diff
@@ -114,8 +114,8 @@ function homePage (
     globalCompliance
   , globalGauge
   , nodeCompliance
-  , nodeComplianceColors
```
fanf, line 117: that's not used anywhere

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/homePage.js
```diff
@@ -183,8 +184,21 @@ function homePage (
     $("#gauge-value").text(globalGauge+"%");
 
     var complianceHColors = nodeCompliance.colors.map(x => complianceHoverColors[x]);
-    doughnutChart('nodeCompliance', nodeCompliance, allNodes, nodeCompliance.colors, complianceHColors);
```
fanf, line 186: the count is now computed directly in `doughnutChart` function

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/homePage.js
```diff
@@ -229,9 +243,10 @@ var complianceHoverColors =
   , "#DA291C" : "#ed1809ff"
   }
 
-function doughnutChart (id,data,count,colors,hoverColors) {
+function doughnutChart (id,data,colors,hoverColors) {
 
   var context = $("#"+id)
+  var count = data.values.length < 1 ? 0 : data.values.reduce((a, b) => a + b, 0);
```
fanf, line 249: in place of count, compute the sum of values here. 
I didn't see a case where it should be ok to have a different count, and all calling place in scala backend use the same source of data to build both

PR #5650, opened 2024-05-03
Fixes #24832: Migrate rudder-web to zio-json

clarktsiory:
https://issues.rudder.io/issues/24832


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
```diff
@@ -184,8 +185,7 @@ class AcceptNode extends Loggable {
    * popuId : the id of the popup
    */
   def details(jsonArrayOfIds: String, template: NodeSeq, popupId: String): JsCmd = {
-    implicit val formats = DefaultFormats
-    val serverList       = parse(jsonArrayOfIds).extract[List[String]].map(x => NodeId(x))
```
clarktsiory, line 188: `parse` could throw an exception but we already have an alter on user-side when the list is empty, so we can do with an empty list

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #5655, opened 2024-05-06
Fixes #24839: We cannot login with a user login containing uppercase letter if the option case-sensitivity is set to false

clarktsiory:
https://issues.rudder.io/issues/24839

To load a user we both need to consider case-sensitivity when : 
* getting the user id from the database => we need to search with `LOWER(id)`
* getting the user name from the users xml file (which is a map with the original file usernames as keys) => we search the key matching a case-insensitive comparison, there will only be one result because we already ignore duplicates when validating the file

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -38,7 +38,9 @@ package com.normation.rudder.users
  */
 
 import cats.data.NonEmptyList
+import com.normation.errors.Inconsistency
 import com.normation.errors.IOResult
+import com.normation.errors.SystemError
 import com.normation.rudder.db.Doobie
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.utils.DateFormaterService
@@ -47,6 +49,7 @@ import doobie.*
 import doobie.free.connection.unit as connectionUnit
 import doobie.implicits.*
 import doobie.postgres.implicits.*
+import doobie.util.invariant.UnexpectedContinuation
 import org.joda.time.DateTime
 import zio.*
 import zio.interop.catz.*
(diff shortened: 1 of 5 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -860,8 +862,9 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
     transactIOResult(s"Error when retrieving user information")(xa => sql.query[UserInfo].to[List].transact(xa))
   }
 
-  override def get(userId: String): IOResult[Option[UserInfo]] = {
-    val sql = sql"""select * from users where id = ${userId}"""
+  override def get(userId: String, isCaseSensitive: Boolean): IOResult[Option[UserInfo]] = {
+    val filter = if (isCaseSensitive) fr"id = ${userId} " else fr"lower(id) = ${userId.toLowerCase()}"
+    val sql    = fr"""select * from users where""" ++ filter
 
     transactIOResult(s"Error when retrieving user information for '${userId}'")(xa => sql.query[UserInfo].option.transact(xa))
```
fanf, line 869: here, you need to manage the case where there is more than one result

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
clarktsiory: That's a well seen edge case !
:white_check_mark: Fixed with https://github.com/Normation/rudder/pull/5655/commits/1c5f6d4bed99d82742587af9372e55afec84d1e8 and https://github.com/Normation/rudder/pull/5655/commits/a325e13d14574a7b3c793c424a1ff497ae00f94f, I made it throw an `Inconsistency` and added test for both in-memory and jdbc repository

fanf (approved):
Good!

PR #5656, opened 2024-05-07
Fixes #24833: Migrate NodeApi to zio-json in rudder-rest and utils

clarktsiory:
https://issues.rudder.io/issues/24833

* First commit : create zio-json datatypes for node details fields. This is just basically creating the same datatype but with the same fields order as in previous serialiser, and doing the mapping using chimney
* Second commit : remove lift-json imports from `utils`
* Remaining : migrating the APIs


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/RestDataSerializer.scala
```diff
@@ -110,12 +108,7 @@ trait RestDataSerializer {
       detailLevel: DetailLevel
   ): JValue
 
-  def serializeServerInfo(srv: Srv, status: String): JValue
-
-  def serializeNodeInfo(nodeInfo: NodeInfo, status: String): JValue
-  def serializeNode(node:         Node): JValue
-
-  def serializeInventory(inventory: FullInventory, status: String): JValue
```
clarktsiory, line 120: :tada: Those are getting removed little by little

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
```diff
@@ -334,14 +371,62 @@ object JsonQueryObjects {
...
+      } else {
+        val base         = {
+          if (fields.contains("minimal")) {
+            MinimalDetailLevel
+          } else {
+            DefaultDetailLevel
+          }
+        }
+        val customFields = fields.filter { field =>
+          field != "minimal" &&
```
fanf, line 392: I'm not sure why you special-case `minimal` and `default` here. From the field's name point of view, they are invalid like "foo" could be, and you already took care of them aboce. 

It won't have any impact on perf either, since `allFields` is not big, the hash resolution will be quick. 
Is there any other reason for doing so I don't see ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -103,6 +135,703 @@ object RenderInheritedProperties {
...
+      implicit val nodeInfo:    NodeInfo               = nodeFact.toNodeInfo
+      implicit val securityTag: Option[SecurityTag]    = nodeFact.rudderSettings.security
+      implicit val software:    Chunk[domain.Software] = nodeFact.software.map(_.toSoftware)
+      // we could keep inventory as Option to make syntax easier to filter empty lists, or write Some(...).filter(_.nonEmpty) everywhere
+      implicit val inventory:   Option[FullInventory]  = Some(nodeFact.toFullInventory)
+      transformer
+    }
+
+    implicit def transformer(implicit
+        nodeInfo:    NodeInfo,
```
fanf, line 285: you need to start from `NodeFact` here. We kept nodeinfo / inventory for compat reason with the existing API, but that's not the goal. We need to take the opportunity to remove it enterily from here.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -103,6 +135,703 @@ object RenderInheritedProperties {
...
+        .withFieldComputed(
+          _.virtualMachines,
+          levelField(_)("virtualMachines")(
+            inventory.map(_.node.vms.transformInto[Chunk[VirtualMachine]]).filter(_.nonEmpty)
+          )
+        )
+        .buildTransformer
+    }
+
+    @jsonDiscriminator("type") sealed abstract class OsDetails(
```
fanf, line 422: I don't get why you are redefining these structure while they already exist in `NodeFact`. 
One of the goal of node facts is to be a "as near as possible" json <-> scala mapper. 
There will be some friction at some points (espcially regarding rudder config / agent / etc) for the rest output, but in most case, we would like to use the same data structures. 
For bios etc, it should be ok.

If it's done for the maybe-different lifecycle of data: right, we need to have stronger garanties for the rest API than for node fact. Is this for that ? If so, can you clearly document it ? Plus, we will likely want a process to know what to do where when nodefact or rest data need to change.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
clarktsiory: This whole portion of code was copied from the previous [RestExtractorService method](https://github.com/Normation/rudder/blob/branches/rudder/8.1/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RestExtractorService.scala#L276) to "parse" the node details level...

Right below we actually need to separate the input set of string into either : 
* special values ("full", "minimal", "default")
* custom values ("bios", etc.), which we need to validate before passing them to `CustomDetailLevel(special, custom)`
=> Maybe what we want here is a smarter constructor for `CustomDetailLevel`, which is the actual structure that needs the splitting

clarktsiory: (and I forgot to remove the stale `RestExtractorService` method which is now unused)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
clarktsiory: :white_check_mark: => https://github.com/Normation/rudder/pull/5656/commits/b8c2bf3e57e6486c5fb19c633a0c75199e84271f

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
fanf: As a general rule of thumb, the less we have code, the better it is. That's less code to maintain, less code to modify (typically if we add a relevant information in inventory for bios, we need to change it at many more place without automatic (as in compiler/tests) way to prevent error, and also - and especially - less code to wonder "why that code exists? What is meaningful about it, is it important or just something to clean-up?"

All this as to be balanced with the need to have strong garanties to public API, and both your consideration are valid. 
So let's see how we do the balancing :)

- in the case of node facts, one of the stated goal is to get a common bi-directionnal json/scala structure. So it's a place where we are ok to sacrifice a little bit of business-code-side agility in favor of global optimisation in code reuse. What means that sacrifices will happen, especially now that we are at the first iteration of node facts more general use (ie the first time it goes out of it's poc place where everything was neatly aligned for it). 
- for adding fields, that's not a problem. This doesn't break any existing code, and actually it's something we want to be able to do routinely. But for a public API, what we don't want, it's empty optionnal field, which clutters a lot the output and json size. Typically, `"description":""` is generally not what is done, we remove the field. It seems that we can do that in the case of NodeFact serialisation too
- for field ordering. Well, it should not matter, since json is not ordered in the spec, but it will, because the world is like that. Still, it seems that the ordering can be done directly in business objects. The risk it then that somewhere, we switch two `Option[String]` without noticing

fanf: given that, I would like that you explore the cost/risk for changing `NodeFact` and related data structure that are identical modulo order/additional fields, and only keep API-specific structure for when there is changes that need to be taken care of. Basically, so that if a structure is redefined, the person reading the code can know/see/read the comment to understand why it exists without having to guess

fanf: Actually, all objects/fields are not equals. For most of them, we don't care of the order: slot, video, etc - if people use them, it's for data mapping, so field name will be used. So, yes, look it's risky/costly to change order in our business object, but if yes, perhaps we will just accept to change their order in public API.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
clarktsiory: I agree with finding the right balance, so I looked at how to minimize additional datastructures.
The resulting change is here => https://github.com/Normation/rudder/pull/5656/commits/04e280cd3c2c35704bc997f465c5c31ed578b570

When reusing all NodeFact structures I still found a main obstacle : the existing serializers use the original field name (`ifAddresses` VS `ipAddresses`, `commandName` VS `name`) in that case we are doomed because we would have to rename the encoded field, which breaks compatibility for existing serialization of node facts, or we would need to encode both fields, which will need another case class anyway...

So, **the ones that need to be copied and changed are the ones which have renamed any field** in the JSON of the node details endpoint. I found out eventually that the only ones from nodefacts whose codec is initially compatible with the endpoint JSON are :
* Processor
* Slot
* Sound
* Video
And sadly, that's all :disappointed:... all others have to be duplicated (I added block comments for each of them though)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -397,174 +397,84 @@ object JsonResponseObjects {
...
     }
 
-    @jsonDiscriminator("type") sealed abstract class OsDetails(
-        val name:          String,
-        val version:       Version,
-        val fullName:      String,
-        val servicePack:   Option[String],
-        val kernelVersion: Version
-    )
-    object OsDetails {
```
clarktsiory: :heavy_check_mark: This has already been factored to a `JsonOsDetails` for the codec definition in `NodeFacts`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -397,174 +397,84 @@ object JsonResponseObjects {
...
-        val id:           Option[MachineUuid],
-        val provider:     Option[VmType],
-        val manufacturer: Option[Manufacturer],
-        val serialNumber: Option[String]
-    )
+    object MachineType extends Enum[MachineType] {
+      case object UnknownMachineType  extends MachineType("Unknown")
+      case object PhysicalMachineType extends MachineType("Physical")
+      case object VirtualMachineType  extends MachineType("Virtual")
+      case object NoMachine           extends MachineType("No machine Inventory")
```
clarktsiory, line 423: I simplified the `MachineInfo` ADT to make it closest to the domain object of the same name i.e. instead of having the type discriminator on an sum type ADT, just map the `machineType` attribute from the case class to this custom `MachineType` one...

(I should have done that from the start :see_no_evil:)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
fanf: I'm not at ease with having diverging serialization because just a field name, especially since for now, nodefact serialization is only used internally for storage of inventory on acceptation/refusal. So in the *worst* case, we can just change the names in 8.1.n + migrate in base and forget about it. 

Still, migration are annoying, so a better solution would be to write for the time being in a common format, but be able to read both (and do that in 8.1, and only have the new format 8.2).

zio-json allows that with `@jsonAliases("...")` so I think we should just do that. 
https://zio.dev/zio-json/configuration#aliases

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -103,6 +135,581 @@ object RenderInheritedProperties {
...
+        .define[NodeInventory, ManagementDetails]
+        .withFieldComputed(_.cfengineKeys, n => Chunk.fromIterable(n.agents.map(_.securityToken)))
+        .withFieldComputed(_.cfengineUser, _.main.rootUser)
+        .buildTransformer
+    }
+
+    final case class DisplayDateTime(value: DateTime) extends AnyVal
+
+    /**
+     * Needed because datetime is serialized differently and therefore needs the DisplayDateTime wrapper
```
fanf, line 510: I'm not sure we want that. Having only one date format for everything is more UX friendlier

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
fanf: Actually with aliases, I'm not even sure we need to do the change in 8.1, we can keep the aliases for ~ forever

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -103,6 +135,581 @@ object RenderInheritedProperties {
...
+        quantity:                 Int,
+        description:              Option[String],
+        manufacturer:             Option[Manufacturer]
+    )
+    object Controller {
+      implicit val transformer: Transformer[domain.Controller, Controller] = Transformer.derive[domain.Controller, Controller]
+    }
+
+    /**
+     * Needed for "swap", which has no mountpoint but the field is not optional in the domain object
```
fanf, line 539: It's ok to use `none` for swap mountpoint, it's what is done in fstab, so we're not going out of standard path here

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
clarktsiory: I made the `Bios` keep the same format of datatime for the `releaseDate` field in the API versions before v20 https://github.com/Normation/rudder/pull/5656/commits/b591687a7572efef195c57d3d289553551157d4c

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
clarktsiory: I added it to the the current API https://github.com/Normation/rudder/pull/5656/commits/b591687a7572efef195c57d3d289553551157d4c

webapp/sources/ldap-inventory/inventory-api/pom.xml
```diff
@@ -55,5 +55,9 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
       <groupId>org.bouncycastle</groupId>
       <artifactId>bcpkix-${bouncycastle-compat}</artifactId>
     </dependency>
+    <dependency>
+      <groupId>com.comcast</groupId>
+      <artifactId>ip4s-core_${scala-binary-version}</artifactId>
+    </dependency>
```
clarktsiory, line 61: it was needed to move the `INetAddress` codec from `NodeFacts` (from rudder-core)

clarktsiory:
PR updated with a new commit

fanf (approved):
:tada:  LGTM :clap: :clap: :clap:

fanf:
OK, merging this PR

PR #5657, opened 2024-05-12
Fixes #24826: Add an API endpoint to get plugin license info

fanf:
https://issues.rudder.io/issues/24826

Add a public API to get information about the installed plugins and their limits. 
The returned JSON structure start by global limits if any, which are computed to present the most restrictive limits (ie, if a plugin limits to 200 nodes and one to 500, then 200 is given). 

Then, we have an array with installed plugins and their specific limits if any. 

Most of the PR is documentation for that API, and for the plugins repo setting API, which was missing. 

There is nothing complicated at all, but two noticable points: 
- I've started using `java.time` API, and added the relevant formatter/etc in `DateTimeFormaterService` ; 
- plugin's license object needed to be migrated from `rudder-web` to `rudder-rest`;

### DIFF
--- a/webapp/sources/api-doc/code_samples/curl/plugins/get-plugins-settings.sh
+++ b/webapp/sources/api-doc/code_samples/curl/plugins/get-plugins-settings.sh
@@ -1,0 +1,1 @@
+curl --header "X-API-Token: yourToken" https://rudder.example.com/rudder/api/latest/plugins/settings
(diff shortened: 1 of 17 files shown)


VinceMacBuche:
There is a wip commit ! can you rebase into one ?

VinceMacBuche (commented):
apart the wip commit, this is good to me and can be self merged

fanf:
OK, merging this PR

PR #5663, opened 2024-05-15
Fixes #24834: When we save an API Token, the UI is grayed out

RaphaelGauthier:
https://issues.rudder.io/issues/24834

I had to replace the use of bootstrap attributes with a "hard-coded" display of modals in Elm. + I've made the content of the modal scrollable, so that the height of the modal doesn't exceed the height of the screen.


clarktsiory:
LGTM, we are no longer stuck with the modal now.

But something weird happens with "static backdrop" modal, there behavior is different when clicking outside of the modal content to close it. For example : 
* I create a token, and click on the "gray" area to close it
* I attempt to create another one but when I click on the creation button, there is a console error and I have to refresh the page to make it disappear : 
Screenshot from 2024-05-15 10-31-29

The same error also occurs when just attempting to regenerate more than one token.
I think it has to do with our Elm modal states...

RaphaelGauthier:
PR updated with a new commit

RaphaelGauthier:
OK, squash merging this PR

PR #5666, opened 2024-05-16
Fixes #24845: Option to force validation of change requests cannot be set via API

clarktsiory:
https://issues.rudder.io/issues/24845

The setting key in the API : `enable_validate_all`


fanf (approved):
Works as expected

fanf:
OK, merging this PR

PR #5668, opened 2024-05-16
Fixes #24869: Node acceptation or refusal is not logged in event logs in 8.0

clarktsiory:
https://issues.rudder.io/issues/24869

For some reason, the "version" i.e. DateTime of the inventory couldn't be retrieved in the _nodefacts_ table after the `HistorizeNodeStateOnChoice` acceptor saved it, maybe it is just the wrong timing or the ordering of acceptors is wrong...

So instead just pass the same datetime into context everywhere and don't rely on that for logging into the event log :grin:...

(I have not tested this PR but it compiles)


fanf:
Not only does it works, it also repair the history as expected: 
image

fanf (approved):

fanf:
OK, merging this PR

PR #5678, opened 2024-05-23
Fixes #24863: when we have compliance right, the rule page shows a weird message

ElaadF:
https://issues.rudder.io/issues/24863

It seems like there is bigger issue under this one.

### Initial issue
The initial problem was that when we go to a rule page, in Elm we initialize the page by fetching many information of which change-request's information that required `administration_read` right.
```
enable_change_message
enable_change_request
enable_self_deployment
enable_self_validation
enable_validate_all
```

### Potential issue discovered
In the initialization of the Rule page, we check if change-request is enabled. This fetching is done through [settings API](https://docs.rudder.io/api/v/19/#tag/Settings/operation/getAllSettings) [GetAllSetings](https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala#L535) that required [adminstration_read right](https://github.com/Normation/rudder/blob/master/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RoleApiMapping.scala#L114)
The main problem with this PR and logic is that an user for example with `validator` role will not see any change-request on a rule, because he doesn't have the administration_read access, and we don't want to great him with admin read because it can see other settings that are only meant for admin users.
 
### Potential solution
One possibility would be to extract the change-request part of the `GetAllSetings` API to create something like `GetChangeRequestSettingds`, add appropriate rights to this in order for a `validator` (maybe other roles) user to retrieve this info

### How I solved the problem

I have added some API calls for the specific settings :
- enable_change_message
- mandatory_change_message
- change_message_prompt
- enable_change_request
- enable_self_deployment
- enable_self_validation
- enable_validate_all

`GET /settings/<setting_name>`
We trigger these API calls if the user have `validator_read` right

:warning: This fix should be well tested, the risks are :
- Hide the change request message even if the user as the right to see the CR message
- Disabled CR on a rule for a user, so he can modify the rule without generating a change-request


VinceMacBuche (approved):

ElaadF:
PR updated with a new commit

ElaadF:
PR updated with a new commit

ElaadF:
PR updated with a new commit

ElaadF:
Commit modified

ElaadF:
Commit modified

PR #5679, opened 2024-05-23
Fixes #24907: The list of plugin in info API is always empty

fanf:
https://issues.rudder.io/issues/24907

So, at some point `ZIO` changed implementation of `ZIO.succeed` to make its parameter call by name, which was not reflected in our syntax add-on, which means that we were capturing the result of the computation and not a real effect.

### DIFF
--- a/webapp/sources/utils/src/main/scala/zio/syntax.scala
+++ b/webapp/sources/utils/src/main/scala/zio/syntax.scala
@@ -5,7 +5,7 @@ import zio.IO
  * .fail and .succeed were removed in RC18
  */
 object syntax {
-  implicit class ToZio[A](a: A) {
+  implicit class ToZio[A](a: => A) {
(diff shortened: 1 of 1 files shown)


webapp/sources/utils/src/main/scala/zio/syntax.scala
```diff
@@ -5,7 +5,7 @@ import zio.IO
  * .fail and .succeed were removed in RC18
  */
 object syntax {
-  implicit class ToZio[A](a: A) {
+  implicit class ToZio[A](a: => A) {
```
clarktsiory, line 8: ~It looks legit, but it doesn't seem to fix the issue, however doing a `ZIO.succeed(...)` explicitly works :exploding_head:~ I had to re-install the `utils` project locally, it works !

clarktsiory (approved):
LGTM !

PR #5683, opened 2024-05-24
Fixes #24865: User management should display tenants, last and previous session

clarktsiory:
https://issues.rudder.io/issues/24865

Change is both in the API and the UI :
* in the API : 
  * add `tenants` as a string, which is sourced from `users.xml` file or the _usersessions_ table. I had no choice but parse it as a regex since from the `tags:[...]` format is not the one we want to display in the table, so API value is either `all`, `none` or `zoneA, zoneB`
  * add `previousLogin`, to be distinguished from `lastLogin` which could be the "current login with still an open session"
* in the UI : 
  * display tenants as is, but consider `all` and `none` as special values 
  * use `previousLogin` everywhere instead of `lastLogin` because that's what we want to display wrt semantic of columns. We will use `lastLogin` information later to infer if the user is currently logged in, and display a green circle (when both dates are different)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/UserManagementApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/UserManagementApi.scala
@@ -201,6 +201,7 @@ class UserManagementApiImpl(
     getAuthBackendsProviders:  () => Set[String]
 ) extends LiftApiModuleProvider[UserManagementApi] {
   api =>
+  import UserManagementApiImpl.*
 
   override def schemas: ApiModuleProvider[UserManagementApi] = UserManagementApi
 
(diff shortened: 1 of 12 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/UserManagementApi.scala
```diff
@@ -628,4 +689,29 @@ class UserManagementApiImpl(
...
+    nodePerms.fold(identity, _.value) match {
+      case tenantsTagsRegex(r) => r
+      case s                   => s
+    }
+  }
+}
+
+object UserManagementApiImpl {
+  // Tenants value encoded in user session of form "tags:[...]" needs to be displayed as list (see .value)
+  private val tenantsTagsRegex = """^tags:\[(.*)\]$""".r
```
fanf, line 712: I don't think you need that, nor you want that. It creates a coupling between rudder-core and something that is out of rudder core, in the tenant plugin. The boundary dependencies need to go the other way around. 

It will be fragile and will need synchronisation between the plugin and rudder, when it was designed to not need one. 

From Rudder, you need to distinguish the case where the plugin is there, or not. If the plugin is not there, then there is only two possibilities: all or none. 
If the plugin is there, then it is *its* responsibility to translate what something is (and even give context: it's likely you don't want to display an ID but the human name of the tenant with the short description on hover or something like that). 

So, from an arch point of view, it seems that: 
- you need in Rudder something that tells you if the tenant plugin is enabled or not. I'm not sûr we have it, but if not I should be possible to model it on the old user extended right way

Then, you need to have a service that from a tenant serialisation form gives you the current (ie, with regards to the current context of plugins) value of tenant right, and from that you will do whatever you want in the api. 
That service will have a default implementation in rudder core which only says all or none, and a better implementation added in the plugin which says none, all, or a list of tenants. It's classicaly implemented with;

``` 
trait MyService { def giveMeWhatINeed(s: String) : IOResult[XXX]
object MyCoreService { def giveMeWhatINeed(s: String) : IOResult[XXX] = all or none }
class MyServiceProvider {
  val myService = Ref.make(MyCoreService)
}
// and in RudderConfig
val myServiceProvider = new MyServiceProvider(...) 
// + expose it in RudderConfig API

// and in tenant plugin
class MyBetterService(dependencies) {   def giveMeWhatINeed(s: String) : IOResult[XXX] = some logic and retrival }
// and in tenant plugin init
RudderConfig.myServiceProvider.myService.setTo(new MyBetterService(dependencies))

fanf (changes requested):
Some tests are not passing, plus I don't think you are solving it right with respect to systems rudder-core <-> tenant plugin boundaries

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/UserManagementApi.scala
clarktsiory: :+1: Indeed, the table should display the current tenants of the user.

In fact we already have the `nodePerms: NodeSecurityContext` of the user as it is parsed from the `users.xml` file, but I just didn't realize I could use it instead of the regex... :sweat_smile: 

So, instead of using the session data for displaying the tenants, I could just use the file data, and it seems consistent with the existing model of tenants for a `RudderUserDetail`.

For the record, I initially thought it was similar to user permissions, that we need to display from previous session data for externally provided users, because we don't have any other way... In the case of tenants, it appears we can always display from the declarations in the users file, and `all` if the user is not in the file

clarktsiory:
PR updated with a new commit

fanf:
The general logic seems ok and all, but you still need to hide the tenant column when the tenant plugin is not there: tha column will be full list of "all" for all user of core, which is just a lost of space and in the worst case misleading. 

I think we can merge like that, it's already a nice enhancement, and come back to that later on. I will need a way to check if tenant feature is present (usually, we build a capacity service like there was for extended roles, but perhaps it would make sense to have only one capacity service with a list of features enabled/disabled, and the possibility for plugins to let the service know they provide the feature - with a runtime check needed, because license and the world is complicated).
Is this OK for you to get that one merged, and open a related issue for the hiddening of the tenant column ?

clarktsiory:
This is a nice suggestion, and yes I think we can merge it like that and come back later ! 
I created an issue for that here : https://issues.rudder.io/issues/24976

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #5684, opened 2024-05-27
Fixes #24439: Clicking on technique in technique tree doesn't do anything

clarktsiory:
https://issues.rudder.io/issues/24439

This follows the problem in groups and directives tree in https://github.com/Normation/rudder/pull/5382 : the same problem appeared for the `techniqueLibraryManagement` page.
* The tree almost has the same structure as the directive tree, but we only have "technique nodes" 
* I refactored duplicate code that alters the `allEvent` global variable


clarktsiory:
Commit modified

VinceMacBuche (approved):

VinceMacBuche (approved):

PR #5688, opened 2024-05-29
Fixes #24930: Compliance score chart for vulnerabilities is all black with X

clarktsiory:
https://issues.rudder.io/issues/24930

* Replace the black color (empty `""`) with gray (`#d8dde5`, same as the gray color for the gauge in the dashboard)
* Display `No ... score found` instead of `X` which is not very user-friendly. With chart.Js, it seems that the legend text is hard to customize, so need to map the label `No ... score found` back to the `X` label when building query filters when redirecting to nodes page...

**Screenshot :**  
Screenshot from 2024-05-29 10-23-42


clarktsiory:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
@clarktsiory I think it should be "No Score" not a sentence needed

VinceMacBuche:
OK, squash merging this PR

PR #5689, opened 2024-05-29
Fixes #24925: Refresh on nodes tables leads to datatable and console error 

clarktsiory:
https://issues.rudder.io/issues/24925

Since we now pass the scores list directly as JS object from the _Scala_ code, we need to move the `allColumns` variables into the scope where we have access to the scores list, as it needs the list.
Changes consist of : 
* moving global variables into `createNodesTable`
* propagate `scores` list where it is needed (when refreshing and reloading tables)


VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #5690, opened 2024-05-29
Fixes #24944: Add pre generation hook

VinceMacBuche:
https://issues.rudder.io/issues/24944

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
@@ -75,6 +75,7 @@ import com.normation.rudder.facts.nodes.CoreNodeFact
 import com.normation.rudder.facts.nodes.NodeFactRepository
 import com.normation.rudder.facts.nodes.QueryContext
 import com.normation.rudder.hooks.HookEnvPairs
+import com.normation.rudder.hooks.HookReturnCode
 import com.normation.rudder.hooks.Hooks
 import com.normation.rudder.hooks.HooksLogger
 import com.normation.rudder.hooks.RunHooks
@@ -108,7 +109,6 @@ import org.joda.time.Period
 import org.joda.time.format.ISODateTimeFormat
 import org.joda.time.format.PeriodFormatterBuilder
 import scala.collection.MapView
-import scala.collection.immutable.Map
 import scala.concurrent.duration.FiniteDuration
 import zio.{System as _, *}
 import zio.syntax.*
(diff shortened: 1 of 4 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -1920,10 +1932,26 @@ trait PromiseGeneration_Hooks extends PromiseGenerationService with PromiseGener
...
   /*
    * Pre generation hooks
    */
   override def runPreHooks(generationTime: DateTime, systemEnv: HookEnvPairs): Box[Unit] = {
+    (for {
+      preHooks <- RunHooks.getHooksPure(HOOKS_D + "/policy-generation-pre", HOOKS_IGNORE_SUFFIXES)
+      res <-
+        RunHooks.syncRun(preHooks, HookEnvPairs.build(("RUDDER_GENERATION_DATETIME", generationTime.toString)), systemEnv) match {
+          case _: HookReturnCode.Success => ().succeed
+          case x : HookReturnCode.Error => Inconsistency(s"${x.msg}\n stdout: ${x.stdout}\n stderr: '${x.stderr}'").fail
```
fanf, line 1945: Since it will interrupt the policy generation and got up to user, I think we need a simpler error message than that ("policy generation interrupted by pre-check hook xxxx, error message: xxxxx" or something like that. 
And still get that error message in log for the ops who will do the migration and need detailled info.

webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-pre/readme.adoc
```diff
@@ -0,0 +1,14 @@
+= policy-generation-pre
+
+== When/What ?
+
+This directory contains hooks executed before policy generation.
+
+Typically, these hooks are used to check the integrity of techniques.
```
fanf, line 7: You need to explain that these hooks behave differently than all the other ones and that the will PREVENT policy generation on error. 
In other cases, an error just stop the hook pipeline

webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-pre/hook-template
```diff
@@ -0,0 +1,18 @@
...
+# For example, 'HOOK_WARN_TIMEOUT' = 1 second (without quote, without text on the line after duration) will emit a warning if
+# the hook takes more than one second to run.
+
+# Parameters are passed by environment variable:
+#
+# - RUDDER_GENERATION_DATETIME: generation datetime: ISO-8601 YYYY-MM-ddTHH:mm:ss.sssZ date/time that identify that policy generation.
+
+# Errors code on hooks are interpreted as follow:
+# - 0     : success, no log (apart if debug one)          , continue to next hook
+# - 1-31  : error  , error   log in /var/log/rudder/webapp/, stop processing
```
fanf, line 14: this need to be adapted. 

I think "stop processing" should be updated in all hook template to be more precise on what is stop (ie just processing of other hooks) ; and on that one, very loudly say that it will prevent policy generation (and so possibly block all policy generation in case of error in the script)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -1920,10 +1932,26 @@ trait PromiseGeneration_Hooks extends PromiseGenerationService with PromiseGener
     traverse(codeHooks.toSeq)(_.beforeDeploymentSync(generationTime)).map(_ => ())
   }
 
+
   /*
    * Pre generation hooks
    */
   override def runPreHooks(generationTime: DateTime, systemEnv: HookEnvPairs): Box[Unit] = {
+    (for {
+      preHooks <- RunHooks.getHooksPure(HOOKS_D + "/policy-generation-pre", HOOKS_IGNORE_SUFFIXES)
```
fanf, line 1941: Perhaps `pre-start`  ? Just `pre` doesn't seems very clear

fanf (commented):
I really think the user doc and error message reporting must be reworked, else users won't understand why this hook behave differently.

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf:
Works as expected, just a problem with tests (spotless) 
image

fanf (approved):

VinceMacBuche:
Commit modified

PR #5691, opened 2024-05-29
Fixes #24942: Properties tab in groups has no content after a save

clarktsiory:
https://issues.rudder.io/issues/24942

The problem was that the Elm app is not mounted when a refresh of the whole DOM occurs on the server side (it leaves the `<div id="nodeproperties-app"></div>` as is).

We just forgot to initialize the app each time it occurs : on success or failure of the node group form update.


fanf (approved):
Works as expected :+1:

VinceMacBuche:
OK, merging this PR

PR #5692, opened 2024-05-29
Fixes #24945: Error when going to group page

fanf:
https://issues.rudder.io/issues/24945

So, the root cause of the bug is that some key is not in the map and we try to get it, which fails with an exception. 
The cheap correction could have been just adding a test "if(allRuleInfo.contains(rule.id)" or something like that. 
A better one would have been to pattern match on `allRuleInfo.get(rule.id)`
But what we really want is to make that thing impossible by construction. 

Here, we can just avoid passing both source of rules as parameter and construct the map before, so the problem can't exist at all.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -893,14 +893,10 @@ class ComplianceAPIService(
       allDirectives:         Map[DirectiveId, (FullActiveTechnique, Directive)],
       nodeFacts:             MapView[NodeId, CoreNodeFact],
       nodeSettings:          Map[NodeId, RudderSettings],
-      rules:                 Seq[Rule],
-      allRuleInfos:          Map[RuleId, (Chunk[NodeId], Option[PolicyMode])],
+      rules:                 Map[RuleId, (Rule, Chunk[NodeId], Option[PolicyMode])],
       level:                 Option[Int],
       isGlobalCompliance:    Boolean
   )(implicit qc: QueryContext): IOResult[ByNodeGroupCompliance] = {
-    val ruleMap = rules.map(r => (r.id, r)).toMap
-    val ruleIds = ruleMap.keySet
-
     for {
       compliance <- getGlobalComplianceMode
       globalMode <- getGlobalPolicyMode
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -893,14 +893,10 @@ class ComplianceAPIService(
       allDirectives:         Map[DirectiveId, (FullActiveTechnique, Directive)],
       nodeFacts:             MapView[NodeId, CoreNodeFact],
       nodeSettings:          Map[NodeId, RudderSettings],
```
fanf, line 895: We have two source of the same elements, this is often a by idea because most of the time, it comes with hidden assumption about what properties hold on both group to not break the following implementation logic

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -893,14 +893,10 @@ class ComplianceAPIService(
...
-      rules:                 Seq[Rule],
-      allRuleInfos:          Map[RuleId, (Chunk[NodeId], Option[PolicyMode])],
+      rules:                 Map[RuleId, (Rule, Chunk[NodeId], Option[PolicyMode])],
       level:                 Option[Int],
       isGlobalCompliance:    Boolean
   )(implicit qc: QueryContext): IOResult[ByNodeGroupCompliance] = {
-    val ruleMap = rules.map(r => (r.id, r)).toMap
-    val ruleIds = ruleMap.keySet
```
fanf, line 902: I think it's not really necessary. It is not from a perf point of view, since keySet is built when the map is built (and is actually a false keySet, it forward operation to the map, which likely mean the map can't be gc'ed if we pass the set around, but that has nothing to do with the story here)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -1181,127 +1193,138 @@ class ComplianceAPIService(
...
-        ZIO.foreach(nodeGroupsInfo.toList) {
-          case (id, (name, serverList, _)) =>
-            (getByNodeGroupCompliance(
-              id,
-              name,
-              serverList,
-              directiveLib.allDirectives,
-              nodeFacts,
-              nodeSettings,
-              globalRulesByGroup(id),
```
fanf, line 1282: we really want to avoid that. It is not very resilient to refactoring. 
Here, we avoid it again by making is unrepresentable with the datastructure

clarktsiory (approved):
LGTM ! It is indeed cleaner and easier to reason about !

VinceMacBuche:
OK, merging this PR

PR #5699, opened 2024-05-31
Enable new Xsource:3-cross compilation flag for scala 2.13.14

mbaechler:


mbaechler:
no longer relevant, the flag is activated on master

PR #5703, opened 2024-06-01
Fixes #24876: Add a compliance-by-feature logic and save compliance in base

fanf:
https://issues.rudder.io/issues/24876

Add a `PolicyTypes` class, which contains a list of `PolicyTypeName`, which are aspects of a policy, like "is system" or "base user technique". Generally, a policy will only have one aspect, but we want to be able in the futur to color policies with several aspects to build ad-hoc view of compliance. 

`PolicyType` superseeds `isSystem` boolean in configuration objects. All object can still use that boolean and the code knows how to translate it, but there's no urgency to rewrite things and migration is transparent. If both `isSystem` and `PolicyTypes` are present on the same object, `PolicyTypes` wins.  

For now, `PolicyTypes` is only added to a minimal subset elements: 
- internal technique, with the XML property of `metadata.xml` `POLICY_TYPES` ;
- active technique in LDAP, with an evolution of LDAP schema,
- internal generation object `PolicyTechnique`, use to hold node policy values and build expected reports. 

Then, the main change is that now, `NodeStatusReport`, ie the main compliance object, used to hold the computed compliance for node given expected reports and a node with reports changed from:

```
final class NodeStatusReport private (
    ...
    val reports:    Set[RuleNodeStatusReport]
    ...
)
```
To:
```
final class NodeStatusReport private (
    ...
    val reports:    Map[PolicyTypeName, AggregatedStatusReport]
    ...
)
```
So now, a node status report has its compliance segmented by policy type (typically, user defined policies, system policies, campaign policies...). 

Note: the meaning of `NodeStatusReport.compliance` was not change, and is still the aggregate of all policy types, but I'm not sure it makes much sense; Two other standard compliance (base and system) are added. 

For now, the upper compliance services were not modified to keep the change small. But a lot of simplification will be possible, since most of compliance services are doing complicated things to split compliance by user/system.

Some of the lower services (policy generation) were still using `NodeInfo`, but mapping if from and to `NodeCoreFact`, so I changed that part first to remove some more `NodeInfo`. 

For now, I'm keeping the commit separated, at least so that the part refactoring nodeInfo can be ignored.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -1237,22 +1237,25 @@ class WoLDAPDirectiveRepository(
...
-                                                )
-                                  } yield {
-                                    moved
-                                  }
-                                }
-                                .chainError("Error when trying to archive automatically the technique move")
+      autoArchive          <-
+        ZIO
+          .when(
+            autoExportOnModify && !moved.isInstanceOf[LDIFNoopChangeRecord] && !movedActiveTechnique.policyTypes.isSystem
```
fanf, line 1243: it's just because that line is longer, nothing else changed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -1291,17 +1294,18 @@ class WoLDAPDirectiveRepository(
...
-                                commiter <- personIdentService.getPersonIdentOrDefault(actor.name)
-                                archive  <- gitATArchiver.archiveActiveTechnique(
-                                              newactiveTechnique,
-                                              parents.map(_.id),
-                                              Some((modId, commiter, reason))
-                                            )
-                              } yield archive
-                            }
+      autoArchive        <-
+        ZIO.when(autoExportOnModify && !saved.isInstanceOf[LDIFNoopChangeRecord] && !newactiveTechnique.policyTypes.isSystem) {
```
fanf, line 1298: same

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDirectiveRepository.scala
```diff
@@ -1329,17 +1333,18 @@ class WoLDAPDirectiveRepository(
...
-                                commiter <- personIdentService.getPersonIdentOrDefault(actor.name)
-                                archive  <- gitATArchiver.archiveActiveTechnique(
-                                              newActiveTechnique,
-                                              parents.map(_.id),
-                                              Some((modId, commiter, reason))
-                                            )
-                              } yield archive
-                            }
+      autoArchive        <-
+        ZIO.when(autoExportOnModify && !saved.isInstanceOf[LDIFNoopChangeRecord] && !newActiveTechnique.policyTypes.isSystem) {
```
fanf, line 1337: same

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -602,7 +602,12 @@ class LDAPEntityMapper(
         id                   <- e.required(A_ACTIVE_TECHNIQUE_UUID)
         refTechniqueUuid     <- e.required(A_TECHNIQUE_UUID).map(x => TechniqueName(x))
         isEnabled             = e.getAsBoolean(A_IS_ENABLED).getOrElse(false)
-        isSystem              = e.getAsBoolean(A_IS_SYSTEM).getOrElse(false)
+        policyTypes        = e(A_POLICY_TYPES) match {
+                                  case Some(json) => json.fromJson[PolicyTypes].getOrElse(PolicyTypes.rudderBase)
+                                  case None       =>
+                                    if (e.getAsBoolean(A_IS_SYSTEM).getOrElse(false)) PolicyTypes.rudderSystem
+                                    else PolicyTypes.rudderBase
+                                }
```
fanf, line 610: For compat, try to read policy Type, and if not here is system. When writting, we only need to persist policy type. 
At some point, we will be able to remove isSystem.

fanf:
PR updated with a new commit

VinceMacBuche:
Don't you think the policy type should be added in event logs ? you changed the is system logic, but maybe it would also need a new field policy type ?

VinceMacBuche:
Do you think we will remove all "isSystem" and replace by policy type ? I think that what we will end up to do

fanf:
> Don't you think the policy type should be added in event logs ? you changed the is system logic, but maybe it would also need a new field policy type ?
(or rules/directives/groupes/etc)

I wanted to keep the change as small as possible to enable the split of compliance by aspects. For that, it appears that it is not needed in anything but techniques (rudder) and their generation object in policies. 

I clearly see other possibilities for the tag for other kind of segregations, like filtering all aspect of rudder only for one `policyType` for a quick overview of what "touch" that policy aspect, or even limiting user capabilities for users by `policyType`. I don't see any limitation for that in the design, it's just out of scope and hard to clearly bound without an exact user story to fulfil. So I didn't do it at all. 

> Do you think we will remove all "isSystem" and replace by policy type ? I think that what we will end up to do

In the long term, `isSystem` isn't powerful enough to what we need, so yes, it looks like it will be removed. Still, a ìsSystem` may have some sense as "the minimum policies needed for rudder to run", but then, I'm not sure about what it brings compared to the flavor "system" for `policyTag`.

clarktsiory (commented):
Looks all good to me ! Except I don't know if we should keep the serialization of `PolicyTypes` as `{"types": [{"value":"rudder.system"}]}` or just `{"types": ["rudder.system"]}`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/policies/PolicyTypes.scala
```diff
@@ -0,0 +1,95 @@
...
+  val rudderBase:   PolicyTypes = PolicyTypes(NonEmptyChunk(PolicyTypeName.rudderBase))
+}
+
+/*
+ * Feature are tags with a name.
+ */
+final case class PolicyTypeName(value: String)
+
+object PolicyTypeName {
+  implicit val codecFeatureTagName: JsonCodec[PolicyTypeName] = DeriveJsonCodec.gen
```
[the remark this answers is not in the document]
fanf: no, I don't think there is. I think at first I explored an other mapping and just didn't come back to it after.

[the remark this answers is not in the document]
fanf: (I think it can't be anyval for compat with scala 3 for now, but it's a detail)

PR #5704, opened 2024-06-02
Fixes #24965: Remove rudder-pkg Python implementation

amousset:
https://issues.rudder.io/issues/24965

This removes the autocompletion which was already incorrect since 8.1. We need to rewrite it, the topic was already identified by @Fdall .


fanf (approved):
:tada:

PR #5706, opened 2024-06-03
Fixes #24851: Migrate NodeApi to zio-json

clarktsiory:
https://issues.rudder.io/issues/24851

Migrating the remaining endpoints of the nodes API : 
* test-driven migration : test lift-json endpoint then migrate to zio-json, then adapt tests (all changes were backward compatible in this PR) 
* there are many endpoint which expect fields in the request (e.g. `nodeId` filter) in `json` or `params` form, so we need a case class and two extract methods in `ZioJsonExtractor` for each request (`JQNodeId`)
* most endpoints used the lift-json DSL to build a new structure that isn't reused anywhere, so we create a case class for the response (`JRNodeX`)
* sometimes the API schema does not follow the public structure (instead of `{"action":"get","data":{...}`, the endpoint just returns `{...}`) and therefore we cannot use the `toLiftResponseOne` but we have to build the lift response by hand (it is often ~3 lines)


clarktsiory:
PR updated with a new commit

webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/DataTypes.scala
```diff
@@ -130,6 +130,30 @@ object SecurityToken {
...
+    PureResult
+      .attempt(
+        new PEMParser(new StringReader(key))
+          .readObject()
+      )
+      .left
+      .map(ex => InventoryError.CryptoEx(s"Key '${key}' cannot be parsed as a public key", ex.cause): InventoryError)
+      .flatMap { obj =>
+        obj match {
+          case _: SubjectPublicKeyInfo  => Right(PublicKey(key))
```
fanf, line 145: we only support certificate now, but it's ok to do it that way - the removing can be an other issue (even in an other version of rudder)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
```diff
@@ -55,11 +55,9 @@ import com.softwaremill.quicklens.*
 import com.typesafe.config.ConfigValue
 import enumeratum.*
 import java.util.regex.Pattern
-import net.liftweb.json.JArray
-import net.liftweb.json.JField
-import net.liftweb.json.JObject
-import net.liftweb.json.JString
-import net.liftweb.json.JValue
```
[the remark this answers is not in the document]
fanf: :)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
```diff
@@ -237,6 +217,26 @@ object Creation {
...
+    implicit val onSaveNodeEncoder:          JsonEncoder[OnSaveNode]                     = JsonEncoder[String].contramap(_.errorMsg)
+    implicit val onAcceptationEncoder:       JsonEncoder[OnAcceptation]                  = JsonEncoder[String].contramap(_.errorMsg)
+
+    // we need to redefine this because we don't want a discriminator, and also the JSON subtype are not the same
+    implicit val encoder: JsonEncoder[CreationError] = new JsonEncoder[CreationError] {
+      override def unsafeEncode(a: CreationError, indent: Option[Int], out: Write): Unit = a match {
+        case v: OnValidation    => JsonEncoder[OnValidation].unsafeEncode(v, indent, out)
+        case v: OnSaveInventory => JsonEncoder[OnSaveInventory].unsafeEncode(v, indent, out)
+        case v: OnSaveNode      => JsonEncoder[OnSaveNode].unsafeEncode(v, indent, out)
+        case v: OnAcceptation   => JsonEncoder[OnAcceptation].unsafeEncode(v, indent, out)
```
fanf, line 237: Why not  `case v: OnAcceptation   => onAcceptationEncoder.unsafeEncode(v, indent, out)` ? 
Since they are just above ?

fanf (commented):
Seems good and working as intented. Just: why not using the encoder you just define? Is there a reason ? The implementation is just calling the object, so it seems to be technically the same, it just doesn't seem as obvious ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/CreateNodeData.scala
clarktsiory: I'm just used to summon the implicit like this whenever possible :grin: 
Because in scala 3 some day it may become nameless : `given JsonEncoder[OnAcceptation]` (even now in scala 2, it can be derived using `@jsonDerive` macro if we want)

clarktsiory:
PR rebased : I had to move tests from `api_node_properties.yml` to `api_nodes.yml` to make them pass since https://github.com/Normation/rudder/pull/5717

clarktsiory:
OK, squash merging this PR

PR #5712, opened 2024-06-06
Fixes #24472: We should rework the targets rules page when we create a directive

RaphaelGauthier:
https://issues.rudder.io/issues/24472

I think this is the simplest solution for bringing the filters back on the target rules (starting with the search bar), while waiting for this interface to be completely rewritten in Elm.

I removed all the dead code linked to the angular application that managed the filters, and replaced it with a simple javascript function, adding an onkeyup event to the search input :

target-rules-search


VinceMacBuche (approved):

RaphaelGauthier:
OK, merging this PR

PR #5713, opened 2024-06-06
Fixes #24971: Search for machine type (virtual, physical) does not work

fanf:
https://issues.rudder.io/issues/24971

So, the root cause of the bug is that we weren't comparing something that existed in 8.1. 
In 8.0 and before, the LDAP attribute for machine type is multivalued and has both "physical/virtual" + the details for vm. 

In 8.1, it directly uses the scala object "kind", which is always more precise that virtual. 
Fortunatly, it's easy to write a dedicated comparator for that. I also wrote the one for the VM kind, which was sort of missing ? (the ldap logic was here, not the corresponding entry... Not sure about what happened). 

Also, it doesn't make sense anymore to put the display name of these VM in a separated file, since now they are scala object, it can just be there. 

Then, for test, I changed our data set, which a lot of impact all around the code. 

And finally, I needed to rewrite the json decoder for machine type, else we wouldn't be able to decode machine that were encoded with the incorrect type name (we must keep the query one, since it can be used group definition).


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
```diff
@@ -101,7 +104,8 @@ class NodeQueryCriteriaData(groupRepo: () => SubGroupComparatorRepository) {
     ObjectCriterion(
       OC_MACHINE,
       Chunk(
-        Criterion("machineType", MachineComparator, NodeCriterionMatcherString(_.machine.machineType.kind.wrap)),
+        Criterion("machineType", MachineComparator, MatchineTypeCriterionMatcherCaseIgnoreString),
```
fanf, line 107: that's the actual bug correction

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
```diff
@@ -1924,14 +1924,21 @@ object NodeFactSerialisation {
...
-    )
+    implicit val codecMachineType:    JsonCodec[MachineType]    = {
+      val encoder: JsonEncoder[MachineType] = JsonEncoder.string.contramap(_.kind)
+      // for compat before correction of https://issues.rudder.io/issues/24971, we need to keep
+      // "physicalMachine". "virtualMachine" wasn't directly used
+      val decoder: JsonDecoder[MachineType] = JsonDecoder.string.map {
+        case s =>
+          s.toLowerCase match {
+            case UnknownMachineType.kind                      => UnknownMachineType
+            case PhysicalMachineType.kind | "physicalMachine" => PhysicalMachineType
```
clarktsiory, line 1935: compatibility would still be broken : `s.toLowerCase` may only be `physicalmachine` or the lowercased kind

clarktsiory: it's the same capitalization problem with the kind of unknown machine type which is still `unknownMachineType`...
Maybe it's the `.toLowerCase` that should be removed

fanf: right, there's a missing toLowerCase / one too many

PR #5717, opened 2024-06-08
Fixes #24985: Migrate rest test from yaml file to zio-test

fanf:
https://issues.rudder.io/issues/24985

So, it was unbeliebaly hard to get something that is not horrible to use. 
Porting the test by itself was OK. I needed to merge some test files because I'm not totally understanding how sequencing works, and so for test whose modification were depending of change in previous tests, I got wrong results. 

It tried to make the diff work in intellij, but it's just extremely brittle, and actually zio-test produce a text trace that add a lot of garbages in the diff view (or make it fail). 
In the end, I finaly found the diffing feature of zio-test "smart assertion" (by finding the bounty for implementing in normal assert)[https://github.com/zio/zio/issues/8664], which is good enought: I used it a lot with a rather enjoyable DX. 
image

The reason why I used it a lot is because I was forced to normalize the json string through a `fromJson.toJson` sequence for both expected and obtained result, else I was not able to get any correct comparison. 

I'm not sure I'm totally happy with that, because it means that we will always go throught a JSON encode/decoder, and that it can hide a bug in them (for example if it ignore one part of the expected output). 
So perhaps it will need more change to really be able to compare the two string at some point. 

Notable points:

-  I had to delete `updatePackageManagementRevision`  because it was causing the git repository to not find a technique with a commit. But it's not used in our current tests, and I tried to remove it in master without any impact... So it seems that the more strict init of zio-test reveals a bad behavior that was hidden before. 

- `TraitTestApiFromYamlFiles` is not a trait anymore, but just an object with method with more parameters than before. Extending it was causing strange "cannot read private field in public class" error in intellij (but not in cli), and I'm not sure it was because of some specifities of intellij, cache, or something else, and I lost patience. 

- other changes will be documented directly in code


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RestTestSetUp.scala
```diff
@@ -1053,6 +1034,40 @@ class RestTest(liftRules: LiftRules) {
...
+        .map(_.apply() match {
+          case Full(a) => Full(LiftRules.convertResponse((a, Nil, S.responseCookies, req)))
+          case r       => r
+        }) match {
+        case Full(x) => tests(x)
+        case eb: EmptyBox => tests(eb)
+      }
+    }
+  }
+
```
fanf, line 1070: these two methods are just the copy/paste of the corresponding methods for specs, but for zio-test.

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RestTestSetUp.scala
```diff
@@ -978,26 +979,6 @@ class RestTestSetUp {
...
-    val mod  = orig.replaceAll("""name="Package version" """, """name="AN OTHER REVISION OF TECHNIQUE PACKAGE MANAGEMENT" """)
-
-    metadata.write(mod)
-    mockGitRepo.gitRepo.git.add().setUpdate(true).addFilepattern(".").call()
-    mockGitRepo.gitRepo.git.commit().setMessage("new revision of packageManagement/1.0 technique").call()
-    metadata.write(orig)
-    mockGitRepo.gitRepo.git.add().setUpdate(true).addFilepattern(".").call()
-    mockGitRepo.gitRepo.git.commit().setMessage("revert to original content for packageManagement/1.0 technique").call()
-  }
-
```
fanf, line 1000: Removed because unused and caused a problem with commit version (exhibited by zio-test init order ?)

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestFromFileDef.scala
```diff
@@ -78,18 +61,47 @@ class TestRestFromFileDef extends TraitTestApiFromYamlFiles with AfterAll {
...
+  val yamlDestTmpDirectory: File   = tmpApiTemplate
+
+  val transformations: Map[String, String => String] = Map(
+    ("api_revisions.yml" -> copyTransformApiRevision(restTestSetUp) _)
+  )
+
+  tmpApiTemplate.createDirectories()
+
+  override def spec: Spec[TestEnvironment with Scope, Any] = {
+    (suite("All REST tests defined in files") {
```
fanf, line 88: zio-test is quite nice to build suite of suite of test in code. Nothing else to do than that.

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestFromFileDef.scala
```diff
@@ -78,18 +61,47 @@ class TestRestFromFileDef extends TraitTestApiFromYamlFiles with AfterAll {
...
+      for {
+        s <- TraitTestApiFromYamlFiles.doTest(
+               yamlSourceDirectory,
+               yamlDestTmpDirectory,
+               restTestSetUp,
+               Nil,
+               transformations
+             )
+        _ <- effectUioUnit(
+               if (java.lang.System.getProperty("tests.clean.tmp") != "false") IOResult.attempt(restTestSetUp.cleanup())
```
fanf, line 99: there is also an "afterAll" annotation, but it feels rather natural to do like that

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TraitTestApiFromYamlFiles.scala
```diff
@@ -137,35 +132,22 @@ object TraitTestApiFromYamlFiles {
...
-  // the base temp directory where are stored yaml test files
-  // It can have a `templates` subdirectory where templates for transformation are store
-  // If you have template, that directory needs to be writable to allows to copy processed template there.
-  // You need to take care of the cleanup by yourself.
-  def yamlDestTmpDirectory: File
-  // transformation are pre-processing of template before running the test, typically to set api version, random things, etc
-  def transformations:      Map[String, String => String]
-
-  // the liftRules to use for API
-  def liftRules: LiftRules
```
fanf, line 156: Since we don't have trait anymore, these things are passed as additionnal function parameters

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TraitTestApiFromYamlFiles.scala
```diff
@@ -290,72 +276,118 @@ trait TraitTestApiFromYamlFiles extends Specification with BoxSpecMatcher with J
...
+                case (eb: EmptyBox, i) =>
+                  val f = eb ?~! s"I wasn't able to run the ${i}th tests in file ${name}"
+                  ApplicationLogger.error(f.messageChain)
+
+                  // erg. There must be a better way
+                  zio.test.test(
+                    s"[${i}] failing test ${i}: can not read description. The yaml description ${i} in ${name} cannot be read: ${f.messageChain}"
+                  )(
+                    assert(true)(Assertion.isFalse)
+                  )
```
fanf, line 337: I wasn't able to find a `test.failWithMessage(....)` method. I'm sure it exists somehow, but in the meantime, that will do.

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TraitTestApiFromYamlFiles.scala
```diff
@@ -290,72 +276,118 @@ trait TraitTestApiFromYamlFiles extends Specification with BoxSpecMatcher with J
...
-                  restTest.execRequestResponse(mockReq)(response => {
-                    val responseBox = response.map(cleanResponse(_))
-                    responseBox.map(_._1) must equalsBoxStrict(test.responseCode)("response code") and (
-                      responseBox.map(_._2) must equalsBoxJson(test.responseContent)("response content")
-                    )
-                  })
-                }
-            }
+        // we can't have "." in test description, it breaks junit test runner in intellij
+        (suite(s"Tests defined in '${name.split("\\.").head}' yaml") {
```
fanf, line 323: a suite of tests for each yaml file

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TraitTestApiFromYamlFiles.scala
```diff
@@ -290,72 +276,118 @@ trait TraitTestApiFromYamlFiles extends Specification with BoxSpecMatcher with J
...
+
+                  // erg. There must be a better way
+                  zio.test.test(
+                    s"[${i}] failing test ${i}: can not read description. The yaml description ${i} in ${name} cannot be read: ${f.messageChain}"
+                  )(
+                    assert(true)(Assertion.isFalse)
+                  )
+
+                case (Full(test), i) =>
+                  zio.test.test(s"[${i}] ${test.description}") {
```
fanf, line 340: this is the way to declare a test. A test for each request.

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TraitTestApiFromYamlFiles.scala
```diff
@@ -290,72 +276,118 @@ trait TraitTestApiFromYamlFiles extends Specification with BoxSpecMatcher with J
...
+  }
+
+  def cleanJson(json: String): Either[String, String] = {
+    import zio.json.*
+    import zio.json.ast.*
+    json.fromJson[Json].map(_.toJsonPretty) match {
+      case Left(err)   =>
+        // here, we need to take care of the case of a pure string. Rule of thumb: if the first char is not json-ok, fallback
+        err match {
+          case "(unexpected 'O')" => Right(json)
```
fanf, line 387: in that case, we assume it's not json, but just a simple string. Return it as is.

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TraitTestApiFromYamlFiles.scala
```diff
@@ -290,72 +276,118 @@ trait TraitTestApiFromYamlFiles extends Specification with BoxSpecMatcher with J
...
+                        clean        <- response.map(cleanResponse).toPureResult.left.map(_.fullMsg)
+                        expectedJson <- cleanJson(test.responseContent)
+                        responseJson <- cleanJson(clean._2)
+                      } yield (clean._1, expectedJson, responseJson)) match {
+                        case Left(s)                         =>
+                          assert(s)(Assertion(TestArrow.make[String, Boolean](s => TestTrace.fail(s))))
+                        case Right((code, expJson, resJons)) =>
+                          assertTrue(
+                            code == test.responseCode,
+                            resJons == expJson
```
fanf, line 366: so, this is magic assert that will do the correct diffing

fanf:
PR rebased

VinceMacBuche (approved):

PR #5724, opened 2024-06-13
Fixes #25008: When node compliance right is missing, we don't want a red error pop-up

fanf:
https://issues.rudder.io/issues/25008

So, when someone does not have the compliance rights, we want to display a nice message in place of the red frightening pop-up. 
Still, we want that someone with the `node_read` right is able to see the "system status" tab, since that compliance is not really compliance, it's just "how rudder is working on that node" - so totally what `node_read` should give access to. 

So with the correction and a `node_all` right, you now get: 

For compliance: 
image

For system status:
(same as before)
image

Code is just an added "if" and a new API right for the system status case.


clarktsiory (approved):
LGTM !

fanf:
OK, merging this PR

PR #5725, opened 2024-06-13
Fixes #24988: Property inheritance of type array doesn't work

clarktsiory:
https://issues.rudder.io/issues/24988

When having an inheritance chain with more than 1 parent/child relationship, properties from child are being merged without really taking the global parameter inherit mode into account...

N.B. : This only fixes the inheritance when there is a global parameter along the chain. By default, the inherit mode for arrays is "override", so it makes sense


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM, correction is quite elegant (and small!)

Thanks for the added unit test, it helped understand

clarktsiory:
OK, squash merging this PR

PR #5735, opened 2024-06-18
Fixes #25035: Process are not saved anymore in 8.1

fanf:
https://issues.rudder.io/issues/25035

So, the problem is that #3899 changed the way processes were stored without changing it in `FullInventory.save`. 
So this was working for new inventories using the `DefaultInventoryAccept` workflow. 
But in 8.1, we changed the way node inventory are persisted to always use  `NodeFactStorage` and thus always using `FullInventoryRepo.save` which didn't have the logic to save processes. 

I added a configurable threshold to switch the behavior between one write request or two write request (many processess), since most of the time, it will be much more efficient to have only one write request. 
The threshold choice at 1000 is the result of a fair negociation with @VinceMacBuche.

~The property is hidden by default, because no user should have a need to know about it, ever. ~ added in `properties.sample`


clarktsiory (approved):
LGTM ! But the new property `inventory.threshold.processes.isolatedWrite` likely needs to be added in the sample properties file at this place : https://github.com/Normation/rudder/blob/branches/rudder/8.1/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample#L264

fanf:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

PR #5743, opened 2024-06-21
Fixes #25050: 

clarktsiory:
https://issues.rudder.io/issues/25050

Escape a bunch of JS strings. 

We need to escape strings and will use the well tested `common-text` `StringEscapeUtils` for it. 

* add dependency to `commons-text` 
* remove use of `scala.xml.Utility.escape` since it is not really maintained anymore and is redondant with new dependency
* use `escapeHtml4` or `escapeEcmaScript` based on use case
* check that ALL JsRaw are OK, and escaped strings that weren't. 
 
There is a non negligible chance that it breaks something somewhere, but a first batch of tests (going to node, directive, rules, group, parameter, clicking on things, adding technique and directive...) didn't show any suspicious behavior nor error in console. Node inventories, too. 

Tested with change validation plugin, too, to check that adapted validation pop-up are still ok.

PR #5748, opened 2024-06-25
Fixes #25050: 

fanf:
https://issues.rudder.io/issues/25050

Supersed https://github.com/Normation/rudder/pull/5743

https://issues.rudder.io/issues/25050

Escape a bunch of JS strings. 

We need to escape strings and will use the well tested `common-text` `StringEscapeUtils` for it. 

* add dependency to `commons-text` 
* remove use of `scala.xml.Utility.escape` since it is not really maintained anymore and is redondant with new dependency
* use `escapeHtml4` or `escapeEcmaScript` based on use case
* check that ALL JsRaw are OK, and escaped strings that weren't. 
 
There is a non negligible chance that it breaks something somewhere, but a first batch of tests (going to node, directive, rules, group, parameter, clicking on things, adding technique and directive...) didn't show any suspicious behavior nor error in console. Node inventories, too. 

Tested with change validation plugin, too, to check that adapted validation pop-up are still ok.


fanf:
OK, merging this PR

PR #5751, opened 2024-06-27
Fixes #25053: Error 500 when we visit for the first time the campaign page

VinceMacBuche:
https://issues.rudder.io/issues/25053

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignRepository.scala
@@ -74,21 +74,25 @@ class CampaignRepositoryImpl(campaignSerializer: CampaignSerializer, path: File,
     extends CampaignRepository {
 
   def getAll(typeFilter: List[CampaignType], statusFilter: List[CampaignStatusValue]): IOResult[List[Campaign]] = {
-    for {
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignRepository.scala
```diff
@@ -74,21 +74,25 @@ class CampaignRepositoryImpl(campaignSerializer: CampaignSerializer, path: File,
...
-                              } yield {
-                                c
-                              }).either.chainError("Error when getting all campaigns from filesystem")
-                            })
-      (errs, campaignRes) = campaigns.partitionMap(identity)
-      _                  <- ZIO.foreach(errs)(err => CampaignLogger.error(err.msg))
-      filteredCampaign    = Campaign.filter(campaignRes, typeFilter, statusFilter)
-    } yield {
-      filteredCampaign
+    if (path.exists) {
```
clarktsiory, line 77: Is the problem that the path does not exist only when the method is called ? I wonder how is the `path` created, and if this solution won't just always return `Nil`...

VinceMacBuche: the path is created when a campaign is created

VinceMacBuche:
Commit modified

PR #5754, opened 2024-06-29
Fixes #25087: Remove unexpected interpretation choice for vars because it's not meaningful anymore

fanf:
https://issues.rudder.io/issues/25087

This is all automatic changes from the deletion of `UnexpectedReportBehavior`


clarktsiory (approved):
LGTM !

PR #5756, opened 2024-07-01
Fixes #25091: Add VARIABLE_NAME in technique parameters so that we can use variable real name

VinceMacBuche:
https://issues.rudder.io/issues/25091

### DIFF
--- a/policies/rudderc/src/backends/metadata.rs
+++ b/policies/rudderc/src/backends/metadata.rs
@@ -96,6 +96,7 @@ impl From<Parameter> for Input {
 
         Self {
             name: p.id.to_string().to_uppercase(),
+            variable_name: p.name.clone()
             description: p.description.unwrap_or(p.name),
             long_description: p.documentation,
             constraint: Constraint {
@@ -135,6 +136,7 @@ impl From<Parameter> for SelectOne {
 
         Self {
             name: p.name.clone(),
+            variable_name: p.name.clone(),
             description: p.description.unwrap_or(p.name),
             long_description: p.documentation,
             item: items,
@@ -375,11 +377,12 @@ struct SectionInput {
 struct Input {
     // actually, the id
     name: String,
+    #[serde(rename = "VARIABLE_NAME")]
     // actually, the name
+    variable_name : String,
     description: String,
     #[serde(rename = "LONGDESCRIPTION")]
     #[serde(skip_serializing_if = "Option::is_none")]
-    // actually, the description
     long_description: Option<String>,
     constraint: Constraint,
 }
@@ -389,11 +392,12 @@ struct Input {
 struct SelectOne {
     // actually, the id
     name: String,
+    #[serde(rename = "VARIABLE_NAME")]
     // actually, the name
+    variable_name : String,
     description: String,
     #[serde(rename = "LONGDESCRIPTION")]
     #[serde(skip_serializing_if = "Option::is_none")]
-    // actually, the description
     long_description: Option<String>,
     item: Vec<SelectItem>,
     constraint: Constraint,
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
@@ -179,6 +179,7 @@ sealed trait VariableSpec {
   type V <: Variable // type of variable linked to that variable spec
 
   def name:            String
+  def variableName:    Option[String]
   def description:     String
   def longDescription: String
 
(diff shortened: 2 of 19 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/CfclerkXmlConstants.scala
```diff
@@ -110,6 +110,7 @@ object CfclerkXmlConstants {
   val VAR_NAME                        = "NAME"
   val VAR_DESCRIPTION                 = "DESCRIPTION"
   val VAR_LONG_DESCRIPTION            = "LONGDESCRIPTION"
+  val VAR_VARIABLE_NAME               = "VARIABLE_NAME"
```
fanf, line 113: I would use `VARIABLENAME` to be consistent with other properties here. Or change for underscore everywhere.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/xmlparsers/VariableSpecParser.scala
```diff
@@ -190,6 +191,12 @@ class VariableSpecParser extends Loggable {
                 )
               }
             }
+            variableName   <- (elt \ VAR_VARIABLE_NAME).toList match {
+                                case var_name :: Nil => Right(Some(var_name.text))
```
fanf, line 195: this is not conventional scala naming scheme, please use `varName`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -117,14 +117,15 @@ object BuildBundleSequence {
...
     def quote(agentEscape: String => String): String
-    def name:  String
-    def value: String
+    def name:         String
+    def value:        String
+    def variableName: Option[String]
   }
   object BundleParam       {
-    final case class SimpleQuote(value: String, name: String) extends BundleParam {
+    final case class SimpleQuote(value: String, name: String, variableName: Option[String]) extends BundleParam {
```
fanf, line 125: why not always using it ? Is it important to use the ID on linux (I understand it is important to use the name on windows)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/domain/VariableAndSectionSpec.scala
```diff
@@ -248,10 +250,10 @@ final case class TrackerVariableSpec(
   override type T = TrackerVariableSpec
   override type V = TrackerVariable
 
-  override val name:        String = TRACKINGKEY
-  override val description: String = "Variable which kept information about the policy"
-
-  override val checked: Boolean = false
+  override val name:         String         = TRACKINGKEY
+  override val description:  String         = "Variable which kept information about the policy"
+  override val variableName: Option[String] = None
```
fanf, line 255: there's something I don't get here: we have : 
- name
- description
- variableName
- longDescription
- id

But we only have in yaml / metadata: 
- id
- name
- description
- documentation

Mapping from/to longdescription <-> documentation is clear ; but I don't get the remaining ones. 

I'm almost sure `name` actually contains `id` (always). And then should be named `id`. 
I'm not sure about what contains `id: Option[String]` nor name. 
And I think there is a difference between a "display name" which is perhaps a description ?

fanf:
I think I'm really missing a schema or document that list the names of all things at all level and evolution proposed here. Especially around usage of id/variable name/display name/description at different places. 

The categories I'm interested in are: 
- the yml technique
- the technique editor UI
- the metadata.xml and techniques (cf, ps1)
- the "parameters" element of technique in directive/technique API and json directive
- the directive UI
- the directive serialisation in LDAP
- the directive serialisation in XML

And not sure it matters (I think it's more variable value than name here):
- expected reports
- the resulting reports in db

And perhaps other things ?

Ideally, it would a unit test, but at least the part from yaml to metadata would be hard to do (perhaps other part are at the frontier of systems, but at least they are mostly in the same language)

policies/rudderc/src/backends/metadata.rs
```diff
@@ -389,11 +392,12 @@ struct Input {
 struct SelectOne {
     // actually, the id
     name: String,
+    #[serde(rename = "VARIABLE_NAME")]
```
amousset, line 395:
```suggestion
```

No need, everything is already uppercased by line `#[serde(rename_all = "UPPERCASE")]` on the struct

policies/rudderc/src/backends/metadata.rs
```diff
@@ -375,11 +377,12 @@ struct SectionInput {
 struct Input {
     // actually, the id
     name: String,
+    #[serde(rename = "VARIABLE_NAME")]
```
amousset, line 380:
```suggestion
```

fanf:
Replaced by rudderc/src/backends/metadata.rs:523:56

PR #5755, opened 2024-06-29
Fixes #25088: Remove unused methods in ReportingService

fanf:
https://issues.rudder.io/issues/25088

Some well needed cleaning in `ReportingService` (first step, other will follow): 

**Getting only user/system compliance**

A lot of code in reporting service is based on the pattern : 

``` 
rules <- getRules(system = false) // only user rules
compliance <- computeComplianceForNodes(limitToRules = rules.map(_.id))
```

Or the same thing for system rule, getting all rules then diffing. 
This is not needed anymore since we have now a `Map[PolicyTypeName, AggregatedStatusReport]` now in `NodeStatusReport`. 

**Class methods when they can be object ones**

`ReportingService` trait has several method that are not linked to service dependencies and can be moved to its companion object. 

**Unused methods/classes**

Some methods were just unused anywhere: `complianceByNode`, `systemComplianceByNode`, `NodeCompliance`, `NodeSystemCompliance`.

PR #5737, opened 2024-06-18
Fixes #24652: Rudder 8.1 slows down over time

fanf:
https://issues.rudder.io/issues/24652

PR #5757, opened 2024-07-01
Fixes #25092: Refactor CachedReportingService to make persistance simpler

fanf:
https://issues.rudder.io/issues/25092

So, it seems that a lot of things are going on here, but that's not so much: 
- the first two commits are linked to pr #5754  and #5755 . So there's only the third commit to consider. 
- that should not change anything on Rudder behavior (bravo François, great commit, does nothing :facepalm: )

This commit splits the big `CachedReportingServiceImpl` into four parts:
- a dumb `NodeStatusReportRepository` that only store/retrieve NodeStatusReports. Also add a simple
  implementation in memory. Latter, we will be able to persist in a psql backend storage
- a new extremely simple `ReportingService` using previous `NodeStatusReportRepository`. With that and the availability of `PolicyType`, it's now apparent that that service is just syntaxic sugor on top of filtering `NodeStatusReports`. 
- a `ComputeNodeStatusReportService`  that get the logic with the queue but change nothing in it for now. It updates `NodeStatusReportRepository` with its updates. Latter on, we will be able to add a check on expiration based on node parameter about how long to keep expired compliance. 
- a `FindNewNodeStatusReports` that only has the logic to retrieve information for new NodeStatusReports
  in base from expected reports and run. The fact that it is now independant from `RepositoryService` method signature allows to show that `RuleId`, `DirectiveId` and `PolicyType` paramters are never used in it. We will be able to simplify it even further more latter on (his only job should be to retieve `RunAndConfigInfo`, and let `ComputeNodeStatusReportService` do the computing). 

For now, we keep the same init logic than we had before the refactoring: there's a bootstrap action that expires all node compliance, so they are all computed back which now init the in-memory `NodeStatusReportRepository`. Of course once we have that in base, we will get it from there, only asking for the missing nodes if any. 

Then, the commit adapt signatures where `ReportingService` was used. 

It alos remove `ReportingServiceTest` which was only comments (very useful) ; and adapt `CachedReportingServiceTest` to show that things still work when splitted in several pieces. One value change in that test because we don't exactly have the same call chain: now, the call to trigger a computation is explicit, and the chain is smarter on what should be updated (following https://issues.rudder.io/issues/24652 / https://github.com/Normation/rudder/pull/5737)

All in all, the new code is much, much simpler and get ride of a lot of oranically-grown-not-refactored code ; and it will be rather easy to add the missing parts for persisting and not expiring compliance.


clarktsiory (approved):
LGTM ! Nice and clear refactoring

PR #5763, opened 2024-07-05
Fixes #25111: 

clarktsiory:
https://issues.rudder.io/issues/25111
To summarize : we must guarantee that anything within `${}` is replaced when generating variables for Windows :
* with support for path accessors (`${path[foo]}`) : is parsed as an `NonRudderVar` (the class is changed to now support the "accessors")
* by parsing any other interpolation with "unknown" syntax or characters into an `UnsafeRudderVar` which is disallowed on Windows and mapped to previous `CharSeq` behavior to keep compatibility with the previous behavior for Linux

As for why we need to reconsider the `NonRudderVar` interpolation : 
* a `${prefix.var[foo]}` was never mapped to a `NonRudderVar` because of the CFEngine indexing syntax (the `foo` accessor). Along with that, as soon as there is an "unknown" character in an `${prefix.var...}` interpolation, the expression is parsed as a `CharSeq`, and is never replaced but used as is when generating policies for Linux. This is due to the change motivated by this PR : https://github.com/Normation/rudder/pull/2946. But it will cause problems on a Windows agent...
* we need to map the CFEngine-like indexing syntax to Nustache paths for Windows (`vars.prefix.var.foo`), so we need to parse the `List("foo")` accessor in the data type
* we also want to specifically disallow `.` in accessors 
* we need to disallow the `UnsafeRudderVar`, which is "a `NonRudderVar` with unsupported syntax"

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
@@ -39,16 +39,23 @@ package com.normation.rudder.ncf
 
 import better.files.File
 import cats.data.NonEmptyList
+import cats.syntax.functor.*
 import com.normation.cfclerk.domain.ReportingLogic
-import com.normation.errors.Inconsistency
-import com.normation.errors.IOResult
-import com.normation.errors.PureResult
-import com.normation.errors.Unexpected
+import com.normation.errors.*
 import com.normation.inventory.domain.AgentType
 import com.normation.inventory.domain.Version
 import com.normation.rudder.domain.policies.PolicyMode
 import com.normation.rudder.ncf.Constraint.CheckResult
 import com.normation.rudder.ncf.Constraint.Constraint
+import com.normation.rudder.services.policies.PropertyParser
+import com.normation.rudder.services.policies.PropertyParserTokens.CharSeq
+import com.normation.rudder.services.policies.PropertyParserTokens.NodeAccessor
+import com.normation.rudder.services.policies.PropertyParserTokens.NonRudderVar
+import com.normation.rudder.services.policies.PropertyParserTokens.Param
+import com.normation.rudder.services.policies.PropertyParserTokens.Property
+import com.normation.rudder.services.policies.PropertyParserTokens.RudderEngine
+import com.normation.rudder.services.policies.PropertyParserTokens.Token
+import com.normation.rudder.services.policies.PropertyParserTokens.UnsafeRudderVar
 import java.util.regex.Pattern
 import zio.ZIO
 import zio.json.SnakeCase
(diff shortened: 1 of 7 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndGlobalParameterLookup.scala
```diff
@@ -374,24 +399,22 @@ class TestNodeAndGlobalParameterLookup extends Specification {
...
-      val s2 = " alkjf fm ^{i àié" // will be split at '$'
-      val s3 = "${rudde ut ùt "
+      val s1 = "plj jmoji h imj " // 16
+      val s2 = " alkjf fm ^{i àié$rudde ut ùt ending"
       test(
         all(_),
-        s1 + "${rudder.node.policyserver.id}" + s2 + s3,
-        List(CharSeq(s1), NodeAccessor(List("policyserver", "id")), CharSeq(s2), CharSeq(s3))
+        s1 + "${rudder.node.policyserver.id}" + s2,
+        List(CharSeq(s1), NodeAccessor(List("policyserver", "id")), CharSeq(s2))
```
clarktsiory, line 407: SPEC has changed : since we changed the parser and introduced a `UnsafeRudderVar` parser, we consider that `${` should always be followed by the enclosing `}`, or else we will end up with a parsing error

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
```diff
@@ -236,17 +247,70 @@ object ParameterType {
...
+
+    private def translateDscHereString(value: String): PureResult[String] = {
+      // Format to DSC specific expression for a known Rudder path variable.
+      def mkDatastate(path: List[String]) = s"""([Rudder.Datastate]::Render('{{' + @'
+                                               |vars.${path.mkString(".")}
+                                               |'@ + '}}'))""".stripMargin
+
+      // Refuse "dots" in individual path keys for DSC. For now, assume we only have names without interpolation itself in a path element
+      def validatedPath(path: List[String]): PureResult[Unit] = {
+        Either.cond(
```
fanf, line 271: oh nice, I didn't know about that one

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndGlobalParameterLookup.scala
```diff
@@ -286,8 +286,33 @@ class TestNodeAndGlobalParameterLookup extends Specification {
       test(all(_), """${something.cfengine}""", List(NonRudderVar("something.cfengine")))
     }
 
-    "parse a non valid cfengine variable as a string" in {
-      test(all(_), """${something . cfengine}""", List(CharSeq("${something . cfengine"), CharSeq("}")))
+    "parse a non rudder param variable with accessor" in {
+      test(all(_), """${something.cfengine[foo]}""", List(NonRudderVar("something.cfengine", List("foo"))))
```
fanf, line 290: Space/return are forbidden know ? I *think* some user had asked to them to be able to put long var on several lines, but I'm not sure and don't find it back. So I would say, not too grave (perhaps even a net win), but will need a note in release note.

fanf:
LGTM, will just wait for knowing what are the changes in matching plugin before merging.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala
```diff
@@ -304,7 +299,7 @@ object ParameterType {
           validatedPath(method).as(mkDatastate("data" :: engine :: method))
       }
 
-      def assembleDscExpr(tokens: List[Token]) = tokens.traverse(translateToken).map {
+      def assembleDscExpr(tokens: List[Token]) = tokens.accumulatePure(translateToken).map {
```
clarktsiory, line 302: We would want to catch all errors if there are more than one

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
```diff
@@ -848,9 +848,10 @@ class PolicyWriterServiceImpl(
       val writeAgent = for {
         // changing `writeAllAgentSpecificFiles.write` to IOResult breaks DSC
         t0 <- currentTimeNanos
-        _  <- IOResult
-                .attempt(writeAllAgentSpecificFiles.write(prepared))
+        _  <- writeAllAgentSpecificFiles
+                .write(prepared)
                 .chainError(s"Error with node '${prepared.paths.nodeId.value}'")
+                .toIO
```
clarktsiory, line 854: The `write` method should not throw at all, all implementations have already been rewritten with `PureResult` instead

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndGlobalParameterLookup.scala
clarktsiory: They are parsed as `UnsafeRudderVar` [here](https://github.com/clarktsiory/rudder/blob/bug_25111/_/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestNodeAndGlobalParameterLookup.scala#L315), which are supported on Linux (old behavior is kept), and not supported in Windows !

fanf (approved):
Perfect! Nice refactoring out of Box, too.

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, squash merging this PR

clarktsiory:
This has been merged with commit 293abba1387b3da9537758824a0d9e6349d1fef3 on master

PR #5764, opened 2024-07-05
Fixes #25032:  Use Content-Security-Policy strict headers in utilities pages

clarktsiory:
https://issues.rudder.io/issues/25032

* :bulb: Replace the regex-based filtering of CSP headers, and use a lift directive instead : `data-lift="with-enabled-csp"` to include in every HTML page for which to enforce strict-CSP
* :hammer: Migrate scripts in Archives page
* :hammer: Migrate scripts in Event logs page
*  :hammer: Reuse the lift directive in the Healthcheck page

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/Archives.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/Archives.scala
@@ -48,6 +48,7 @@ import com.normation.rudder.git.GitArchiveId
 import com.normation.rudder.git.GitCommitId
 import com.normation.rudder.repository.*
 import com.normation.rudder.users.CurrentUser
+import com.normation.rudder.web.snippet.WithNonce
 import com.normation.utils.DateFormaterService
 import net.liftweb.common.*
 import net.liftweb.http.*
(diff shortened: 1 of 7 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/Archives.scala
```diff
@@ -383,24 +384,28 @@ class Archives extends DispatchSnippet with Loggable {
...
-            """enableIfNonEmpty("%s", "%s");$("#%s").prop("disabled",true);"""
-              .format(archiveDateSelectId, restoreButtonId, restoreButtonId)
-          ) // JsRaw ok, all const values of actionFormBuilder
-        )
+      WithNonce.scriptWithNonce(
+        Script(
+          OnLoad(
+            JsRaw(
+              """enableIfNonEmpty("%s", "%s");$("#%s").prop("disabled",true);"""
+                .format(archiveDateSelectId, restoreButtonId, restoreButtonId)
```
fanf, line 392: why keeping that format ? Interpolation wouldn't be clearer ?

fanf (approved):
It does the job, and the explicit control is good during migration. 
But it will be very easy to forget, so we should try the opposite in next version: 
- create a tag that disable csp checking
- put it on content that breaks with csp checking only (ie, enabled  by default). 

Given there is a *bunch* of other things to test in that version I think we're too late to do it now, but it can be done as one of the first step of 8.3 (and then use 8.3 dev cycle to add the "without-csp" tag where it is needed)

PR #5765, opened 2024-07-10
Fixes #25120: Use node properties to decide how long compliance is valid

fanf:
https://issues.rudder.io/issues/25120

This PR add the possibility to delay node compliance expiration based on a settings given via a node property. 
It implements the bottom service `ComplianceExpirationService` on the schema.

The main changes are: 

- add a new `KeepLastCompliance` run info case 
- change nothin in `ExecutionBatch` - the analysis of weither we keep or not last compliance when we have a `NoReport` is done latter on and is not based on a run/expect report analysis (obviously)
- add the logic to check is a node is configured to keep compliance in `ComputeNodeStatusService` based on `{"rudder": { "keep_compliance_duration": "duration" }}` node property
- the duration format is the standard one for scala: an integer followed by an unit, ie something like `2 days` or `48 hours`
- in the repo, actually atomically do the check is a past compliance exists and should be kept. Return modified compliance so that `scoringService` can be plugged and update with the real changes. 
    - note: that means that we kind of reverted #5737 but normally, the condition to do so are now correct

Globally, things remain simple and well separated, it's rather nice. 
Some comments added in the code for noticeable things.

Current limits: 
- the property MUST BE  at node level. It does take into account inherited properties. This will be addressed in a latter PR (it will be the `ResolvedPropertiesCache` or something alike on the schema)
- there is no cold persistance. So when rudder reboot, all old compliance that can't be computed again (because the corresponding run is too old and was deleted) are lost. This will be addressed in a latter PR (it will be the NodeLastCompliance DB connexion on the left of the schema, perhaps with an additionnal StorageService layer in between)

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/NodeComplianceExpiration.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/NodeComplianceExpiration.scala
@@ -1,0 +1,33 @@
+package com.normation.rudder.domain.reports
+
+import scala.concurrent.duration.Duration
+
+/*
+ * This file defines data structures to inform what we should display when a node compliance
+ * is expired (keep compliance for a longer time, etc)
+ */
+
+sealed trait NodeComplianceExpiration {
+  def id: String
+}
+
+object NodeComplianceExpiration {
+
+  /*
+   * Historical expiration based on a 2 agent run duration + a grace period.
+   * But if the node is in status `NoReportInInterval`, then display that
+   */
+  case object ExpireImmediately extends NodeComplianceExpiration {
+    val id = "expireImmediately"
+  }
+
+  /*
+   * Keep the last know compliance for given additional time.
+   * At some point, we will certainly make that property "by-policyType", but
+   * for now it is the same for all compliance (and it's ok since we are computed
+   * at the same time for now).
+   */
+  case class KeepLast(duration: Duration) extends NodeComplianceExpiration {
+    def id = "keepLast"
+  }
+}
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -106,12 +106,12 @@ object RunComplianceInfo {
   final case class PolicyModeInconsistency(problems: List[PolicyModeError]) extends RunComplianceInfo
 }
 
-final class NodeStatusReport private (
-    val nodeId:     NodeId,
-    val runInfo:    RunAndConfigInfo,
-    val statusInfo: RunComplianceInfo,
-    val overrides:  List[OverridenPolicy],
-    val reports:    Map[PolicyTypeName, AggregatedStatusReport]
+final case class NodeStatusReport(
+    nodeId:     NodeId,
+    runInfo:    RunAndConfigInfo,
+    statusInfo: RunComplianceInfo,
+    overrides:  List[OverridenPolicy],
+    reports:    Map[PolicyTypeName, AggregatedStatusReport]
 ) extends StatusReport {
   // for compat reason, node compliance is the sum of all aspects
   lazy val compliance: ComplianceLevel = ComplianceLevel.sum(reports.values.map(_.compliance))
@@ -132,14 +132,14 @@ final class NodeStatusReport private (
       case Some(r) => Map((t, r))
       case None    => Map.empty[PolicyTypeName, AggregatedStatusReport]
     }
-    new NodeStatusReport(nodeId, runInfo, statusInfo, overrides, r)
+    NodeStatusReport(nodeId, runInfo, statusInfo, overrides, r)
   }
 }
 
 object NodeStatusReport {
   // To use when you are sure that all reports are indeed for the designated node. RuleNodeStatusReports must be merged
   // Only used in `getNodeStatusReports`
-  def apply(
+  def buildWith(
       nodeId:     NodeId,
       runInfo:    RunAndConfigInfo,
       statusInfo: RunComplianceInfo,
@@ -155,7 +155,7 @@ object NodeStatusReport {
     )
     // group map and aggregate
     val aggregates = reports.groupBy(_.complianceTag).map { case (tag, reports) => (tag, AggregatedStatusReport(reports)) }
-    new NodeStatusReport(nodeId, runInfo, statusInfo, overrides, aggregates)
+    NodeStatusReport(nodeId, runInfo, statusInfo, overrides, aggregates)
   }
 
   /*
@@ -167,7 +167,7 @@ object NodeStatusReport {
    */
   def filterByRules(nodeStatusReport: NodeStatusReport, ruleIds: Set[RuleId]): NodeStatusReport = {
 
-    new NodeStatusReport(
+    NodeStatusReport(
       nodeStatusReport.nodeId,
       nodeStatusReport.runInfo,
       nodeStatusReport.statusInfo,
@@ -178,7 +178,7 @@ object NodeStatusReport {
 
   def filterByDirectives(nodeStatusReport: NodeStatusReport, directiveIds: Set[DirectiveId]): NodeStatusReport = {
 
-    new NodeStatusReport(
+    NodeStatusReport(
       nodeStatusReport.nodeId,
       nodeStatusReport.runInfo,
       nodeStatusReport.statusInfo,
@@ -534,6 +534,10 @@ object NodeStatusReportSerialization {
       case NoReportInInterval(e, _)                 =>
         (("type"              -> "NoReportInInterval")
         ~ ("expectedConfigId" -> e.nodeConfigId.value))
+      case KeepLastCompliance(e, exp, keep, r)      =>
+        (("type"              -> "KeepLastCompliance")
+        ~ ("expectedConfigId" -> e.nodeConfigId.value)
+        ~ ("runConfigId"      -> r.map(_._2.nodeConfigId.value)))
       case ReportsDisabledInInterval(e, _)          =>
         (("type"              -> "ReportsDisabled")
         ~ ("expectedConfigId" -> e.nodeConfigId.value))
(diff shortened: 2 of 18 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/NodeComplianceExpiration.scala
```diff
@@ -0,0 +1,33 @@
+package com.normation.rudder.domain.reports
+
```
fanf, line 2: this new class is the holder for the configuration setting for a node (based on what the property holds)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -106,12 +106,12 @@ object RunComplianceInfo {
   final case class PolicyModeInconsistency(problems: List[PolicyModeError]) extends RunComplianceInfo
 }
 
-final class NodeStatusReport private (
-    val nodeId:     NodeId,
-    val runInfo:    RunAndConfigInfo,
-    val statusInfo: RunComplianceInfo,
-    val overrides:  List[OverridenPolicy],
-    val reports:    Map[PolicyTypeName, AggregatedStatusReport]
+final case class NodeStatusReport(
```
fanf, line 109: I see not longer benefits to make that constructor private: the `reports` type is now different for when we need to compute it with the apply method and what is in the class. 

Plus, it will ease the usage.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -132,14 +132,14 @@ final class NodeStatusReport private (
...
-    new NodeStatusReport(nodeId, runInfo, statusInfo, overrides, r)
+    NodeStatusReport(nodeId, runInfo, statusInfo, overrides, r)
   }
 }
 
 object NodeStatusReport {
   // To use when you are sure that all reports are indeed for the designated node. RuleNodeStatusReports must be merged
   // Only used in `getNodeStatusReports`
-  def apply(
+  def buildWith(
```
fanf, line 142: impact of changing NodeStatusReport to a case class

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/execution/ReportsExecutionService.scala
```diff
@@ -144,13 +139,6 @@ class ReportsExecutionService(
     }
   }
 
-  def updateScore(nodeId: NodeId, report: NodeStatusReport): IOResult[Unit] = {
```
fanf, line 147: moved back in `ComputeNodeStatusReportService` which contradicts https://github.com/Normation/rudder/pull/5737 
It is done that way because we really want the queue to manage the I/O load, it is done FOR THAT. That's its point. And now, the services are better slipted appart, so the root cause of the problem in quoted PR should be resolved. 
Still, this will need a lot of attention and test in scaling machine.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeChangesService.scala
```diff
@@ -315,7 +315,9 @@ class CachedNodeChangesServiceImpl(
    * important than responsiveness.
    */
   def update(lowestId: Long, highestId: Long): Box[Unit] = {
-    addUpdate(ChangesUpdate.For(lowestId, highestId)).toBox
+    if (lowestId < highestId) {
```
fanf, line 318: So, this one is interesting and might be a genuine independant bug. We weren't checking for the case when nothing happened (lowestId == highestId), and so we were alway adding an empty event, and doing a lot of work for nothing. 
I'm more at ease to do that in 8.2 because everything is complicated in that part of Rudder, but it *could* be backported in 8.1 since it's a trivial change to revert in the worst case.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
```diff
@@ -98,11 +105,39 @@ class DummyNodeStatusReportRepository(initialReports: Ref[Map[NodeId, NodeStatus
 
   override def saveNodeStatusReports(
       reports: Iterable[(NodeId, NodeStatusReport)]
-  )(implicit cc: ChangeContext): IOResult[Unit] = {
+  )(implicit cc: ChangeContext): IOResult[Chunk[(NodeId, NodeStatusReport)]] = {
     ReportLoggerPure.Repository.debug(
       s"Add NodeStatusReports to repository for nodes: [${reports.map(_._1.value).mkString(", ")}]"
     ) *>
-    cache.update(_ ++ reports)
+    cache.modify { m =>
```
fanf, line 112: the part where we actually keep last computed latency if asked.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
```diff
@@ -122,6 +126,24 @@ trait ComputeNodeStatusReportService extends InvalidateCache[CacheComplianceQueu
 
 }
 
+trait NodeStatusReportUpdateHook {
+  def name:     String
+  def onUpdate: Iterable[(NodeId, NodeStatusReport)] => UIO[Unit]
+}
```
fanf, line 132: A class to handle result of saved NodeStatusReport changed - with below the only implem for now, for scoring service.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
```diff
@@ -241,6 +279,41 @@ class ComputeNodeStatusReportServiceImpl(
     })
   }
 
+  /*
+   * Filter out reports that are marked as "NoReportInInterval" but have a policy to keep them longer
+   */
```
fanf, line 284: this is the main logic that decides what needs to be done for reports that are in `NoRepertInInterval`

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/NodeComplianceExpiration.scala
clarktsiory: OK, does it need the license header ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComplianceExpirationService.scala
```diff
@@ -0,0 +1,120 @@
...
+      }.toMap)
+  }
+}
+
+object NodePropertyBasedComplianceExpirationService {
+  def getPolicy(p: NodeProperty, propKey: String, propName: String, debugId: NodeId): NodeComplianceExpiration = {
+    try {
+      // direct access to implementation details is so-so, but alternative force to go
+      // to json and is quite convoluted, and ConfigValue is horrible.
+      val v = p.config.getString("value." + propName)
```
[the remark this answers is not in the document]
clarktsiory: where does the "value" string come from ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -183,6 +183,16 @@ final case class NoReportInInterval(
     expirationDateTime: DateTime
 ) extends NoReport
 
+/*
+ * When we want to keep compliance even in case of NoReportInterval
+ */
+final case class KeepLastCompliance(
+    expectedConfig:     NodeExpectedReports,
+    expiredSince:       DateTime,
+    expirationDateTime: DateTime,
```
clarktsiory, line 192: I'm struggling to understand the difference between the two fields only from the naming, what is `expirationDateTime` if it's not the date since which compliance expire ?

clarktsiory: is it not `expirationTime: DateTime, keepUntil: DateTime` ?

clarktsiory: or `expiredSince: DateTime, keepUntil: DateTime` ?

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/reports/CachedFindRuleNodeStatusReportsTest.scala
```diff
@@ -231,8 +235,37 @@ class CachedFindRuleNodeStatusReportsTest extends Specification {
...
+
+    // let a chance for zio to exec
+    Thread.sleep(1000)
+
+    // now node was ask, it will return all nodes, even expired, see: https://issues.rudder.io/issues/16612
+    val n2 = repo.getNodeStatusReports(finder.reports.keySet).runNow
+
+    (n1 must beEqualTo(Map())) and
+    (n2 must beEqualTo(computed))
+  }
```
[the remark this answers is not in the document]
fanf: ah yes, I just copy/pasted what was in the other tests. I think the tests should be reworked even more profondly, but I didn't want to do that in that PR to keep the same oracle

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: it's what I did initially, but `expirationDateTime` has a special meaning provided by the trait `ExpiringStatus` and is used by the compute engine to know when the node need to be checked again if nothing happened to it. 
So in the case of `KeepLastCompliance`, it need to be the time on which the compliance that we kept should not be kept anymore. 

But I wanted to also have at hand the time on which the node should have become grey to make informative message, like "that node stopped responding at $expiredSince, but that's ok because it is configured to have extra grace period until $expirationDateTime"

clarktsiory: ok ! it's makes sense for me now !

clarktsiory (approved):
LGTM ! I had a few questions and remarks that can be addressed in the future

fanf:
OK, squash merging this PR

PR #5774, opened 2024-07-18
Fixes #25164: When agent version is missing in inventory, we get a security token error

fanf:
https://issues.rudder.io/issues/25164

Check for missing agent version in parsing to get a correct error message. 

Relax a bit the check on node fact transformation to make it easier to spot that kind of problem. 

image

### DIFF
--- a/webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionInventoryParser.scala
+++ b/webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionInventoryParser.scala
@@ -440,12 +440,13 @@ class FusionInventoryParser(
                 "could not parse agent security token (tag AGENT_CERT), which is mandatory"
               ).fail
           }
-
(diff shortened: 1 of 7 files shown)


webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionInventoryParser.scala
```diff
@@ -440,12 +440,13 @@ class FusionInventoryParser(
                 "could not parse agent security token (tag AGENT_CERT), which is mandatory"
               ).fail
           }
-
+        version        <-
+          findAgent(inventory.applications, agentType).notOptional(
+            s"Agent is not present in software list and so we can't get its version. This is not supported anymore"
```
amousset, line 445:
```suggestion
            s"Agent is not present in software list and so we can't get its version. This is not supported anymore."
```

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #5775, opened 2024-07-18
Fixes #25161: Node inventory hooks are are not working properly

ElaadF:
https://issues.rudder.io/issues/25161

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
@@ -70,24 +70,28 @@ import zio.syntax.*
  */
 class PostCommitInventoryHooks[A](
     HOOKS_D:               String,
-    HOOKS_IGNORE_SUFFIXES: List[String]
+    HOOKS_IGNORE_SUFFIXES: List[String],
+    nodeFactRepo:          NodeFactRepository
 ) extends PostCommit[A] {
+  import QueryContext.todoQC
   import scala.jdk.CollectionConverters.*
-
   override val name = "post_commit_inventory:run_node-inventory-received_hooks"
 
   override def apply(inventory: Inventory, records: A): IOResult[A] = {
     val node  = inventory.node.main
     val hooks = (for {
       systemEnv <- IOResult.attempt(java.lang.System.getenv.asScala.toSeq).map(seq => HookEnvPairs.build(seq*))
-      hooks     <- if (node.status == PendingInventory) {
-                     RunHooks.getHooksPure(HOOKS_D + "/node-inventory-received-pending", HOOKS_IGNORE_SUFFIXES)
-                   } else if (node.status == AcceptedInventory) {
-                     RunHooks.getHooksPure(HOOKS_D + "/node-inventory-received-accepted", HOOKS_IGNORE_SUFFIXES)
-                   } else {
-                     Inconsistency(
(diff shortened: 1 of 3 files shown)


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
```diff
@@ -70,24 +70,28 @@ import zio.syntax.*
...
       systemEnv <- IOResult.attempt(java.lang.System.getenv.asScala.toSeq).map(seq => HookEnvPairs.build(seq*))
-      hooks     <- if (node.status == PendingInventory) {
-                     RunHooks.getHooksPure(HOOKS_D + "/node-inventory-received-pending", HOOKS_IGNORE_SUFFIXES)
-                   } else if (node.status == AcceptedInventory) {
-                     RunHooks.getHooksPure(HOOKS_D + "/node-inventory-received-accepted", HOOKS_IGNORE_SUFFIXES)
-                   } else {
-                     Inconsistency(
-                       s"node-inventory-received-* hooks are not supported for node '${node.hostname}' [${node.id.value}] wiht status '${node.status.name}'"
-                     ).fail
+      hooks     <- nodeFactRepo.getStatus(node.id).flatMap { s =>
```
fanf, line 84: why all this is needed ? Inventory doesn't have the correct status ? Else, perhaps we should just pass a nodefact ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NewNodeManager.scala
```diff
@@ -185,7 +182,7 @@ class NewNodeManagerHooksImpl(
     .runNow
 
   override def afterNodeAcceptedAsync(nodeId: NodeId)(implicit qc: QueryContext): IOResult[Unit] = {
-    codeHooks.get.flatMap(hooks => ZIO.foreach(hooks)(h => IOResult.attempt(h.run(nodeId)))).unit
+    codeHooks.get.flatMap(hooks => ZIO.foreach(hooks)(h => IOResult.attempt(h.run(nodeId).forkDaemon.runNow))).unit
```
fanf, line 185: you can can't `.runNow` in a foreach. 
But I think that the root cause: `.run` is now an IOResult, so it's just a no-op in the `attempt`. `IOResult.attempt` need to be removed.

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/inventory/PostCommits.scala
ElaadF: Something is missing, it seems like `FactRepositoryPostCommit[A]` was updating correctly the status, but it is not used any more (I don't know why we keep it though?)
https://github.com/Normation/rudder/pull/5775/files#diff-060592aa1d65bfa1ec057158e1139794df7ad9136723a72ecd4e9695e98891b7L1990

fanf: OK, it will need to be understood why it is like that, and clean everything / change the parameters of that methods which are meaningless.

fanf (approved):

ElaadF:
PR rebased

PR #5786, opened 2024-07-25
Fixes #25203: The pending nodes API now returns array of arrays of nodes instead of an array of nodes

ElaadF:
https://issues.rudder.io/issues/25203
image


ElaadF:
Commit modified

fanf:
Seems ok, can you add an unit test in `node_api.yml` so that we don't get bitten by that again?

fanf (changes requested):
Seems ok, can you add an unit test in node_api.yml so that we don't get bitten by that again?

ElaadF:
See https://github.com/Normation/rudder/pull/5799#pullrequestreview-2227352345

PR #5790, opened 2024-07-27
Fixes #25225: Improve SVG image handling in the repo

amousset:
https://issues.rudder.io/issues/25225

This PR introduces a difference between:

* Source files, for edition
  * All are current sources are SVG files, to be edited with Inkscape.
* Production files, derived automatically from the sources
  * Optimized SVG (size reduced by 40-90%)
  * Optimized bitmap (PNG/ICO)

Sources:
* Move all our SVG sources at the same place (`webapp/sources/rudder/rudder-web/src/main/svg`)
  * Remove duplicate existing logos
  * Contains logo sources from `/logo` and all SVG sources from the webapp
* Add a new `notext` source logo (required for the favicon)

Build:
* Build optimized SVG at webapp build time using gulp
* Add a script to rebuild the favicon from sources, but don't use it at build time (too many dependencies and the logo does not change often)
* Add a prebuilt set of optimized SVG files for external usage in `/logo`

Additionally:

* Remove an unused (ugly) background JPEG and the matching unused CSS rule.
* Remove the `shortcut` in the `rel` attribute of a `link` which should not be used:

> Note: The shortcut link type is often seen before icon, but this link type is non-conforming, ignored and web authors must not use it anymore.
(https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#icon)


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

PR #5791, opened 2024-07-27
Fixes #25219: Cache for node properties and hierarchies

fanf:
https://issues.rudder.io/issues/25219

This PR add a `PropertiesRepository` that stores the computed/resolved `GroupProperties` and `NodeProperties`. 

The repository is very simple: it is just an pure in-memory one, back by a pair of `Ref[Map[objId,  Chunk[NodePropertyHierarchy]]`.
It is used in place of interactive inheritance computation in group/node API that get inherited props, and in policy generation. 
The part is policy generation is the riskiest change. Moreover, I think we are changing a bit the way things are done, because we use to do: 
- interpolate global props, 
- do inheritance resolution for node props with the result, 
- interpolate resolved node props. 

Now, we directly get inherited props and interpolate that. I *think* it's the same, and no tests show that it's different, but I'm sure our tests would exhibite the kind of case where a problem can happen, which would be something leading to interpolated global props not being overridden in the same way that non interpolated ones. 

The repository is only update at two moment thanks to the action of `NodePropertiesService.updateAll()`: 
- when rudder starts to get initial values, 
- after dynamic group computation. 

I *think* it should be enough, since global properties, group properties, and node properties change all lead to a dyn group computation, but that's point will need to be toroughtly tested. 

For tests, I needed to add callback for global prop change and node change, because we don't have a real dyngroup update mocked-up. 

I also moved things related to property inheritance/overriding into a new package: `com.normation.rudder.properties`. I didn't move the `General/Global/Group/NodeProperty` class declaration into it, because it would have mean a massive refactoring, and it would have totally killed the understandability of that one, with repercussion on plugins etc.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/UpdateDynamicGroups.scala
@@ -48,13 +48,16 @@ import com.normation.rudder.domain.logger.ScheduledJobLogger
 import com.normation.rudder.domain.nodes.NodeGroupId
 import com.normation.rudder.facts.nodes.ChangeContext
 import com.normation.rudder.facts.nodes.QueryContext
+import com.normation.rudder.properties.NodePropertiesService
 import com.normation.rudder.services.queries.*
 import com.normation.rudder.utils.ParseMaxParallelism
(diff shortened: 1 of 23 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #5797, opened 2024-08-07
Fixes #25282: Postinst script are not run during upgrades

amousset:
https://issues.rudder.io/issues/25282

The problem here is that the uninstallation rremoves the script after preinst run. So the postinst never actually runs.


amousset:
PR updated with a new commit

Fdall (approved):

amousset:
OK, squash merging this PR

PR #5801, opened 2024-08-08
Fixes #25095: Dashboard information are not updated when deleting a node

clarktsiory:
https://issues.rudder.io/issues/25095


clarktsiory:
Commit modified

PR #5802, opened 2024-08-08
Fixes #25095: Dashboard information are not updated when deleting a node

clarktsiory:
https://issues.rudder.io/issues/25095

Indeed, if it is a `SessionVar`, nodes updates only get displayed the same as when logging in.
Previously, there has not been any `Var`, the node info was always requested anyway, it should be fetched from cache so it should not cause performance issues here


ElaadF:
I would never have found :grin:

ElaadF (approved):

PR #5803, opened 2024-08-08
Fixes #21668: tooltips in the rules page persist when switching content

clarktsiory:
https://issues.rudder.io/issues/21668

We just need to backport `initTooltips` command from later versions or Rudder, in which we don't have the issue
(in 7.3, we didn't even initialize tooltips on first page render, but only when we open a Rule details :scream_cat:)


VinceMacBuche (approved):

clarktsiory:
OK, merging this PR

PR #5808, opened 2024-08-09
Fixes #24976: Tenants column in user-management should depend on plugin activation

clarktsiory:
https://issues.rudder.io/issues/24976

Change consists of adding the `tenantsEnabled` field in the API (the variable already exists), and conditionally displaying the table header and rows for the "Tenants" column based on a new boolean variable in the Elm model.


fanf:
LGTM, works as expected

fanf (approved):

PR #5810, opened 2024-08-12
Fixes #25301: User management table has the wrong count of entries with deleted users

clarktsiory:
https://issues.rudder.io/issues/25301

The text is just using the entire list of users instead of the filtered on to display  `Showing 1 to 20 of 20 entries`.
We don't paginate the table, so that text does not really make sense, it should rather be `Showing 20 of 20 entries` and `Showing 5 of 20 entries` when filtered with the search bar.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/UserManagement/View.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/UserManagement/View.elm
@@ -21,7 +21,7 @@ view model =
                 text "Waiting for data from server..."
 
             else
-                displayUsersConf model model.users
+                displayUsersConf model
 
         deleteModal =
             if model.ui.openDeleteModal then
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/UserManagement/View.elm
```diff
@@ -514,8 +515,8 @@ displayRightPanel model user =
         ]
 
 
-displayUsersConf : Model -> Users -> Html Msg
-displayUsersConf model u =
+displayUsersConf : Model -> Html Msg
+displayUsersConf model =
```
clarktsiory, line 519: This was unused

clarktsiory:
Commit modified

ElaadF (approved):

clarktsiory:
PR rebased

PR #5811, opened 2024-08-12
Fixes #25302: API documentation examples are not valid bash scripts

clarktsiory:
https://issues.rudder.io/issues/25302

The _.sh_ files need to be valid Bash scripts and using multiline curl arguments for the JSON data is a way to make it a valid command


fanf (approved):
:+1:

PR #5812, opened 2024-08-13
Fixes #25300: Only remaining user managed by file cannot be deleted 

clarktsiory:
https://issues.rudder.io/issues/25300

The case with `setExistingUsers(Nil)` should not be `ZIO.unit` because it means that we want to remove all users !


clarktsiory:
Commit modified

clarktsiory:
Commit modified

fanf (approved):
LGTM

PR #5813, opened 2024-08-13
Fixes #25303: Database tests have non temporary DDL that prevents running them twice

clarktsiory:
https://issues.rudder.io/issues/25303

There is no such `CREATE TEMP TYPE score` in PostgreSQL, even if there is `CREATE TEMP TABLE`.
The solution here is quite radical : use a [schema](https://www.postgresql.org/docs/current/ddl-schemas.html) for tests, all tables, sequence, etc. creation will be under the schema, and it could be dropped at the end of tests, along with any structure that was declared in tests because they are configured to be within the schema.

An alternative solution which could be harder to maintain as we add new Types, is to explicitly drop the Type.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/db/DBCommon.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/db/DBCommon.scala
@@ -79,8 +79,14 @@ trait DBCommon extends Specification with Loggable with BeforeAfterAll {
    * everything is temporary
    */
   def sqlInit: String = {
-    val is      = this.getClass().getClassLoader().getResourceAsStream("reportsSchema.sql")
-    val sqlText = Source
+    val is              = this.getClass().getClassLoader().getResourceAsStream("reportsSchema.sql")
+    // This is a way to create and use all DDL under a schema, that will be dropped at the end of tests
+    val setTestNamepace = {
+      """
+        |SET search_path TO pg_temp
+        |""".stripMargin
+    }
+    val sqlText         = Source
       .fromInputStream(is)
       .getLines()
       .toSeq
@@ -88,15 +94,10 @@ trait DBCommon extends Specification with Loggable with BeforeAfterAll {
         val s = line.trim(); s.isEmpty || s.startsWith("/*") || s.startsWith("*")
       }
       .map(s => {
-        s
-          // using toLowerCase is safer, it will always replace create table by a temp one,
-          // but it also mean that we will not know when we won't be strict with ourselves
-          .toLowerCase
-          .replaceAll("create table", "create temp table")
-          .replaceAll("create sequence", "create temp sequence")
-          .replaceAll("alter database rudder", "alter database test")
+        // we work on a "test" database which needs to be created, and which is cleaned (see clean methods)
+        s.replaceAll("alter database rudder", "alter database test")
       })
-      .mkString("\n")
+      .mkString(s"${setTestNamepace};\n", "\n", "")
     is.close()
 
     sqlText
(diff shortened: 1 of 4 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/db/DBCommon.scala
```diff
@@ -125,8 +127,13 @@ trait DBCommon extends Specification with Loggable with BeforeAfterAll {
...
 
-  def cleanDb(): Unit = {
-    if (sqlClean.trim.size > 0) doobie.transactRunEither(xa => Update0(sqlClean, None).run.transact(xa)) match {
+  final def cleanDb(): Unit = {
+    doobie.transactRunEither(xa => {
+      val finalizer = fr"DROP SCHEMA temp_schema CASCADE".update.run.transact(xa)
+      ZIO.when(sqlClean.trim.size > 0)(
+        Update0(sqlClean, None).run.transact(xa)
+      ) *> finalizer
+    }) match {
```
clarktsiory, line 136: From what I understand this should not really be overriden, because it may contain strict finalizers in the supertrait. And the `sqlClean` was added to make it possible to still execute sql in inheritors.  
This should be the same for `initDb` but it has a more complex override in [TestDbMigration_5_6](https://github.com/Normation/rudder/blob/b5ce7c2f3f645b77512b60ec9db7f8f391d163bf/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/migration/TestDbMigration.scala#L125), which needs to be rewritten.

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestMigrateChangeValidationEnforceSchema.scala
```diff
@@ -63,19 +63,6 @@ class TestMigrateChangeValidationEnforceSchema extends DBCommon with IOChecker {
...
-        DROP TABLE IF EXISTS ${tempWorkflowTable};
-        DROP TABLE IF EXISTS ${tempCRTable};
-        DROP SEQUENCE IF EXISTS ChangeRequestId_temp;
-      """.update.run.transact(_)
-    ) match {
-      case Right(_) => ()
-      case Left(ex) => throw ex
-    }
-  }
-
```
clarktsiory, line 78: Since the test schema is dropped anyway, every table/sequence/etc. created during tests is guaranteed to also be dropped

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/db/DBCommon.scala
```diff
@@ -79,24 +79,26 @@ trait DBCommon extends Specification with Loggable with BeforeAfterAll {
...
    */
   def sqlInit: String = {
-    val is      = this.getClass().getClassLoader().getResourceAsStream("reportsSchema.sql")
-    val sqlText = Source
+    val is              = this.getClass().getClassLoader().getResourceAsStream("reportsSchema.sql")
+    // This is a way to create and use all DDL under a schema, that will be dropped at the end of tests
+    val setTestNamepace = {
+      """
+        |CREATE SCHEMA temp_schema;
+        |SET search_path TO temp_schema;
```
fanf, line 87: what happens if tests are interrupted and the `DROP SCHEMA temp_schema CASCADE` is never done ? Do we hand up with that schema being used on that database by default, even in new connections? 
If so, is there a way to say "SET search_path TO temp_schema" but only for current connection (and so revert to the default schema in any case once the connection is terminated ?)

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/db/DBCommon.scala
clarktsiory: Yes the schema would still remain after tests have executed.
Provided that we can create a schema-per-connection, we should indeed do as you say. 
And it seems to be the case (https://dba.stackexchange.com/a/76661), so I changed it and it is even simpler : https://github.com/Normation/rudder/pull/5813/commits/5193ce3116f7d003ef4b29e037b2171840e95645

fanf:
LGTM, simpler and nicer than before, very good :)

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #5816, opened 2024-08-16
Fixes #25307: 

clarktsiory:
https://issues.rudder.io/issues/25307

We need to check on every request if the current user is still active or not (Lift does not provide a way to match our user sessions with the ones for all users, because there are session fixations from spring security happening at login so session ids are changed). So we use a Spring filter for that.

To be able to that efficiently, especially to avoid querying the users status on every request, we want to cache users status.
And updating the cache is done when the users file is reloaded after a disable/active in `user-management` (to work, it then requires changes from https://github.com/Normation/rudder-plugins/pull/744).

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
@@ -1,0 +1,105 @@
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+package bootstrap.liftweb
+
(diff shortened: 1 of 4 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -0,0 +1,102 @@
...
+ */
+class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailListProvider: FileUserDetailListProvider)
+    extends OncePerRequestFilter {
+  private val userStatuses = Ref
+    .make(Map.empty[String, UserStatus])
+    .map(ref => {
+
+      userDetailListProvider.registerCallback(
+        RudderAuthorizationFileReloadCallback(
+          "userSessionInvalidation",
```
fanf, line 65: there is inconsistency in the naming scheme for these `RudderAuthorizationFileReloadCallback`: we have either camelCase like here, or snake-case for `update-pg-users-on-xml-file-reload` => can you homogeneize ?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -0,0 +1,102 @@
...
+    val auth    = SecurityContextHolder.getContext.getAuthentication
+    if (auth != null) {
+      val userDetails = auth.getPrincipal.asInstanceOf[RudderUserDetail]
+      val status      = userStatuses.get.map(
+        _.getOrElse(userDetails.getUsername, UserStatus.Deleted)
+      ) // user do no longer exist is the same as if was deleted
+
+      status.runNow match {
+        case UserStatus.Disabled | UserStatus.Deleted =>
+          session.invalidate()
```
fanf, line 91: (not important, but can you let it right of the arrow so that it's consistant with () below ?)

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -0,0 +1,102 @@
...
+    })
+    .runNow
+
+  @throws[ServletException]
+  @throws[IOException]
+  override def doFilterInternal(request: HttpServletRequest, response: HttpServletResponse, filterChain: FilterChain): Unit = {
+    val session = request.getSession(false)
+    val auth    = SecurityContextHolder.getContext.getAuthentication
+    if (auth != null) {
+      val userDetails = auth.getPrincipal.asInstanceOf[RudderUserDetail]
```
fanf, line 84: You need to pattern match on user detail here, at least to avoid a class cast exception in the futur, and to make apparent what is done in that case (like for when `auth` is null I believe, ie nothing - with a comment)

fanf (commented):
It takes some time to take effect so it's a bit strange for a few seconds, but it works as expected. Once the little comments are addressed, it can be merged (same for the linked plugin PR)

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
clarktsiory: There was another camelCase, but I switched all to kebab-case because it is used in error logs and is easier to search for as a string

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #5818, opened 2024-08-16
Fixes #25123: Authentication happens twice with same session id

clarktsiory:
https://issues.rudder.io/issues/25123

We just ignore the `insert` in that case, but it then needs to be done at the last operation of the database transaction because a sql error interrupts the whole transaction (hence the change of order).
Doobie provides a nice API for SQL errors, we just use it to log a user-friendly warning.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -37,6 +37,7 @@ package com.normation.rudder.users
  *************************************************************************************
  */
 
+import cats.ApplicativeError
 import cats.data.NonEmptyList
 import com.normation.errors.Inconsistency
 import com.normation.errors.IOResult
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -616,10 +619,16 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
...
-    }).unit
+        _   <- lastLogin.update.run
+        err <- session.update.run
+                 .as(none[String])
+                 .onUniqueViolation(s"User session for ${userId} has already been saved".some.pure[ConnectionIO])
+      } yield err).transact(xa)
+    }).flatMap {
+      case None      => ZIO.unit
+      // it should not be fatal if a login already happened, but it could have security implications
+      case Some(err) => ApplicationLoggerPure.warn(err)
```
fanf, line 630: I don't think we want that. 
The warning here will be non actionnable for users, but still source of worry for people who monitor rudder logs and escalate on warning and errors. Especially if they found the comment `but it could have security implications` - they will want to know exactly which, and how to prevent them. 

Would it be possible to force the user to completely reautent in the case of a stalled sessionid ? IIUC, that happen only when the browser kept a session cookie that should have been destroyed because Rudder lost that context (restarted or anything). 
So can we force a full reauth (even if it means calling "user.logout" on sql error here) and put a login error message like "Rudder authentication service restarted during your login. Plean log again" ?

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #5819, opened 2024-08-16
Fixes #24955: One user is created each case change even if case sensivity if false

clarktsiory:
https://issues.rudder.io/issues/24955

We need to only insert the user when it does not already exist with lowered id : 
```sql
INSERT INTO users (id, ...)
SELECT ?,...,?
WHERE NOT EXISTS (
  SELECT 1 from users WHERE LOWER(id) = LOWER(?)
)
```
is the query that does that conditional insertion.

The user are not created in that case, so we need also a `RETURNING` clause to get back inserted users and compare with requested ones, to log non-created users as warning.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -106,8 +106,16 @@ trait UserRepository {
    * (locked won't be touched).
    * For user whose status is changed, also add the status update in its history.
    * User from other authenticators are not changed.
+   * When users with the same lower-cased id should not be allowed, case sensitivity should be set to false.
+   *
+   * @return A set of users that could not be updated.
    */
-  def setExistingUsers(origin: String, users: List[String], trace: EventTrace): IOResult[Unit]
+  def setExistingUsers(
+      origin:          String,
+      users:           List[String],
+      trace:           EventTrace,
+      isCaseSensitive: Boolean = true
+  ): IOResult[Set[String]]
 
   /*
    * Disable users based on filter.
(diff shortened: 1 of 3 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

fanf:
It should have a log when the user is not inserted because it already exists in the inser select case.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -693,18 +709,58 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
...
             maybeUpdate       <- updatable.query[UserInfo].to[Vector]
             (zombies, managed) = maybeUpdate.partition(_.status == UserStatus.Deleted)
             updated            = UserRepository.computeUpdatedUserList(users, origin, trace, toMap(zombies), toMap(managed))
-            _                 <- update(updated.values.toVector)
-          } yield ()).transact(xa)
+            dbUpdated         <- update(updated.values.toVector)
+            // compute users that may have not been inserted nor updated because of the case-sensitivity clause in the query
+            diff               = updated.keySet.diff(dbUpdated.toSet)
+          } yield diff).transact(xa).flatMap { nonUpdated =>
+            ZIO
```
fanf, line 754: this bit is complicated (not complex, but there's a lot that can go wrong, now and on futur update of it). Can you add a unit test for it ? I see there's a test that check that nothing change when the user is there, but it's different than "we get the correct values for error message", and it's all the ops will see. 

I think you should make an utility method alike `setExistingUsers` but return the non updated ones, or change signature of `setExistingUsers` to do that, and then check that you have the correct returned non updated user.

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
clarktsiory: :+1: That is right, I changed the `setExistingUsers` to return the non updated users, also previously it logged both conflicting users : "AlIcE" and "alice" instead of only the one which is passed for the update ("AlIcE" in the unit test)

fanf (approved):

clarktsiory:
PR rebased

clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #5826, opened 2024-08-22
Fixes #25334: Node property usage search seems to be limited to 10 elements

clarktsiory:
https://issues.rudder.io/issues/25334

We want the `limit` to be passed as REST query parameter instead of `limit:x` backend search : for now the `Query` case class is passed to the quicksearch backend param to transform into low-level query (LDAP, etc.), but this information is not actually used in any backend.
We only do the `take(limit)` at last when all results from a backend have been returned.
So the changes here removes the support introduced in https://github.com/Normation/rudder/pull/5778 and adds the support of the query parameter in place.

API tests have been bootstrapped for the Quicksearch API, and it was useful to understand and snapshot the behavior of the search (will also be useful when moving from lift-json to zio-json later).

Now it sees the number of requested elements : 
image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
@@ -68,6 +68,7 @@ import java.util.regex.Pattern
 import net.liftweb.common.Full
 import net.liftweb.common.Loggable
 import scala.util.control.NonFatal
+import zio.syntax.*
 
 /**
  * Correctly quote a token
@@ -430,14 +431,13 @@ object QSLdapBackend {
     val returnedAttributes =
       (query.attributes.map(_.ldapName).toSeq ++ Seq(A_OC, A_HOSTNAME, A_NAME, A_UUID, A_PARAMETER_NAME, A_IS_SYSTEM)).distinct
 
-    for {
-      connection <- ldap
-      entries    <- connection.search(rudderDit.BASE_DN, Sub, filter, returnedAttributes*)
-    } yield {
-
-      if (ocFilter.isEmpty || attrFilter.isEmpty) { // nothing to search for in that backend
-        Seq()
-      } else {
+    if (ocFilter.isEmpty || attrFilter.isEmpty) { // nothing to search for in that backend
+      Seq.empty[QuickSearchResult].succeed
+    } else {
+      for {
+        connection <- ldap
+        entries    <- connection.search(rudderDit.BASE_DN, Sub, filter, returnedAttributes*)
+      } yield {
         // transformat LDAPEntries to quicksearch results, keeping only the attribute
         // that matches the query on the result and no system entries but nodes.
         // Also, only keep nodes that exists.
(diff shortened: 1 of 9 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchBackendImpl.scala
```diff
@@ -430,14 +431,13 @@ object QSLdapBackend {
...
-      if (ocFilter.isEmpty || attrFilter.isEmpty) { // nothing to search for in that backend
-        Seq()
-      } else {
+    if (ocFilter.isEmpty || attrFilter.isEmpty) { // nothing to search for in that backend
+      Seq.empty[QuickSearchResult].succeed
+    } else {
+      for {
+        connection <- ldap
+        entries    <- connection.search(rudderDit.BASE_DN, Sub, filter, returnedAttributes*)
+      } yield {
```
clarktsiory, line 440: The for-yield is unnecessary when we have nothing to search for...
This change allows for passing everything which is related to LDAP as `null` for unit tests

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/quicksearch/QuickSearchService.scala
```diff
@@ -86,10 +90,14 @@ class FullQuickSearchService(implicit
...
                      .tap(results => logger.debug(s"  - [${b}] found ${results.size} results"))
+                     .map(limit match {
+                       case Some(value) => _.take(value)
+                       case None        => identity
+                     })
                      .chainError(s"Error with quicksearch bachend ${b}")
                  })
     } yield {
-      results.toSet.flatten.take(query.limit)
+      results.flatten.toSet
```
clarktsiory, line 100: This will be supposed to be more efficient, as results from different backends are unique, and flattening a seq is more efficient

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestQuicksearch.scala
```diff
@@ -142,17 +142,19 @@ class RestQuicksearch(
...
-        val unique  = set.map(x => (x.id, x)).toMap.values.toSeq.sortBy(_.name)
-        // on take the nth first, sorted by name
-        val summary = ResultTypeSummary(tpe.name, unique.size, unique.size)
+        // distinct by id, sorted by name
+        val unique   = set.map(x => (x.id, x)).toMap.values.toSeq.sortBy(_.name)
+        // take the nth first
+        val returned = unique.take(maxByKind)
 
-        (tpe, (summary, unique))
+        val summary = ResultTypeSummary(tpe.name, unique.size, returned.size)
```
clarktsiory, line 155: still take the default behavior of MAX_RES_BY_KIND = 10 elements when no limit is specified, it is the supposed main usage of Quicksearch

ElaadF (approved):

fanf:
OK, squash merging this PR

fanf:
OK, merging this PR

PR #5827, opened 2024-08-22
Fixes #25136: Nodes table sorting is not obvious and we need to guess the sort column and direction 

RaphaelGauthier:
https://issues.rudder.io/issues/25136

Because the columns in the node table are customisable and this table is still generated by datatable.js, the .head class cannot easily be added to the <tr> tag in the table header (without causing any side-effects). 

The simplest way is to apply the same css rule to this kind of table header

Now this header is identical to that of the other tables :
header-node


clarktsiory (approved):
LGTM !

PR #5834, opened 2024-08-23
Fixes #25348: Deleting CVE group is possible even if it is a system group

clarktsiory:
https://issues.rudder.io/issues/25348

We need to forbid the deletion of system groups and system group categories in the REST API (this PR adds the error handling and API tests for that).
However, in the UI we now want to allow cloning a system group which has a group target and an ID.

For the CVE groups, we now have them as system in https://github.com/Normation/rudder-plugins-private/pull/720, so we now redirect to CVE details page to delete them :
Peek 2024-08-28 16-04

Illustration of the cloning part : 
Peek 2024-08-28 16-06

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
@@ -86,6 +86,7 @@ import com.normation.rudder.domain.RudderDit
 import com.normation.rudder.domain.archives.ParameterArchiveId
 import com.normation.rudder.domain.archives.RuleArchiveId
 import com.normation.rudder.domain.nodes.*
+import com.normation.rudder.domain.nodes.NodeGroup
 import com.normation.rudder.domain.policies.*
 import com.normation.rudder.domain.properties.AddGlobalParameterDiff
 import com.normation.rudder.domain.properties.DeleteGlobalParameterDiff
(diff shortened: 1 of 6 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf:
I'm not sure of the general consistency rules of the whole: 
- we have system or normal categories, system or normal groups
- system categories implies some specific restriction (no deletion), appart root category which is different, 
- some other subtilities 

Why no go for the simple semantic: system things are things managed by Rudder (and the users can't directly, they need to rely on specific interface for that), and user things are things managed by rudder (and they can't have specific restriction, else they are system and need a dedicated API to manage them) ? 
I fail to see where that model doesn't work in our case. CVE groups are system groups because they are group created and managed by (a plugin of) Rudder, not directly by the user. 
Root category is system because it is managed by Rudder, and users don't have the right to change them. 
Etc.

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):
Great :)

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #5835, opened 2024-08-26
Fixes #25350: no reports in campaign page even if the node sent the data

VinceMacBuche:
https://issues.rudder.io/issues/25350


fanf:
Replaced with #5837

PR #5836, opened 2024-08-27
Fixes #25352: Event log rollback action is prevented from CSP headers

clarktsiory:
https://issues.rudder.io/issues/25352

There was some `onclick` event handlers that were dynamically assigned (`$(...).append('<button onclick=.../>')`), but the would never be executed since we have strict CSP headers.
We have to replace them by hand and bind within a `script` tag on the server template.


clarktsiory:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -109,7 +110,7 @@ class EventLogDetailsGenerator(
       val name = (x.details \ "rule" \ "displayName").text
       Text("Rule ") ++ {
         if (id.serialize.length < 1) Text(name)
-        else <a href={ruleLink(id)} onclick="noBubble(event);">{name}</a> ++ actionName
+        else <a href={ruleLink(id)}>{name}</a> ++ actionName
```
clarktsiory, line 113: This `onclick` also causes issues, removing it still make it work fine

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

RaphaelGauthier (approved):

clarktsiory:
OK, squash merging this PR

PR #5837, opened 2024-08-27
Fixes #25350: no reports in campaign page even if the node sent the data

clarktsiory:
https://issues.rudder.io/issues/25350

This replaces #5835 

There is no error handling on the `loop` block, so when it has defects (uncaught runtime errors the fiber just terminates).
We need to use `catchAllCause` instead of `catchAll` for that purpose to catch other possible runtime errors, this process is really meant to run forever and retried every 5 seconds if needed.


clarktsiory:
PR updated with a new commit

fanf (approved):
Lgtm

clarktsiory:
OK, squash merging this PR

PR #5839, opened 2024-08-28
Fixes #25365: Scheduled processes could stop on error in ZIO

clarktsiory:
https://issues.rudder.io/issues/25365

`UIO[?]` does not mean that all errors have been caught : only `E` ones, when the initial type is `ZIO[R, E, A]` and often `RudderError` in the case of `IOResult`. There could also be [defects](https://zio.dev/reference/error-management/expected-and-unexpected-errors#unexpected-errors-1) (runtime errors within the fiber) which would have been uncaught and would interrupt the currently running fiber.
This PR is for preventing such error from interrupting a repeated process in ZIO (with a delay and running forever), and for logging errors and continuing instead.

The type is not an absolute reference because we could do : 
* `IOResult.attempt: IOResult[A]` which already include error handling of all defects
but doing : 
* `ioResult#catchAll: UIO[A]` is not enough to include error handling of all defects, we need to `catchAllDefects` (or do both catches at once with `catchFailureOrCause` as in #5837)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeDeletedInventories.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeDeletedInventories.scala
@@ -68,7 +68,7 @@ class PurgeDeletedInventories(
       logger.debug(
         s"***** starting batch that purge deleted inventories older than ${TTL} days, every ${updateInterval.toString()} *****"
       )
-      val prog = purgeDeletedNodes
+      val prog: UIO[Unit] = purgeDeletedNodes
         .purgeDeletedNodesPreviousDate(DateTime.now().withTimeAtStartOfDay().minusDays(TTL))
         .either
         .flatMap(_ match {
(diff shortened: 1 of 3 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/PurgeDeletedInventories.scala
```diff
@@ -85,8 +85,18 @@ class PurgeDeletedInventories(
...
             ScheduledJobLoggerPure.error(error.fullMsg)
         })
+        .unit
       import zio.Duration.fromScala as zduration
-      ZioRuntime.unsafeRun(prog.delay(zduration(updateInterval)).repeat(Schedule.spaced(zduration(updateInterval))).forkDaemon)
+      // Effect should be running forever and every defect caught into the UIO
+      ZioRuntime.unsafeRun(
+        prog
+          .catchAllDefect(err =>
+            ScheduledJobLoggerPure.error(s"Error in process or purging deleted nodes inventories older than ${TTL} days", err)
```
fanf, line 94: If we go there, the fiber died. I believe it's the one with the "unit of work for that iteration", ie just one purge delete, and thus the next schedule will just happen. Is that so ?

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #5841, opened 2024-08-28
Fixes #25369: Add an optional visibility attribute on node property 

fanf:
https://issues.rudder.io/issues/25369

This PR add a new `visibility` attribute on properties (global, group, node) that can be either `hidden` or `displayed` (default value). 
That attribute direct the fact that a property is returned or not in API/UI/Archive. Even if `hidden`, a property will still be used for policy generation and transferred to node `properties.d/prorperties.json` file. 

That attribute is inherited, so that if the `benchmark_rsc01` property is defined at global level as `hidden`, then it is also `hidden` when inherited in groups or nodes. 

From a code point of view, it's just one more key in the config value for a properties. 
Contrary to other cases, I choose to never have the default value (`displayed`) stored. The semantic is "if the `visibility` key is not here, then its the default value". So when we update, we actually remove the property. It looks cleaner, perhaps we could do that with other property with a clear default value (provider for ex). 

The other change in code are rather simple:
- most of the PR is about testing the different case of filtering own or inherited properties,
- the actual changes for GET are adding a `filter(visibility == Displayed)` when we transform group/node properties to json for API
- for PUT we put back group/node hidden properties in addition to user provided one. If the user use a property with the same name than a hidden one, then the user property overrides the hidden one.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
@@ -56,27 +56,7 @@ import com.normation.inventory.ldap.core.InventoryDitService
 import com.normation.inventory.ldap.core.InventoryDitServiceImpl
 import com.normation.inventory.services.core.ReadOnlySoftwareDAO
 import com.normation.rudder.MockNodes.allNodeFacts
-import com.normation.rudder.campaigns.Campaign
-import com.normation.rudder.campaigns.CampaignDetails
-import com.normation.rudder.campaigns.CampaignEvent
-import com.normation.rudder.campaigns.CampaignEventId
-import com.normation.rudder.campaigns.CampaignEventRepository
-import com.normation.rudder.campaigns.CampaignId
-import com.normation.rudder.campaigns.CampaignInfo
-import com.normation.rudder.campaigns.CampaignParsingInfo
-import com.normation.rudder.campaigns.CampaignRepository
-import com.normation.rudder.campaigns.CampaignSerializer
-import com.normation.rudder.campaigns.CampaignStatusValue
-import com.normation.rudder.campaigns.CampaignType
-import com.normation.rudder.campaigns.DayTime
-import com.normation.rudder.campaigns.Enabled
-import com.normation.rudder.campaigns.Finished
-import com.normation.rudder.campaigns.JSONTranslateCampaign
-import com.normation.rudder.campaigns.MainCampaignService
-import com.normation.rudder.campaigns.Monday
-import com.normation.rudder.campaigns.Running
-import com.normation.rudder.campaigns.Scheduled
-import com.normation.rudder.campaigns.WeeklySchedule
+import com.normation.rudder.campaigns.*
 import com.normation.rudder.configuration.ConfigurationRepositoryImpl
 import com.normation.rudder.configuration.DirectiveRevisionRepository
 import com.normation.rudder.configuration.GroupRevisionRepository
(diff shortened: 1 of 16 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
```diff
@@ -2837,7 +2801,8 @@ class MockNodeGroups(mockNodes: MockNodes, mockGlobalParam: MockGlobalParam) {
         None,
         None
       )
-      .getOrElse(throw new RuntimeException(s"error group prop: jsonParam")) // for test
+      .getOrElse(throw new RuntimeException(s"error group prop: jsonParam")), // for test
+    GroupProperty("hiddenProp", GitVersion.DEFAULT_REV, "hidden prop value".toConfigValue, None, None).withVisibility(Hidden)
```
clarktsiory, line 2805: I do not really understand why there is no `visibility: Visibility` in the constructor of `NodeProperty`, `GroupProperty`, even if there already is a `.withVisibility` method. Is it to avoid propagating the changes in plugins and APIs everywhere ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
```diff
@@ -1358,20 +1359,21 @@ class GroupApiService14(
...
-                                    .checkValueTypes(matchingChildProperties.map(NodePropertyHierarchy(p.prop, _)))
-                                    .isLeft
-
-                                  (
-                                    p,
-                                    matchingChildProperties.flatten,
-                                    hasConflicts
-                                  )
+                                parentProperties.collect {
+                                  case p if p.prop.visibility == Displayed =>
```
clarktsiory, line 1363: I wonder why this is not also done in the target JSON mapper/transformer (here to `JRGroupProperty`), as done with the properties for `JRGroup` elsewhere : 
```
.withFieldComputed(_.properties, _.properties.filter(_.visibility == Displayed).map(JRProperty.fromGroupProp(_)))
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
```diff
@@ -752,14 +754,16 @@ class NodeApiInheritedProperties(
...
+   * Full list of node properties, including inherited ones for a node but
+   * without hidden property.
    */
   def getNodePropertiesTree(nodeId: NodeId, renderInHtml: RenderInheritedProperties)(implicit
       qc: QueryContext
   ): IOResult[JRNodeInheritedProperties] = {
     for {
       properties       <- propRepo.getNodeProps(nodeId).notOptional(s"Node with ID '${nodeId.value}' was not found.'")
       propertiesDetails = properties
+                            .filter(_.prop.visibility == Displayed)
```
clarktsiory, line 766: Why not filter at the level of the target JSON class i.e. in `JRNodeInheritedProperties.fromNode(...)` ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ParametersApi.scala
```diff
@@ -481,7 +481,9 @@ class ParameterApiService14(
   }
 
   def listParameters(): IOResult[Seq[JRGlobalParameter]] = {
-    readParameter.getAllGlobalParameters().map(_.sortBy(_.name).map(JRGlobalParameter.fromGlobalParameter(_, None)))
+    readParameter
+      .getAllGlobalParameters()
+      .map(_.filter(_.visibility == Visibility.Displayed).sortBy(_.name).map(JRGlobalParameter.fromGlobalParameter(_, None)))
```
clarktsiory, line 486: Same question about filtering inside `JRGlobalParameter.fromGlobalParameter(...)` instead

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
fanf: Yes, and for now, it looks like an exceptionnal attribute. 
If it becomes common in the future (possible, since properties are so powerful, we are likely to want to use them more and more, for thing that will be intenal to rudder behavior), we will change that.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
fanf: I think I don't understand where you would want to put it ? I need to filter from the list of properties, not when I transform just one ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
fanf: filter are better done early, it makes less work for the following steps.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ParametersApi.scala
fanf: same answer, it's sooner here.

fanf:
I will merge as that because of the big weekend and since the feedbacks doesn't look like blocking. Of course we will be able to come back to that if I misunderstood a comment.

fanf:
OK, squash merging this PR

fanf:
OK, squash merging this PR

fanf:
OK, squash merging this PR

fanf:
OK, squash merging this PR

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
clarktsiory: I am asking because in the case of `JRNodeDetailLevel` and `JRGroup` the filtering was done at the level of mapping to the JSON class for serialization. I expected it to be the same everywhere else because otherwise it would not guarantee that the JSON will not contain hidden ones if we just forget to filter somewhere before the final serialization step

PR #5844, opened 2024-08-29
Fixes #25380: Score enum values are not updated before 8.1~rc1

clarktsiory:
https://issues.rudder.io/issues/25380


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableScore.scala
```diff
@@ -77,7 +82,7 @@ class CheckTableScore(
     );"""
 
     transactIOResult(s"Error with 'score' type creation")(xa =>
-      checkType.query[Int].option.transact(xa).flatMap(i => if (i.isDefined) ().succeed else sqlType.update.run.transact(xa))
+      checkType.query[Int].option.transact(xa).flatMap(i => (if (i.isDefined) sqlTypeUpdate else sqlType).update.run.transact(xa))
```
clarktsiory, line 85: CREATE or UPDATE depending on the presence of the type

PR #5846, opened 2024-08-29
Fixes #25383: Worst report takes the worst component instead of block

clarktsiory:
https://issues.rudder.io/issues/25383

Adding a new worst reporting logic type "by percent" :
image

The reporting computation that results takes the worst with respect to computed compliance (using default compliance precision) : 
Screenshot from 2024-09-05 16-44-58

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -39,7 +39,6 @@ package com.normation.rudder.domain.reports
 
 import com.normation.cfclerk.domain.ReportingLogic
 import com.normation.cfclerk.domain.ReportingLogic.*
-import com.normation.cfclerk.domain.WorstReportReportingLogic
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.policies.*
 import com.softwaremill.quicklens.*
@@ -357,7 +356,7 @@ final case class BlockStatusReport(
       // simple weighted compliance, as usual
       case WeightedReport => ComplianceLevel.sum(subComponents.map(_.compliance))
       // worst case bubble up, and its weight can be either 1 or the sum of sub-component weight
-      case worst: WorstReportReportingLogic =>
+      case worst: WorstReportWeightedReportingLogic =>
         val worstReport = ReportType.getWorseType(subComponents.map(_.status))
         val allReports  = getValues(_ => true).flatMap(_.messages.map(_ => worstReport))
         val kept        = worst match {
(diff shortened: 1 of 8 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/pom.xml
```diff
@@ -431,7 +431,7 @@ limitations under the License.
          upgrade without much care and correcting how we parse XML doc (xml
          parser option - see https://github.com/scala/scala-xml/releases/tag/v2.1.0 -->
     <scala-xml-version>1.3.0</scala-xml-version>
-    <lift-version>4.0.0-M1</lift-version>
+    <lift-version>4.0.0-alpha01</lift-version>
```
fanf, line 434: Nope, the official release is m1

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -357,14 +356,27 @@ final case class BlockStatusReport(
...
+            // block compliance will be turned into a percent, and the worst will be taken
+            b.getValues(_ => true).flatMap(_.messages.map(m => m.reportType))
+          }
+          case o: ValueStatusReport => {
+            // value has 1 count
+            List(ReportType.getWorseType(o.getValues(_ => true).flatMap(_.messages.map(_ => o.status))))
+          }
+        }
+        val worst      = allReports.minBy(ComplianceLevel.compute(_).computePercent().compliance)
+        ComplianceLevel.compute(worst)
```
fanf, line 379: I think it's the correct (only one that makes sense) thing to do. But we need to document it much better (internally, but also here)

At least a comment on each object déclaration, perhaps with some ASCII art of an example so that the next dev get the intent intuition (and then can check it with the unit tests)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -357,14 +356,27 @@ final case class BlockStatusReport(
...
         ComplianceLevel.compute(kept)
+      case WorstReportByPercent =>
+        // Get reports of sub-components to find the worst by percent
+        val allReports = subComponents.map {
+          case b: BlockStatusReport =>
+            // Convert block compliance to percent, take the worst
+            b.getValues(_ => true).flatMap(_.messages.map(m => m.reportType))
+          case o: ValueStatusReport =>
+            // Value has 1 count
+            List(ReportType.getWorseType(o.getValues(_ => true).flatMap(_.messages.map(_ => o.status))))
```
fanf, line 375: I'm not sure about the semantic of that one. Shouldn't we keep the weitght ? (ie is it more like weighted or like sum1 worst?)
It matters only for upper component, not directly that one.

fanf (approved):
I'm wondering what weight we want for the percent on value, but let's test like that, and if we want something else, we will open an other ticket.

clarktsiory:
PR rebased

PR #5848, opened 2024-08-29
Fixes #25351: Windows KB don't show up in the inventory anymore

clarktsiory:
https://issues.rudder.io/issues/25351

Previous inventory class before introduction of `SoftwareFact` already sees the version as optional.
Just add the option, make it compile and add a test to get the software from ldap (we will likely need to see how tests in 8.2 behave, there are many of them and for json).


clarktsiory:
PR updated with a new commit

fanf (approved):
:+1:

clarktsiory:
OK, merging this PR

PR #5850, opened 2024-08-29
Fixes #25388: Add log for error in JSON campaigns on FS

fanf:
https://issues.rudder.io/issues/25388

Also add zio test junit runner which is needed for making our tests visible


fanf:
Self merging, since it looks inocuitious (famous last words)

fanf:
OK, merging this PR

PR #5853, opened 2024-08-30
Fixes #25311: Remove expected reports from NodeStatusReport

fanf:
https://issues.rudder.io/issues/25311

Remove the big structure `NodeExpectedReports` that are kept in `NodeStatusReport`, in the `RunAndConfigInfo` part (ie the part where we keep information about the run, because we keep information about the run config information, ie config id and expected reports). Plus, current `RunAndConfigInfo` is too complexe for magnolia serialisation in any case. 

We don't need them appart for information about debugging/message that can be constructed without the whole structure. 

We also want to make the minimum changes and absoltly not touch at the compliance computation logic done in `ExcecutionBatch`. 

For that: 
- we duplicate the `NodeStatusReport` into a `NodeStatusReportInternal` so that nothing change in it, 
- that structure is translated into a newer, leaner `NodeStatusReport` where `RunAndConfigInfo` is replaced with a dumb case class `RunAnalysis`, easely serialisable, and not big ; the translation happens at the end of `ExecutionBatch#getNodeStatusReports` once we don't need the config info

```
final case class RunAnalysis(
    kind:                RunAnalysisKind,
    expectedConfigId:    Option[NodeConfigId],
    expectedConfigStart: Option[DateTime],
    expirationDateTime:  Option[DateTime],
    expiredSince:        Option[DateTime],
    lastRunDateTime:     Option[DateTime],
    lastRunConfigId:     Option[NodeConfigId],
    lastRunExpiration:   Option[DateTime]
)
```

The translation from old `RunAndConfigInfo` to corresponding `RunAnalysis` is done in each `RunAndConfigInfo` leaf case classes. I chose to keep the names of attributes in `RunAnalysis` instanciation to try to make it somehow readable. 

Then, it's "just" weaving the change everywhere, trying to not break the semantic of any of the optional values in `RunAnalysis`. 

Some part loose a bit of precision but avoid some duplication. For ex, we used to need to pattern match each case individially to build the debug log. Now, we just have one or two different ones, and build the string with `.fold("unknown")(...)` for each optional values. 

Tests pass, Rudder starts, but it will need real testing to check that it looks ok.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/DebugComplianceLogger.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/DebugComplianceLogger.scala
@@ -44,6 +44,7 @@ import com.normation.NamedZioLogger
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.reports.NodeConfigIdInfo
 import com.normation.rudder.domain.reports.NodeExpectedReports
+import com.normation.rudder.domain.reports.RunAnalysis
 import com.normation.rudder.reports.ChangesOnly
 import com.normation.rudder.reports.ComplianceMode
 import com.normation.rudder.reports.ResolvedAgentRunInterval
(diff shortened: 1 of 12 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ComplianceRepository.scala
```diff
@@ -123,37 +124,22 @@ class ComplianceJdbcRepository(
...
+          case _ =>
+            (r.runInfo.lastRunDateTime, r.runInfo.expirationDateTime) match {
+              case (Some(a), Some(b)) =>
+                Some(RunCompliance.from(a, b, r))
+              case (Some(a), None)    =>
+                r.runInfo.lastRunExpiration match {
+                  case Some(b) => Some(RunCompliance.from(a, b, r))
+                  case None    => Some(RunCompliance.from(a, a.plusMinutes(10), r))
+                }
+              case _                  => None
```
clarktsiory, line 141: We lost the comment that explains the `10`. Is it the same value as in `ReportsDisplayer` ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ComplianceRepository.scala
```diff
@@ -188,10 +174,10 @@ class ComplianceJdbcRepository(
       Nil
     }
 
-    val saveComplianceDetails = if (getSaveComplianceDetails().getOrElse(false)) {
+    val saveComplianceDetails: ConnectionIO[Unit] = if (getSaveComplianceDetails().getOrElse(false)) {
       val queryCompliance = """insert into nodecompliance (nodeid, runtimestamp, endoflife, runanalysis, summary, details)
                               | values (?, ?, ?, ?, ?, ?)""".stripMargin
-      Update[RunCompliance](queryCompliance).updateMany(runCompliances)
+      Update[RunCompliance](queryCompliance).updateMany(runCompliances).map(_ => ())
```
clarktsiory, line 180:
```suggestion
      Update[RunCompliance](queryCompliance).updateMany(runCompliances).void
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/DebugComplianceLogger.scala
```diff
@@ -86,6 +87,48 @@ object ComplianceDebugLogger extends Logger {
...
+        s"expected NodeConfigId: ${c.logId}|" +
+        s" last: ${c.logRun} ; expired but will be kept until ${c.expirationDateTime.fold("unknown")(_.toIsoStringNoMillis)}"
+
+      case ReportsDisabledInInterval =>
+        s"expected NodeConfigId: ${c.logId}|" +
+        s" last run: none available (compliance mode is reports-disabled)]"
+
+      case _ =>
+        s"expected NodeConfigId: ${c.logId} |" +
+        s" last run: ${c.logRun}"
```
clarktsiory, line 126: The last one has a space before `|`, all the other ones don't, which make regex harder than they already are... :innocent:

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/DebugComplianceLogger.scala
```diff
@@ -86,6 +87,48 @@ object ComplianceDebugLogger extends Logger {
...
+    def logId:  String = {
+      s"${c.expectedConfigId.fold("unknown id")(_.value)}/[${c.expectedConfigStart.fold("unknown start time")(
+          _.toIsoStringNoMillis
+        )}-${c.expirationDateTime.fold("unknown expiration date")(_.toIsoStringNoMillis)}]"
+    }
+    def logRun: String = {
+      s"${c.lastRunConfigId.fold("none or too old")(i => s"configId: ${i.value}")} received at: ${c.lastRunDateTime.fold("unknown")(
+          _.toIsoStringNoMillis
+        )}${c.lastRunExpiration.fold("")(r => s" | expired at ${r.toIsoStringNoMillis}")}"
+    }
```
clarktsiory, line 102: It's quite hard to see the pattern and the separating char with all the folds, maybe there should be local variables to makes things clearer ?

clarktsiory (commented):
LGTM, only a few remarks on code clarity

clarktsiory (approved):

PR #5857, opened 2024-09-04
Fixes #25409: User identity does not show up in bar when user has last login

clarktsiory:
https://issues.rudder.io/issues/25409

The name can be saved as an empty string, so we need to reject `Some("")` and display the _id_ instead

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/UserInformation.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/UserInformation.scala
@@ -67,7 +67,7 @@ class UserInformation extends DispatchSnippet with DefaultExtendableSnippet[User
       case Some(u) =>
         val displayName = userRepo.get(u.getUsername).runNow match {
           case None       => u.getUsername
-          case Some(info) => info.name.getOrElse(info.id)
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/UserInformation.scala
```diff
@@ -67,7 +67,7 @@ class UserInformation extends DispatchSnippet with DefaultExtendableSnippet[User
       case Some(u) =>
         val displayName = userRepo.get(u.getUsername).runNow match {
           case None       => u.getUsername
-          case Some(info) => info.name.getOrElse(info.id)
+          case Some(info) => info.name.filter(_.nonEmpty).getOrElse(info.id)
```
fanf, line 70: should test with `strip()` before `nonEmpty` no?

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/UserInformation.scala
clarktsiory: Absolutely :see_no_evil:

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #5863, opened 2024-09-07
Fixes #25429: When a mandatory variable has not value but a default one is configured, use it

fanf:
https://issues.rudder.io/issues/25429

This one use https://github.com/Normation/rudder/pull/5862

The main idea is that when we produce policy from a draft, we first try to assign default value to missing mandatory ones. 

This exposes a failing test that wasn't making sense: we have one test where we have null values for non-optionnal case. 
This test used to pass because we were only checking `v.values.exists(_ == "")` and not also for null values. I added "may be empty" for that test. 
(it should not change anything in real rudder, because we should never have a null value, and it is likely a scare from the past. Still, the test was false, the code was false, and the two problem were compensating each others)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
@@ -582,37 +582,56 @@ final case class BoundPolicyDraft(
    */
   def toPolicy(agent: AgentType): Either[String, Policy] = {
     PolicyTechnique.forAgent(technique, agent).flatMap { pt =>
-      expandedVars.values.collectFirst { case v if (!v.spec.constraint.mayBeEmpty && v.values.exists(_ == "")) => v } match {
-        case Some(v) =>
-          Left(
-            s"Error for policy for directive '${directiveName}' [${id.directiveId.debugString}] in rule '${ruleName}' [${id.ruleId.serialize}]: " +
-            s"a non optional value is missing for parameter '${v.spec.description}' [param ID: ${v.spec.name}]"
-          )
-        case None    =>
-          Right(
-            Policy(
-              id,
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -582,37 +582,56 @@ final case class BoundPolicyDraft(
    */
   def toPolicy(agent: AgentType): Either[String, Policy] = {
     PolicyTechnique.forAgent(technique, agent).flatMap { pt =>
-      expandedVars.values.collectFirst { case v if (!v.spec.constraint.mayBeEmpty && v.values.exists(_ == "")) => v } match {
```
fanf, line 585: here, we should test for null values too, because we have a test for null values. 
I'm almost sure we don't authorize null values since something like Rudder 2.4, but sill, I corrected the check to make it agree with the test

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -1110,7 +1110,7 @@ class TestNodeConfiguration(
   lazy val clockVariables: Map[ComponentId, Variable] = {
     val spec = clockTechnique.getAllVariableSpecs.map(s => (s.name, s)).toMap
     Seq(
-      spec("CLOCK_FQDNNTP").toVariable(Seq("true")),
+      spec("CLOCK_FQDNNTP").toVariable(Seq("")), // testing mandatory value with default "true"
```
fanf, line 1113: this is the main test: all the written policy files are the same while we provide an empty value for the mandatory variable which as a default constrain.

VinceMacBuche (approved):

PR #5864, opened 2024-09-07
Fixes #25430: Property sub-key is used in place of name for lookup

fanf:
https://issues.rudder.io/issues/25430

So, there was a lot of small things that got unaligned during the last changes. 
- I added a lot of ops logs that help understand what really happen on prod without a debugger on hand, 
- I renamed the PROP_KEY to PROP_NAME, because in rudder we use "property name" to mean the key (and so PROP_NAME becomes PROP_SUB_NAME)
- which lead to an error, I was looking for the property sub name in place of name to check the expiration policy, whiich was always empty (added a full integration test for that)
- I change the expiration date for computed `NoReportInInterval` in `ExecutionBatch` to never use "now", even when we don't know the date. Since we are persisting things in base now, we were continuously make the expiration date a bit latter, and so node with no reports were always OK if they had a grace duration congiured ; at each check, we were giving them a bit more time 
- there was a loop, because the semantic of what need to be chech for expiration changed: now, only good things need that, because only good things get altered with time and become worse. The bad status remains bad, they need an actual change of context (new run, new config)
- clean up a bit our default logback.xml 
Also, I tried to normalize date format by using only the human one for HTML view, and the rfc one for logs.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
@@ -39,6 +39,7 @@ package com.normation.rudder.services.reports
 
 import com.normation.errors.*
 import com.normation.inventory.domain.NodeId
+import com.normation.rudder.domain.logger.ComplianceLoggerPure
 import com.normation.rudder.domain.logger.ReportLogger
 import com.normation.rudder.domain.logger.ReportLoggerPure
 import com.normation.rudder.domain.logger.TimingDebugLoggerPure
@@ -57,6 +58,7 @@ import com.normation.rudder.reports.execution.RoReportsExecutionRepository
 import com.normation.rudder.repository.*
 import com.normation.rudder.score.ComplianceScoreEvent
 import com.normation.rudder.score.ScoreServiceManager
+import com.normation.utils.DateFormaterService
 import com.normation.zio.*
 import com.softwaremill.quicklens.*
 import org.joda.time.*
@@ -280,6 +282,9 @@ class ComputeNodeStatusReportServiceImpl(
   def saveUpdatedCompliance(newReports: Iterable[(NodeId, NodeStatusReport)]) = {
     for {
       now     <- currentTimeMillis
+      _       <- ReportLoggerPure.Cache.trace(
+                   s"Reports to save: ${newReports.map { case (id, r) => s"${id.value}: ${r.runInfo}" }.mkString("\n  ")}"
+                 )
       kept    <- updateKeepCompliance(new DateTime(now), newReports)
       updated <- nsrRepo.saveNodeStatusReports(kept)(ChangeContext.newForRudder())
       // exec hooks
@@ -287,6 +292,9 @@ class ComputeNodeStatusReportServiceImpl(
       _       <- ZIO.foreachDiscard(hooks)(_.onUpdate(updated))
       _       <- ReportLoggerPure.Cache.debug(
                    s"Compliance recomputed, saved and score updated for nodes: ${updated.map(_._1.value).mkString(", ")}"
+                 )
+      _       <- ReportLoggerPure.Cache.trace(
+                   s"Reports actually saved: ${updated.map { case (id, r) => s"${id.value}: ${r.runInfo}" }.mkString("\n  ")}"
                  )
     } yield ()
   }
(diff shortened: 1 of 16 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -794,14 +794,10 @@ object ExecutionBatch extends Loggable {
...
-                      // expiration time is based on now as run is invalid and we can't get proper data
-                      val complianceComputationExpirationTime =
-                        computeExpirationDate(now, currentConfig.n.agentRun, currentConfig.complianceMode)
-
                       // take care of the potential case where currentConfig != runConfig in the log message
                       runType(
-                        s"Last run at ${t} is for the configId ${runConfig.nodeConfigId.value} but a new one should have been sent for configId ${currentConfig.nodeConfigId.value} before ${expirationTime}",
-                        NoReportInInterval(currentConfig, complianceComputationExpirationTime)
+                        s"last run at ${t} is for the configId ${runConfig.nodeConfigId.value} but a new one should have been sent for configId ${currentConfig.nodeConfigId.value} before ${expirationTime}",
+                        NoReportInInterval(currentConfig, expirationTime)
```
fanf, line 800: never depend on "now" for expiration date

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -759,17 +759,17 @@ object ExecutionBatch extends Loggable {
...
-                // expiration time is based on now as run is invalid and we can't get proper data
-                val complianceComputationExpirationTime =
-                  computeExpirationDate(now, currentConfig.n.agentRun, currentConfig.complianceMode)
                 // If the run has expired, consider that no report were sent
                 if (expirationTime.isBefore(now)) {
                   runType(
-                    s"Last run at ${t} is for the configId ${rv.value} but a new one should have been sent for configId ${currentConfig.nodeConfigId.value} before ${expirationTime}",
-                    NoReportInInterval(currentConfig, complianceComputationExpirationTime)
+                    s"last run at ${t} is for the configId ${rv.value} but a new one should have been sent for configId ${currentConfig.nodeConfigId.value} before ${expirationTime}",
+                    NoReportInInterval(currentConfig, expirationTime)
```
fanf, line 766: never depend on "now" for expiration date

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
```diff
@@ -391,18 +414,25 @@ class ComputeNodeStatusReportServiceImpl(
...
           compliance.runInfo.kind match {
-            // here, we have a special case for unexpected version: it is useless to recompute compliance until we don't have a new run,
-            // ie the node config was changed elsewhere. It means that "unexpected version" wins above "No report in interval",
-            // ie that that error is bigger.
-            case RunAnalysisKind.UnexpectedVersion => false
-            // other expiring status
-            case _                                 =>
+            // here, we only want to check for compliance that expires when time pass and we
+            // were in a good state. Bad states need an external action (new config, new runs)
+            // to change back to good state
```
fanf, line 424: this is the main loop-breaker. 
Now that things are retrieved from base as a cache, we really just have two kind of things : 
- the one that change just with time - and unfortunatelly, things only get worse with time, it doesn't magically repair a bad case, 
- things that need an external action (new config, new run) to change. 
Here, we only care for the first one, because we want to check that we didn't let things rotting in base. 
Also, I enumerated all cases, because it will be a point of attention to have if we ever had a new status, I don't want to let it be caught by a "*" where it doesn't belong.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
```diff
@@ -161,17 +161,7 @@ class NodeStatusReportRepositoryImpl(
                              case Some(e) =>
                                e.runInfo.kind match {
                                  case RunAnalysisKind.ComputeCompliance => // keep existing compliance, change run info
-                                   Some(
-                                     (
-                                       id,
-                                       e.modify(_.runInfo.lastRunExpiration)
```
fanf, line 167: oupsy, forgot the kind. So in fact, we just want to keep of the run analysis.

webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
```diff
@@ -310,17 +310,6 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
...
-  <!--
-    Cache (nodes, compliance)
-    =========================
-    Information on cache, interesting for Rudder 3.0 and up
-    Set to debug or trace to have information.
-  -->
-  <logger name="com.normation.rudder.services.reports.CachedReportingServiceImpl" level="info" />
-  <logger name="com.normation.rudder.services.reports.CachedNodeChangesServiceImpl" level="info" />
-  <logger name="com.normation.rudder.services.nodes.NodeInfoServiceCachedImpl" level="info"/>
-
```
fanf, line 322: these things don't exists anymore

webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
```diff
@@ -356,6 +345,15 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
...
   </logger>
 
+
+  <!--
+    Cache (nodes, compliance)
+    =========================
+    Information on cache, interesting for Rudder 3.0 and up
+    Set to debug or trace to have information.
+  -->
+  <logger name="compliance" level="info" />
```
fanf, line 355: this one does exist now

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
```diff
@@ -177,10 +173,8 @@ response:
                   ]
                 },
                 "origval" : {
-                  "rudder" : {
```
fanf, line 180: that error is not really relevant (the value is created by hand in `MockService` above), but I wanted it to look like the real property, so that if people look at the test, they see something good

clarktsiory (approved):
LGTM !

PR #5866, opened 2024-09-09
Fixes #25407: 

clarktsiory:
https://issues.rudder.io/issues/25407

We previously had a cache of user statuses only, that we processed upon every request (in the `doFilter`) because the status value is sufficient to guess if the user session should be invalidated.

But now, the state management is more complex, since we need to handle a *change* of password and rights, instead of their value alone. The diff would needs to be computed and processed before storing into the cache, hence another cache model with `ValidUser` and `InvalidUser` within the cache, the invalidity being declined into 3 different reasons : invalid status, password change and rights change.
Also, since multiple changes could occur at once, before the session of the user in question is checked, we need to process changes and guarantee to always keep the final invalid state with the most up-to-date field values (see the unit tests).

We also enforce that admin users (admin_write/edit and any_rights) are not affected and remain logged

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
@@ -60,24 +60,26 @@ import zio.syntax.*
  */
 class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailListProvider: FileUserDetailListProvider)
     extends OncePerRequestFilter {
-  private val userStatuses = Ref
-    .make(Map.empty[String, UserStatus])
-    .map(ref => {
+  import UserSessionInvalidationFilter.*
 
-      userDetailListProvider.registerCallback(
-        RudderAuthorizationFileReloadCallback(
-          "user-session-invalidation",
-          (users: ValidatedUserList) => {
-            userRepository
-              .getStatuses(users.users.keys.toList)
-              .flatMap(ref.set)
-          }
+  private val userCache = {
+    Ref
+      .make(Map.empty[String, Int])
+      .map(ref => {
+
+        userDetailListProvider.registerCallback(
+          RudderAuthorizationFileReloadCallback(
+            "user-session-invalidation",
+            (users: ValidatedUserList) => {
+              userRepository.getStatuses(users.users.keys.toList).flatMap(updateUsers(ref, _)(users.users))
+            }
+          )
         )
-      )
 
-      ref
-    })
-    .runNow
+        ref
+      })
+      .runNow
+  }
 
   @throws[ServletException]
   @throws[IOException]
@@ -89,24 +91,25 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
         val userDetails = auth.getPrincipal
         userDetails match { // we should only do session invalidation in specific cases : status is disabled/deleted, user is unknown
           case user: RudderUserDetail =>
-            val status =
-              userStatuses.get.map(_.getOrElse(user.getUsername, UserStatus.Deleted)) // unknown user : falls back to deleted
+            val username   = user.getUsername
+            val cachedUser = userCache.get.map(_.get(username)).runNow
+            implicit val userDetail: RudderUserDetail = user
 
-            (status.flatMap {
-              case status if status.in(UserStatus.Disabled, UserStatus.Deleted) =>
-                val endSessionReason = s"Session invalidated for ${status.value} user"
+            (cachedUser match {
+              case checkUser(reason) =>
+                val endSessionReason = s"Session invalidated because ${reason}"
                 IOResult
                   .attempt(session.invalidate())
                   .foldZIO(
                     {
                       case SystemError(_, _: IllegalStateException) =>
                         ApplicationLoggerPure.info(
-                          s"User session for user '${user.getUsername}' is already invalidated for user with status '${status.value}'"
+                          s"User session for user '${username}' is already invalidated because : ${reason}"
                         )
                       case err                                      =>
                         err
                           .copy(msg = {
-                            s"User session for user '${user.getUsername}' could not be invalidated. " +
+                            s"User session for user '${username}' could not be invalidated for reason : ${reason}. " +
                             s"Please contact Rudder developers with the following explanation : ${err.fullMsg}"
                           })
                           .fail
@@ -118,11 +121,11 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
                         endSessionReason
                       ) *>
                       ApplicationLoggerPure.info(
-                        s"User session for user '${user.getUsername}' is invalidated because user has status '${status.value}'"
+                        s"User session for user '${username}' is invalidated because : ${reason}"
                       )
                     }
                   )
-              case _                                                            =>
+              case _                 =>
                 ZIO.unit
             }).runNow
           case _ => ()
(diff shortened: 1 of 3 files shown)


clarktsiory:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -132,3 +138,136 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
   override protected def shouldNotFilterErrorDispatch = false
 
 }
+
+private[liftweb] object UserSessionInvalidationFilter {
+
+  // Avoid copy method because changing state may cause invalidation
+  sealed abstract class LoggedUser(val status: UserStatus, val password: String) {
+    def withStatus(newStatus:     UserStatus): LoggedUser
```
fanf, line 146: You should not keep a clear form of the password in ram, it's too easy to get access to it (for some value of easy, but let say it's a good practice to not hold that on ram)

fanf: (saying that, I think we already have it in memory for the user details).

clarktsiory: Yes we already had it for the User account also. What we could do here is to make a value class, and also override the toString method just in case it is logged

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
clarktsiory: => https://github.com/Normation/rudder/pull/5866/commits/ddd30387f42ccc96481710b0cf6e52c723390097

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -132,3 +139,227 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
...
+   *
+   * It must exist, and the order is a conscious choice, because many changes can occur at once,
+   * so to avoid keeping an array of reasons of invalidity with still all the states, we just choose one
+   * (there are as many states i.e. fields to keep, as the number of invalidity reasons, with more it could become tedious).
+   * When updating states, the "status" is always checked because the criteria is absolute.
+   *
+   * Implementation of state update assumes that states are updated in the same order.
+   * The absolute check for the "status" is evaluated every time
+   */
+  sealed trait InvalidUser { self: LoggedUser =>
```
clarktsiory, line 215: A bit more details of the ADT hierarchy here

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
fanf: Why do we keep it for User account ? Once authenticated, we don't care anymore. 
But ok, other subject. At least, we should not add more of them at all everywhere.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -60,24 +60,24 @@ import zio.syntax.*
...
-          (users: ValidatedUserList) => {
-            userRepository
-              .getStatuses(users.users.keys.toList)
-              .flatMap(ref.set)
-          }
-        )
+  import UserSessionInvalidationFilter.*
+
+  private val userCache = {
+    val ref: Array[(String, Int)] = new Array(USERS_REF_SIZE) // an array of users, there are not that much users in Rudder
```
fanf, line 66: Actually you should use a Chunk. It's an array in disguise, but the sizing and rising is done for you, and you don't have to use arbitrary limits that will explode one day and we will be very sad, because a new version of rudder will be needed for that

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -60,24 +60,24 @@ import zio.syntax.*
...
+    // avoid nulls, no user has empty login anyway
+    (0 until USERS_REF_SIZE).foreach(ref(_) = ("", 0))
```
fanf, line 68: Iirc chunk as zeroing init function.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
fanf: And actually : if the array is not of a simple Java type, it won't be memory compact, because Java is a sad place that needs to wait for the Valhalla. So you can use a map, it's basically the same thing, an array of pointers, and it will easier to understand the code (and I don't think we need to go the path of two arrays and use index for lookup before actually measuring the thing as an hot spot)

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -132,3 +132,86 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
...
+
+  /**
+    * Match an user in cache : if no user, it means its cache has changed, with a known or unknown cause of invalidity
+    * if some user, it's ok, unless the last known status is invalid
+    */
+  object checkUser {
+    def unapply(user: Option[(String, Int)]): Option[String] = {
+      user match {
+        case None                                => Some("user access to Rudder has been updated")
+        case Some((_, HASH_USER_INVALID_STATUS)) => Some("user status has been updated to an invalid one")
```
fanf, line 148: Shouldn't it be any 0 or negative value different from -1?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -132,3 +132,86 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
...
+        } else {
+          None
+        }
+    }
+    // we just need to have another array that would be copied to the final target
+    val sourceArray    = (fileUsers ++ validatedUsers).toArray
+    IOResult.attempt {
+      Array.copy(sourceArray, 0, ref, 0, sourceArray.length)
+    }.chainError(
+      s"User sessions cache may have been corrupted, having more than ${USERS_REF_SIZE} users in Rudder is not supported"
```
fanf, line 213: We never want that kind of case, the cost of having someone out reaching it is too high.
if really it's important, at least you need to use a file/jvm based property that can be changed without a recompilation. The user need to have the agency without you to workaround the limit.

fanf (commented):
That's much better, but I think we can simplify it even more

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
clarktsiory: :+1: I changed it back to a `Ref[Map[String, Int]]` which is indeed a lot easier to manipulate !

clarktsiory: https://github.com/Normation/rudder/pull/5866/commits/22713762017fed7cc9007b19867c39bdbc81ee74

clarktsiory:
PR updated with a new commit

fanf (approved):
:heart:

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/UserSessionInvalidationFilter.scala
```diff
@@ -146,13 +149,89 @@ class UserSessionInvalidationFilter(userRepository: UserRepository, userDetailLi
...
+   * Match an user in cache : if no user, it means its cache has changed, with a known or unknown cause of invalidity
+   * if some user, it's ok, unless the last known status is invalid
+   */
+  object checkUser {
+    def unapply(user: Option[Int])(implicit userDetail: RudderUserDetail): Option[String] = {
+      user match {
+        case None                                                   => Some("user is unknown")
+        case Some(HASH_USER_INVALID_STATUS)                         => Some("user status has been updated to an invalid one")
+        // user hash has changed
+        case Some(hash) if hashRudderUserDetail(userDetail) != hash => Some("user access to Rudder has been updated")
```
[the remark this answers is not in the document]
clarktsiory: not really because the admin hash itself is stored when cache is reloaded. In fact the case for the admin would be merged with the last one (resulting in `None`), but to mark it as special I kept it.
As discussed, I put right below the status case instead : https://github.com/Normation/rudder/pull/5866/commits/2023ab520b281a3d4251e4b6721a0c45465e9187

clarktsiory:
OK, squash merging this PR

PR #5872, opened 2024-09-12
Fixes #25456: Update diesel to mitigate RUSTSEC-2024-0365 

VinceMacBuche:
https://issues.rudder.io/issues/25456


VinceMacBuche:
To update diesel we also need to update rust version to at least 1.78.1 in 8.1 we use 1.74, I don't think we can change rust version like this ? if we can't our only way is to ignore it for version 8.1 (diesel is up to date in 8.2)

amousset (approved):

VinceMacBuche:
OK, merging this PR

PR #5875, opened 2024-09-12
Fixes #25463: Session is still invalidated instantly for admin and does not redirect to login

clarktsiory:
https://issues.rudder.io/issues/25463

* admin check is just more important than status check every time : now admin can still pursue with their session, even new admins 
* forcing redirect to login avoids weird behavior, but also prevents notifications to happen, there is something to change later in our setup of `noCometSessionCmd`


fanf (approved):
Lgtm

PR #5876, opened 2024-09-13
Fixes #25461: A user should not be able to disable themselves in user management 8.2

clarktsiory:
https://issues.rudder.io/issues/25461

Same as https://github.com/Normation/rudder-plugins/pull/752, with a simpler `userManagement.html` without unused classes


fanf (approved):

clarktsiory:
PR rebased

PR #5878, opened 2024-09-16
Fixes #25469: Users from external providers manage to log in but their session is invalidated instantly 

clarktsiory:
https://issues.rudder.io/issues/25469

The users that log in into rudder may come from the users file or from external providers (in which case their passwords are roles are not known until login in Rudder).
We need to update the users cache with all known users from the database, by getting all users with their status, in the existing file reload callback.
Also, for newly provisioned users, the cache will need to be updated with the latest user definition, independently from the file reload callback : it needs to be done at the end of the _authenticate_ mechanism


fanf (approved):
LGTM!

PR #5880, opened 2024-09-17
Fixes #25479: Users cleanup configuration is too strict on disabled users

clarktsiory:
https://issues.rudder.io/issues/25479

* Add an appropriate log with the `application.authentication` name and use it in the `CleanupUsers` file
* Improve logs globally : add some missing logs (for disable/purged) and use `zio.Duration#render` to display the date
* Only apply deletion of inactive only on disabled users

As for the properties file :
* Allow to parse "never" as `zio.Duration.Infinity`, when used in the config the existing logic of the code is still valid (eventually the code is disabling/deletion is never run)
* Improve the comments and update default to `90 days` for `disableAfterLastLogin`


fanf (approved):
Lgtm.
in an other pr, we will need to merge all logs toward "application.authentication".

PR #5883, opened 2024-09-18
Fixes #25478: Normalize authentication logs

clarktsiory:
https://issues.rudder.io/issues/25478

Just removing `ApplicationLoggerPure.Authz` and `ApplicationLoggerPure.User` definitions, and replacing all usages by `ApplicationLoggerPure.Auth` to make it compile and use the `application.authentication` logger

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
@@ -416,6 +416,12 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
     <appender-ref ref="STDOUT" />
   </logger>
 
+  <!--
+    Log information about user authentication and user-related access to Rudder 
+    (rights, session, user management in general).
+  -->
+  <!-- Uncomment to have precise log about user management and authentication -->
+  <!-- <logger name="application.authentication" level="debug"/> -->
 
   <!--
     Schechuled Jobs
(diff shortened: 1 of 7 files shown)


fanf:
you also need to update the provided `logback.xml` and the comments in it

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
```diff
@@ -416,6 +416,13 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
     <appender-ref ref="STDOUT" />
   </logger>
 
+  <!--
+    Log information about user authentication and user-related access to Rudder 
+    (rights, session, user management in general).
+  -->
+  <logger name="application.authentication" level="info" additivity="false">
+    <appender-ref ref="STDOUT" />
```
fanf, line 424: you don't need the complicated part when you only give stdout, since it is the default appender. 
Plus for logger with parents (logback is like log4j & all hierachized by logger doted names), you inherite parent conf. 

So just `<logger name="application.authentication" level="info" />` is enough and likely the config behavior we want.

fanf: Plus, we should comment it out by default, so that it follows the level of `application`, which is consistant with what we want

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
clarktsiory: Right, it is now commented by default to  "debug" level : https://github.com/Normation/rudder/pull/5883/commits/657e2d019d03ad1bb58d8ea53ba2a8fac022a398

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #5893, opened 2024-09-20
Fixes #25476: API don't include acceptedSince information

fanf:
https://issues.rudder.io/issues/25476

Add the `acceptanceDate` (when the node was accepted in Rudder) in the default pool of node details properties. 

Add `scheduleOverride` in the `managementTechnologyDetails` with details of the overridden values. 
It's in details, and only for the overriding case, because it was simpler technically to do so (and upmerge will be complicated), it doesn't seem to be asked a lot, and it will be possible to later add the run schedule in a more general section if we want (and in 8.2 and up, it will be easier to do so). 

I used `nodeFact` since we have it in 8.2, but in any cases, the upmerge will be interesting.


clarktsiory (approved):
LGTM ! :crossed_fingers: for the upmerge with zio-json !

fanf:
OK, merging this PR

PR #5900, opened 2024-09-26
Fixes #25549: Technique library is only reloaded once after boot and not periodically

clarktsiory:
https://issues.rudder.io/issues/25549

The lift actor scheduler is a "Java Future scheduler", it only limited to doing it once and never again...
We should schedule with zio instead


clarktsiory:
It was reloaded but logs needed to be at trace level to see it.
The actor schedules to message itself every update interval

PR #5903, opened 2024-09-26
Fixes #25555: 

amousset:
https://issues.rudder.io/issues/25555

The problem was that the log message used the raw token and not the typed `ApiToken` (which has a custom `toString`). But it is useful to be able to identify an unknown token, and we can afford to lose a bit of randomness.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
@@ -60,6 +60,9 @@ final case class ApiAccountName(value: String) extends AnyVal
 /**
  * The actual authentication token.
  *
+ * TODO: Once support for plain text tokens is dropped, make separate types for plain and hashed tokens.
+         Current situation is confusing, and hence a bit risky.
+ *
  * There are two versions of tokens:
  *
  * * v1: 32 alphanumeric characters stored as clear text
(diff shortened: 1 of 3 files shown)


amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -80,6 +83,13 @@ case class ApiToken(value: String) extends AnyVal {
   // Avoid printing the value in logs, regardless of token type
   override def toString: String = s"[REDACTED ApiToken]"
 
+  // For cases we need to print a part of the plain token for debug.
+  // Show the first 4 chars: enough to disambiguate, and preserves 166 bits of randomness.
+  def exposeSecretBeginning: String = {
+    val acceptableToReveal = 4
+    value.take(acceptableToReveal) + '*'.toString() * (value.size - acceptableToReveal)
```
fanf, line 90: You don't actually need to use the real length of * here, it might be already a valuable information. A fixed 6 * or whatever would be better.

amousset: Yes, I wondered about it. As it is a Rudder token is should always be 32 chars long, and I wanted to avoid provided "false information" about the length that would hamper debugging. I'll try to find a suitable format.

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
amousset: Updated with a bit different solution

fanf (approved):

fanf:
OK, squash merging this PR

PR #5905, opened 2024-09-27
Fixes #25527: List techniques with compilation failure in bar for generation status

clarktsiory:
https://issues.rudder.io/issues/25527

* Create a `TechniqueCompilationStatusService` that **reloads** the global compilation status : 
  * it is done every 5 minutes by default (`rudder.batch.techniqueLibrary.updateInterval` property)
  * it is reloaded at startup with a new bootstrap check
  * it is updated on individual techniques after every update of a techniques from a REST API call
* We call the lift actor that is responsible for updating the UI directly in the method that uploads the status that is computed from a cache. The architecture is a bit weird, but it prevents having to manually call the actor at every place which is also weird (call an UI actor in a REST API method ?). Later we would be exposing the status information in a dedicated endpoint so that we no longer need a stateful actor in our layers. 

The architecture of components looks like this : 
```mermaid
graph LR
   subgraph Technique_hooks
     CheckTechniqueCompilationStatus
     CheckTechniqueLibrary
     MigrateJsonTechniquesToYaml
   end
   Technique_hooks --get()-->TCSS
   
	 subgraph Techniques_core
     TCSS[TechniqueCompilationStatusService] --update--> TCEC[TechniqueCompilationErrorsCache]
   end
   TCEC --. ! UpdateCompilationStatus--> ADA
   
   subgraph Actors
     ADA((AsyncDeploymentActor)) --. ! AsyncDeploymentActorCreateUpdate--> AD((AsyncDeployment))
   end
   
   AD --> UI(UI)
```

UI : 
Peek 2024-09-30 12-02

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
@@ -49,6 +49,10 @@ import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.eventlog.*
 import com.normation.rudder.domain.logger.PolicyGenerationLogger
 import com.normation.rudder.domain.logger.PolicyGenerationLoggerPure
+import com.normation.rudder.domain.logger.StatusLoggerPure
+import com.normation.rudder.ncf.CompilationStatus
+import com.normation.rudder.ncf.CompilationStatusAllSuccess
+import com.normation.rudder.ncf.CompilationStatusErrors
 import com.normation.rudder.services.eventlog.EventLogDeploymentService
 import com.normation.rudder.services.marshalling.DeploymentStatusSerialisation
 import com.normation.rudder.services.policies.PromiseGenerationService
@@ -135,6 +139,15 @@ final case class DeploymentStatus(
     processing: DeployerState
 )
 
+case class UpdateCompilationStatus(
+    status: CompilationStatus
+)
+
+final case class AsyncDeploymentActorCreateUpdate(
+    deploymentStatus:  DeploymentStatus,
+    compilationStatus: CompilationStatus
+)
+
 /**
  * Async version of the deployment service.
  */
@@ -173,6 +186,7 @@ final class AsyncDeploymentActor(
   private case class NewDeployment(id: Long, modId: ModificationId, started: DateTime, actor: EventActor, eventLogId: Int)
 
   private var lastFinishedDeployement: CurrentDeploymentStatus = getLastFinishedDeployment
+  private var compilationStatus:       CompilationStatus       = CompilationStatusAllSuccess
   private var currentDeployerState:    DeployerState           = IdleDeployer
   private var currentDeploymentId = lastFinishedDeployement match {
     case NoStatus => 0L
@@ -202,7 +216,9 @@ final class AsyncDeploymentActor(
   /**
    * Manage what we send on other listener actors
    */
-  override def createUpdate: DeploymentStatus = DeploymentStatus(lastFinishedDeployement, currentDeployerState)
+  override def createUpdate: AsyncDeploymentActorCreateUpdate = {
+    AsyncDeploymentActorCreateUpdate(DeploymentStatus(lastFinishedDeployement, currentDeployerState), compilationStatus)
+  }
 
   private def WithDetails(xml: NodeSeq)(implicit actor: EventActor, reason: Option[String] = None) = {
     EventLogDetails(
@@ -434,6 +450,27 @@ final class AsyncDeploymentActor(
       }
       // update listeners
       updateListeners()
+    }
+
+    //
+    // Updating state from technique compilation output result
+    //
+    case UpdateCompilationStatus(status)                                                   => {
+      StatusLoggerPure.Techniques.logEffect.trace(
+        s"Compilation status has a new update command with payload : ${status}"
+      )
+      if (compilationStatus != status) {
+        compilationStatus = status
+        status match {
+          case CompilationStatusErrors(techniquesInError) =>
+            StatusLoggerPure.Techniques.logEffect.warn(s"Status of technique compilation has ${techniquesInError.size} errors")
+          case CompilationStatusAllSuccess                =>
+            StatusLoggerPure.Techniques.logEffect.debug(
+              s"Status of technique compilation has changed to success from previous : ${compilationStatus}"
+            )
+        }
+        updateListeners()
+      }
     }
 
     //
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompilationCache.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompilationCache.scala
@@ -1,0 +1,224 @@
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.ncf
+
+import com.normation.errors.IOResult
+import com.normation.inventory.domain.Version
+import com.normation.rudder.batch.UpdateCompilationStatus
+import com.normation.rudder.domain.logger.StatusLoggerPure
+import net.liftweb.common.SimpleActor
+import zio.*
+
+sealed trait CompilationResult
+object CompilationResult {
+  case object Success                extends CompilationResult
+  case class Error(errorMsg: String) extends CompilationResult
+}
+
+case class EditorTechniqueCompilationResult(
+    id:      BundleName,
+    version: Version,
+    name:    String,
+    result:  CompilationResult
+)
+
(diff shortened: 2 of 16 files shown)


clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ConfigurationStatusLogger.scala
```diff
@@ -0,0 +1,47 @@
...
+ */
+package com.normation.rudder.domain.logger
+
+import com.normation.NamedZioLogger
+
+/**
+ * Applicative pure logger for configuration status that may have vary in Rudder, 
+ * from success to errors in different configuration management parts of Rudder
+ */
+object ConfigurationStatusLoggerPure extends NamedZioLogger {
```
fanf, line 45: it should be a sublogger of the generation one I believe, more than of status

clarktsiory: the thing is, the generation loggers start with `policy.`

I thought that I would write logs under the same logger for node properties errors, but I was not convinced that all node properties error and technique compilation errors should be logged with the `policy.` namespace...

(Also there is already [a `node.properties` logger](https://github.com/Normation/rudder/blob/branches/rudder/8.2/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/logger/ApplicationLogger.scala#L144), maybe this should be the one used for node properties errors ?)

clarktsiory: Also I was already think about a separate web page that would have all the global information about policies, errors in nodes, techniques, etc. and I would have named it "Configuration status", that is why I chose the same name for the logger

fanf: Loggers are by applicative units. A "status" logger makes sense if we have a part of rudder dealing with statuses. 

So ok, let say we have status as an application unit, different from policy generation, and that it deals only with the status related logic, not what the status is for. 

Still, you should have either a general `status`, or a `status.techniques` (with a `status.generation` and a `status.properties`), and not configuration, that is the whole thing of managing configurations.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
```diff
@@ -86,29 +86,13 @@ trait TechniqueWriter {
...
- *
- * Finally, when `rudderc` is used, the webapp will look for its return code and generated files.
- * When the return code is XXXX, or if generated files doesn't pass a sanity check, then
- * the webapp is used to generate things. We also update the `compilation.yaml` to reflect the fact
- * that the last compilation used a fallback app and the error (if any)
- *
- * In that two last cases, then a big red persistent message is displayed in the technique editor
- * with the fact that a fallback was used.
- * In the last case, in addition, we need to display to the user the problem that they need to report
- * to rudder dev.
```
fanf, line 104: why removing the whole comment ? If (parts of) is are outdate, you should rewrite and enhanced them with the real state of things, not just remove.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -405,7 +405,7 @@ class TechniqueArchiverImpl(
 
     (for {
       res       <- techniqueCompiler.migrateCompileIfNeeded(techniquePath)
-      _         <- ZIO.when(res.resultCode != 0) {
+      _         <- ZIO.when(res.isError) {
```
fanf, line 408: :+1:

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -886,14 +885,12 @@ class TechniqueAPIService14(
...
     import zio.json.*
-    import zio.json.yaml.DecoderYamlOps
     import TechniqueCompilationIO.codecTechniqueCompilationOutput
     import com.normation.zio.*
-    val outputFile = techniqueCompiler.getCompilationOutputFile(editorTechnique)
-    val json       = (for {
-      content <- IOResult.attempt("error when reading compilation output")(outputFile.contentAsString(StandardCharsets.UTF_8))
-      out     <- content.fromYaml[TechniqueCompilationOutput].toIO
-      json    <- out.toJsonAST.toIO
+    val output = techniqueCompiler.getCompilationOutput(editorTechnique)
```
fanf, line 890: I don't understand why output is declared like that ? Why not in the for ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompilationCache.scala
```diff
@@ -0,0 +1,167 @@
...
+/**
+  * Service to update global technique compilation status and get the latest one from filesystem
+  */
+class TechniqueCompilationStatusService(
+    readTechniques:           EditorTechniqueReader,
+    techniqueCompiler:        TechniqueCompiler,
+    compilationStatusService: WriteTechniqueCompilationStatusService
+) extends ReadTechniqueCompilationStatusService {
+
+  override def get(): IOResult[CompilationStatus] = {
```
fanf, line 86: I'm not at ease with that design decision. Here, you have something that is neither functional (ie a kind of pure data pipe function, even if it can fail) or purely stateful (ie with a side effect). 
It means that future usage won't ever know if they need to find of the use place for that service and pipe its result, add a service in the constructor, can use it without changing the world, etc. 

I now get the meaning of that service is "look-up for all compilation output, apply some logic to find the error ones, and notify the interested parties". Not, as I though earlier, "an aggregation of things to build a more complicated logic, perhaps getting data from different points". 
It's almost the second (the implementation is a shiny `for` in disguise), but not exactly, because each time it's called, it has a side effect. So it's not a "read", it's a repository that holds a state - indirectly through the write service, but the write service is mandatory and very specific. 

So, two ways to get out of that: 

- option 1: inverse the read/write relationship (ie write has in its constructor , and make the read just a "gather data / transform data / produce new data" - each time the method is called. 
- option 2: add callback to read, get return `IOResult[Unit]`, write is just one callback

I think the first is better in that case. The second is better when you have a real repository.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompilationCache.scala
```diff
@@ -0,0 +1,167 @@
...
+  * This cache is a simple in-memory one which only saves errors, 
+  * so it saves only compilation output stderr message in case the compilation failed.
+  * It notifies the lift actor on update of the compilation status.
+  * 
+  * It is used mainly to get errors in the Rudder UI, and is updated when API requests to update techniques are made,
+  * when technique library is reloaded
+  */
+class TechniqueCompilationErrorsCache(
+    actor: SimpleActor[UpdateCompilationStatus],
+    ref:   Ref[Map[(BundleName, Version), EditorTechniqueError]]
```
fanf, line 132: for properties part of an API like to constructor attribute, try to name things for what they model, not what they are: here, it's the `errorBase` (even if it's implemented with a ref)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
clarktsiory: I didn't see the `rudderc` part and I thought the documentation in the parent interface was already sufficient, but yes after double-checking, it's missing things about the FS like compilation-output.yaml

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
clarktsiory: because... the compiler didn't complain :see_no_evil:

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueWriter.scala
clarktsiory: :white_check_mark: https://github.com/Normation/rudder/pull/5905/commits/ed28eef83afc9fb4aef84d7d020a68ce36b4d63f

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompilationCache.scala
clarktsiory: :white_check_mark: https://github.com/Normation/rudder/pull/5905/commits/ed28eef83afc9fb4aef84d7d020a68ce36b4d63f

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
clarktsiory: :white_check_mark: https://github.com/Normation/rudder/pull/5905/commits/ed28eef83afc9fb4aef84d7d020a68ce36b4d63f

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
```diff
@@ -436,6 +452,27 @@ final class AsyncDeploymentActor(
...
+    //
+    case UpdateCompilationStatus(status)                                                   => {
+      StatusLoggerPure.Techniques.logEffect.trace(
+        s"Compilation status has a new update command with payload : ${status}"
+      )
+      if (compilationStatus != status) {
+        compilationStatus = status
+        status match {
+          case CompilationStatusErrors(techniquesInError) =>
+            StatusLoggerPure.Techniques.logEffect.info(s"Status of technique compilation has ${techniquesInError.size} errors")
```
fanf, line 466: I think it should be error level. Or at least warn. 
Yes, warn. It's not a rudder bug, but something that is not normal (ie not the normal error flow of rudder, technique should always compile).

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/AsyncDeploymentActor.scala
clarktsiory: yes that is right :white_check_mark: https://github.com/Normation/rudder/pull/5905/commits/394d64c80db4aa00c84fe5793449093c3b48ab76

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM

fanf:
OK, squash merging this PR

PR #5906, opened 2024-09-27
Fixes #25557: Spaces as a default value provided by node properties are now refused

fanf:
https://issues.rudder.io/issues/25557

So, this is a bit complicated, because the generation process is a bit complicated and an accumulation of decades of code rewrite and underlying model change. 

The problem was that at some point (in `buildTechniqueBundles`), we had a `optionVariableValue.getOrElse("")`. This was late in the process and bypassed the other constraint checks. Perhaps that it used to be ok, since "" and "not defined" are roughtly the same, but it breaks on edge cases.  
And when I added a check for the default value testing on `isBlank` for the default value here: https://github.com/Normation/rudder/pull/5863/files#diff-e3d1394b948db4d6e9788c933762300d3e99930f208f2e5ca98735aa329cf7abR594 , the default value was not used and so the "" was used in a mandatory variable.

_**The big understanding is: we should never have reached that point and use `getOrElse`. We should never get pass `BoundPolicyDraft.toPolicy` and have any constrain not validated.**_

This method is the one we call once we have done all the node configuration merge (including system variable and everything). So after that point, variable values are not modified anymore appart for actually writing them where agent specific escaping can happen. 

So, that PR is here to clean up a bit our decade old mess and: 
- add a method `getAllVariableSpecs: Map[ComponentId, VariableSpec]` in `Technique` to get ALL variable specs with their component ID. It was computed at several place, not always the same (especially in tests)
- centralize in `BoundPolicyDraft.toPolicy` function all the checks on mandatory variable, missing one, and fallback when default value is provided
- remove the same checks in other places

After a lot of unraveling, everything seems to work: 
- there is no change appart the one wanted in all the tests writing policies or building configs
- actually, some tests were re-enabled and corrected
- I hurt my feet in all the possible combination of missing system variables, having too many variable compared to spec, missing tracking key, not having the correct values, so it looks like our tests have a good coverture now. 
- I tested in my dev env to generate policies with that change and in a standard 8.1 branch, putting in git the generated files, and the only changes are dates and configId. 
- I think it's the first time I understand the whole variable generation process. It took only 15y :shrug: And it's starting to be ok simple. 

NOTE: I still don't use the ComponentId to merge variables, but the variable spec name. It's How It Is Done In Rudder For Now. 
But now, everything is their to switch to a more robust, namespace-based identification of variables, and I'm pretty confident it will just work by changin the key for comparison in  `toPolicy`. 

So all in all, I'm rather happy with the change.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DeploymentService.scala
```diff
@@ -1748,12 +1748,12 @@ object RuleExpectedReportBuilder extends Loggable {
...
         case (variables, m) if (m.isEmpty)                                                                                =>
-          PolicyGenerationLogger.warn(
+          PolicyGenerationLogger.debug(
             s"Somewhere in the expansion of variables, the bounded variable ${boundingVar} for ${vars.trackerVariable.spec.name} in Directive ${directiveId.serialize} appeared, but was not originally there"
           )
           (variables.values.flatMap(_.values).toSeq, variables.values.flatMap(_.values).toSeq) // this is an autobounding policy
         case (variables, originalVariables)                                                                               =>
-          PolicyGenerationLogger.warn(
+          PolicyGenerationLogger.debug(
             s"Expanded and unexpanded values for bounded variable ${boundingVar} for ${vars.trackerVariable.spec.name} in Directive ${directiveId.serialize} have not the same size : ${variables.values} and ${originalVariables.values}"
```
fanf, line 1757: this can now happen often since we correct earlier the real content of tracking key

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/RuleValService.scala
```diff
@@ -78,16 +78,15 @@ class RuleValServiceImpl(
...
       variableSpecs.map {
         case (parents, spec) =>
-          // be carefull, component id are not defined at the varspec level, but in its parent spec one
+          // be careful, component id are not defined at the varspec level, but in its parent spec one
           val reportId = spec.id.orElse(parents.headOption.flatMap(_.id))
           context.get(spec.name) match {
             case None            =>
-              spec.toVariable()
-              (ComponentId(spec.toVariable().spec.name, parents.map(_.name), reportId), spec.toVariable())
+              (ComponentId(spec.name, parents.map(_.name), reportId), spec.toVariable())
```
fanf, line 85: this was useless, I really don't get why the `toVariable().spec` is here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -363,60 +365,75 @@ class BuildBundleSequence(
     val name = Promiser(policy.ruleName + "/" + policy.directiveName)
 
     // and for now, all bundle get the same reportKey
```
fanf, line 367: All the following code is just a rewrite of the same algo in a failing for loop returning `PureResult` since at that point we should not modify anymore any variable value. If we don't have what we expect, it's a problem that should be exposed.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -357,54 +357,9 @@ class PrepareTemplateVariablesImpl(
     // we can't do it when we built node configuration because some (system variable at least (note: but only? If so, we could just check
     // them here, not everything).
```
fanf, line 359: all the following checks are now useless. Or more likely: they were already semi-useless, but now, everything is done earlier in `.toPolicy`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -357,54 +357,9 @@ class PrepareTemplateVariablesImpl(
...
-                             case None    =>
-                               Inconsistency(
-                                 s"[${agentNodeProps.nodeId.value}:${policy.technique.id.debugString}] Misssing mandatory value for tracker variable: '${x.name}'"
-                               ).fail
-                             case Some(v) =>
-                               stVariableFromVariable(v, agentVariableHandler, agentNodeProps.nodeId, policy.technique.id)
-                                 .map(Some(_))
-                           }
-                         case x: SystemVariableSpec  =>
-                           systemVars.get(x.name) match {
```
fanf, line 376: system variables are part of the `technique.getAllVariableSpec` method, so missing one are checked

webapp/sources/rudder/rudder-core/src/test/resources/configuration-repository/techniques/ncf_techniques/Create_file/1.0/metadata.xml
```diff
@@ -69,6 +69,13 @@
         <DESCRIPTION>file_name</DESCRIPTION>
         <LONGDESCRIPTION></LONGDESCRIPTION>
       </INPUT>
+      <INPUT>
+        <NAME>3021FC4F-DA33-4D84-8991-C42EBAB2335F</NAME>
+        <DESCRIPTION>test_space</DESCRIPTION>
+        <CONSTRAINT>
+          <DEFAULT> </DEFAULT>
```
fanf, line 76: adding a variable to see if a default at XML level is taken into account.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -1016,31 +1017,39 @@ class TestNodeConfiguration(
...
+      variables: List[(String, Seq[String])]
   ): Map[ComponentId, Variable] = {
-    variables
-      .map(name => {
-        spec
-          .get(name)
+    val specs = technique.getAllVariableSpecs
+    // add system variable as a base, we don't build them correctly because we skip NodeConfiguration set-up
+    // in that test, see TestBuildNodeConfiguration to see how it should be done
+    specs.collect { case (cid, s) if (s.isSystem) => (cid, s.toVariable(Seq(s"dummy_system_${s.name}"))) } ++
```
fanf, line 1028: as the comment says

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -579,58 +579,114 @@ final case class BoundPolicyDraft(
...
    * This is the place where we can check for variable consistency too, since we have the final
    * techniques and expanded variables (like: check that a mandatory variable was expanded to "").
+   *
+   * It's also the place where we add missing variables that are well defined, and where we check for
+   * constrains.
+   *
+   * If there is variable defined here but not in the technique, they are removed.
    */
   def toPolicy(agent: AgentType): Either[String, Policy] = {
     PolicyTechnique.forAgent(technique, agent).flatMap { pt =>
```
fanf, line 589: this is the main change. Now we check for all variable here, and not only the one which were set.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -579,58 +579,114 @@ final case class BoundPolicyDraft(
...
-          )
-        })
-        .left
+      // check that all variables from root section are here and well defined
+      // the lookup for variables is a bit complicated. We have unique variable by name, so for now, we must to only match on that.
+      // Since we use the same technique to build section path, it should be the same (apart in tests) but
+      // I prefer to do the lookup on name
+      val techVarSpecs = technique.getAllVariableSpecs.map { case (cid, s) => (cid.value, (cid, s)) }
+
+      val allVars = expandedVars + (ComponentId(trackerVariable.spec.name, Nil, None) -> trackerVariable)
```
fanf, line 596: tracker variable is not part of the standard ones, but we need to check that it has a value (it's mandatory)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -579,58 +579,114 @@ final case class BoundPolicyDraft(
...
+                                                     s"a non optional value is missing for parameter '${variable.spec.description}' [param ID: ${variable.spec.name}]"
+                                                   )
+                                                 )
+                                             }
+
+                                         }
+                                     }
+                                 }
+        missing                = techVarSpecs -- checkedExistingValues.map(_._1.value).toSet
+        // add missing if possible
```
fanf, line 651: here, we check that the variables that are declared in the technique spec but didn't have a value in the draft are either optional or have an usable default value. And we add them, so that all variable defined in the technique exists (the optional just don't have any value).

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -960,14 +959,16 @@ class TestNodeConfiguration(
   val commonTechnique:                                                      Technique                  =
     techniqueRepository.unsafeGet(TechniqueId(TechniqueName("common"), TechniqueVersionHelper("1.0")))
   def commonVariables(nodeId: NodeId, allNodeInfos: Map[NodeId, NodeInfo]): Map[ComponentId, Variable] = {
```
fanf, line 961: most of changes here look like that: I needed to add the component id as a key to be consistent with the new `technique.getAllVariableSpecs` method.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/NodeConfigData.scala
```diff
@@ -1167,7 +1179,8 @@ class TestNodeConfiguration(
     id,
     "",
     None,
-    ""
+    "",
+    _isEnabled = true
```
fanf, line 1183: the directive was disable :face_exhaling: A luck it's not used much (appart for the node configuration build)

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/policies/TestBuildNodeConfiguration.scala
```diff
@@ -99,9 +99,9 @@ class TestBuildNodeConfiguration extends Specification {
   def newNode(i: Int): CoreNodeFact =
     fact1.modify(_.id).setTo(NodeId("node" + i)).modify(_.fqdn).setTo(s"node$i.localhost")
```
fanf, line 101: this test wasn't testing anything, which is a bit sad since it's a nice test that exercize all the first steps of policy generation.

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_directives.yml
```diff
@@ -251,7 +251,7 @@ response:
             "longDescription":"",
             "techniqueName":"Create_file",
             "techniqueVersion":"1.0",
-            "parameters":{"section":{"name":"sections","sections":[{"section":{"name":"Directory create","vars":[{"var":{"name":"expectedReportKey Directory create","value":"directory_create_/tmp/foo"}}]}},{"section":{"name":"File create","vars":[{"var":{"name":"expectedReportKey File create","value":"file_create_/tmp/foo/bar"}}]}},{"section":{"name":"Technique parameters","vars":[{"var":{"name":"1AAACD71-C2D5-482C-BCFF-5EEE6F8DA9C2","value":"\"foo"}}]}}]}},
+            "parameters":{"section":{"name":"sections","sections":[{"section":{"name":"Directory create","vars":[{"var":{"name":"expectedReportKey Directory create","value":"directory_create_/tmp/foo"}}]}},{"section":{"name":"File create","vars":[{"var":{"name":"expectedReportKey File create","value":"file_create_/tmp/foo/bar"}}]}},{"section":{"name":"Technique parameters","vars":[{"var":{"name":"1AAACD71-C2D5-482C-BCFF-5EEE6F8DA9C2","value":"\"foo"}},{"var":{"name":"3021FC4F-DA33-4D84-8991-C42EBAB2335F","value":" "}}]}}]}},
```
fanf, line 254: Yeah ! We have our value!

webapp/sources/rudder/rudder-core/src/test/resources/configuration-repository/expected-share/node-cfe-with-two-directives/rules/cfengine-community/rudder-directives.cf
```diff
@@ -33,8 +33,8 @@ bundle agent run_16d86a56_93ef_49aa_86b7_0d10102e4ea9
   methods:
       "50-rule-technique-ncf/Create a file" usebundle => rudder_reporting_context_v4("16d86a56-93ef-49aa-86b7-0d10102e4ea9","208716db-2675-43b9-ab57-bfbab84346aa","Create_file","","","16d86a56-93ef-49aa-86b7-0d10102e4ea9208716db-2675-43b9-ab57-bfbab84346aa");
       "50-rule-technique-ncf/Create a file" usebundle => enable_reporting;
-      "50-rule-technique-ncf/Create a file" usebundle => Create_file("\"foo");
-      "50-rule-technique-ncf/Create a file" usebundle => Create_file_rudder_reporting("\"foo");
+      "50-rule-technique-ncf/Create a file" usebundle => Create_file("\"foo"," ");
+      "50-rule-technique-ncf/Create a file" usebundle => Create_file_rudder_reporting("\"foo"," ");
```
fanf, line 37: Hourra!

fanf:
PR updated with a new commit

VinceMacBuche (approved):
LGTM

fanf:
OK, squash merging this PR

PR #5909, opened 2024-09-30
Fixes #25579: System variable can be empty early on the generation process

fanf:
https://issues.rudder.io/issues/25579

So, our management of system variable is not very good. We need to idenfy the one that are late bound and allow only these one to be void, but that's a big change. So for now, we are just relaxing the check on system var. It's ok, we never have a failure because of that for system var (which would be a bug) if we don't add/remove system var or change their generation. We will need to correct it in next patch, though.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
@@ -616,9 +616,10 @@ final case class BoundPolicyDraft(
                                            )
                                          }
                                          // check if there is mandatory variable with missing/blank values
+                                         // NOTE : it's OK that system variables are not filled here yet
                                          (
-                                           variable.spec.constraint.mayBeEmpty,
-                                           variable.values.exists(v => v == null || v.isEmpty)
+                                           variable.spec.constraint.mayBeEmpty || variable.spec.isSystem,
+                                           variable.values.isEmpty || variable.values.exists(v => v == null || v.isEmpty)
                                          ) match {
                                            // simple case: it's optional, we don't care if empty or not.
                                            case (true, _) => Right((cid, variable))
@@ -651,7 +652,8 @@ final case class BoundPolicyDraft(
         // add missing if possible
         added                 <- missing.accumulatePure {
                                    case (_, (cid, spec)) =>
-                                     (spec.constraint.mayBeEmpty, spec.constraint.default) match {
+                                     // if mayBeEmpty or if it's a system variable, just add it - empty.
+                                     (spec.constraint.mayBeEmpty || spec.isSystem, spec.constraint.default) match {
                                        case (true, _)        => Right((cid, spec.toVariable()))
                                        case (false, Some(d)) => Right((cid, spec.toVariable(Seq(d))))
                                        case (false, None)    =>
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -616,9 +616,10 @@ final case class BoundPolicyDraft(
                                            )
                                          }
                                          // check if there is mandatory variable with missing/blank values
+                                         // NOTE : it's OK that system variables are not filled here yet
                                          (
-                                           variable.spec.constraint.mayBeEmpty,
-                                           variable.values.exists(v => v == null || v.isEmpty)
+                                           variable.spec.constraint.mayBeEmpty || variable.spec.isSystem,
+                                           variable.values.isEmpty || variable.values.exists(v => v == null || v.isEmpty)
```
fanf, line 622: if there is no value, it's also a problem actually.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -616,9 +616,10 @@ final case class BoundPolicyDraft(
                                            )
                                          }
                                          // check if there is mandatory variable with missing/blank values
+                                         // NOTE : it's OK that system variables are not filled here yet
                                          (
-                                           variable.spec.constraint.mayBeEmpty,
-                                           variable.values.exists(v => v == null || v.isEmpty)
+                                           variable.spec.constraint.mayBeEmpty || variable.spec.isSystem,
```
fanf, line 621: skip system variable here, because they can be late bound. They will go in the "true" case, ie like if they are optionnal.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -353,11 +353,8 @@ class PrepareTemplateVariablesImpl(
       systemVars:           Map[String, Variable]
   ): IOResult[Seq[STVariable]] = {
 
-    // we want to check that all technique variables are correctly provided.
-    // we can't do it when we built node configuration because some (system variable at least (note: but only? If so, we could just check
```
fanf, line 357: I should have read that more carefully in the previous change

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PrepareTemplateVariables.scala
```diff
@@ -353,11 +353,8 @@ class PrepareTemplateVariablesImpl(
...
-    // them here, not everything).
-
-    ZIO.foreach(policy.trackerVariable :: (systemVars.values ++ policy.expandedVars.values).toList)(v =>
+    // here, we want to be sure to get the system variable that were late defined in priority.
+    ZIO.foreach(policy.trackerVariable :: (policy.expandedVars ++ systemVars).values.toList)(v =>
```
fanf, line 357: here, we just spent some time in other part of the class ensuring that systemVars are well defined. So we really want to use them. The previous thing was working a bit by luck I think (because string template takes the first variable that has the correct name, and so in case of duplicate, system variables where first...)

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5914, opened 2024-10-03
Fixes #25600: Compliance reporting with type worst-case-percent is not known by rudderc

amousset:
https://issues.rudder.io/issues/25600


clarktsiory:
PR updated with a new commit

Fdall (approved):

PR #5916, opened 2024-10-03
Fixes #25517: System rules are now always shown in directives compliance

fanf:
https://issues.rudder.io/issues/25517

Similar to https://github.com/Normation/rudder/pull/5915. We have a method that can only have user rules but we were collecting all rule types from node reports.


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #5917, opened 2024-10-04
Fixes #25604: Reporting type should be focus-worst instead of worst-case-percent

clarktsiory:
https://issues.rudder.io/issues/25604

Based on https://github.com/Normation/rudder/pull/5913 and expects changes from https://github.com/Normation/rudder/pull/5914

This is a change to have a better name by considering existing semantics, so we need to change the name of the type in our model.
Also :

* Change the serialized value :warning: existing techniques will need to be replaced by changing yml with `sed -ie "s/worst-case-percent/focus-worst/"` technique.yml` then reloading the technique
* Update Elm to support the new value, add a new text with the better description
* Update the ASCII doc

Resulting UI : 
Screenshot from 2024-10-04 09-59-21


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

fanf (approved):

PR #5920, opened 2024-10-04
Fixes #25332: When changing node properties the old table is sometimes still displayed

clarktsiory:
https://issues.rudder.io/issues/25332

Awaiting for the resolution of properties in the cache after a group update.
It is no big deal : this will benefit for the Rudder UI which queries the inherited properties, that are already resolved in the cache after the `updateAll()`


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

fanf (approved):
Good !

clarktsiory:
OK, squash merging this PR

PR #5921, opened 2024-10-04
Fixes #25528: List node with properties inheritance error in status bar for generation

clarktsiory:
https://issues.rudder.io/issues/25528

This is very similar to the part in #5905 where we use a separate service to sync the newly updated properties with the UI and add the error to the status : 
image
**AND**
Screenshot from 2024-10-04 19-08-31

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
@@ -40,8 +40,14 @@ package com.normation.rudder.web.comet
 import bootstrap.liftweb.FindCurrentUser
 import bootstrap.liftweb.RudderConfig
 import bootstrap.liftweb.RudderConfig.clearCacheService
+import com.normation.inventory.domain.NodeId
 import com.normation.rudder.AuthorizationType
 import com.normation.rudder.batch.*
+import com.normation.rudder.domain.properties.FailedNodePropertyHierarchy
+import com.normation.rudder.domain.properties.ResolvedNodePropertyHierarchy
+import com.normation.rudder.domain.properties.SuccessNodePropertyHierarchy
+import com.normation.rudder.facts.nodes.MinimalNodeFactInterface
+import com.normation.rudder.facts.nodes.QueryContext
 import com.normation.rudder.ncf.CompilationStatus
 import com.normation.rudder.ncf.CompilationStatusAllSuccess
 import com.normation.rudder.ncf.CompilationStatusErrors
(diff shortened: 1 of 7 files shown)


clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
```diff
@@ -250,26 +266,73 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
...
+      case (_, _: SuccessNodePropertyHierarchy)     => None
+      case (nodeId, f: FailedNodePropertyHierarchy) => allNodes.get(nodeId).map(_ -> f.getMessage)
+    }
+    nodesErrors.toList match {
+      case Nil    => NodeSeq.Empty
+      case errors =>
+        <li class="card border-start-0 border-end-0 border-top-0">
+          <div class="card-body">
+            <h5 class="card-title d-flex justify-content-between">
+              Node configuration errors
```
fanf, line 294: It must be "Node properties errors". The term "properties" is part of the ubiquitous langage for that thing in Rudder (and tools alike).

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
```diff
@@ -250,26 +266,73 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
...
+              <span class="badge bg-danger">{errors.size}</span>
+            </h5>
+            <ul class="list-group">
+            {
+          NodeSeq.fromSeq(
+            errors.map {
```
fanf, line 300: I think you must limits the number of rendered nodes here, else it won't scale and the rendering will be poor/broken/lead to performance problemps. Even with just 4, it's already a lot: 
image


Perhpas: 
- keep the total number of impacted nodes in the red square
- if it's 3 or less, let it like that
- if it's more, add a sentence `only displaying the first nodes in errors:`

It won't change much in UX, because in the case where there is a lot of nodes in error, it is *extremely* likely that all the errors are linked to the same inheritance problem, and so that resolving on one will resolve everywhere. 
And in addition, even if there is two root causes for the error, the user can only solve one at a time.

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
clarktsiory: But some errors, e.g. the DAG error (when there is a cycle in the groups) are not supposed to affect "properties" only, to me it is inconsistent mostly with the criteria of groups, not only the properties. 
Properties computation is the context that helps us detect such problem for now in Rudder, and we may as well name it "dynamic group" because the properties computation happens within dynamic groups computation too.

But yes, the only "configuration" part of nodes seems to be "properties" so it is still valid.
It is not the case for groups though, the configuration is broader because of criteria, so "Group properties error" with a detail that "a group from the criteria is missing" would not make sense.

So for consistency I suggest that we keep "configuration" for both nodes and groups

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
```diff
@@ -250,26 +266,73 @@ class AsyncDeployment extends CometActor with CometListener with Loggable {
...
+              case (n, err) =>
+                <li class="list-group-item d-flex justify-content-between align-items-center">
+                  <button id={nodeBtnId(n.id)} class="btn btn-link text-start">
+                    {StringEscapeUtils.escapeHtml4(n.fqdn)} <i class="fa fa-external-link"/>
+                  </button>
```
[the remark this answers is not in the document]
clarktsiory: no we would need additional logic in both frontend (here to add the a url hash, and in the Elm app for groups), it would not be trivial, we could do it later in another PR

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/comet/AsyncDeployment.scala
clarktsiory: OK it seems good to me

I suggest that we improve the UX and handle specific cases for the expected UX of each section (Policy generation, Groups, Nodes, Techniques) in another PR, where we would also add the possibility to collapse each section

fanf (approved):
The feedback will be addressed in https://issues.rudder.io/issues/25624

clarktsiory:
OK, squash merging this PR

PR #5923, opened 2024-10-08
Fixes #25632: Resolved properties need more tests and logging for success values

clarktsiory:
https://issues.rudder.io/issues/25632

* Remove FIXME comments and unused code
* Add more precise tests for DAGError, and one for group (missing parent group when computing group properties) 
* Add details of success properties even in case of error in debug logs

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/properties/Properties.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/properties/Properties.scala
@@ -1058,10 +1058,6 @@ object NodePropertyError         {
   // The message can be very broad or specific here, it is known from the DAG resolution
   case class DAGError(override val message: String) extends NodePropertyError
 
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/properties/Properties.scala
```diff
@@ -1058,10 +1058,6 @@ object NodePropertyError         {
   // The message can be very broad or specific here, it is known from the DAG resolution
   case class DAGError(override val message: String) extends NodePropertyError
 
-  // The model can end up with having an empty list of properties, so a this subtype may replace the Option API
-  case object NotFound extends NodePropertyError {
-    def message: String = "not found"
-  }
```
clarktsiory, line 1064: This is just unused, it was forgotten in the PR that introduced it

clarktsiory:
Commit modified

fanf (approved):
good

PR #5925, opened 2024-10-08
Fixes #25635: Chunk appending when resolving node ids from union target can be optimized

clarktsiory:
https://issues.rudder.io/issues/25635

`++` is specialized for Chunk, `concat` is on Iterable


fanf (approved):
There was only one, good catch !

PR #5926, opened 2024-10-08
Fixes #25637: Sometimes tests with lift actor fails

clarktsiory:
https://issues.rudder.io/issues/25637

It's weird that it fails sometimes even within ZIO fibers, the `!` method in the lift mock actor has a synchronized lock on the mutable list of messages it receives...
Our common solution for these synchronization issues in tests is `eventually` 
BUT here it does seem to take much time, so using a lock is the fastest solution

**Edit : ** the test above did some change in the state of the cache so we need to take these into account, and make tests reproducible by making them "sequential"


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #3615, opened 2021-05-03
Fixes #19213: Put node cache info into a file in place of LDAP

fanf:
https://issues.rudder.io/issues/19213

PR #5929, opened 2024-10-08
Fixes #3615: IP search field with slash notation

fanf:
https://issues.rudder.io/issues/3615

Now that we have access to `CoreNodeFact`, having IP matching on CIDR is easy. But we can only do it for IP in "node summary", not in the "Network Interface" part, which is a bit sad.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
```diff
@@ -519,6 +523,50 @@ final case class NodeCriterionMatcherString(extractor: CoreNodeFact => Chunk[Str
...
+
+  /*
+   * From an input 'value' that can be a cidr or an ip, build a matcher on values
+   */
+  def buildMatcher(value: String, facts: Chunk[String], m: (Chunk[IpAddress], Cidr[IpAddress]) => Boolean): IOResult[Boolean] = {
+    Cidr
+      .fromString(value)
+      .orElse(IpAddress.fromString(value).map {
+        case ip: Ipv4Address => Cidr(ip, 32)
+        case ip: Ipv6Address => Cidr(ip, 128)
```
fanf, line 554: when we have an IP and not a CIDR, we build the corresponding CIDR with a 0 free bits mask, which is the same as the IP address.

fanf:
PR updated with a new commit

fanf:
PR replaced by https://github.com/Normation/rudder/pull/5930

PR #5930, opened 2024-10-09
Fixes #3615: IP search field with slash notation

fanf:
https://issues.rudder.io/issues/3615

Replacing previous PR: https://github.com/Normation/rudder/pull/5929

Now that we have access to CoreNodeFact, having IP matching on CIDR is easy. But we can only do it for IP in "node summary", not in the "Network Interface" part, which is a bit sad.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
@@ -37,6 +37,10 @@
 
 package com.normation.rudder.domain.queries
 
+import com.comcast.ip4s.Cidr
+import com.comcast.ip4s.IpAddress
+import com.comcast.ip4s.Ipv4Address
+import com.comcast.ip4s.Ipv6Address
 import com.normation.errors.*
 import com.normation.inventory.domain.AgentType
 import com.normation.inventory.domain.MachineType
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
```diff
@@ -519,6 +523,50 @@ final case class NodeCriterionMatcherString(extractor: CoreNodeFact => Chunk[Str
...
+object NodeCriterionMatcherIpaddress extends NodeCriterionMatcher {
+  private val extractor     = (n: CoreNodeFact) => n.ipAddresses.map(_.inet)
+  private val stringMatcher = NodeCriterionMatcherString(extractor)
+
+  override def matches(n: CoreNodeFact, comparator: CriterionComparator, value: String): IOResult[Boolean] = {
+    comparator match {
+      case Equals    =>
+        NodeCriterionMatcherIpaddress.buildMatcher(value, extractor(n), (ips, cidr) => ips.exists(ip => cidr.contains(ip)))
+      case NotEquals =>
+        NodeCriterionMatcherIpaddress.buildMatcher(value, extractor(n), (ips, cidr) => ips.forall(ip => !cidr.contains(ip)))
```
fanf, line 540: equals is `exists` (at least one matches), not equals is `!forall` (non value matches)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
```diff
@@ -519,6 +523,50 @@ final case class NodeCriterionMatcherString(extractor: CoreNodeFact => Chunk[Str
...
+
+  /*
+   * From an input 'value' that can be a cidr or an ip, build a matcher on values
+   */
+  def buildMatcher(value: String, facts: Chunk[String], m: (Chunk[IpAddress], Cidr[IpAddress]) => Boolean): IOResult[Boolean] = {
+    Cidr
+      .fromString(value)
+      .orElse(IpAddress.fromString(value).map {
+        case ip: Ipv4Address => Cidr(ip, 32)
+        case ip: Ipv6Address => Cidr(ip, 128)
```
fanf, line 554: when we have an IP and not a CIDR, we build the corresponding CIDR with a 0 free bits mask, which is the same as the IP address.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaData.scala
clarktsiory: I find the `not exists` with the same body as for equality i.e. `!ips.exists(ip => cidr.contains(ip)))` a lot clearer than introducing another predicate `forall` here

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/domain/queries/NodeQueryCriteriaDataTest.scala
```diff
@@ -0,0 +1,70 @@
+package com.normation.rudder.domain.queries
```
clarktsiory, line 1: Do we need License header in test files ?

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM !

fanf:
OK, squash merging this PR

PR #5935, opened 2024-10-09
Fixes #25650: Node in with no reports, pending and keep compliance lead to computation loop

fanf:
https://issues.rudder.io/issues/25650

In persistance, we need to cover all case of "expiring by themself" kind: we were missing `Pending`. So if a pending reached the database, then we were always updating it, always pushing ahead its expiration time. 

Other changes are just additionnal debug info, because I still wasn't able to find quickly what was done.


clarktsiory (approved):
LGTM !

PR #5937, opened 2024-10-10
Fixes #25530: Group properties wrongly appear empty when group has no member nodes

clarktsiory:
https://issues.rudder.io/issues/25530

We needed to get the node properties for a group, in order to get errors that applied in them (the type of the properties, to make a warning if it is not the same) 

Now, when the list of nodes is empty for a group, and we apply the same check (including only parent groups and global properties then), so we need to factor it out


clarktsiory:
PR rebased

PR #5941, opened 2024-10-10
Fixes #25534: Debug information download is broken by anti-CSRF header

clarktsiory:
https://issues.rudder.io/issues/25534

The redirect is now impossible, we need a frontend button with an ajax call, and to keep the same UX we need to download the file from the browser itself, within the context of the ajax call where the file is in-memory.
(If debug info get bigger we would likely need a dedicated non REST API path to serve generated debug files and just open the link to them)


fanf (approved):
LGTM, works as expected, and I don't see any security problem.

PR #5942, opened 2024-10-10
Fixes #25589: Lost manufacturer & serial for BIOS  between 7.3 and 8.0

ElaadF:
https://issues.rudder.io/issues/25589
I simply add a `Manufacturer` and `Serial number` field in the `BIOS` object, these fields were previously present in `HARDWARE` section of the inventory but was cleaned because it was not map to anything.
Since these values are optional, it should not break anything, I add the missing filters in criteria when creating group for the BIOS too

The main strategy was to add the missing value `Manufacturer` and `Serial number` and let the compiler do the work


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
OK, merging this PR

PR #5946, opened 2024-10-14
Fixes #25670: "Started since" time in status zone is not correct

clarktsiory:
https://issues.rudder.io/issues/25670

Because the date can take too much space the date is moved to the "help" section when on the same day.
The duration is still not refreshed (that would require some JS code) 

image


clarktsiory:
Commit modified

fanf (approved):

PR #5949, opened 2024-10-15
Fixes #25652: Default settings for new nodes are not applied on a accepted node

ElaadF:
https://issues.rudder.io/issues/25652
There were missing logic to modify the node according to the default global setting for new node. I simply add a `unitAcceptors` that is not really a unit... That fetch the node and modify his policy mode and his state.
It's working, but it can be confusing since theses should me `unitAcceptors` and for pre accepting :grimacing: 

see @clarktsiory comments https://github.com/Normation/rudder/pull/5949#discussion_r1802652374

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NewNodeManager.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NewNodeManager.scala
@@ -44,6 +44,8 @@ import com.normation.inventory.domain.AcceptedInventory
 import com.normation.inventory.domain.NodeId
 import com.normation.inventory.ldap.core.LDAPConstants.*
 import com.normation.rudder.domain.logger.NodeLoggerPure
+import com.normation.rudder.domain.nodes.NodeState
+import com.normation.rudder.domain.policies.PolicyMode
 import com.normation.rudder.domain.queries.CriterionLine
 import com.normation.rudder.domain.queries.DitQueryData
 import com.normation.rudder.domain.queries.Equals
@@ -305,7 +307,6 @@ class ComposedNewNodeManager[A](
   ////////////////////////////////////////////////////////////////////////////////////
 
   def accept(id: NodeId)(implicit cc: ChangeContext): IOResult[CoreNodeFact] = {
-
     // validate pre acceptance for a Node, if an error occurs, stop everything on that node.
     def passPreAccept(nodeFact: CoreNodeFact) = {
       ZIO.foreachDiscard(unitAcceptors) { a =>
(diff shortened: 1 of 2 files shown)


ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

ElaadF:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/nodes/NewNodeManager.scala
```diff
@@ -445,3 +446,23 @@ class AcceptHostnameAndIp(
...
+  override def preAccept(cnf: CoreNodeFact)(implicit cc: ChangeContext): IOResult[Unit] = {
+    for {
+      policyMode <- defaultPolicyModeOnAccept
+      nodeState  <- defaultStateOnAccept
+      newCnf      = cnf
+                      .modify(_.rudderSettings.state)
+                      .setTo(nodeState)
+                      .modify(_.rudderSettings.policyMode)
+                      .setTo(policyMode)
+      _          <- nodeFactRepo.save(newCnf)
```
clarktsiory, line 465: Since there is a `repo.save` it is no longer a "unitcheckaccept", it mutates the state depending on the inventory...
What we want could even be something else than a `preAccept`, maybe adding a new trait with the `postAccept` method (and also include hooks that are already run with `NewNodeManagerHooks#afterNodeAcceptedAsync` to be within the post-acceptors list)

fanf: given current infra, it looks like we can't do better. 
We *could* have a chain of "pre accept mutator" that would take a CoreNodeFact and check/change it and then pass the new CoreNodeFact to the next link, so that you would change in your step settings and have them saved in the end. At the end of the pre-accept, if the cnf is different from the one got at the begining, save it before going to accept phase. 
But that would be a massive change. 

The current whole acceptation logic is: 
```
    for {
      // Get inventory og the node
      cnf       <- nodeFactRepo.get(id)(cc.toQuery, SelectNodeStatus.Pending).notOptional(s"Missing inventory for node with ID: '${id.value}'")
      // Pre accept it
      preAccept <- passPreAccept(cnf) 
      // Accept it
      _         <- nodeFactRepo.changeStatus(id, AcceptedInventory)
      // Update hooks for the node
      _         <- hooksRunner
                     .afterNodeAcceptedAsync(id)(cc.toQuery)
                     .catchAll(err => {
                       NodeLoggerPure.PendingNode.error(
                         s"Error when executing post-acceptation hooks for node '${cnf.fqdn}' " +
                         s"[${cnf.id.value}]: ${err.fullMsg}"
                       )
                     })
    } yield cnf.modify(_.rudderSettings.status).setTo(AcceptedInventory)
```

So, it's ok to change the core node fact in a pre accept, since it's that mutated version that will be accepted. 
And since it's the last step, the other step will have a correct cnf. 
The only problem is that we return the cnf at the end. It looks like an API design mistake with current tooling (and it would make sense with the pipeline approach).

fanf: OK, one easy thing to do: 
- get again the node fact after acceptation and return that. 
The data will be correct, we don't lie with a change done by hand for the status, and the cost of getting the core node fact is extremelly small since it's in cache.

fanf (changes requested):
Can you update the main acceptation loop to retrieve again the the cnf just before returning it in the yield ? So we are sure to have the correct data in following callers. Else, seems fine to me.

ElaadF:
PR updated with a new commit

fanf (approved):
LGTM !

ElaadF:
OK, squash merging this PR

PR #5953, opened 2024-10-17
Fixes #25687: Same as parent ticket for branch 8.1

clarktsiory:
https://issues.rudder.io/issues/25687

Add `nonEmpty` condition similarly to https://github.com/Normation/rudder/pull/5937/files, to return the list of group properties when there are no nodes

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
@@ -1371,49 +1371,57 @@ class GroupApiService14(
 
       nodes <- nodeFactRepo.getAll().map(_.filterKeys(serverList.contains(_)).values.toList)
 
-      params           <- paramRepo.getAllGlobalParameters().map(_.map(p => (p.name, p)).toMap)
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/GroupsApi.scala
```diff
@@ -1367,61 +1369,68 @@ class GroupApiService14(
...
+          }
 
+          success ++ error
+        }
+      }
+      groupPropertyStatuses = checkedNodeProperties match {
+                                case Nil       => // no Node is group : just use resolved group properties
+                                  parentProperties match {
+                                    case s: SuccessNodePropertyHierarchy =>
+                                      s.resolved.map(InheritedPropertyStatus.from(_))
```
clarktsiory, line 1421: I just realized we should check that the resolved property is "Displayed" before transforming it in 8.2 (the "Visibility" is only from 8.2 so it simplifies things here)

clarktsiory:
Commit modified

fanf (approved):
seems to work as expected !

clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #5956, opened 2024-10-18
Fixes #25709: OpenAPI documentation is invalid with duplicate /nodes/pending endpoints

clarktsiory:
https://issues.rudder.io/issues/25709

The linting of the openapi was misplaced, it should be done on split YAML files because loading them into a single `main` replaces duplicate keys silently (it is the behavior of the yaml loader from the Python library).

Also the `relay` could not build unless the [no-server-example-com](https://redocly.com/docs/cli/rules/no-server-example-com#no-server-example.com) rule is set to 'warn', we should at some point use another URL than _localhost_ for our examples.

### DIFF
--- a/webapp/sources/api-doc/openapi.src.yml
+++ b/webapp/sources/api-doc/openapi.src.yml
@@ -255,9 +255,10 @@ paths:
   "/nodes/status":
     $ref: paths/nodes/status.yml
   "/nodes/pending":
-    $ref: paths/nodes/pending-all.yml
-  "/nodes/pending":
-    $ref: paths/nodes/pending-update.yml
+    get:
+      $ref: paths/nodes/pending-all.yml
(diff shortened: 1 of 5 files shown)


webapp/sources/api-doc/openapi.src.yml
```diff
@@ -255,9 +255,10 @@ paths:
   "/nodes/status":
     $ref: paths/nodes/status.yml
   "/nodes/pending":
-    $ref: paths/nodes/pending-all.yml
-  "/nodes/pending":
-    $ref: paths/nodes/pending-update.yml
+    get:
+      $ref: paths/nodes/pending-all.yml
```
amousset, line 259: why moving this?

amousset (approved):

PR #5959, opened 2024-10-23
Fixes #25713: ReportsExecution doesn't have timezone on all fields

fanf:
https://issues.rudder.io/issues/25713

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableReportsExecutionTz.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableReportsExecutionTz.scala
@@ -1,0 +1,120 @@
+package bootstrap.liftweb.checks.migration
+
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+import bootstrap.liftweb.*
+import com.normation.errors.IOResult
+import com.normation.rudder.db.Doobie
+import com.normation.zio.*
+import doobie.Update0
+import doobie.implicits.*
+import doobie.util.fragment.Fragment
+import zio.interop.catz.*
+
+/*
+ * When we created table ReportsExecution, we didn't use "timestamp with time zone" for column
+ * 'insertiondate' and 'compliancecomputationdate'
+ */
+class CheckTableReportsExecutionTz(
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableReportsExecutionTz.scala
```diff
@@ -0,0 +1,120 @@
...
+
+    val sql = {
+      fr"""
+        DO $$$$ BEGIN
+          IF NOT EXISTS (
+            select column_name, data_type
+            from information_schema.columns
+            where table_name = """ ++ toQuotedFr(table) ++ fr" and column_name = " ++ toQuotedFr(
+        column
+      ) ++ fr""" and data_type = 'timestamp with time zone'
```
clarktsiory, line 110: The table and column does not need to be constant fragment but can simply be interpolated, since they are standard parameterized arguments for the query, and not part of the SQL DML syntax,
for instance : https://github.com/Normation/rudder/blob/a900c33686260e9b935f49e3983e9caa44dad193/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/DbCommonMigration.scala#L13 (the License header is missing from that file :see_no_evil:)

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableReportsExecutionTz.scala
```diff
@@ -0,0 +1,120 @@
...
+  val complianceComputationDateColumn: String = "compliancecomputationdate"
+
+  // overrideChange is only used for tests
+  def query(table: String, column: String, overrideChange: Option[Fragment] = None): Update0 = {
+    def toQuotedFr(s: String) = Fragment.const("'" + s + "'")
+    def toFr(s:       String) = Fragment.const(s)
+
+    val alter: Fragment = fr"ALTER TABLE " ++ toFr(table) ++ fr" ALTER COLUMN " ++ toFr(
+      column
+    ) ++ fr" TYPE TIMESTAMP WITH TIME ZONE USING " ++ toFr(column) ++ fr" AT TIME ZONE 'UTC';"
```
clarktsiory, line 100: I'm not sure for the 'UTC' timezone, in our [Doobie Meta](https://github.com/Normation/rudder/blob/branches/rudder/8.2/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala#L182) we write dates using no specific timezone and I thought the PG session current timezone is used (`SHOW TIMEZONE;`), so maybe 'UTC' is a special test value here ?
I could not test the behavior because I have no reports locally but I believe that if the timezone is something else than UTC we could end up assigning wrong values to the columns

clarktsiory (changes requested):
The `CheckTableReportsExecutionTz` appears to be missing from the checks at boot, otherwise changes look fine.
There is also a question : why is 'UTC' the timezone we have assumed to be written in the column, and not the PG session timezone ?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableReportsExecutionTz.scala
fanf: we don't any information about what was used to write these timezone, this is lost for ever. The server config can have changed. And since it's not really use, it doesn't not matter. UTC is a as correct value, as other and in case of problem it will be much more easy to guess what happened if we don't have to check for what was the timezone when the migration was done.

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestMigrateTableReportsExecution.scala
```diff
@@ -0,0 +1,143 @@
...
+    "migrate all columns successfully" in {
+
+      checkTableReportsExecutionTz.migrationProg(testTable).runNow // await for the migration before running assertions test
+
+      val sql = {
+        fr"""SELECT DISTINCT column_name, data_type FROM INFORMATION_SCHEMA.COLUMNS
+          """ ++ whereAnd(
+          fr"table_name = " ++ Fragment.const("'" + testTable + "'"),
+          fr"column_name in (" ++ Fragment.const("'" + insertionDateColumn + "'")
+        ) ++ fr", " ++ Fragment.const("'" + complianceComputationDateColumn + "'") ++ fr")"
```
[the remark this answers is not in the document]
fanf: well, yes, but not in the ALTER TABLE part, so I wanted to make it very clear that we can't use the same things in both place by using the same construct with a different escaping rule.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckTableReportsExecutionTz.scala
fanf: Added and fixed the quoting convention - we always use `${}`

fanf:
PR updated with a new commit

clarktsiory:
OK, merging this PR

PR #5960, opened 2024-10-23
Fixes #25691: Full archive export does not work anymore

clarktsiory:
https://issues.rudder.io/issues/25691

The problem is the same as #5941 : we can no longer redirect to a secure API endpoint since they need CSRF headers, so we need to execute AJAX to download the file content (the AJAX consists in saving the archive itself from the browser).
* In this PR the JS functions that do the download are put in global scope (`rudder.js`) they may be reused in more that one place in Rudder (and I checked btw that this one is the last case of redirect).
* The FS mappings for archive types (rules, directives, etc.) have been put in an attribute of an enum type, which helps avoiding the "String" API.
* Also the previous tests about this part were always testing an error because the test setup with commit ids of archive were causing a null pointer exception, I enforced the tests


fanf (approved):
Good, works as expected. 
OK for the other refactoring too.

clarktsiory:
OK, merging this PR

PR #5963, opened 2024-10-24
Fixes #25720: Technical and event logs get log from browser timezone instead of server timezone

clarktsiory:
https://issues.rudder.io/issues/25720

The `initDatePicker` function is adapted with an additional `timezone` argument (a IANA one, that can be used to ["convert" a JS date in the right timezone](https://stackoverflow.com/a/53652131) i.e. it has the right value in the datepicker with a selected timezone).

I used the server timezone because all server information is returned in the server timezone so it makes more sense.


RaphaelGauthier (approved):

clarktsiory:
OK, merging this PR

clarktsiory:
OK, merging this PR

PR #5967, opened 2024-10-25
Fixes #25742: Add LDAP synchronous mode config param

fanf:
https://issues.rudder.io/issues/25742

Make LDAP request synchronous by default (configurable via a rudder properties parameter). 
This make async request impossible (ie when we query LDAP with a listener and then asked for answers), but we don't use any of them (AFAIK). 
Making request synchronous make the pool usage more predictable, error more clearly linked to the request leading to them, and general enhance performance. 
See https://ldapwiki.com/wiki/Wiki.jsp?page=Tips%20using%20UnboundID%20LDAP%20SDK#section-Tips+using+UnboundID+LDAP+SDK-LDAPConnectionPoolsAndGetConnection

Also add options to: 
- release the connection to the pool as early as possible in case of error, 
- use information from our LDAP Schema and don't try to refresh them (since in Rudder, the schema doesn't change without a restart of Rudder)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
@@ -87,6 +87,10 @@ ldap.port=389
 ldap.authdn=cn=manager,cn=rudder-configuration
 ldap.maxPoolSize=2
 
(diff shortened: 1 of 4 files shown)


clarktsiory (commented):
LGTM but it should be merged after the current release patch in preparation, marking this as "can't merge"

fanf:
Replaced by #6045

PR #5968, opened 2024-10-28
Fixes #25745: db error about pid column

amousset:
https://issues.rudder.io/issues/25745


amousset:
PR updated with a new commit

PR #5981, opened 2024-10-31
Fixes #23959: When removing files from a technique, it might get disabled unexpectedly

VinceMacBuche:
https://issues.rudder.io/issues/23959


VinceMacBuche:
Commit modified

fanf (commented):
Looks good but i think a second look would be good

fanf (approved):
LGTM, plus seems to work as expected (but rather hard to live test more than for trivial cases). 
Thanks for the added tests.

PR #5986, opened 2024-11-02
Fixes #25780: Clean-up support for rsa key for inventory signature

fanf:
https://issues.rudder.io/issues/25780

Based on top of #5985 . 

Nothing special, I didn't try to change code logic now that there is just one kind of security token, I didn't want to break things of create a migration for the expected Json format (for ex).


fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
PR updated with a new commit

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
```diff
@@ -1065,30 +1065,33 @@ class InventoryMapper(
...
+            ok => Some(ok).succeed
+          )
+          .flatMap {
+            // we must have exactly one agent
+            case Some(a :: Nil) => List(a).succeed
+            case Some(several)  =>
+              InventoryDataLogger.error(
+                s"Error: inventory for node '${}' has several (${several.size}) different agents defined, which is not supported. Ignoring all agents."
+              ) *> Nil.succeed
+            case None           => Nil.succeed
```
clarktsiory, line 1093: Okay there is some additional checks, but the whole logic could all be under the body of the `foldZIO` ? we can easily miss the `map(_.distinct)` (I'm also curious why it's necessary) and the whole process when it is split up like that

clarktsiory (commented):
LGTM, the cleaning is relieving, there is a comment about some usage of combinators

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
fanf: I think it's purelly historical aggregation, I don't see specific reasons for not merging everything. 
The distinct is because once upon a time, you could have several agent. Then only one, but you didn't want to have a duplicate xml agent tag leading to a false error. I'm not sure we can still have duplicate xml part, but in doubts...

fanf:
Merging by hand because of unrelated relayd test failing

fanf:
OK, merging this PR

PR #5996, opened 2024-11-06
Fixes #25820: On new groups, adding a prop leads to "No query defined for group" error

fanf:
https://issues.rudder.io/issues/25820

Correct the case where updating a dynamic group with empty query was in error and also make a bit clearer the case where there is no criterion in the UI: 

When group is created: 
image

When we click on the plus: 
image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/SearchNodeComponent.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/SearchNodeComponent.scala
@@ -157,7 +157,7 @@ class SearchNodeComponent(
 
   def buildQuery(isGroupsPage: Boolean)(implicit qc: QueryContext): NodeSeq = {
 
-    if (None == query) query = Some(Query(NodeReturnType, And, ResultTransformation.Identity, List(defaultLine)))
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/SearchNodeComponent.scala
```diff
@@ -321,7 +323,11 @@ class SearchNodeComponent(
         ("id"    -> "SubmitSearch"),
         ("class" -> "btn btn-primary ")
       ) &
-      "#query_lines *" #> criteria.zipWithIndex.flatMap { case (cl, i) => displayQueryLine(cl, i, criteria.size > 1) }).apply(
+      "#query_lines *" #> (if (criteria.isEmpty) <div><span>Add a first criterion line to define your group: </span>{
```
amousset, line 326:
```suggestion
      "#query_lines *" #> (if (criteria.isEmpty) <div><span>Add a first criterion to define your group: </span>{
```

clarktsiory (commented):
LGTM and nominal case works, but I think the "Search" button needs to be hidden, or else making a search without any criteria and saving will make the group have "all nodes"

fanf:
PR updated with a new commit

fanf:
The search button is now hidden when group is created.

fanf:
PR rebased

PR #5998, opened 2024-11-06
Fixes #25825: Deleted OIDC user are reactivated after new provisioning

clarktsiory:
https://issues.rudder.io/issues/25825

* Don't add deleted users back : the `computeUpdatedUserList` method which is reused when "setting existing users" has logic to handle "deleted" and "disabled" user differently, deleted users should not be passed back to existing users (the logic of the method will make already disabled ones remain disabled)
* Add unit tests, 2 tests cases are added : 
  * a simple one to disable Bob
  * another one to check that adding a new user Xavier does not change the state of others : Bob -> disabled, William -> deleted
  Every other changes in the test file is just about moving the values used within the test cases right above the block of the test, the diff cannot just guess it

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/users/UserRepositoryTest.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/users/UserRepositoryTest.scala
@@ -268,154 +268,6 @@ trait UserRepositoryTest extends Specification with Loggable {
       })
     }
 
-    // BOB REMOVED
-    val userFileBobRemoved  = users.filterNot(_ == "bob")
-    val dateBobRemoved      = DateTime.parse("2023-09-02T02:02:02Z")
-    val traceBobRemoved     = EventTrace(actor, dateBobRemoved)
-    val userInfosBobRemoved = {
-      /*
-       * When bob is removed (ie not in the list of user from the files):
-       * - Bob is in the list, but with a status deleted and a new history line
-       * - other users are not changed
-       */
-      userInfosInit.map {
-        case u if (u.id == "bob") =>
-          u.modify(_.status)
-            .setTo(UserStatus.Deleted)
-            .modify(_.statusHistory)
-            .using(StatusHistory(UserStatus.Deleted, traceBobRemoved) :: _)
-
-        case u => u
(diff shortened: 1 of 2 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/users/UserRepositoryTest.scala
```diff
@@ -373,12 +447,62 @@ trait UserRepositoryTest extends Specification with Loggable {
...
+        case u => u
+      }
+    }
+
+    "some user is disabled" >> {
+      val disabled = repo.disable(List("bob"), None, List.empty, traceBobDisabled).runNow
+      (disabled must containTheSameElementsAs(List("bob"))) and (
+        repo.getAll().runNow.toUTC must containTheSameElementsAs(userInfosBobDisabled)
+      )
+    }
```
clarktsiory, line 475: Additional test case no. 1

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/users/UserRepositoryTest.scala
```diff
@@ -373,12 +447,62 @@ trait UserRepositoryTest extends Specification with Loggable {
...
+        StatusHistory(UserStatus.Active, traceXavierOidc) :: Nil,
+        Json.Obj()
+      )
+    }
+
+    "Some user is added in remote with some other remote users disabled/deleted" >> {
+      // Bob: disabled, William: deleted, adding a user should not change the state of existing ones
+      val isAdded = repo.addUser(AUTH_PLUGIN_NAME_REMOTE, "xavier", traceXavierOidc).runNow
+      isAdded must beTrue and (repo.getAll().runNow.toUTC must containTheSameElementsAs(userInfosXavierOidc))
+    }
```
clarktsiory, line 498: Additional test case no. 2

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #6001, opened 2024-11-07
Fixes #25762: Sometimes too long properties values move out actions buttons from window

RaphaelGauthier:
https://issues.rudder.io/issues/25762

This bug only appears on Firefox, so I had to add a simple css rule to prevent it.

> Note: In contrast to [word-break](https://developer.mozilla.org/en-US/docs/Web/CSS/word-break), overflow-wrap will only create a break if an entire word cannot be placed on its own line without overflowing.
https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap


clarktsiory (approved):
LGTM !

PR #6007, opened 2024-11-10
Fixes #25842: scoreService and scoreServiceManager are not initialized where they should

fanf:
https://issues.rudder.io/issues/25842

General ideas and hypothesis: 
- we want to reduce the number of `runNow` (each one create more risk that there is a chain of `.runNow/.attemp/.runNow` with the potential for deadlock)
- it's easier to remember to init a service near where it is declared
- we want to keep all service init at the same place if possible

I added a (mutable) buffer in our `def init` method, into which we can add ZIO effect related to services init, and `runNow` it (in parallel, effect result discarded) at the end of the method. 
I used the same construct for post-plugin init.


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #6009, opened 2024-11-14
Fixes #25847: The setup initial page should not be displayed if rudder-pkg.conf has been modified

clarktsiory:
https://issues.rudder.io/issues/25847

When settings are "defined" we should not redirect.
We are also on a hot path so we need to mark the setup as done to prevent reading the file at every webpage request

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/SetupRedirect.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/SetupRedirect.scala
@@ -38,36 +38,39 @@
 package com.normation.rudder.web.snippet
 
 import bootstrap.liftweb.RudderConfig
-import com.normation.box.*
 import com.normation.rudder.AuthorizationType.Administration
 import com.normation.rudder.users.CurrentUser
+import com.normation.zio.UnsafeRun
 import net.liftweb.common.*
 import net.liftweb.http.CurrentReq
 import net.liftweb.http.DispatchSnippet
 import net.liftweb.http.js.JsCmd
 import net.liftweb.http.js.JsCmds.*
 import scala.xml.NodeSeq
+import zio.syntax.*
 
 class SetupRedirect extends DispatchSnippet with Loggable {
 
-  private val configService = RudderConfig.configService
+  private val pluginSettingsService = RudderConfig.pluginSettingsService
 
   def dispatch: PartialFunction[String, NodeSeq => NodeSeq] = {
     case "display" => _ => WithNonce.scriptWithNonce(Script(display()))
   }
 
   def display(): JsCmd = {
-
-    configService.rudder_setup_done().toBox match {
-      case Full(false)
-          if CurrentUser.checkRights(Administration.Write) && !CurrentReq.value.request.url.contains("administration/setup") =>
-        RedirectTo("/secure/administration/setup")
-      case Full(_) =>
-        Noop
-      case eb: EmptyBox =>
-        val msg = eb ?~! "Could not get 'setup done' property"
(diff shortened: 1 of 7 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

amousset:
> Since https://github.com/Normation/rudder/pull/6010 the rudder settings are empty by default, we want to redirect to the page in the case that : either the file does not exist or the settings are empty

It is only changed starting from 8.3.

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/SetupRedirect.scala
```diff
@@ -38,36 +38,45 @@
...
   }
 
   def display(): JsCmd = {
-
-    configService.rudder_setup_done().toBox match {
-      case Full(false)
-          if CurrentUser.checkRights(Administration.Write) && !CurrentReq.value.request.url.contains("administration/setup") =>
+    (for {
+      isSetupDone     <- configService.rudder_setup_done()
+      setupFileExists <- pluginSettingsService.settingsFileExists()
```
fanf, line 64: this check should not be done each time. It's costlier than a setting check, and should only be done in the case the setting service answer "not done".

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/SetupRedirect.scala
```diff
@@ -38,36 +38,45 @@
...
+        !isSetupDone && (!setupFileExists || !settings.isDefined) && CurrentUser
+          .checkRights(Administration.Write) && !CurrentReq.value.request.url.contains("administration/setup")
+      ) {
         RedirectTo("/secure/administration/setup")
-      case Full(_) =>
-        Noop
-      case eb: EmptyBox =>
-        val msg = eb ?~! "Could not get 'setup done' property"
-        logger.error(msg.messageChain)
+      } else {
```
fanf, line 72: In the case where the file exists and the setup prop is not set to true, we need to also set the setting to true. Else the API will be lying (the init is done, and the API says it's not)

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/SetupRedirect.scala
clarktsiory: Yes, indeed !

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/SetupRedirect.scala
clarktsiory: yes ! I used a flatMap instead, and in fact I could remove the `settingsFileExists` check, it can be superseded with the more powerful check on the settings being defined

clarktsiory:
PR updated with a new commit

fanf (changes requested):
Much better encapsulated, just a log problem. 
Now it's easier to add the missing unit tests on `pluginSettingService` :)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):
Nice :) Good usage of Matthieu practice

clarktsiory:
OK, merging this PR

PR #6017, opened 2024-11-14
Fixes #25727: Locale issue: date picker should be in yyyy/MM/dd in group criteria

clarktsiory:
https://issues.rudder.io/issues/25727


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -513,7 +513,9 @@ case object OrderedStringComparator extends TStringComparator {
 
 case object DateComparator extends LDAPCriterionType {
   override val comparators: Seq[CriterionComparator] = OrderedComparators.comparators.filterNot(c => c == Regex || c == NotRegex)
-  val fmt = "dd/MM/yyyy"
+  val oldFmt = "dd/MM/yyyy" // format for which we need to keep compatibility
+  val fmt    = "yyyy/MM/dd"
+  val frenchOldFmt:                       DateTimeFormatter = DateTimeFormat.forPattern(oldFmt).withLocale(Locale.FRANCE)
   val frenchFmt:                          DateTimeFormatter = DateTimeFormat.forPattern(fmt).withLocale(Locale.FRANCE)
```
amousset, line 519: Why the French locale in the "correct" case?

fanf: And with a second though, why `yyyy/MM/dd` ? It a local standard of a few coutries, but not anything Standard. Why not `yyyy-MM-dd` so that we at least some consistency with RFC-3339?

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
clarktsiory: It seems the locale has no effect because it's all numbers, in fact it could be removed :sweat_smile:

In fact, yes, the RFC format should be the default one, and we can easily support other formats for convenience for the users.

In the frontend, the user could also have the choice of the format when initializing their date picker, we would just need to support them (see https://github.com/Normation/rudder/pull/5201/files#diff-5192ead8ee6e8bbdfc1c1de061965986ff9d79c1e6216b2dc3ac06ba68e61582R582).
We just have to be clear about the order of `MM` and `dd`.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/SearchNodeComponent.scala
```diff
@@ -868,7 +868,7 @@ object SearchNodeComponent {
           // init a jquery datepicker
           override def initForm(formId: String):    JsCmd = OnLoad(JsRaw("""var init = $.datepicker.regional['en'];
        init['showOn'] = 'focus';
-       init['dateFormat'] = 'dd/mm/yy';
+       init['dateFormat'] = 'yy-mm-dd';
```
fanf, line 871: the dev who thought it was a good idea to have `yy` for 4 digits year... *sigh*

fanf (approved):
LGTM, seems fine

clarktsiory:
OK, squash merging this PR

PR #6019, opened 2024-11-15
Fixes #25786: Add timezone field in campaign schedule

clarktsiory:
https://issues.rudder.io/issues/25786

Add `tz: Option[ScheduleTimeZone]` which is the ID of a IANA timezone (offsets are not supported). Also when saving a campaign the tz is the current server timezone. So, `None` means for old campaigns that no timezone was saved and the server one should be used when scheduling campaigns.  

A lot of unit tests were added, mostly to cover different cases for all schedules, and see that the variants with a timezone still work.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/campaign/CampaignSchedulerTest.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/campaign/CampaignSchedulerTest.scala
@@ -37,113 +37,105 @@
 
 package com.normation.rudder.campaign
 
-import com.normation.rudder.campaigns.DayTime
-import com.normation.rudder.campaigns.First
-import com.normation.rudder.campaigns.Last
-import com.normation.rudder.campaigns.MainCampaignScheduler
-import com.normation.rudder.campaigns.Monday
-import com.normation.rudder.campaigns.MonthlySchedule
-import com.normation.rudder.campaigns.Second
-import com.normation.rudder.campaigns.SecondLast
-import com.normation.rudder.campaigns.Sunday
-import com.normation.rudder.campaigns.Third
-import com.normation.zio.*
+import com.normation.errors.*
+import com.normation.rudder.campaigns.*
+import com.normation.rudder.campaigns.WeeklySchedule
 import org.joda.time.DateTime
+import org.joda.time.DateTimeZone
+import org.joda.time.format.ISODateTimeFormat
 import org.junit.runner.RunWith
+import org.specs2.matcher.Matcher
 import org.specs2.mutable.*
 import org.specs2.runner.JUnitRunner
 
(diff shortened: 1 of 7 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/campaign/CampaignSchedulerTest.scala
```diff
@@ -37,174 +37,752 @@
...
+  private def haveScheduleEnd(date: DateTime)              = {
+    beRight(beSome(be_===(date.toString(format)))) ^^ ((t: PureResult[Option[(DateTime, DateTime)]]) =>
+      t.map(_.map { case (_, end) => end.toString(format) })
+    )
+  }
+  private def haveSchedule(start: DateTime, end: DateTime) = {
+    haveScheduleStart(start) and haveScheduleEnd(end)
+  }
+
+  private def haveNoSchedule: Matcher[PureResult[Option[(DateTime, DateTime)]]] = beRight(beNone)
```
fanf, line 787: :exploding_head:

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/CampaignApiTest.scala
```diff
@@ -135,27 +167,83 @@ class CampaignApiTest extends Specification with AfterAll with Loggable {
...
     }
 
-    "save one campaign" in {
+    val c1json = {
+      s"""{"info":{
+         |"id":"c1",
+         |"name":"second campaign",
+         |"description":"a test campaign present when rudder boot",
+         |"status":{"value":"enabled"},
+         |"schedule":{"start":{"day":1,"hour":3,"minute":42},"end":{"day":1,"hour":4,"minute":42},"tz":"${tz}","type":"weekly"}
```
fanf, line 176: I'm not sure it's enough. We want: 
- the tz iso offset (which is the only relevant thing for computing times from our pov)
- the ts human name (to display the correct one to the human who selected it)

clarktsiory: okay so we really want to accept a known timezone ID, and nothing else (not just an offset) : an offset only could be wrong since there are rules for timezone change in some countries ([example in France](https://timezonedb.com/time-zones/Europe/Paris), there is DST and sometimes other future changes in offset)
So I guess we need to refuse everything else in the API, and parse the `ScheduleTimeZone` as a known ID...

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #6020, opened 2024-11-16
Fixes #25884: Migrate SecurityToken ldap json serialisation to ZIO

fanf:
https://issues.rudder.io/issues/25884

This PR is based on https://github.com/Normation/rudder/pull/5986 and need it to be merged before itself. 

The main goad of that PR is to migrate `SecurityToken` (so, only `Certificate` once #5986 is merged) serialization from lift to zio-json. 
In the process, the check for valid certificate data is enforced, so it adds a lot of certificate data string in tests. 
The other interesting changes: 
- don't use `SecurityToken` type in node api output, since the serialization is different. Just say it's a sting. 
- the mapping from/to postgres was using an additional step from/to `JSON` type and then mapping to `SecurityToken` (or whatever). This is inefficient and does not take into account the various `jsonField` and `jsonNameAlias` that could exist on `SecurityToken` json serialization
- there was some confusion between how much \n must be escaped. The rule is: 
    - in JSON, we need to have exactly the 2 chars '\n' printed, not the control char, so in scala it means """blabla\nblabla""" (and not s"""...""" which has other escaping rules IIRC)
    - in scala objects, it means using 
```
"""blabla
   |blabla""".stripMargin
``` 
(easiest way). 
    

(note : I wonder if the check on unserialization should not be replaced by a success + dummy value "invalide certificate". I'm thinking at the case where we init root server and before first inventory is done, I wonder if it wouldn't mean having no root server, which is not good) 
EDIT: I removed the check when decoding JSON. It's not the place to do it, and the fact to have a non inverse encode/decode pair is really asking for  trouble. The certificate check is done where it needs (API creation, new inventory, etc)


fanf:
PR rebased

webapp/sources/ldap-inventory/inventory-api/src/test/scala/com/normation/inventory/domain/AgentTypesTest.scala
```diff
@@ -71,7 +71,7 @@ class AgentTypesTest extends Specification {
    */
 
   val key =
-    "-----BEGIN RSA PUBLIC KEY-----\nMIIBCgKCAQEAtDTtbfw2ic1pcUludrs1I+HGGdRL0r5dKjMzLhUGahCrVaH7H3ND\n/py+XPZKY/Iolttgf1RriQAazEVbFEWXozistMTXtWJu/5IxV47QNqbS82KrhQNp\ns4abfqraGOlYbYS5BXCaHYrKI2VzvAwwvCsE7vmhnO1Br4AueagrFU+itjr/0gMd\nu58xYDiAADXqGDzES75NIxCZelv5vefMfpEMlBmztKmgY+iT+Q8lhf42WUsZ9OBl\nRDRfQ9VCW+8336C1JEpcHAcSElF4mn4D0GN7RvxNOSGpuLjxAvp5qFVbp4Xtd+4q\n8DaGe+w8MplwMVCFTyEMS3E1pS4DdctdLwIDAQAB\n-----END RSA PUBLIC KEY-----"
+    """-----BEGIN RSA PUBLIC KEY-----\nMIIBCgKCAQEAtDTtbfw2ic1pcUludrs1I+HGGdRL0r5dKjMzLhUGahCrVaH7H3ND\n/py+XPZKY/Iolttgf1RriQAazEVbFEWXozistMTXtWJu/5IxV47QNqbS82KrhQNp\ns4abfqraGOlYbYS5BXCaHYrKI2VzvAwwvCsE7vmhnO1Br4AueagrFU+itjr/0gMd\nu58xYDiAADXqGDzES75NIxCZelv5vefMfpEMlBmztKmgY+iT+Q8lhf42WUsZ9OBl\nRDRfQ9VCW+8336C1JEpcHAcSElF4mn4D0GN7RvxNOSGpuLjxAvp5qFVbp4Xtd+4q\n8DaGe+w8MplwMVCFTyEMS3E1pS4DdctdLwIDAQAB\n-----END RSA PUBLIC KEY-----"""
```
fanf, line 74: since it's for use in a json string, we need triple-double quote (whithout "s") so that the \n is really a written down \n (and not a line break), and has only one backslash

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -2117,10 +2092,7 @@ trait RudderJsonEncoders {
 
   implicit val vmTypeEncoder: JsonEncoder[VmType] = JsonEncoder[String].contramap(_.name)
 
-  implicit val agentTypeEncoder:         JsonEncoder[AgentType]           = JsonEncoder[String].contramap(_.displayName)
```
fanf, line 2120: this one is moved into the chimney transformer directly for the case of node api

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -2117,10 +2092,7 @@ trait RudderJsonEncoders {
...
-  implicit val agentVersionEncoder:      JsonEncoder[AgentVersion]        = JsonEncoder[String].contramap(_.value)
   implicit val nodeKindEncoder:          JsonEncoder[NodeKind]            = JsonEncoder[String].contramap(_.name)
-  implicit val securityTokenEncoder:     JsonEncoder[SecurityToken]       = JsonEncoder[String].contramap(_.key)
```
fanf, line 2123: these two are moved along in the corresponding companion objects of their classes

webapp/sources/rudder/rudder-core/src/test/resources/node-facts/node-001-machine-physical-v2.json
```diff
@@ -26,8 +26,8 @@
...
...MRVgtm2+I\\nw5ng1mujcaa/2iHEmenjK6mLcOh+IXrTEXRoV5JBGmTxEYvdKn9Lvv0LJUwc3WFs\\nBR6/AHBAhvd0U2IHMuxki3AaTM+A6atKUo4Hk2MC967Ecz3Tg06YIoZxJPqo2Qij\\n3ib2izzRFuzuVuECO9DNI51OewIDAQABo4GwMIGtMAwGA1UdEwQFMAMBAf8wHQYD\\nVR0OBBYEFIjIpuhURlIay8ffm0w4enOJvoeHMHEGA1UdIwRqMGiAFIjIpuhURlIa\\ny8ffm0w4enOJvoeHoTqkODA2MTQwMgYKCZImiZPyLGQBAQwkOTMwZGNjYjAtNWIw\\nNi00Mzg1LWFiN2EtODBhN2U1NmViOTY5ghQk7n+bw9lF6xJlmMDwv3NJG+RpeTAL\\nBgNVHQ8EBAMCArwwDQYJKoZIhvcNAQELBQADggIBAJRQHHK9RD6ByGOFJ+8w3yH0\\n8QJ78uZe3K+N2dWwbLo6XSV9nvtkfyuUxsaBTTgXSmvkSlGNrCFnWW5vaea/R1Rq\\nw4zGlyS4UzBD31rc2kKl5/yIljYiufYsbMmYqh8Sy/U75msp4inRzVlyxX1FtkgW\\nDYr8nrcHbD94xuT7HTFX8HFAUOsGhEGDYXGvY94iJ4Jx7p5bWiVDbIME7Bf72+Ql\\ncWUapVCXVbuqWW6qw048EGcwjmMXJXNJgBth39yX2cEWklkSENuYwSLAloetUmYY\\nuSl6Y20pTKjSdUTBGIu0uN+0740LuiMlQ1nbdrBBasgLxH/1TfAp927zuQCPKANA\\n1O7OuhrF3gwLOVDjkycAhQb9nynZgpexmFkOMklXVYNWKp6s88thB3gTRCT01GFx\\ntY5WzobFShmv6AEmUWYutFG5k7fwNKGI6VEUVoQlNiHeDW4VV4AFrfK+/vNnJepf\\nqbSyRhS8QBgqt4KbwMdarVEZTvVrc6JhfnTSzXig6dO8bVA1EVLp0xyCd5eoplEV\\nHzhAz2MOrfOUmBjo2HcHA4tpcuE+ImNWuQiDfTdjqkZWRO5pBwbvnVLOGCqtaH1C\\nsGM8wv+QA1Y2NhgzADIb6u4PjXcBy6hqyAFKtvJ2J1pBL1Vtu4rTf9zhZ62HSqBd\\nRlYSlie1YMo1Yy2jzZhP\\n-----END CERTIFICATE-----"
```
fanf, line 30: we could have read kind/token, but the test then serialize again, and the preference goes to type/value (to keep what is in LDAP), and to the back comparison fails.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFact.scala
```diff
@@ -668,7 +668,7 @@ object NodeFact {
   }
 
   def defaultRudderAgent(localAdmin: String): RudderAgent = {
-    RudderAgent(AgentType.CfeCommunity, localAdmin, AgentVersion("unknown"), PublicKey("not initialized"), Chunk.empty)
+    RudderAgent(AgentType.CfeCommunity, localAdmin, AgentVersion("unknown"), Certificate("not initialized"), Chunk.empty)
```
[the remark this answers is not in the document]
fanf: EDIT: not relevant anymore

clarktsiory (commented):
A few comments, there seem to be a non-iso change.
Also the `\\n` vs `\n` thing frightens me, is it really compatible to use the latter one ?

fanf:
for `\n` thing:

- in json, line break are textually written with two chars, `\` then `n` : `{"foo": "bar\nbaz"}`. 
- This is then decoded as the js/scala string: 
```
bar
baz
```
- in scala, the simple quote `\n` means the control char line break, not the two chars. But the triple-quoted one means `\` then `s`. And again, not the s-triple quoted one, which has the same semantic than the simple quoted. So that we have: 
```
 scala> s"""foo\nbar""" == """foo\nbar"""
val res0: Boolean = false

scala> s"""foo\nbar""" == "foo\nbar"
val res1: Boolean = true

scala> s"""foo\nbar""" == """foo
     | bar"""
val res2: Boolean = true
```
- in certificate, for the PEM format, we need to have line break at specific line lenght, so the actual string: 

``` cat certificate.pem
-----BEGIN CERTIFICATE-----
MIIFTjCCAzagAwIBAgIUfa0+S+CyJahRzuwNNOFLNQQjIH8wDQYJKoZIhvcNAQEL
BQAwFzEVMBMGCgmSJomT8ixkAQEMBW5vZGU3MB4XDTI0MDMwMTExMDIxM1oXDTM0
MDIyNzExMDIxM1owFzEVMBMGCgmSJomT8ixkAQEMBW5vZGU3MIICIjANBgkqhkiG
9w0BAQEFAAOCAg8AMIICCgKCAgEAsmBUYI1A5vqkOTCW24m/mQhBaRu4WaAUXDAr
ArdIAAMN0KyHhEXP1/X32Flw90E0VjtUH4P+DYLBozXYWhJrUxdWLyn3TRSbv3Kr
...
```
So we need that string when giving to bouncycastle java security methods with actuall line break control chars (simple quote or s-triple quote). 
But in JSON, we need the string with "\" then "s", ie: 
``` some json content
"""-----BEGIN CERTIFICATE-----\nMIIFTjCCAzagAwIBAgIUfa0+S+CyJahRzuwNNOFLNQQjIH8wDQYJKoZIhvcNAQEL\nBQAwFzEVMBMGCgmSJomT8ixkAQEMBW5vZGU3MB4XDTI0MDMwMTExMDIxM1oXDTM0\nMDIyNzExMDIxM1owFzEVMBMGCgmSJomT8ixkAQEMBW5vZGU3MIICIjANBgkqhkiG\n9w0BAQEFAAOCAg8AMIICCgKCAgEAsmBUYI1A5vqkOTCW24m/mQhBaRu4WaAUXDAr\nArdIAAMN0KyHhEXP1/X32Flw90E0VjtUH4P+DYLBozXYWhJrUxdWLyn3TRSbv3Kr....
```
So we use scala triple-quote without s.

webapp/sources/ldap-inventory/inventory-api/src/main/scala/com/normation/inventory/domain/DataTypes.scala
```diff
@@ -64,15 +65,24 @@ final case class Manufacturer(name: String) extends AnyVal
...
+sealed abstract class SecurityToken(val kind: String) {
+  def key:   String
+  def value: String
 }
 
 object SecurityToken {
+  implicit val securityTokenEncoder: JsonEncoder[SecurityToken] =
+    DeriveJsonEncoder.gen[JsonEncSecurityToken].contramap(t => JsonEncSecurityToken(t.kind, t.value))
+  implicit val securityTokenDecoder: JsonDecoder[SecurityToken] =
+    DeriveJsonDecoder.gen[JsonDecSecurityToken].mapOrFail(t => Right(Certificate(t.value)))
```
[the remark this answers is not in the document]
fanf: Since we can't have anything but `Certificate` since 7.0, it isn't useful to test for the kind, we only have one. If something wasn't migrated since a long time, it should have failing inventories for ages, and the result will "just" be that the content won't be a valide certificate, so generation will fail for that node, which is what we want. 

I didn't do the whole migration that could remove security token trait and only keep certificate and remove a bunch of pattern matcher and etc, because here, it was mostly for the json. 

We can make an other pass of cleaning once we're sure that nothing break here - but it's less important or urgent than moving away from lift-zon.

fanf:
Self merging because of relayd errors

fanf:
OK, merging this PR

PR #6022, opened 2024-11-17
Fixes #25886: Migrate Environment variable from lift-json to zio-json

fanf:
https://issues.rudder.io/issues/25886


webapp/sources/ldap-inventory/inventory-repository/src/test/scala/com/normation/inventory/ldap/core/TestNodeUnserialisation.scala
```diff
@@ -155,13 +155,55 @@ class TestNodeUnserialisation extends Specification {
...
       |inventoryDate: 20180717000031.000Z
       |receiveDate: 20180717000527.050Z
       |lastLoggedUserTime: 20000714084300.000Z
-      |softwareUpdate: {"name":"rudder-agent","version":"7.0.0-realease","from":"yum","arch":"x86_64","kind":"none","description":"Local privilege escalation in pkexec","severity":"low","date":"2022-01-26T00:00:00Z","ids":["RHSA-2020-4566","CVE-2021-4034"]}
+      |softwareUpdate: {"name":"rudder-agent","version":"7.0.0-realease","arch":"x86_64","from":"yum","kind":"none","description":"Local privilege escalation in pkexec","severity":"low","date":"2022-01-26T00:00:00Z","ids":["RHSA-2020-4566","CVE-2021-4034"]}
```
fanf, line 201: the change was needed because of the idempotency tests, but the previous line still works

fanf:
PR updated with a new commit

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
```diff
@@ -844,7 +840,7 @@ class InventoryMapper(
   // map process from node
   def processesFromNode(node: NodeInventory): Seq[String] = {
     // convert the processes
-    node.processes.map(x => Serialization.write(x))
+    node.processes.map(_.toJson)
```
clarktsiory, line 843: yes ! this could have already been changed since a while now :sweat_smile:

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
```diff
@@ -1030,8 +1026,8 @@ class InventoryMapper(
 
       lastLoggedUser     = entry(A_LAST_LOGGED_USER)
       lastLoggedUserTime = entry.getAsGTime(A_LAST_LOGGED_USER_TIME).map(_.dateTime)
-      ev                 = entry.valuesFor(A_EV).toSeq.map(Serialization.read[EnvironmentVariable](_))
-      process            = entry.valuesFor(A_PROCESS).toSeq.map(Serialization.read[Process](_))
+      ev                 = entry.valuesFor(A_EV).toSeq.map(_.fromJson[EnvironmentVariable]).collect { case Right(ev) => ev }
+      process            = entry.valuesFor(A_PROCESS).toSeq.map(_.fromJson[Process]).collect { case Right(p) => p }
```
clarktsiory, line 1030: Since the `Serialization.read` method throws, the equivalent would instead be a traverse/ZIO.foreach here instead of failing silently ?

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
```diff
@@ -51,7 +51,6 @@ import com.normation.ldap.sdk.schema.LDAPObjectClass
 import com.softwaremill.quicklens.*
 import com.unboundid.ldap.sdk.{Version as _, *}
 import java.net.InetAddress
-import net.liftweb.json.*
```
clarktsiory, line 54: :fire:

clarktsiory (commented):
LGTM apart from the non-iso error handling ! I'm eagerly waiting for the merge :innocent:

webapp/sources/ldap-inventory/inventory-repository/src/main/scala/com/normation/inventory/ldap/core/InventoryMapper.scala
fanf: the previous case was just a bug I think. 
So what do we want ? Not a full error and an unreadable inventory because a mear process / env var was uncorectly written in base and can't be read back, because it's completely unactionnable by the user. It's even possible that just sending an other inventory doesn't solve the problem by overriding datas, since we read existing one before writting (which means, we may have other problems). Ok, let's write a test on that. 

I can log before ignoring, but I'm pretty sure that log either won't be read ever, or will cause major error like filling disks.

fanf: Added logs and a test to ensure we can deserialize inventories in presence of incorrect data. 
Alos, did the same for custom properties which were ignored too.

webapp/sources/ldap-inventory/inventory-repository/src/test/scala/com/normation/inventory/ldap/core/TestNodeUnserialisation.scala
```diff
@@ -195,5 +275,114 @@ class TestNodeUnserialisation extends Specification {
...
+        p  = mapper.processesFromNode(n)
+        _  = e.resetValuesTo(LDAPConstants.A_PROCESS, p*)
+      } yield e
+
+      val entry = ZioRuntime.unsafeRun(prog)
+      val diff  = com.unboundid.ldap.sdk.Entry.diff(nodeEntry.backed, entry.backed, false, true)
+      diff.toArray() must beEmpty
+    }
+
+    "correctly unserialize a node even with malformed env var, process, soft update" in {
```
clarktsiory, line 345: okay, I guess if someone is having this problem we haven't heard about it yet :innocent:

clarktsiory (approved):
OK then the change looks good to me

PR #6023, opened 2024-11-17
Fixes #25887: Migrate from lift-json to zio-json in LDAPEntityMapper 

fanf:
https://issues.rudder.io/issues/25887

Cleaning up LDAPEntityMapper from lift-json. 
These is 3 commit: 
- the first one is the most involved (nothing hard) and deal with AgentRunInfo. It has two aspect, either from/to LDAP or from/to ELM. I tried to make that clearer 
- HeartbeatConfiguration is trivial
- AcceptationTime has nothing hard, it's just a map of fancy types.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
```diff
@@ -65,28 +68,53 @@ final case class HeartbeatConfiguration(
...
+final case class ElmAgentRunInterval(
+    overrides:   Option[Boolean], // true, false, null because used in flag
+    interval:    Int,
+    startHour:   Int,
+    startMinute: Int,
+    splayHour:   Int,
+    splayMinute: Int
+)
+object ElmAgentRunInterval    {
+  implicit val codecJsonAgentRunInterval: JsonCodec[ElmAgentRunInterval] = DeriveJsonCodec.gen
```
clarktsiory, line 85: I am not a fan of the `Elm` naming here, moreover I understand that the "Decoder" part is unused in prod, this case class is used to "split" the `splaytime` into `splayHour` and `splayMinute` for encoding, which is not specific to `Elm` at all, we may want to have it available via API...
And now since the decoder is provided with the declared codec, we are tempted to use it in tests... maybe we should limit the case class to the current "split splaytime and encode" use case and only provide the encoder part ?
(The visibility of the class may even be `private`, only accessible with through the `jsonWithSplay` method)

clarktsiory (commented):
LGTM but I'm not sure about the "Elm" naming, maybe at least the name should be more generic ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/reports/ReportingConfiguration.scala
fanf: Ok for the codec, I though I had change it already. 
For the elm part, it's really because it's the Elm semantic for JS to get a `Maybe[Boolean]`, there is no other reason to use the null case without that, our convention everywhere are to not have optionnal fields in JSON in the `None` case. 
Here, we only need it because: 
- it's used in the init part of the application
- there's not dedicated decoder for it in elm
- so it relies on Elm automatic mapping. 

I didn't want to change things in the Elm part because it's not the goal of these PR, the goal is just to remove lift-json without altering existing interface if possible - which it is, here. 
We can totally do a latter refactoring to change the elm init method of that app, and make it use a correct decoder, so that the `overrides: null` case can go away.

fanf: OK made it clearer by moving all things specific to `AgentScheduleEditForm` into that class.  
And really `AgentScheduleEditForm` and corresponding elm app are ripped for refactoring:
- the decoding is done with lift-json parsing and is not clearly isomorph to the encoding,
- the init is bound to lift
- other strange things

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -532,34 +528,20 @@ class LDAPEntityMapper(
...
+    (for {
+      l <- value.fromJson[Map[String, String]]
+      t <- l.toList.traverse {
+             case (a, b) =>
+               for {
+                 tv <- TechniqueVersion.parse(a)
+                 dt <- GeneralizedTime.parse(b).toRight(s"Error when parsing '${b}' as a generalized time'")
+               } yield (tv, dt.dateTime)
+           }
+    } yield t.toMap).left.map(s => Unexpected(s))
```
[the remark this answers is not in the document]
fanf: I didn't do it because I'm really not sure it worth it. 

There is several saillant differences from standard object that make it less nice, almost bad: 
- there is no standard place to put the codec. Contrary to a case class where it lives in the the companion object and it's the standard place to look at, here, we need to put it somewhere and when we need it, find the correct class where it was defined ; (see at the end of `LDAPEntityMapper`, its companion object)
- the quantity of plumbing is more than the direct implementation, 
- we need to encapsulate the `DateTime` codec in a sub-object to avoid importing them elsewhere because of the `import LDAPEntityMapper.*`
- but more dangerously, for a newcomer, looking for decoding `Map[TechniqueVersion, DateTime]`, the use will very likely look for `TechniqueVersion` decoder, find it, a `DateTime` decoder, find it, and map the result to map. Especially someone who doesn't know `zio-json` special case for map keys. And the result will be an incorrect decoding. 

All that let me think that: the root problem is that decoding `Map[A, B]` by themselve is not idiomatic at all in `zio-json`. Either you decode the whole englobing object and in the that object companion, you put the specific map codec, or you encapsulate the Map into an other object and make the decoder in its companion (circling back to previous case), 

I wanted to avoid changing the API of active technique, so I also avoided going to the full "encapsulate map into a `AcceptationTime` object" even if it's clearly the correct, principled solution, because it was not the goal of that refactoring. 

But since I wrote all the plumbing to show you the problems, I will do the last miles in a subsequent commit.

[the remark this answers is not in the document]
clarktsiory: I agree that it may be significantly harder to make something out of all the declarative values, than just a "mapper" class, but the boilerplate is reduced to almost the same cost I think specifically when creating the "opaque type" like you did (we only need to import zio.json.* instead of instantiating the mapper).
Many lib also use the same declarative style so it would not be hard to migrate (I hope we won't need to do that !)

clarktsiory:
LGTM ! The new style is very idiomatic ! 
It can be merged (but rebased and squashed first !)

PR #6025, opened 2024-11-18
Fixes #25841: Create an instance id on each root server

clarktsiory:
https://issues.rudder.io/issues/25841

We need a migration to create the instance ID file, but we also want the instance ID to be loaded from the file on Rudder init : 
* create a service that reads the file and that just expose the instance id, if it does not exist a random value should be created 
* the check takes the service and does the file creation if needed


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -3305,13 +3305,18 @@ object RudderConfigInit {
     lazy val mainCampaignService = new MainCampaignService(campaignEventRepo, campaignRepo, uuidGen, 1, 1)
     lazy val jsonReportsAnalyzer = JSONReportsAnalyser(reportsRepository, propertyRepository)
 
+    lazy val instanceUuidPath    = root / "opt" / "rudder" / "etc" / "instance-id"
+    lazy val instanceIdGenerator = new InstanceIdGeneratorImpl()
+    lazy val instanceIdService   = InstanceIdService.make(instanceUuidPath, instanceIdGenerator).runNow
```
fanf, line 3310: We will want to expose it in the `RudderConfigApi` (in `rci = init()` and then add one line with trait signature and the corresponding service.

fanf (changes requested):
Looks good, it just need to be exposed in `RudderConfigApi`

clarktsiory:
PR rebased

PR #6026, opened 2024-11-18
Fixes #25889: Broken system-updates module build

amousset:
https://issues.rudder.io/issues/25889


amousset:
PR updated with a new commit

PR #6021, opened 2024-11-16
Fixes #25885: Migrate CustomProperties form lift-json to zio-json

fanf:
https://issues.rudder.io/issues/25885

Nothing special, it was quite automatic. 

EDIT: rebased on top of #6020 to ease consistency with data test and imports

PR #6027, opened 2024-11-18
Fixes #25894: Migrate CmdbQuery to zio-json

fanf:
https://issues.rudder.io/issues/25894

In the same vein as #6020, #6021, #6022, etc

In this one, a part of the refactoring is about putting ADT values in the companion object of the sealed root. 
It also needs to use an intermediary for `CriterionLine` and chimney mapping from it because that case class is just too complicated for the simple encoding only we want. If one day we build a full codec, we will be able to delete `JsonCriterionLine` and the plumbing around. 

A needed to `liftjson.parse(query.toJson)` in group serialization because it awaits a `JValue`. It's not a problem, it will be removed once we have migrated group API.


clarktsiory (approved):
LGTM !

PR #6031, opened 2024-11-19
Fixes #25903: Refactor API tokens after clear-text removal

amousset:
https://issues.rudder.io/issues/25903

A similar change needs to be done in the api-authorizations plugins for user tokens.

## Backend

* Since #6030 the old tokens are not usable anymore, this pull request takes advantage of this to refactor this part.

### Data types

Split the `ApiToken` type, which used to contain both clear-text and hashed values into two separate types to prevent confusions and prevent misusage as much as possible:

* `ApiTokenSecret`: The token value, to be sent to the creator and not stored on the server
* `ApiTokenHash`: The token hash, as stored (either in LDAP or memory)
  * Make no effort to prepare for a v3 (appart from storage), it would only be needed in case sha2 is not considered safe anymore. Future changes to the API access would likely be for something quite different (OAUTH/JWT/...), with a different architecture.

Also modifying the account types:

* `ApiAccount` now contains a `ApiTokenHash`
* A new `NewApiAccount` is created, to be used after creation, and contains a `ApiTokenSecret`

Notes:
* The system token now also uses a hash, and the clear text value is not kept after the token files creation.
* Make the access to the secret and hash values explicit with `expose()` methods (and make the value private)

### Usage

The JSON serialization of the accounts, used by the API tokens Web interface, is modified:

* For new accounts, the token value is sent
* For existing account, the token hash version only is sent

## Frontend

* Display the list of unsupported tokens along with a global warning
* Hide the enable/disable button for unsupported tokens (as it could not be actionnable)

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
@@ -72,8 +72,9 @@ final case class ApiAccountName(value: String) extends AnyVal
  * All tokens are 32 alphanumeric characters, optionally
  * followed by a "-system" suffix, indicating a system token.
  *
+ * Use the [[exposeSecret()]] method to retrieve the held value
  */
-final case class ApiTokenSecret(private val secret: String) {
+final class ApiTokenSecret private (secret: String) {
   // Avoid printing the value in logs, regardless of token type
   override def toString: String = "[REDACTED ApiTokenSecret]"
 
@@ -95,9 +96,11 @@ final case class ApiTokenSecret(private val secret: String) {
 object ApiTokenSecret {
   private val tokenSize = 32
 
+  def apply(secret: String) = new ApiTokenSecret(secret)
+
   def generate(tokenGenerator: TokenGenerator, suffix: String = ""): ApiTokenSecret = {
     val completeSuffix = if (suffix.isEmpty) "" else "-" + suffix
-    ApiTokenSecret(tokenGenerator.newToken(tokenSize) + completeSuffix)
+    new ApiTokenSecret(tokenGenerator.newToken(tokenSize) + completeSuffix)
   }
 }
 
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
@@ -40,7 +40,6 @@ import cats.data.*
 import cats.implicits.*
 import com.normation.errors.Inconsistency
 import com.normation.errors.PureResult
-import com.normation.rudder.api.ApiToken.prefixV2
 import com.normation.rudder.facts.nodes.NodeSecurityContext
 import enumeratum.*
 import java.nio.charset.StandardCharsets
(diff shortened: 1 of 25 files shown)


amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

amousset:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -58,53 +57,102 @@ final case class ApiAccountId(value: String) extends AnyVal
...
+ * followed by a "-system" suffix, indicating a system token.
  *
- * There are two versions of tokens:
+ */
+final case class ApiTokenSecret(private val secret: String) {
+  // Avoid printing the value in logs, regardless of token type
+  override def toString: String = "[REDACTED ApiTokenSecret]"
+
+  // Prevent comparison
+  override def equals(obj: Any) = false
```
clarktsiory, line 71: this is confusing as since the Object equality and hashcode in Java follows some rules, here an object would not be equal to itself !

I think the good practice would be to avoid using the default `equals` method, and use another `equalsToken()`, or a `def isEqual(this: ApiTokenSecret, that: ApiTokenSecret)` instead. Many examples apply this, such as for date time types that are also subject to custom comparison logic.
When doing that we should document that the equals method should not be used, it all relies on code convention !

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -58,53 +57,102 @@ final case class ApiAccountId(value: String) extends AnyVal
...
-    value.take(4) + "[SHORTENED ApiToken]"
+final case class ApiTokenHash(private val value: String) {
+  override def toString: String = "[REDACTED ApiTokenHash]"
+
+  // Constant time comparison
+  override def equals(obj: Any): Boolean = {
+    obj match {
+      case ApiTokenHash(other) => MessageDigest.isEqual(value.getBytes(), other.getBytes())
+      case _                   => false
+    }
```
clarktsiory, line 125: Same issue about custom `equals` (here it is even more obvious why we need a custom equality method, `MessageDigest` has it, we should too !)

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -58,53 +57,102 @@ final case class ApiAccountId(value: String) extends AnyVal
...
   }
 }
 
-object ApiToken {
-  private val tokenSize = 32
-  private val prefixV2  = "v2:"
+object ApiTokenHash {
+  val prefix     = "v2"
+  // Guaranteed to never match
+  val neverMatch = ApiTokenHash.build("not-matching")
```
clarktsiory, line 143: It could work like this, but the "not-matching" value is kind of magic...
(also I guess that "match" means that it should not "equal" any other ApiTokenHash ?)

The solution is often to work around with types, the `final case class ApiTokenHash` may be in fact not only a data container but also the _"token that is deprecated"_), so a simple solution would be changing it into a sealed trait instead : 
```scala
trait CompareToken {
  def equalsToken(other: ApiTokenHash): Boolean
}
sealed trait ApiTokenHash extends CompareToken {
  override def equalsToken(token: ApiTokenHash) = (this, token) match {
    case (Deprecated, Deprecated) => true
    case (V2(a), V2(b)) if MessageDigest.isEqual(a.getBytes(), b.getBytes()) => true
    case _ => false
  }
}
object ApiTokenHash {
  private case object Deprecated extends ApiTokenHash
  private case class V2(value: String) extends ApiTokenHash
  // V2 is the default 
  private def build(value: String): ApiTokenHash = {
    V2(prefix + ":" + value)
  }

  ...
}
```

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlSerialisationImpl.scala
```diff
@@ -536,7 +536,7 @@ class APIAccountSerialisationImpl(xmlVersion: String) extends APIAccountSerialis
       (
         <id>{account.id.value}</id>
        <name>{account.name.value}</name>
-       <token>{account.token.value}</token>
+       <token>{account.token.exposeHash()}</token>
```
clarktsiory, line 539: Since in an `Option` is returned, it will write `Some("thehash")`, and this would break. It should be a String, so `.getOrElse("")`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/User.scala
```diff
@@ -123,7 +123,7 @@ case class RudderUserDetail(
 
   override val (getUsername, getPassword, getAuthorities) = account match {
     case RudderAccount.User(login, password) => (login, password, RudderAuthType.User.grantedAuthorities)
-    case RudderAccount.Api(api)              => (api.name.value, api.token.value, RudderAuthType.Api.grantedAuthorities)
+    case RudderAccount.Api(api)              => (api.name.value, api.token.exposeHash(), RudderAuthType.Api.grantedAuthorities)
```
clarktsiory, line 126: this will not compile now that we have an option, we need a getOrElse

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -58,53 +57,130 @@ final case class ApiAccountId(value: String) extends AnyVal
...
+  }
+
+  /*
+  Deprecated or invalid hashes are not returned.
+   */
+  def exposeHash(): Option[String] = {
+    this match {
+      case ApiTokenHash.V2(h) => Some(h)
+      case _                  => None
+    }
```
clarktsiory, line 137: very nice !

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonDecoder.elm
```diff
@@ -54,9 +55,19 @@ decodeAcl =
         |> required "verb" string
 
 
+parseToken : String -> ( Token )
+parseToken str =
+  case str of
+    "2" -> Token.Hashed
+    "1" -> Token.ClearText
+    _   -> Token.New str
+
```
clarktsiory, line 64: okay, I would have used another "tokenVersion" attribute to avoid confusion on this information sensitive field, but since it is an internal endpoint I think it is okay

clarktsiory (commented):
Great change !
There only is some compilation problems

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlSerialisationImpl.scala
clarktsiory: It compiles now but this will cause a problem without the getOrElse : https://scastie.scala-lang.org/h2QlFY3BSzqlEBruNceAvA

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -1142,7 +1142,7 @@ class LDAPEntityMapper(
     mod.resetValuesTo(A_API_UUID, principal.id.value)
     mod.resetValuesTo(A_NAME, principal.name.value)
     mod.resetValuesTo(A_CREATION_DATETIME, GeneralizedTime(principal.creationDate).toString)
-    mod.resetValuesTo(A_API_TOKEN, principal.token.exposeHash())
+    mod.resetValuesTo(A_API_TOKEN, principal.token.exposeHash().getOrElse(""))
```
fanf, line 1145: I don't think that will work. The API token is a mandatory field in the schema for `apiAccount`. 
Perhaps we want to be able to have that case, but it seems to be a hard hypothesis in the lower layout. We can perhaps workaround it with a marker value that would be extracted as none.

fanf: or even `v2:`

fanf: Actually, this will be possible with https://github.com/Normation/rudder/pull/6156

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -68,53 +67,102 @@ object ApiAccountId {
...
+ * The actual authentication token, in clear text.
  *
- * TODO: Once support for plain text tokens is dropped, make separate types for plain and hashed tokens.
-         Current situation is confusing, and hence a bit risky.
+ * All tokens are 32 alphanumeric characters, optionally
+ * followed by a "-system" suffix, indicating a system token.
  *
- * There are two versions of tokens:
+ */
+final case class ApiTokenSecret(private val secret: String) {
```
fanf, line 76: Shoukd it be a case class uf it doesn't have any of its property?
(Answer:  nope, make it a class with a factory if you want to avoid the `new`)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -68,53 +67,102 @@ object ApiAccountId {
...
  */
-case class ApiToken(value: String) extends AnyVal {
-  // Avoid printing the value in logs, regardless of token type
-  override def toString: String = "[REDACTED ApiToken]"
 
-  // For cases we need to print a part of the plain token for debug.
-  // Show the first 4 chars: enough to disambiguate, and preserves 166 bits of randomness.
-  def exposeSecretBeginning: String = {
-    value.take(4) + "[SHORTENED ApiToken]"
+final case class ApiTokenHash(private val value: String) {
```
fanf: Same here

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
clarktsiory: Indeed, it is not supposed to be used for comparison (with hash/equality : the hash is in fact the ApiTokenHash class), nor for serialization.
And it's value is not exposed as a simple value but from an explicit getter `exposeSecret`.

I changed this to a `final class`, and added a comment : https://github.com/Normation/rudder/pull/6031/commits/35e7511c0197107c6c9ed2a036abd39368172046

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
clarktsiory: This has changed to a sealed trait with private subtypes in the final version

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
clarktsiory: Rather this one : https://github.com/Normation/rudder/pull/6031/commits/b3e2b3d00102ea5514d1ff9255a9c4334618af16
The value is private and we can have an apply constructor

amousset:
OK, squash merging this PR

PR #6040, opened 2024-11-22
Fixes #25935: 

clarktsiory:
https://issues.rudder.io/issues/25935

We now have to consider the errored/success value of the `unsafe-hashes` attribute in the boot check, so we need to refactor reading the attribute into a method, similarly to `hash`.
Then we read the attribute and when there is an error reading either of them, we `enforceModern` i.e. we use bcrypt and disallow unsafe hashes.

N.B. : this change only affects the boot check, later when we will want to apply the enforcing at every file reload, we should be backporting that in 8.1

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckUsersFile.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckUsersFile.scala
@@ -45,38 +45,53 @@ import com.normation.rudder.users.UserFileProcessing
 import com.normation.rudder.users.UserFileSecurityLevelMigration
 import com.normation.rudder.users.UserManagementIO
 import com.normation.zio.UnsafeRun
-import zio.*
+import zio.ZIO
 
 class CheckUsersFile(migration: UserFileSecurityLevelMigration) extends BootstrapChecks {
 
   override def description: String = "Check if hash algorithm for user password is a modern one, if not enable unsafe_hashes"
 
-  def allChecks(currentSecurityLevel: SecurityLevel): IOResult[Unit] = {
-    for {
-      _ <- currentSecurityLevel match {
-             case SecurityLevel.Modern => ZIO.unit
-             case SecurityLevel.Legacy => migration.allowLegacy(UserManagementIO.getUserFilePath(migration.file))
-           }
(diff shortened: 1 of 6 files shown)


clarktsiory:
Commit modified

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestCheckUsersFile.scala
```diff
@@ -28,29 +23,53 @@ class TestCheckUsersFile extends Specification with AfterAll {
...
+    "migrate unknown hash to bcrypt with unsafe-hashes set to false" in withMigrationCtx(unknownHashFile) {
+      case (getFile, checkUsersFile) =>
+        checkUsersFile.prog.runNow
+        getFile() must (haveHash("bcrypt") and haveUnsafeHashes(Some("false")))
+    }
+
+    "migrate non-boolean unsafe-hashes to false" in withMigrationCtx(unknownUnsafeHashesFile) {
+      case (getFile, checkUsersFile) =>
+        checkUsersFile.prog.runNow
+        getFile() must (haveHash("bcrypt") and haveUnsafeHashes(Some("false")))
```
clarktsiory, line 57: these two test cases have been added

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckUsersFile.scala
```diff
@@ -45,38 +45,53 @@ import com.normation.rudder.users.UserFileProcessing
...
+                         case (Left(unknownValue), _) =>
+                           BootstrapLogger.warn(
+                             s"Error when reading users file hash in ${migration.file.name}, value '${unknownValue}' is unknown, falling back to secure authentication method"
+                           ) *>
+                           migration.enforceModern(userFile)
+                         case (_, Left(unknownValue)) =>
+                           BootstrapLogger.warn(
+                             s"Error when reading users file unsafe-hashes in ${migration.file.name}, value '${unknownValue}' is unknown, falling back to safe hashes"
+                           ) *>
+                           migration.enforceModern(userFile)
```
clarktsiory, line 82: this corresponds to the change of semantics

PR #6041, opened 2024-11-22
Fixes #25941: Spurious failing test for inherited properties rest API

fanf:
https://issues.rudder.io/issues/25941

The current "test all files" REST test is not very user friendly because we only have one `restTestSetUp` for all tests. 
It's good for speed, because it means we only do one time the set up, which is quite long. 
But it also means that each file is bounded to others, and the test are not orthogonal at all. 
This means a dev has to think not only to the previous tests in the file (that's OK), but also to all other tests (which means each new test has an additionnal cost, when we want a constant, or reduced cost for adding tests). 

I can't do the change easily since it needs to change the API, and so it implies change in ALL plugins - so it's an other PR. But I think we will need to do it, even if it means one rest set-up by file. 

In the meantime, the "solution" is to for a sequential ordering of file, so AT LEAST the order are consistant between all environnements running tests, and so the sequence of change is the same .

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/test/resources/api/api_nodes.yml
+++ b/webapp/sources/rudder/rudder-rest/src/test/resources/api/api_nodes.yml
@@ -1322,28 +1322,6 @@ response:
                   "mode" : "expire_immediately"
                 }
               }
-            },            
-            {
-              "name" : "someJson2",
-              "value" : "string",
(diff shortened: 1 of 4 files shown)


clarktsiory (approved):
LGTM !

fanf:
OK, squash merging this PR

PR #6042, opened 2024-11-24
Fixes #25945: Add and Remove/deprecate API for Rudder 8.3

fanf:
https://issues.rudder.io/issues/25945

Based on #6023 to takes benefits from already migration Lift stuff. 

Along with the deletion of old api version, I did a clean-up of our existing APIs on 3 points: 
- 1/ normalize pattern matching of API implementation, 
- 2/ remove pre-v14 API alternative (seriously)
- 3/ clean-up all unused services or class parameters I could find. 

# 1/ Normalize pattern matching of API implementation

We have a rule of pattern-matching the API declared endpoints so that we're sure no implementation is missing, since we can't enforce that point by types (at least I never found a reasonnable way to do it). We had quite a few ways of doing it, I normalized with the shortest form, which also seems the clearest for that reason (and of course keep the exhautivness check): 

```
class UserManagementApiImpl(...) extends LiftApiModuleProvider[UserManagementApi] {
  override def schemas: ApiModuleProvider[UserManagementApi] = UserManagementApi

  override def getLiftEndpoints(): List[LiftApiModule] = {
    UserManagementApi.endpoints.map {
      case UserManagementApi.GetUserInfo     => GetUserInfo
      case UserManagementApi.ReloadUsersConf => ReloadUsersConf
      case UserManagementApi.AddUser         => AddUser
      ....
    }
  }
```

# 2/ remove pre-v14 API

In several API, we still had the alternative for pre-/post-zio introduction in v14. I removed that it. Concerned API: 
- rules
- parameters
- directives
- groups
- techniques

I also renamed the "xxxxV14" in "xxxx", because V14 isn't very telling anymore (but it's well that API implementation that I kept). 

In the same step, it removes most usage of `lift-json` in these API. 

# 3/ remove unused class parameters in API

Following step 2, but also because we missed some unused class parameters in the past, we had a lot of injected dependencies via class parameters which were not used anymore in API class. I cleaned-up all I saw.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ArchiveApi.scala
@@ -61,8 +61,6 @@ import com.normation.rudder.apidata.implicits.*
 import com.normation.rudder.batch.AsyncDeploymentActor
 import com.normation.rudder.batch.AutomaticStartDeployment
 import com.normation.rudder.configuration.ConfigurationRepository
-import com.normation.rudder.domain.appconfig.FeatureSwitch
-import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.rudder.domain.nodes.NodeGroup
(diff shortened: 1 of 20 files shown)


clarktsiory (approved):
Very helpful cleaning, I removed a few dead lift-json imports and methods along the way in two APIs !
LGTM !

clarktsiory:
OK, squash merging this PR

PR #6044, opened 2024-11-25
Fixes #25948: Score breakdown has inconsistent numbers

VinceMacBuche:
https://issues.rudder.io/issues/25948

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
@@ -40,24 +40,30 @@ package com.normation.rudder.score
 import com.normation.errors.IOResult
 import com.normation.inventory.domain.InventoryError.Inconsistency
 import com.normation.inventory.domain.NodeId
+import com.normation.rudder.facts.nodes.NodeFactRepository
+import com.normation.rudder.facts.nodes.QueryContext
 import com.normation.zio.*
 import zio.*
 import zio.syntax.ToZio
 
 trait ScoreService {
-  def getAll(): IOResult[Map[NodeId, GlobalScore]]
-  def getGlobalScore(nodeId:  NodeId): IOResult[GlobalScore]
-  def getScoreDetails(nodeId: NodeId): IOResult[List[Score]]
+  def getAll()(implicit qc:   QueryContext): IOResult[Map[NodeId, GlobalScore]]
+  def getGlobalScore(nodeId:  NodeId)(implicit qc: QueryContext): IOResult[GlobalScore]
+  def getScoreDetails(nodeId: NodeId)(implicit qc: QueryContext): IOResult[List[Score]]
   def clean(): IOResult[Unit]
-  def cleanScore(name:          String): IOResult[Unit]
-  def update(newScores:         Map[NodeId, List[Score]]): IOResult[Unit]
-  def registerScore(newScoreId: String, displayName: String): IOResult[Unit]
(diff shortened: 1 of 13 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -48,32 +49,33 @@ trait ScoreService {
...
   private val availableScore: Ref[List[(String, String)]] =
     Ref.make((ComplianceScore.scoreId, "Compliance") :: (SystemUpdateScore.scoreId, "System updates") :: Nil).runNow
 
-  def init(): IOResult[Unit] = {
+  def init()(implicit qc : QueryContext): IOResult[Unit] = {
     for {
+      nodeIds <- nodeFactRepo.getAll().map(_.keySet)
       globalScores <- globalScoreRepository.getAll()
-      _            <- cache.set(globalScores)
+      _            <- cache.set(globalScores.filter(n => nodeIds.contains(n._1)))
```
fanf, line 72: here, `nodeIds` only contains node as seen by the person issuing the query context. I think we want all (active) nodes at that point, so directly using the system query context here, and not as a parameter (even if it's called with that, I don't think it should be a parameter at all).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -136,19 +138,21 @@ class ScoreServiceImpl(globalScoreRepository: GlobalScoreRepository, scoreReposi
...
+  def cleanScore(name: String)(implicit qc: QueryContext): IOResult[Unit] = {
     for {
       _         <- cache.update(_.map { case (id, gscore) => (id, gscore.copy(details = gscore.details.filterNot(_.scoreId == name))) })
       newScores <- scoreCache.updateAndGet(_.map { case (id, scores) => (id, scores.filterNot(_.scoreId == name)) })
       _         <- update(newScores)
     } yield {}
   }
 
-  def update(newScores: Map[NodeId, List[Score]]): IOResult[Unit] = {
+  def update(newScores: Map[NodeId, List[Score]])(implicit qc : QueryContext): IOResult[Unit] = {
```
fanf, line 149: same for the queryContext, I don't think we want that to dependant of a tenant view, else the cache will become inconsistant

fanf:
Nice catch. Can you add a unit test to capture the two invariant ? (after init, repo only contains existing nodes and service only save existing nodes)?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
VinceMacBuche: I don't know it gives access to score of all Nodes even those you cannot, like a leak of data to another tenant

VinceMacBuche: Sorry i haven't seen it was in init, it should be systemQC always (and it is always used with it)

VinceMacBuche:
Still needs tests

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):
:rocket: seems to work as expected :+1:

VinceMacBuche:
OK, merging this PR

PR #6045, opened 2024-11-26
Fixes #25892: Error when accepting 2 or more nodes

VinceMacBuche:
https://issues.rudder.io/issues/25892

PR #6047, opened 2024-11-27
Fixes #25892: Error when accepting 2 or more nodes

fanf:
https://issues.rudder.io/issues/25892

This PR replace #6045. It adds some more change in the LDAP pool ergonomics: 
- the main change is still `LDAP_CREATE_IF_NECESSARY` set to `false` which means that the max pool size is really the max pool size, and that yes we want a pool to reuse existing connections, seriously
- the minimum for pool size can be lower than 10. It could likely be even lower than 7, but it seems to work well, even in small servers, and up to big one (10k nodes). 
- add `LDAP_MINIMUM_AVAILABLE_CONNECTION_GOAL` for the number of min alive connexion. Set to 2, which should be OK everywhere. 
- `LDAP_MAX_WAIT_TIME` is the time code will wait for a connection from the pool. It was to 0, which is strange for a pool. Use `30 seconds` like in most pool. 
- deconnect and connect again connection after `LDAP_MAX_CONNECTION_AGE` = 30 minutes. I'm not sure it's necessary, but it's often seen as a good practice to refresh connections regularly. 
- in consequence, also set `LDAP_MIN_DISCONNECT_INTERVAL` which ensure that connection are not all refreshed at the same time, with a default interval of 5s.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/RudderDatasourceProvider.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/RudderDatasourceProvider.scala
@@ -42,6 +42,8 @@ import com.zaxxer.hikari.HikariDataSource
 import java.io.Closeable
 import javax.sql.DataSource
 import net.liftweb.common.Loggable
+import scala.concurrent.duration.Duration
+import scala.concurrent.duration.DurationInt
 
 /**
  * A wrapper around defaut data source provider to allow for several
@@ -51,11 +53,12 @@ import net.liftweb.common.Loggable
  * http://blog.trustiv.co.uk/2014/06/battle-connection-pools
  */
 class RudderDatasourceProvider(
-    driver:      String,
-    url:         String,
-    username:    String,
-    password:    String,
-    maxPoolSize: Int
+    driver:               String,
+    url:                  String,
+    username:             String,
+    password:             String,
+    maxPoolSize:          Int,
+    getConnectionTimeout: Duration = 250.millis
 ) extends Loggable {
   Class.forName(driver)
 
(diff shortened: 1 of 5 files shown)


webapp/sources/scala-ldap/src/main/scala/com/normation/ldap/sdk/LDAPConnectionProvider.scala
```diff
@@ -299,11 +309,16 @@ class RWAnonymousConnectionProvider(
...
-    override val useSchemaInfos: Boolean = false,
-    override val poolSize:       Int = 2
+    override val host:                           String = "localhost",
+    override val port:                           Int = 389,
+    override val ldifFileLogger:                 LDIFFileLogger = new DefaultLDIFFileLogger(),
+    override val useSchemaInfos:                 Boolean = false,
+    override val poolSize:                       Int = 2,
+    override val createIfNecessary:              Boolean = true,
+    override val minimumAvailableConnectionGoal: Int = 2,
+    override val maxWaitTime:                    Duration = 30.seconds,
```
fanf, line 319: the default values are needed since we already have parameters with default values before them.

fanf:
In loadtesting, we now have sometime: 
```2024-11-27 14:40:27+0000 ERROR nodes - Error when executing callback 'node-fact-cec: register even log' on node change event [updatedAccepted] node 'RUDDER-TEST-4671' (23e616de-ec7d-401c-a392-d9a2bffa36f1)': SystemError: Error when persisting event log NodeModified; cause was: java.sql.SQLTransientConnectionException: HikariPool-1 - Connection is not available, request timed out after 250ms (total=25, active=25, idle=0, waiting=2)```

It is likely a non existing problem in real life, but we will make the timeout configurable in unforseen cases.

**Comment by fanf** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/RudderDatasourceProvider.scala [hunk 2]
```diff
@@ -51,11 +53,12 @@ import net.liftweb.common.Loggable
...
-    url:         String,
-    username:    String,
-    password:    String,
-    maxPoolSize: Int
+    driver:               String,
+    url:                  String,
+    username:             String,
+    password:             String,
+    maxPoolSize:          Int,
+    getConnectionTimeout: Duration = 250.millis
```
fanf, line 61: using a default value to not break compilation for plugins

fanf:
Merging after validation on load test machine + it's the same base than overridden PR.

fanf:
OK, squash merging this PR

PR #6050, opened 2024-11-28
Fixes #25669: 

clarktsiory:
https://issues.rudder.io/issues/25669

Adding a `rudder.auth.{provider}.disableRest` boolean option for each known provider passed to the `rudder.auth.provider` list : just read the configuration for the provider into our case class for provider properties, and keep compatibility with previous behavior by using the `false` value (meaning : DO NOT disable REST authentication for this provider)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
+++ b/webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
@@ -642,6 +642,20 @@ rudder.users.cleanup.sessions.purgeAfter=30 days
 rudder.auth.provider=file
 
 #
+# To disable access to the REST API using user-generated API token
+# via the api-authorizations plugin, you can use change this option to "disabled"
+#
+rudder.auth.userRestToken=enabled
+
+#
+# To disable access to the REST API using user-generated API token
+# via the api-authorizations plugin for users from a specific provider (declared in rudder.auth.provider),
+# you can use uncomment this property. You can also set it to "enabled"
+# (the default value is the one for rudder.auth.userRestToken).
+#
+#rudder.auth.oidc.userRestToken=disabled
+
+#
 # Inactivity timeout for user sessions
 #
 # The default value is 30 minutes.
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -40,6 +40,7 @@ package bootstrap.liftweb
 import com.normation.errors.*
 import com.normation.rudder.Role
 import com.normation.rudder.api.*
+import com.normation.rudder.domain.appconfig.FeatureSwitch
 import com.normation.rudder.domain.logger.ApplicationLogger
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.rudder.facts.nodes.NodeSecurityContext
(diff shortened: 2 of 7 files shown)


clarktsiory:
PR updated with a new commit

fanf:
I think we need at least one global switch to totally disable the rest user token feature, something like: `rudder.auth.rest.userToken=enabled/disabled`

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -641,6 +641,19 @@ rudder.users.cleanup.sessions.purgeAfter=30 days
...
+# via the api-authorizations plugin, you can use change this option to "disabled"
+#
+rudder.auth.userRestToken=enabled
+
+#
+# To disable access to the REST API using user-generated API token
+# via the api-authorizations plugin for users from a specific provider (declared in rudder.auth.provider),
+# you can use set this option to "disabled"
+#
+rudder.auth.file.userRestToken=enabled
```
fanf, line 655: I think the example should be "oidc" because it's the most common case for using it. 
Plus, I think we should comment out these two properties so that an user without the plugin is not let thinking he can enable these options.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -158,10 +166,20 @@ class AppConfigAuth extends ApplicationContextAware {
...
       }
-      val properties: Option[(String, AuthBackendProviderProperties)] = x.name match {
+      val restTokenFeatureSwitch: FeatureSwitch                                   = {
+        // special feature switch depends on global one.
+        restTokenGlobalFeatureSwitch match {
+          case FeatureSwitch.Disabled => FeatureSwitch.Disabled
+          case FeatureSwitch.Enabled  =>
+            try {
+              FeatureSwitch.parse(config.getString(s"rudder.auth.${x.name}.userRestToken")).getOrElse(defaultEnableRestToken)
+            } catch { case ex: ConfigException.Missing => defaultEnableRestToken }
```
fanf, line 176: same here: info application log in the case where the value is different from the global one, debug level when it's the same

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -633,7 +659,12 @@ class AuthBackendProvidersManager() extends DynamicRudderProviderManager {
   // add default providers
   this.addProvider(defaultAuthBackendsProvider)
   this.addProviderProperties(
-    Map(DefaultAuthBackendProvider.FILE -> AuthBackendProviderProperties(ProviderRoleExtension.WithOverride))
+    Map(
+      DefaultAuthBackendProvider.FILE -> AuthBackendProviderProperties(
+        ProviderRoleExtension.WithOverride,
+        restTokenFeatureSwitch = FeatureSwitch.Enabled
```
fanf, line 665: I don't get why we have an `Enabled` here?

fanf: All providers must behave the same. 
The only one that must be different is the "super user" provider (the one account that can be configured in properties with `rudder.auth.admin.login`/`rudder.auth.admin.password` : this one must *ALWAYS* be disabled (ie, it can't use API at all).

fanf:
When the feature is disabled for a user, he must not see the menu for personnal token, but a message saying that the personal token feature is disabled. 

image

webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
clarktsiory: yes indeed

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -142,6 +143,13 @@ class AppConfigAuth extends ApplicationContextAware {
     // the list of authentication backend configured by the user
     val configuredAuthProviders = AuthenticationMethods.getForConfig(config)
 
+    // by default enable user REST token authentication for compatibility
+    val defaultEnableRestToken = FeatureSwitch.Enabled
+    val restTokenGlobalFeatureSwitch: FeatureSwitch = {
+      try { FeatureSwitch.parse(config.getString(s"rudder.auth.userRestToken")).getOrElse(defaultEnableRestToken) }
+      catch { case ex: ConfigException.Missing => defaultEnableRestToken }
+    }
```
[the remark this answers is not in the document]
clarktsiory: yes and a warning when the value exists but has a typo, and cannot be parsed into a FeatureSwitch

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
clarktsiory: this one will be overridden (the config by map by provider and we always traverse values from `rudder.user.provider`), I think the operation is irrelevant and can be removed here, but I will test that

clarktsiory:
Yes the display according to the config value is handled in the api-authorizations plugin, I will make a PR for that

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -641,6 +641,20 @@ rudder.users.cleanup.sessions.purgeAfter=30 days
...
+#
+# To disable access to the REST API using user-generated API token
+# via the api-authorizations plugin, you can use change this option to "disabled"
+#
+rudder.auth.userRestToken=enabled
+
+#
+# To disable access to the REST API using user-generated API token
+# via the api-authorizations plugin for users from a specific provider (declared in rudder.auth.provider),
+# you can use uncomment this property. You can also set it to "enabled"
```
fanf, line 653: s/use//

clarktsiory:
PR updated with a new commit

fanf (approved):
:rocket:

clarktsiory:
OK, squash merging this PR

PR #6051, opened 2024-11-28
Fixes #25973: Acceptation error with new node on hostname duplication is not hightlitghed

fanf:
https://issues.rudder.io/issues/25973

Highlight with a danger callout, put text in strong to make it pop more (only the callout is too light)

image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
@@ -138,9 +138,10 @@ class AcceptNode extends Loggable {
       accept match {
         case f:     Failure  =>
           S.error(
-            <span>
-            {f.messageChain}
-          </span>
+            <div class="alert alert-danger">
+              <i class ="fa fa-warning"></i>
+              <strong>{f.messageChain}</strong>
+            </div>
           )
         case empty: EmptyBox =>
           val errorMsg = s"Add new node '$id.value' lead to Failure."


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
```diff
@@ -138,9 +138,9 @@ class AcceptNode extends Loggable {
       accept match {
         case f:     Failure  =>
           S.error(
-            <span>
-            {f.messageChain}
-          </span>
+            <div class="callout-fade  callout-danger">
+              <strong>{f.messageChain}</strong>
+            </div>
```
RaphaelGauthier, line 143: I'd prefer to use the .alert class rather than .callout :
```
<div class="alert alert-danger">
  <i class ="fa fa-warning"></i>
  <strong>{f.messageChain}</strong>
</div>
```

 It's not very clear which class to use in this kind of situation (we sometimes use .callout, sometimes .alert), it lacks consistency. There's no good reason for this, we should make a final decision on this.

Until then, the .alert style would allow this message to be even more noticeable than the information messages that precede it, and we tend to use this style of message in our latest interfaces
alert-danger-preview

RaphaelGauthier (changes requested):

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/node/AcceptNode.scala
```diff
@@ -138,9 +138,9 @@ class AcceptNode extends Loggable {
       accept match {
         case f:     Failure  =>
           S.error(
-            <span>
-            {f.messageChain}
-          </span>
+            <div class="alert alert-danger">
```
RaphaelGauthier, line 141: The only thing missing is a warning icon : `<i class ="fa fa-warning"></i>`, and that's it for me :)

RaphaelGauthier (approved):

fanf:
Merging by hand, the CI error is unrelated

PR #6055, opened 2024-11-28
Fixes #25705: search on event log must also search on event type

clarktsiory:
https://issues.rudder.io/issues/25705

This part of query is within a larger set of condition, it needs parenthesis, and it knows that the `eventtype` column is there. Also we don't need to consider the exact case even for the user query (EventType is stored as PascalCase). 

The drawback of this is that the user should know the EventType. And we have no way of statically checking for known event types before applying the filter.


clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM

clarktsiory:
OK, squash merging this PR

PR #6059, opened 2024-11-28
Fixes #25978: CustomProperties (inventory properties) are not always correctly retrieved from inventory

fanf:
https://issues.rudder.io/issues/25978

We were not correctly mapping back attributes from NodeInventory to Node in the case where we were retrieving both node and inventory info from storage. 

Rebased on top of #6063 which is around the same problems

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
@@ -59,6 +59,7 @@ import com.normation.rudder.domain.logger.NodeLogger
 import com.normation.rudder.domain.logger.NodeLoggerPure
 import com.normation.rudder.domain.nodes.MachineInfo
 import com.normation.rudder.domain.nodes.NodeInfo
+import com.normation.rudder.domain.properties.NodeProperty
 import com.normation.rudder.facts.nodes.LdapNodeFactStorage.needsSoftware
 import com.normation.rudder.git.GitItemRepository
 import com.normation.rudder.git.GitRepositoryProvider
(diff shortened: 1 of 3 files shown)


fanf:
Hum, no, there is something strange, the init isn't done with the full inventory data, we should not have custom properties with my understanding

fanf:
OK, it's just that we were not retrieving custom properties in a case

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
```diff
@@ -821,8 +822,8 @@ class LdapNodeFactStorage(
       } yield {
         optInvS.map {
           case (inv, _) =>
-            val info = NodeInfo(
-              node,
+            val info       = NodeInfo(
+              node.modify(_.properties).using(p => inv.node.customProperties.map(NodeProperty.fromInventory(_)) ::: p),
```
fanf, line 826: this is the actual correction

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
```diff
@@ -836,7 +837,11 @@ class LdapNodeFactStorage(
               inv.node.ram,
               inv.node.timezone
             )
-            NodeFact.fromCompat(info, Right(inv), softs, None)
+            val softUpdate = inv.node.softwareUpdates match {
+              case Nil  => None
+              case list => Some(Chunk.fromIterable(list))
+            }
+            NodeFact.fromCompat(info, Right(inv), softs, softUpdate)
```
fanf, line 844: we were missing also software update

clarktsiory (approved):
LGTM !

PR #6060, opened 2024-11-29
Fixes #25979: Migrate custom properties along with node properties for consistency

fanf:
https://issues.rudder.io/issues/25979

Ported from first correction of #6059 

So, we have two kind of properties for nodes: 
- `customProperty` comes from inventory and are stored in `ou=[Pending, Accepted] Inventories` branch, 
- `serializedNodeProperty` comes from plugin, users, etc and are stored in `ou=Nodes` branch. 

Before 8.0, the separation was mandatory because node entry under `ou=Nodes` was only created at acceptation time. This limitation was removed in Rudder 8.0. 

It makes not architectural sense to have the split persists. We can merge them together in `ou=Nodes` entries.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckMigrateCustomProperties.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckMigrateCustomProperties.scala
@@ -1,0 +1,123 @@
+/*
+ *************************************************************************************
+ * Copyright 2024 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package bootstrap.liftweb.checks.migration
+
+import bootstrap.liftweb.BootstrapChecks
+import com.normation.errors.*
+import com.normation.inventory.domain.NodeId
+import com.normation.inventory.ldap.core.InventoryDit
+import com.normation.inventory.ldap.core.LDAPConstants.*
+import com.normation.ldap.sdk.BuildFilter
+import com.normation.ldap.sdk.LDAPConnectionProvider
+import com.normation.ldap.sdk.RwLDAPConnection
+import com.normation.rudder.domain.logger.MigrationLoggerPure
+import com.normation.rudder.facts.nodes.ChangeContext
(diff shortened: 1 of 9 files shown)


webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestMigrateCustomProperties.scala
```diff
@@ -0,0 +1,114 @@
...
+    /*
+     * Node1 contains:
+     * - custom properties 'datacenter' and 'from_inv'
+     * - node property 'foo'
+     */
+
+    "find some properties to migrate" in {
+      (migration.findNodeNeedingMigration().runNow must containTheSameElementsAs(List(NodeId("node1")))) and
+      (getNode1PropNamesAsSeenByLdap() must containTheSameElementsAs(List("foo", "datacenter", "from_inv"))) and
+      (getNode1PropNamesAsSeenByRepo() must containTheSameElementsAs(List("foo", "datacenter", "from_inv")))
```
clarktsiory, line 87: I forgot but is this really guaranteed to be true i.e. does the other test that adds `datacenter`, `from_inv` always execute before this one ?

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestMigrateCustomProperties.scala
```diff
@@ -0,0 +1,114 @@
...
+    "not find any nodes to migrate after migration" in {
+      (migration.findNodeNeedingMigration().runNow must beEmpty) and
+      (getNode1PropNamesAsSeenByLdap() must containTheSameElementsAs(List("foo", "datacenter", "from_inv"))) and
+      (getNode1PropNamesAsSeenByRepo() must containTheSameElementsAs(List("foo", "datacenter", "from_inv")))
+    }
+  }
+
+  def getNode1PropNamesAsSeenByLdap(): Seq[String] = {
+    (for {
+      _ <- errors.effectUioUnit(println(s"****** ldap ********"))
```
clarktsiory, line 103: I don't see the meaning of the println, for debugging purposes even is it really providing information because I don't see nothing else than `**** ldap ****` and `**** repos ****` printed out ?

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckMigrateCustomProperties.scala
```diff
@@ -0,0 +1,123 @@
...
+import com.normation.rudder.facts.nodes.NodeFactRepository
+import com.normation.zio.*
+import zio.*
+
+/*
+ * This migration check looks if there is still nodes with custom properties
+ * in the ou=Accepted Inventories branch, and if so, it save back them so that
+ * they are moved along with other properties.
+ * Added in rudder 8.1.8 (https://issues.rudder.io/issues/25978)
+ * Can be removed in Rudder 8.3
```
clarktsiory, line 59: This PR is in 8.3, the comment could be misleading, why not removing this file then ?

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestMigrateCustomProperties.scala
fanf: it's in the source data

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/migration/TestMigrateCustomProperties.scala
fanf: debugging remaining :)

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/migration/CheckMigrateCustomProperties.scala
fanf: yes, I was not sure if we should put it in 8.1 or not. But there is 0 reasons to do so: 
- it's an architectural change, 
- that doesn't correct any direct bugs

That goes to 8.3. I updated the text accordingly.

fanf:
Given https://github.com/Normation/rudder/pull/6063, we will need to think back how to do the migration: 
- now, `customProperties` are not saved (deleted) when saving the core node fact, since it uses the new inventory method `saveUserSetAttributes` which does not delete missing attributes. 
- so we could add `customProperties` to the mapped attributes. It won't do anything appart for the migration
- or we can handle the migration in an other way. 

Perhas we should add `customProperties`, and also migrate `keyStatus` in node part, since it's a user-settable think, in one go.

fanf:
OK, so after more thought, I believe we just don't want to do that anymore. Plus, it would be merging `fact`, ie something that is imposed by the context (for inventory properties), and `configuration`, ie something that is chosen by the user, and there's no reason to force that now.

PR #6061, opened 2024-11-29
Fixes #25982: Disable metrics pop-up and form

fanf:
https://issues.rudder.io/issues/25982

Only hide the menu part, plus ensure that the system variable always get "no" - but keep the config service, API, etc logic so that it's easy to reinstanciate in 8.3. 

It looks like the pop-up logic was removed and not replaced in bb01e23b8eb48962e37bef93894fc2517b08dcd4 
The pop-up code still exists, it can be clean-up in Rudder 8.3 along with the script etc.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/secure/administration/policyServerManagement.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/secure/administration/policyServerManagement.html
@@ -630,8 +630,8 @@
               </div>
             </div>
 
-            <div class="inner-portlet">
-              <h3 class="page-title">Usage survey</h3>
+            <div class="inner-portlet d-none">
+              <h3>Usage survey</h3>
               <div class="portlet-content">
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-web/src/main/webapp/secure/administration/policyServerManagement.html
```diff
@@ -630,8 +630,8 @@ <h3 class="page-title">Debug information</h3>
               </div>
             </div>
 
-            <div class="inner-portlet">
-              <h3 class="page-title">Usage survey</h3>
+            <div class="inner-portlet d-none">
```
VinceMacBuche, line 633: why not removing it completely ?

VinceMacBuche: ok it will be done in 8.3

fanf:
Merging by hand since the CI error is not related.

fanf:
OK, merging this PR

PR #6066, opened 2024-12-03
Fixes #25980: Migrate Group category API endpoints to zio-json

clarktsiory:
https://issues.rudder.io/issues/25980

Migrating remaining lift-json endpoints in the GroupsApi file : endpoints for rule categories.
It ends up being very similar to the already migrated Rules categories API.
I added some tests for validation error (on the `name`) along the way, I removed what could be removed from the evil  `RestExtractorService`...

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
@@ -67,6 +67,7 @@ import com.normation.rudder.domain.properties.Visibility.Hidden
 import com.normation.rudder.domain.queries.Query
 import com.normation.rudder.domain.queries.QueryReturnType
 import com.normation.rudder.domain.queries.QueryReturnType.NodeReturnType
+import com.normation.rudder.repository.FullNodeGroupCategory
 import com.normation.rudder.rule.category.RuleCategory
 import com.normation.rudder.rule.category.RuleCategoryId
 import com.normation.rudder.services.policies.PropertyParser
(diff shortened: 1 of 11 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
```diff
@@ -359,6 +360,57 @@ object JsonQueryObjects {
       _isEnabled:  Option[Boolean]
   )
 
+  final case class JQGroupCategory private[apidata] (
+      id:          Option[NodeGroupCategoryId] = None,
+      name:        Option[String] = None,
+      description: Option[String] = None,
+      parent:      Option[NodeGroupCategoryId] = None
+  ) {
```
clarktsiory, line 368: simply moved from the old `RestGroupCategory` data type, the handling logic is the same

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonResponseObjects.scala
```diff
@@ -1914,7 +1914,111 @@ object JsonResponseObjects {
...
+      parent:                                 NodeGroupCategoryId,
+      @jsonField("categories") subCategories: List[JRFullGroupCategory],
+      groups:                                 List[JRGroup],
+      @jsonField("targets") targetInfos:      List[JRRuleTargetInfo]
+  )
+
+  object JRFullGroupCategory {
+    /*
+     * Somehow the chimney transformer derivation does not provide a way to pass the parent due to recursive type,
+     * so we have to make an inline transformer.
```
clarktsiory, line 1942: I could not figure a solution to make a `implicit lazy val transformer: Transformer[JQGroupCategory, FullNodeGroupCategory]` declaration otherwise than inline (as in Rules), the culprit is having the `parent` being passed down so the parent value is relative but it is not an attribute of the `JQGroupCategory`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -271,25 +272,25 @@ object GroupApi       extends Enum[GroupApi] with ApiModuleProvider[GroupApi] {
     val z: Int = implicitly[Line].value
     val description    = "Get information about the given group category"
     val (action, path) = GET / "groups" / "categories" / "{id}"
-    override def dataContainer: Some[String] = Some("groupCategories")
+    override def dataContainer: None.type = None
```
fanf, line 275: isn't that changing the resulting JSON ? I think it will lead to have one less `{ ..., "groupCategories": [ ... content ...] }`

If so, you can't do that. We don't have the right to change external APIs without a fallback path for their consumer. Typically, if a script consume it for extracting info about the category, it will break, and it can't be repaired by using previous API version in that case. If it changes nothing in the output, you can ignore that message :)

fanf:
Appart for the endpoint structural change, LGTM - and such a good change :)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
clarktsiory: That is a caveat of the migration : the `dataContainer` is used in the plumbings of the API only to return an ARRAY within the data : `{...,data: { groupCategories : [] }}`, and using `toLiftResponseOne` will make an array with a single element...
However here there is an existing test which shows that we have `{...,data: { groupCategories : {} }}` with an object instead, so we have to suppress the `dataContainer` and wrap the Json within an object under a `groupCategories` field (a wrapper `case class Wrapper(groupCategories: GroupCategory)`) :
image

clarktsiory: (the case you addressed was covered by this test, which has not changed)

fanf (approved):

PR #6069, opened 2024-12-05
Fixes #25824: Properties tab : Non-rearrangement of the pop-ups

RaphaelGauthier:
https://issues.rudder.io/issues/25824

Several changes have been made:

- First, this bug has been fixed: tooltips are now reinitialized after the table has been sorted.

- Property tooltips can only be opened by clicking on their label (the mouse cursor is different from other labels when hovering these labels). This means we can scroll freely through their contents.
To close the tooltip, simply click outside (a js event has been added to the <body> for this, as Bootstrap doesn't offer this option by default).

- Tooltips width now adapts to content, and their maximum width has been increased by 100px to make large content easier to read

- A line break is forced when the width of a < pre > in a tooltip exceeds its container

- I've added some missing css classes that were breaking the display of some tooltips (like the ‘overriden’ ones)

Here is a preview :

tooltip-click


RaphaelGauthier:
Commit modified

VinceMacBuche (approved):

RaphaelGauthier:
OK, merging this PR

PR #6071, opened 2024-12-05
Fixes #26022: Status should not be on error when there is technique compilator error on disable techniques

clarktsiory:
https://issues.rudder.io/issues/26022

When the actor receives the message, we want to ignore disabled techniques, and only process the enabled one for display in the Lift frontend actor.
This adds the `status` field to the data in the cache, so that the frontend actor can ignore disabled one. This also means we have to "sync" the `status`, especially upon changing the status in the UI. The status is still looked up every time we "sync all techniques", so the API is also changed a bit to add the new `TechniqueActiveStatus` data type.


clarktsiory:
PR rebased

fanf:
Merging by hand after rebase

fanf:
OK, merging this PR

PR #6073, opened 2024-12-06
Fixes #26017: Migrate old internal EventLog APIs to zio-json

clarktsiory:
https://issues.rudder.io/issues/26017

This is the migration of an old internal API : the one for event logs with 3 endpoints.
API YAML tests were added for all endpoints, and the only complexity apart from adding tests is to ensure that the json model are translated correctly in existing doobie queries.

The API endpoints schema were migrated as is, even though they may be changed (`GET /eventlog/id/details/rollback?action=before` is a really weird one, hopefully it's only for internal use in the JS ajax client)


clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

fanf (approved):
Rebased after group PR migration. 

LGTM :) ApiAccount looks next :grin:

PR #6079, opened 2024-12-12
Fixes #26056: Non admin users in file get disabled even if last login parameter does not exceed the scheduled date

clarktsiory:
https://issues.rudder.io/issues/26056

The bug is because of an `OR` instead of `AND` in a `SELECT` query...

The change only impacts the disabling of users, this is the only place where the `select()` is used with both filters, so changing the semantics of the method is safe (and also more consistent).

Tests are added to the Jdbc repository, to validate the filtering behaves as we want

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -673,18 +673,10 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
       authenticatorName: String,
       date:              DateTime
   ): IOResult[Unit] = {
-    val session   = {
-      sql"""insert into usersessions (sessionid, userid, creationdate, authmethod, permissions, authz, tenants)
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -876,12 +888,7 @@ class JdbcUserRepository(doobie: Doobie) extends UserRepository {
       excludeFromOrigin: List[String],
       andSelectFragment: Option[Fragment] // refined selection
   ): ConnectionIO[List[(String, List[StatusHistory])]] = {
-    val selectUsers = (predicateUser(userIds), predicateLogDate(notLoggedSince)) match {
-      case (Some(a), Some(b)) => Some(Fragments.or(a, b))
-      case (Some(a), None)    => Some(a)
-      case (None, Some(b))    => Some(b)
-      case (None, None)       => None
-    }
+    val selectUsers = Fragments.andOpt(predicateUser(userIds), predicateLogDate(notLoggedSince))
```
clarktsiory, line 891: The main change : it should not be an `OR` but it should be `AND` instead

fanf:
lgtm

fanf (approved):

PR #6082, opened 2024-12-12
Fixes #26002: The directive page is much slower with Rudder 8.1 than in 7.3

fanf:
https://issues.rudder.io/issues/26002

So, we were calling an ajax method, in the return there was a `window.location.hash = xxxx`, but in https://github.com/Normation/rudder/pull/5760 we added an event listener on window location hash which triggers an ajax call for the same update. 
I'm not sure why it stops after two round, it could have looped for ever.


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM, it works.
The other event triggers the display when the hash changes only, that may be the reason why it does not loop infinitely :innocent:

fanf:
OK, squash merging this PR

PR #6083, opened 2024-12-12
Fixes #26064: Group query select attribute has different name in create and update payload

clarktsiory:
https://issues.rudder.io/issues/26064

This is breaking but the previous case was actually not what was in [the documentation](https://docs.rudder.io/api/v/20/#tag/Groups/operation/createGroup) : the previous `returnType` needed to be the default zio-json sealed trait payload with the type discriminator as key : `"returnType":{"NodeAndRootServerReturnType":""}`.


VinceMacBuche (approved):

clarktsiory:
OK, merging this PR

PR #6085, opened 2024-12-15
Fixes #26070: Add tests and clean-up JsDataLine structures

fanf:
https://issues.rudder.io/issues/26070

We want to get ride of `JsDataLine` and `JsDataTable` structure. 

This PR: 
- add an util class, `SimpleStatus` (`Enabled`, `Disabled`) that could (should) be used each time we have a `isEnabled: Boolean` parameter. 
- correct all `overriden` into `overridden` (that adds quite a lot of noise, it should have been done in an other PR)
- remove dead code around `ComplianceData` and other structures using `JsDataLine` - there was a lot
- in `JsTableData`, transform the `def json: JObject` into a `def json(freshName: () => String): JObject` so that we can define the "next fresh name" in tests,
- refactor some code to make clearer that we just want to test the serialization of `XXXLine`, not all the surrounding logic. I didn't do it for `ComplianceData`, it would have been too much. 
- change the compliance mock serivice to use a technique with a block, because that wasn't tested (and change the billions API tests impacted)

Generaly, there is a big mismatch in these objects between business logic and presentation / serialisation. 
There is *a ton* of business code that should be tested here, and don't reuse existing services. This is a first step in a more disciplined direction.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
@@ -39,7 +39,6 @@ package com.normation.rudder.web.components
 
 import bootstrap.liftweb.RudderConfig
 import com.normation.box.*
-import com.normation.cfclerk.domain.Technique
 import com.normation.eventlog.ModificationId
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.eventlog.RudderEventActor
@@ -56,6 +55,7 @@ import com.normation.rudder.web.services.ComputePolicyMode
 import com.normation.rudder.web.services.JsTableData
 import com.normation.rudder.web.services.JsTableLine
 import com.normation.utils.Control.traverse
+import com.normation.utils.SimpleStatus
 import com.normation.zio.*
 import net.liftweb.common.*
 import net.liftweb.http.*
@@ -89,7 +89,6 @@ object DisplayColumn {
 
 class RuleGrid(
     htmlId_rulesGridZone: String, // JS callback to call when clicking on a line
-
     detailsCallbackLink:  Option[(Rule, String) => JsCmd],
     showCheckboxColumn:   Boolean,
     directiveApplication: Option[DirectiveApplicationManagement],
@@ -97,20 +96,7 @@ class RuleGrid(
     graphRecentChanges:   DisplayColumn
 ) extends DispatchSnippet with Loggable {
 
-  sealed private trait Line { val rule: Rule }
-
-  private case class OKLine(
-      rule:              Rule,
-      applicationStatus: ApplicationStatus,
-      trackerVariables:  Seq[(Directive, ActiveTechnique, Technique)],
-      targets:           Set[RuleTargetInfo]
-  ) extends Line
-
-  private case class ErrorLine(
-      rule:             Rule,
-      trackerVariables: Box[Seq[(Directive, ActiveTechnique, Technique)]],
-      targets:          Box[Set[RuleTargetInfo]]
-  ) extends Line
+  import RuleGrid.*
 
   private val getFullNodeGroupLib      = RudderConfig.roNodeGroupRepository.getFullGroupLibrary _
   private val getFullDirectiveLib      = RudderConfig.roDirectiveRepository.getFullDirectiveLibrary _
@@ -233,7 +219,7 @@ class RuleGrid(
               ruleCompliances = {};
               recentChanges = {};
               recentGraphs = {};
-              refreshTable("${htmlId_rulesGridId}", ${newData.json.toJsCmd});
+              refreshTable("${htmlId_rulesGridId}", ${newData.toJson.toJsCmd});
               ${futureCompliance.toJsCmd}
               ${futureChanges.toJsCmd}
           """) // JsRaw ok, escaped
@@ -290,7 +276,7 @@ class RuleGrid(
         val onLoad              = {
           s"""createRuleTable (
                   "${htmlId_rulesGridId}"
-                , ${tableData.json.toJsCmd}
+                , ${tableData.toJson.toJsCmd}
                 , ${showCheckboxColumn}
                 , ${showActionsColumn}
                 , ${showComplianceAndChangesColumn}
@@ -421,6 +407,25 @@ class RuleGrid(
 
   }
 
+  private def getRulePolicyMode(
+      rule:         Rule,
+      nodes:        Set[NodeId],
+      nodeFacts:    MapView[NodeId, CoreNodeFact],
+      directiveLib: FullActiveTechniqueCategory,
+      globalMode:   GlobalPolicyMode
+  ) = {
+    val nodeModes = nodes.flatMap(id => nodeFacts.get(id).map(_.rudderSettings.policyMode))
+    // when building policy mode explanation we look into every directive applied by the rule.
+    // But some directive may be missing so we do 'get', we them skip the missing directive and only use existing one ( thanks to flatmap)
+    // Rule will be disabled somewhere else, stating that some objects are missing and you should enable it again to fix it
+    ComputePolicyMode.ruleMode(
+      globalMode,
+      rule.directiveIds.map(directiveLib.allDirectives.get(_)).flatMap(_.map(_._2)),
+      nodeModes
+    )
+
+  }
+
   /*
    * Get Data to use in the datatable for all Rules
    * First: transform all Rules to Lines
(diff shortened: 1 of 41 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -233,7 +219,7 @@ class RuleGrid(
               ruleCompliances = {};
               recentChanges = {};
               recentGraphs = {};
-              refreshTable("${htmlId_rulesGridId}", ${newData.json.toJsCmd});
+              refreshTable("${htmlId_rulesGridId}", ${newData});
```
fanf, line 222: for now, `newData` returns a string but we will rewrite it with `toJson` from zio

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -45,84 +45,19 @@ import com.normation.rudder.domain.policies.*
...
- *   Javascript object containing all data to create a line about a change in the DataTable
- *   { "nodeName" : Name of the node [String]
- *   , "message" : Messages linked to that change [String]
- *   , "directiveName" : Name of the directive [String]
- *   , "component" : Component name [String]
- *   , "value": Value of the change [String]
- *   , "executionDate" : date the report was run on the Node [String]
- *   }
- */
-final case class ChangeLine(
```
fanf, line 76: not used anywhere

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -144,16 +79,16 @@ final case class RuleComplianceLine(
     modeExplanation: String,
     tags:            JValue
 ) extends JsTableLine {
-  val json: js.JsObj = {
+  def json(freshName: () => String): js.JsObj = {
```
fanf, line 82: we need to be able to predict test value

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -207,39 +142,6 @@ final case class DirectiveComplianceLine(
...
- *   Javascript object containing all data to create a line in the DataTable
- *   { "node" : Node name [String]
- *   , "id" : Node id [String]
- *   , "compliance" : array of number of reports by compliance status [Array[Float]]
- *   , "compliancePercent" : Compliance percentage [Float]
- *   , "details" : Details of Directive applied by the Node [Array of Directive values ]
- *   , "jsid"    : unique identifier for the line [String]
- *   }
- */
-final case class NodeComplianceLine(
```
fanf, line 220: not used anywhere

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -329,79 +230,33 @@ final case class ComponentValueComplianceLine(
...
 
 }
 
 object ComplianceData extends Loggable {
 
-  /*
-   * For a given rule, display compliance by nodes.
-   * For each node, elements displayed are restraint
-   */
-  def getRuleByNodeComplianceDetails(
```
fanf, line 344: not used anywhere

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ComplianceData.scala
```diff
@@ -473,52 +327,26 @@ object ComplianceData extends Loggable {
...
     }
 
-    overridesData.toList
+    overridesData
   }
 
   //////////////// Directive Report ///////////////
 
-  // From Rule Point of view
-  def getRuleByDirectivesComplianceDetails(
```
fanf, line 493: not used anywhere

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/LogDisplayer.scala
```diff
@@ -172,34 +172,55 @@ class LogDisplayer(
...
+        .toBox
+    }
+
+    def getRuleName(ruleId: RuleId): Box[String] = {
+      ruleRepository.get(ruleId).map(x => x.name).toBox
+    }
+
+    val data = LogDisplayer.getReportsLineForNode(reports, getDirectiveName, getRuleName)
+
+    OnLoad(JsRaw(s"""refreshTable("${StringEscapeUtils.escapeEcmaScript(tableId)}",${data});"""))
```
clarktsiory, line 189: `${data}` will represent the `data` as string but it is a case class, however we want it's json representation with `data.json.toJsCmd` !

webapp/sources/rudder/rudder-web/src/test/scala/com/normation/rudder/web/services/ReportLineTest.scala
```diff
@@ -0,0 +1,132 @@
...
+      executionDate: DateTime,
+      ruleId:        String,
+      directiveId:   String,
+      nodeId:        String,
+      reportId:      String,
+      component:     String,
+      keyValue:      String,
+      message:       String
+  ) {
+    def toT: (DateTime, RuleId, DirectiveId, NodeId, String, String, String, DateTime, String) = (
```
clarktsiory, line 63: with such method names I prefer when the case class is private (even within a `private object ReportLineTest` companion object)

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
clarktsiory: as in a preview comment, `${newData}` will use the default `toString` on a case class, `JsTableLine` here, so it's not a valid json, we should still keep the `newData.json.toJsCmd`

webapp/sources/rudder/rudder-web/src/test/scala/com/normation/rudder/web/components/RuleLineTest.scala
```diff
@@ -0,0 +1,104 @@
...
+  val mockRules  = new MockRules()
+  val mockNodes  = new MockNodes
+  val mockGroups = new MockNodeGroups(mockNodes, new MockGlobalParam)
+
+  import com.normation.utils.SimpleStatus.*
+
+  val callback: Rule => Option[AnonFunc] = (r: Rule) => {
+    val ajax = SHtml.makeAjaxCall(JsRaw("'" + "cb" + "=' + encodeURIComponent(" + RedirectTo(r.id.serialize).toJsCmd + ")"))
+    Some(AnonFunc("action", ajax))
+  }
```
clarktsiory, line 67: for tests, we may actually avoid using the horrendous JS API, and make a simple `AnonFunc(s"return ${r.id.serialize}")` instead...

webapp/sources/rudder/rudder-web/src/test/scala/com/normation/rudder/web/services/ComplianceLineTest.scala
```diff
@@ -0,0 +1,285 @@
...
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |, [0, 0.0]
+      |]
+      |, "compliancePercent": 100.0, "id": "r1", "details": [{"directive": "25. Testing blocks", "id": "directive-techniqueWithBlocks", "techniqueName": "technique with blocks", "techniqueVersion": "1.0", "compliance": [[0, 0.0]
```
clarktsiory, line 104: I would prefer a json semantic equality check (`equalsJsonSemantic`) and having a prettified JSON without margin in tests, it is very hard to edit,copy/paste and diff otherwise

clarktsiory (changes requested):
Overall it looks good to me, there are two breaking changes related to string interpolation that need to be fixed

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
fanf: ah yes, I changed return to `JsTableLine` back since comment, and forgot to correct all places

webapp/sources/rudder/rudder-web/src/test/scala/com/normation/rudder/web/components/RuleLineTest.scala
fanf: I really want to have the test as near as possible from the use case, to check that what we test is really what is used.

It should go away in latter refactoring and migration

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -290,7 +276,7 @@ class RuleGrid(
         val onLoad              = {
           s"""createRuleTable (
                   "${htmlId_rulesGridId}"
-                , ${tableData.json.toJsCmd}
+                , ${tableData}
```
clarktsiory, line 279: This breaking change remains, but I think this is the last one : 
```suggestion
                , ${tableData.toJson.toJsCmd}
```

fanf:
PR updated with a new commit

clarktsiory (approved):

fanf:
OK, squash merging this PR

PR #6087, opened 2024-12-16
Fixes #26071: Default configuration for user API tokens should be disabled for OIDC and OAuth2 

clarktsiory:
https://issues.rudder.io/issues/26071

For OIDC, the default value is not the global one : the "global values map" should instead be a mapping of default values.
And instead of concatenating the global map with the user-defined properties, we should just use our mapping of default values now.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -157,20 +157,26 @@ class AppConfigAuth extends ApplicationContextAware {
       value
     }
 
-    // prepare specific properties for each configuredAuthProviders - we need system properties for spring
-    import scala.jdk.CollectionConverters.*
-    // provider properties need to be set to global config before being overridden by specific config by provider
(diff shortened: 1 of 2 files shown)


clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -209,20 +218,27 @@ class AppConfigAuth extends ApplicationContextAware {
...
                     value
                 }
-              } catch { case ex: ConfigException.Missing => defaultEnableRestToken }
+              } catch {
+                case _: ConfigException.Missing =>
+                  if (defaultValue != restTokenGlobalFeatureSwitch) {
+                    logger.info(
+                      s"User access to REST API via token for provider ${x.name} is by default configured to be ${defaultValue.name}, overriding the global configuration"
+                    )
+                  }
```
clarktsiory, line 227: without this the user will not know that the Rudder default value is `disabled` for OIDC

fanf (approved):
Works as expected

clarktsiory:
OK, squash merging this PR

PR #6097, opened 2024-12-19
Fixes #26075: Trying to save a group with empty criteria removes all entries

clarktsiory:
https://issues.rudder.io/issues/26075

* LDAP : any invalid stored query in the LDAP causes errors when reading them, we should not care to validate if comparator values is not empty, validation should happen during write,read **is not strict**. So we need to provide a version of parser that does not apply validations to LDAP mapper
* API : we should enforce that having value as empty string is also not valid, since this was the initial error, the parser was missing an invalid case, it should **be strict** against validating `""`
* UI : the save button is currently disabled only when there are form errors. But we need to disable the button also when the a criterion line has been added (or removed), this prevents from saving the query with an empty string : screenshot with the disabled button 
image

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
+++ b/webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
@@ -1122,6 +1122,78 @@ response:
       }
     }
 ---
+description: Update a node group query with empty query comparator value
+method: POST
+url: /api/latest/groups/00000000-cb9d-4f7b-abda-ca38c5d643ea
+headers:
+  - "Content-Type: application/json"
+body: >-
+  {
+    "id": "00000000-cb9d-4f7b-abda-ca38c5d643ea",
+    "query": {
+      "select": "nodeAndPolicyServer",
+      "composition": "and",
+      "transform": "invert",
+      "where":[
+        {
+          "objectType": "node",
+          "attribute": "nodeId",
+          "comparator": "eq",
+          "value": ""
+        }
+      ]
+    }
+  }
+response:
+  code: 200
+  content: >-
+    {
+      "action":"updateGroup",
+      "id":"00000000-cb9d-4f7b-abda-ca38c5d643ea",
+      "result":"success",
+      "data":{
+        "groups":[
+          {
+            "id":"00000000-cb9d-4f7b-abda-ca38c5d643ea",
+            "displayName":"clone from api of debian group",
+            "description":"Some long description",
+            "category":"GroupRoot",
+            "query": {
+              "select": "nodeAndPolicyServer",
+              "composition": "and",
+              "transform": "invert",
+              "where":[
+                {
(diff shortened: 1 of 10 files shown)


fanf:
I don't think this is the correct correction. This correction is just handling the "UX should not be annoying" part, but it doesn't correct the cause, which is: we saved something in LDAP, and we aren't able to read it back to retrieve the same state. 
That invariant must be kept. I don't know where we lose it, but whatever we put in LDAP must be unserializable back. 

Then, we would at least get back the same list of criteria in the group - which in the the case of `AND hostname = ""` will lead to an empty list of node obviously, but that's something the user can act on by themself. 

And only at that point, we add the UX layer, which force the user to see the consequences of their criteria on the node list before saving.

fanf (changes requested):
Please make the group unserialisable back

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
```diff
@@ -1122,6 +1122,37 @@ response:
...
+url: /api/latest/groups/00000000-cb9d-4f7b-abda-ca38c5d643ea
+headers:
+  - "Content-Type: application/json"
+body: >-
+  {
+    "id": "00000000-cb9d-4f7b-abda-ca38c5d643ea",
+    "query": {
+      "select": "nodeAndPolicyServer",
+      "composition": "and",
+      "transform": "invert",
```
fanf, line 1136: why invert ? (really to know: it's not the default queries, and not relevant to that problem, so I'm wondering)

clarktsiory: it's to have it at least covered in test somewhere, I don't think the value is tested anywhere.
I think the value for "transform" is validated before the criteria

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
```diff
@@ -1122,6 +1122,37 @@ response:
...
+      "where":[
+        {
+          "objectType": "node",
+          "attribute": "nodeId",
+          "comparator": "eq",
+          "value": ""
```
fanf, line 1142: I'm really not sure that we want to forbid that. Things could be the empty string and being valid. And it could be different than "present/absent". Typically in overriding case, where you specifficaly want to override a description to empty string.

clarktsiory:
PR updated with a new commit

fanf (approved):
Perhaps we will want to totally remove the raw parser in the future, the strcit seems to be a better fit for what we want. But for now, we will remain safe for the patch release.

clarktsiory:
OK, squash merging this PR

clarktsiory:
OK, merging this PR

PR #6098, opened 2024-12-19
Fixes #26088: NPE in archive generation

fanf:
https://issues.rudder.io/issues/26088

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/ZipUtils.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/ZipUtils.scala
@@ -59,7 +59,14 @@ import zio.syntax.*
 
 object ZipUtils {
 
-  final case class Zippable(path: String, useContent: Option[(InputStream => IOResult[Any]) => IOResult[Any]])
+  // we need to ensure that path is never null, which can happen because we interact with Java
+  final case class Zippable(path: String, useContent: Option[(InputStream => IOResult[Any]) => IOResult[Any]]) {
+    if (null == path) {
+      throw new IllegalArgumentException(
+        s"A zippable can not be created with an 'null' path (useContent ${if (useContent.isEmpty) "is empty" else "is defined"})"
+      )
+    }
+  }
 
   object Zippable {
     def make(path: String, content: Option[IOManaged[InputStream]]): Zippable = {
@@ -193,7 +200,12 @@ object ZipUtils {
     }
     val base = file.getParentFile.toURI
 
-    def getPath(f: File) = base.relativize(f.toURI).getPath
+    def getPath(f: File): Option[String] = {
+      base.relativize(f.toURI).getPath match {
+        case null => None
+        case p    => Some(p)
+      }
+    }
 
     def recZippable(f: File, existing: Seq[Zippable]): Seq[Zippable] = {
       // sort accordingly to the required convention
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/ZipUtils.scala
```diff
@@ -59,7 +59,14 @@ import zio.syntax.*
...
 
-  final case class Zippable(path: String, useContent: Option[(InputStream => IOResult[Any]) => IOResult[Any]])
+  // we need to ensure that path is never null, which can happen because we interact with Java
+  final case class Zippable(path: String, useContent: Option[(InputStream => IOResult[Any]) => IOResult[Any]]) {
+    if (null == path) {
+      throw new IllegalArgumentException(
+        s"A zippable can not be created with an 'null' path (useContent ${if (useContent.isEmpty) "is empty" else "is defined"})"
+      )
+    }
+  }
```
clarktsiory, line 69: Because there is a factory `Zippable#make`, is could be more idiomatic to put the null check in the constructor instead (and make the constructor private).
Also if we return an `Option[Zippable]`, we could also handle the error at the place where the zippable object needs to be used

clarktsiory (commented):
A few suggestions to make it more look like Scala than Java

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/ZipUtils.scala
```diff
@@ -193,7 +200,12 @@ object ZipUtils {
     }
     val base = file.getParentFile.toURI
 
-    def getPath(f: File) = base.relativize(f.toURI).getPath
+    def getPath(f: File): Option[String] = {
+      base.relativize(f.toURI).getPath match {
+        case null => None
+        case p    => Some(p)
+      }
```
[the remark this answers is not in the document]
fanf: I specifically avoided that one because it is not clear at all and a java thing. It's really hard to read that bit and infer in what case you get what. 
(one of scala API design failt - `Option.fromNullable` would have been so much better)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/ZipUtils.scala
fanf: I played with that way, but it was meh : 
- what is idiomatic is to have the general contract that things are not null. Almost nowhere, we assume that a parameter can be null, because we do scala
- so the general usage is standard scala `case class`, with public constructor and caller not doing wrong things. 
- since we do have prior art of things passing null at that point, we add a guard to show that it's a developper error, and that it shall not be done like that. 
- the correct ansmwer is to have `Nullable` as a separed aspect of types and scala API forbidding it, but it's an ecosystem matter. 

For the factory part, I'm not sure why I reverted the "`Option` return statement. I think I was all-in in the mood of "if you return null, you have to deal with that, no defensive code that complexify everything everywhere"

fanf: Ah, yes: the problem is that the factory just doesn't know what to do with a null. It's already too late in the code to handle it correctly, so it's really a dev error, and so an exception is the correct way to signal that the dev did something very bad: 
image

clarktsiory (approved):
Okay, I get the point... Java being a null-oriented language is really a doom for us...

PR #6105, opened 2024-12-30
Fixes #26116: Add Scala backend for technique loops

VinceMacBuche:
https://issues.rudder.io/issues/26116

Just adding two fields to Yaml and Editor techniques classes

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/ncf/TestEditorTechniqueWriter.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/ncf/TestEditorTechniqueWriter.scala
@@ -458,7 +458,9 @@ class TestEditorTechniqueWriter extends Specification with ContentMatchers with 
           "any",
           "Customized component",
           disabledReporting = false,
-          policyMode = None
+          policyMode = None,
+          None,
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/ncf/TestEditorTechniqueWriter.scala
```diff
@@ -623,7 +641,7 @@ class TestEditorTechniqueWriter extends Specification with ContentMatchers with
     EditorTechnique(
       BundleName("testing_variables_in_conditions"),
       new Version("1.0"),
-      "Testing variables in conditions",
+      "Testing vçariables in conditions",
```
Fdall, line 644: ?

VinceMacBuche:
Commit modified

fanf (approved):
OK to merge that as this but we will need to: 
- avoid string-typing and give these structure some shape when working on the UI
- add the missing unit tests that demo what is expected in parsing / serialisation and ensure we keep that API stable in the future and question what invariant we have on them (esp on the key/value set for iteration, how we deal we empty/missing values, does order matter - at least for serialisation/UI consistency, etc)

PR #6106, opened 2024-12-30
Fixes #26120: Make sure that bundle arity is always strict when resolving loops

Fdall:
https://issues.rudder.io/issues/26120

### DIFF
--- a/policies/rudderc/src/backends/unix/ncf/method_call.rs
+++ b/policies/rudderc/src/backends/unix/ncf/method_call.rs
@@ -180,15 +180,9 @@ pub fn method_call(
             call_parameters.push(cfengine_canonify_condition(condition.as_ref()));
             method_parameters.push("method_call_condition".to_string())
         }
-        Condition::NotDefined => {
+        Condition::NotDefined | Condition::Defined => {
             if m.resolved_foreach_state.is_some() {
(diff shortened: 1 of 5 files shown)


Fdall:
PR updated with a new commit

policies/rudderc/src/backends/unix/ncf/method_call.rs
```diff
@@ -173,9 +173,25 @@ pub fn method_call(
...
         "class_prefix".to_string(),
     ];
-    if let Condition::Expression(_) = condition {
-        call_parameters.push(cfengine_canonify_condition(condition.as_ref()));
-        method_parameters.push("method_call_condition".to_string())
+    // If the item is a result of a foreach loop, we must assume that one of the branch could define
+    // a condition, and so, each branch should call the bundle using the method_call_condition
+    match condition {
+        Condition::Expression(_) => {
+            call_parameters.push(cfengine_canonify_condition(condition.as_ref()));
```
amousset, line 180: You should be able to to just remove the if and keep the same content. `Condition`'s `Display` implementation already handle the true/false cases.

Fdall:
PR updated with a new commit

amousset (approved):

PR #6107, opened 2024-12-30
Fixes #26055: Change rudder 8.3 main menu

fanf:
https://issues.rudder.io/issues/26055

A lot of changes which can be grouped into: 
- changing menu item position and related URLs, 
- changing the structure of page: Node, Settings, Maintenance (new) to embed old pages as tabs

image

# Menu items

Most of the work goes in `Boot.scala`. 
The main changes are: 
- changing item position (numbers etc)
- removing `LocGroup` which is a feature to build breadcrump and the like that we don't use, 
- factoring out and normalizing things, like the security check, or always redirecting to `/secure/index`

Some URL needed to be changed: 
- `/secure/security/` -> `/secure/patch` for CVE related things
- `/secure/plugins/` doesn't exist anymore
- `/secure/utilities/` doesn't exist anymore
- `secure/access/` is a new one for user management / api tokens

# Page with tabs

We want to group several page (from rudder core or modules) into one main page with tabs: 
- `nodes` for the node list and node search, 
- `settings` for general settings and plugins settings
- `maintenance` for pages that used to be under utilities like health check, report database info, etc

The first one is simple since everything is in Rudder Core. 
For the other two, we need to build a tab menu that can be extended by plugins, like for node details. 

The general process for `settings` is (it's the same for `maintenance`) : 
- create a snippet in `snippet/administration/Settings` with `DefaultExtendableSnippet` trait and the corresponding HTML file in `webapp/secure/administration/settings`. 
    - the HTML provides the general skeleton for tabs and the first tab (which is special, it needs to be here and activate from start) 
    - the snippet is very simple and mainly provide utility methods to help caller add tabs
- then, for each page that becomes a tab: 
    - migrate the HTML from `webapp/secure` to `webapp/templates-hidden`
    - change the root HTML from using `common-layout` to using `<xml:group><component-body>`
    - if the page is in rudder core, just include it in the corresponding main snippet with a `ChooseTemplate`
    - if the page is in a plugin (or for `RudderCompanyAccount`, used for demo of extensible snippet in core): 
        - create a snippet class extending trait `SnippetExtensionPoint[Settings]` (or `PluginExtensionPoint[Settings]`)
        - in the compose, make it uses the `Settings.addTab` method
        - register the extension with `snippetExtensionRegister.register(new RudderCompanyAccount())`

See https://github.com/Normation/rudder-plugins/pull/782 and https://github.com/Normation/rudder-plugins-private/pull/883 for impact on plugins (a lot, too)


webapp/sources/rudder/rudder-web/src/main/webapp/secure/nodeManager/nodes.html
```diff
@@ -25,17 +25,57 @@ <h1>
...
-          <div class="main-container">
-            <div class="main-details table-container">
-              <div data-lift="node.Nodes.table"></div>
+      <div id="nodes_tab">
+        <div class="one-col-main">
+          <div class="template-main">
+            <div class="main-container">
+              <div class="main-details">
+                <div class="main-container">
+                  <div class="main-details">
```
clarktsiory, line 45: there is a little problem here

Also when opening the `Nodes` menu, I get the following (if I click on tabs, the content is actually split by tab) : 
Screenshot from 2025-01-03 10-52-52

webapp/sources/rudder/rudder-web/src/main/style/rudder/rudder-menu.scss
```diff
@@ -788,31 +788,24 @@ header.main-header .logo-lg img{
 
 /* -- SIDEBAR MENU CUSTOMIZATION -- */
 
-.sidebar ul.sidebar-menu > li.treeview:nth-child(2) > ul.treeview-menu ,
-.sidebar ul.sidebar-menu > li.treeview:nth-child(3) > ul.treeview-menu {
+.sidebar ul.sidebar-menu > li.treeview:not(:nth-last-child(-n+3)):not(:nth-last-child(-n+2)) > ul.treeview-menu {
```
clarktsiory, line 791: I'm not really sure about what this means...

clarktsiory (changes requested):
The change is clear and seems exhaustive, the menu makes more sense.

LGTM, except that there is a display issue in the Nodes menu

webapp/sources/rudder/rudder-web/src/main/style/rudder/rudder-menu.scss
fanf: I'm better than you at css \o/
apply that, but "not the n-3 and n-2 child' (ie main cat) of the menu, ie not to account and administration (there is a last menu entry).

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/webapp/secure/access/index.html
```diff
@@ -1,4 +1,4 @@
 <lift:surround with="common-layout" at="content">
   <!-- just a redirect -->
-  <div data-lift="Index.utilities"></div>
+  <div data-lift="Index.access"></div>
```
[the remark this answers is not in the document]
fanf: :+1:

webapp/sources/rudder/rudder-web/src/main/webapp/secure/nodeManager/nodes.html
fanf: I yes, both have the class "active" in the template

fanf:
PR updated with a new commit

PR #6108, opened 2025-01-02
Fixes #26119: All rule show up in directive compliance

clarktsiory:
https://issues.rudder.io/issues/26119

The directive compliance reports are using rules compliance reports, but currently an empty list of rule reports is just flattened further down as a rule with 0 compliance.
It should instead be deduced from the `Option`, obtained when getting directive reports

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -631,6 +631,7 @@ class ComplianceAPIService(
                        } else {
                          Seq().succeed
                        }
+      ruleIds        = rules.map(_.id).toSet
       t2            <- currentTimeMillis
       _             <- TimingDebugLoggerPure.trace(s"getByDirectivesCompliance - getAllRules in ${t2 - t1} ms")
       nodeFacts     <- nodeFactRepos.getAll()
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -684,34 +690,43 @@ class ComplianceAPIService(
...
       } yield {
         // We will compute compliance for each rule for the current Directive
-        val rulesCompliance = for {
-          (ruleId, ruleReports) <- reportsByRule.toSeq
-
-          // We will now gather our report by component for the current Directive
-          reportsByComponents = (for {
-                                  ruleReport       <- ruleReports
-                                  nodeId            = ruleReport.nodeId
-                                  directiveReports <- ruleReport.directives.get(directive.id).toList
```
clarktsiory, line 694: This is likely the culprit : the `option.toList` is hidden in the for-comprehension.
It exactly indicates if the rule has reports for the directive i.e. if the compliance report of this rule is needed or not.

fanf:
can you please add a unit test ? It's clearly a edge case we want to have on chek

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_compliance.yml
```diff
@@ -2312,3 +2312,59 @@ response:
...
+            {
+              "id" : "br4",
+              "name" : "R4",
+              "compliance" : 100.0,
+              "policyMode" : "enforce",
+              "complianceDetails" : {
+                "successAlreadyOK" : 100.0
+              },
+              "components" : []
+            }
```
clarktsiory, line 2365: in the previous version, it returned rules that did not event target this directive (only `br4` has this directive) : 
```
        +        {
        +          "id" : "br2",
        +          "name" : "R2",
        +          "compliance" : 0.0,
        +          "policyMode" : "enforce",
        +          "complianceDetails" : {},
        +          "components" : []
                 }
```

This test ensures that this no longer happens, only rules with the directive are listed in the directive compliance

fanf (approved):
seems to do as intended :+1:

clarktsiory:
OK, squash merging this PR

PR #6112, opened 2025-01-03
Fixes #26135: Instance ID should be added as a variable for nodes

clarktsiory:
https://issues.rudder.io/issues/26135


clarktsiory:
PR updated with a new commit

fanf (approved):
Great !

PR #6115, opened 2025-01-08
Fixes #26149: Handle the plugin metadata for license requirement

amousset:
https://issues.rudder.io/issues/26149

* Adds a test ensuring JSON output stability now it will be used.
* Add the `requires_license` parameter to metadata and JSON output.
* Adapt human output to mention missing licenses.

The change is minimal, we could improve other parts like preventing installation of plugins the user does not have a license for, instead of failing with a 404. This would make `rudder-package` usable in Rudder core contexts.


amousset:
PR updated with a new commit

Fdall (approved):

PR #6119, opened 2025-01-10
Fixes #26153: Add a user interface for managing method/block loops in the techniques editor

RaphaelGauthier:
https://issues.rudder.io/issues/26153

:warning: work in progress

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor.elm
@@ -140,9 +140,15 @@ subscriptions model =
                   )
         ]
 
+defaultMethodUiInfo : Maybe MethodCall -> MethodCallUiInfo
+defaultMethodUiInfo call =
+  let
+    (foreachName, foreach) = case call of
+      Just c -> (c.foreachName, c.foreach)
+      Nothing -> (Nothing, Nothing)
+  in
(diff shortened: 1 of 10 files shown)


RaphaelGauthier:
Here's what it looks like at the moment :

If no foreach has been declared: the ‘Foreach’ tab contains a greyed-out counter :
foreach-disabled

If a foreach has been declared, this badge is highlighted to show this information, and a banner above the method is displayed with the name of the foreach and the number of iterations so that this information is always visible, even when the method is collapsed:
foreach-enabled

To declare a ‘foreach’, user must enter the name and keys of this iterator, then validate to activate the feature :
foreach-creation

Once the foreach has been declared, the user can modify the name and keys, and add new items on which to iterate.
:warning: For the moment, this interface isn't finished, so here's a look at how it looks now, compared with the result we're aiming for :
foreach-edit-WIP
foreach-edit-goal

Here is the list of things to finish:
- enable iterator editing (name, keys and items)
- do the same for Blocks
- add a copy-to-clipboard on the banner
- decode, store and encode the foreach list correctly
- add a warning before deleting the foreach

RaphaelGauthier:
PR updated with a new commit

RaphaelGauthier:
PR updated with a new commit

fanf (approved):
Rebased to branche 8.3. 
Merging, since it works as designed. Enhancement will be done during alpha.

PR #6121, opened 2025-01-13
Fixes #26167: Make API authentication pluggable

fanf:
https://issues.rudder.io/issues/26167

The only meaningful changes are: 
- adde a `name` to the http security filter for API, so that it can be referenced in the plugin, 
- make the current Rudder API filter, `RestAuthenticationFilter`, just pass the request away when the authentication is already done

Other changes are new logger/trying to make msg construction clearer. 
Look a the change with `?w=1` .


clarktsiory (approved):
LGTM !

PR #6120, opened 2025-01-10
Fixes #26162: Install remove and change status of plugins from API

clarktsiory:
https://issues.rudder.io/issues/26162

* Add API tests to previous PR
* Add API endpoints, there are endpoints to install, remove and change status
* Add tests for errors computation 
* Along the way, added the error when a plugin is required as mentioned in https://github.com/Normation/rudder/pull/6114#discussion_r1909262123

PR #6123, opened 2025-01-14
Fixes #26169: License information and credentials management in plugins API

clarktsiory:
https://issues.rudder.io/issues/26169

## :warning: Based on #6120

* Adds license information to the plugins API (see yml tests), by refactoring the existing structures 
* Adds credential management on the 'plugins list' endpoint : we need to handle the command-line error message, up to our API response, by adding utility methods to convert an Either with an error into a lift response.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
@@ -39,8 +39,10 @@ package com.normation.plugins
 
 import better.files.File
 import com.normation.errors.*
+import com.normation.plugins.RudderPackageService.*
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.rudder.hooks.Cmd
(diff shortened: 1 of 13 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
```diff
@@ -536,6 +579,26 @@ trait RudderPackageService {
...
 
+object RudderPackageService {
+
+  val ERROR_CODE: Int = 1
+
+  // see PluginSettings : the url and credentials configuration could cause errors :
+  final case class CredentialError(msg: String) extends RudderError
+
+  object CredentialError {
+    private val regex = "^.*ERROR.* (Received an HTTP 401 Unauthorized error.*)$".r
```
fanf, line 590: this should be a dedicated error code from the cli

PR #6134, opened 2025-01-20
Fixes #26207: Plugin management operations need confirmation popup and license error info

clarktsiory:
https://issues.rudder.io/issues/26207

Adding modals to confirm actions on plugins :
image

Fixing also some other problems :
* in case of missing license, the plugin is not supposed to be selected 
* disable and enable action were not working due to the command being "enabled/disabled" instead of "enable/disable"
* sorting the plugins by their status, then by alphabetical order, using the [elm-ordering](https://package.elm-lang.org/packages/matthewsj/elm-ordering/latest/) lib which is very handy for that
* minor SCSS adjustments


clarktsiory:
PR updated with a new commit

RaphaelGauthier (approved):

fanf:
OK, squash merging this PR

PR #6135, opened 2025-01-21
Fixes #26235: License update needs a button and we need all license end dates

clarktsiory:
https://issues.rudder.io/issues/26235

:arrow_down:  We now have a "Check for updates" buttons that calls `rudder package update` to refresh the license.
We now also show a list of end dates for groups of plugins expiring on the same date :

image

:arrow_down:  We can also updates licenses in case there is no global license yet : 
image

## Changes 
* we need to change the datatype for aggregated license to have a mapping from the license `endDate` to the number of plugins ending with this date, both in Scala and Elm  
* (unrelated :) migration of some lift-json to zio-json, see `JsonPluginFile`


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/RudderPluginJson.scala
```diff
@@ -52,17 +52,38 @@ import zio.syntax.*
...
 // this is only used internally but liftweb can't hand it if not top level
-final protected case class JsonPluginFile(plugins: Map[String, JValue])
+final protected case class JsonPluginFile(plugins: Map[String, JsonPluginRaw])
+protected object JsonPluginFile {
+  implicit val decoder: JsonDecoder[JsonPluginFile] = DeriveJsonDecoder.gen[JsonPluginFile]
+}
 
-// this is only used internally but liftweb can't hand it if not top level
+// json format of plugin, optional fields are there to obtain empty lists when mapping none
+@jsonMemberNames(KebabCase)
```
fanf, line 61: if it's a json interface class, please don't use that, else: 
- we can't find attribute name seen in JSON easily with a search
- it's easier to introduce divergence. 

Plus, if it's used only internally, we must change it to use the general standard. 
And if it's used publicly, we need to keep the name in a past version, and change them in new version to keep naming consistency.

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_plugins.yml
```diff
@@ -13,7 +13,9 @@ response:
             "test-licensee"
           ],
           "startDate" : "2025-01-10T21:53:20+01:00",
-          "endDate" : "2025-01-10T21:53:20+01:00",
+          "endDates" : {
+            "2025-01-10T21:53:20+01:00" : 1
```
fanf, line 17: I'm really not fan about that encoding. Often, it makes compat with other part of systems more complicated for just a bit of compactness - but really, if we are chasing that, there's a lot of other things to do, like using long for epoch at GMT, only one letter keys, arrays, etc. 

Can you use something like 
```
"endDates": [
  { 
    "date": "2025-01-10T21:53:20+01:00",
    "count": 1
  }
]
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/RudderJsonResponse.scala
```diff
@@ -353,6 +353,43 @@ object RudderJsonResponse {
       ): LiftResponse = {
         toLiftResponseOneEither(params, ResponseSchema.fromSchema(schema), SuccessIdTrace(id))(JsonEncoder[B], ev)
       }
+
+      def toLiftResponseZeroEither(
```
fanf, line 357: can you add a comment for that one?

fanf (changes requested):
Overall seems OK, but some things should be made clearer/simpler (dumber) to simplify future maintenance

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_plugins.yml
clarktsiory: Sure, it is simpler to handle within the code especially in the frontend !
(The real reason for which I used this is that the Map naturally encodes to key-values as a JSON object)

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/RudderPluginJson.scala
clarktsiory: Okay, I added a better comment.
It is really not the zio-json mapping of member names that we want, but chimney mapping in that case, but we'd have to wait for Scala 3 to use it (if there are libs for mapping to known cases, or if we make our own) : https://chimney.readthedocs.io/en/stable/supported-transformations/#defining-custom-name-matching-predicate

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
```diff
@@ -433,8 +433,84 @@ object PluginId                  {
...
+      Some(Map((endDate, DateCount(endDate, 1)))),
+      info.maxNodes
+    )
+  }
+
+  private def from(licenses: NonEmptyChunk[PluginLicenseInfo]): JsonPluginsLicense = {
+    val res                     = licenses.reduceMapLeft(fromLicenseInfo) { case (lim, lic) => lim.combine(fromLicenseInfo(lic)) }
+    val sortedDistinctLicensees = res.licensees.map(_.sorted.distinct).flatMap(NonEmptyChunk.fromChunk)
+    res.copy(licensees = sortedDistinctLicensees)
+  }
```
fanf, line 494: you have that exact code two time, one for `JsonPluginsLicense`, one for `JsonGlobalPluginLimits`, can you factor it out ? (yes, only 2 when we generally wait for three, but it's the exact same code, applying for all variant of the object, it cries for having the common pattern somewhere else)

fanf (commented):
Some details. They can be corrected in a further PR, no need to change that one. If so, just mark the PR as "ready to merge"

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
```diff
@@ -433,8 +433,84 @@ object PluginId                  {
...
+    }
+  }
+}
+object JsonPluginsLicense        {
+  import DateFormaterService.JodaTimeToJava
+
+  final case class DateCount(date: ZonedDateTime, count: Int)
+
+  // encoded as json list but using a Map for unicity when grouping by date
+  type DateCounts = Map[ZonedDateTime, JsonPluginsLicense.DateCount]
```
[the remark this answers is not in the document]
clarktsiory: okay I also thought that an opaque type would be great, and that we could refactor when needed (and it would be a great thing to test when Scala 3 will come) but I guess it is not that complicated to change from a case class either

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
clarktsiory: agreed, so I will look up how to do this in the most simple but generic waay

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginSettings.scala
```diff
@@ -93,73 +94,153 @@ final case class JsonPluginsDetails(
...
+      def fromZonedDateTime(date: ZonedDateTime): ZonedDateTime = date
+    }
+
+    implicit val dateCounts: ToEndDate[DateCounts] = new ToEndDate[DateCounts] {
+      override def fromZonedDateTime(date: ZonedDateTime): DateCounts = DateCounts.one(date)
+    }
+  }
+
+  final case class DateCount(date: ZonedDateTime, count: Int)
+  // Opaque type for counts by date. Encoded as json list but using a Map for unicity when grouping by date
```
fanf, line 177: that's not opaque at all ;)
Can you use "dedicated data structure" until we really switch to scala 3 and are able to use opaque-type-in-the-scala-meaning to avoid confusion ?

fanf (approved):
There's a comment that could be improved, but that's more a general stance than blocking that PR. 
Works as expected, and the code is nice.

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

PR #6137, opened 2025-01-22
Fixes #26244: License repository errors should result in different exit codes

clarktsiory:
https://issues.rudder.io/issues/26244

The main can return an `ExitCode`, we only have to map it from the result of running the CLI : we intercept the repository errors and associate specific error codes depending on the HTTP error. For now the exit code are 2 and 3.

### DIFF
--- a/relay/sources/rudder-package/src/lib.rs
+++ b/relay/sources/rudder-package/src/lib.rs
@@ -22,11 +22,13 @@ use std::process;
 use std::{
     fs::create_dir_all,
     path::{Path, PathBuf},
+    process::ExitCode,
 };
 
 use anyhow::{anyhow, bail, Context, Result};
 use clap::Parser;
 use cli::Args;
+use repository::RepositoryError;
 use rudder_cli::custom_panic_hook_ignore_sigpipe;
 use tracing::{debug, error, info, warn};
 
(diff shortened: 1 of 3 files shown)


clarktsiory:
PR updated with a new commit

amousset (commented):
:+1: for the return code handling and the usage of a custom error type!

This is good on paper, but we will actually almost never reach a 403. The way the authentication code works is by hiding the existence of unauthorized files, due to technical reasons, so the 403 and 404 are mostly indistinguishable.
Additionally, the 403 error very likely means the credentials are valid but the content forbidden, so the error message is not right.

Furthermore, I think the `RepositoryError` should not expose raw HTTP semantics but rather a "Rudder-level" error and keep the matching private.

Maybe something like:

```rust
#[derive(Debug)]
pub enum RepositoryError {
    InvalidCredentials(anyhow::Error),
    Unauthorized(anyhow::Error),
}
```

While keeping the details in the error itself with something like:

```rust
let e = anyhow!("Invalid credentials, please check your credentials in the configuration. (received HTTP {status}");
RepositoryError::InvalidCredentials(e)
```

clarktsiory:
PR updated, the changes have been made !
Now there is one error code per enum case (so there are only 2 of them)

clarktsiory:
PR rebased

amousset (approved):

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

PR #6140, opened 2025-01-23
Fixes #26206: Directive will not be displayed when clicking on a technique then on directive in tree

clarktsiory:
https://issues.rudder.io/issues/26206

As we don't display techniques based on the window hash, we can just set the hash to empty when clicked (it may leave `#` but it is not an issue, and does not seem to refresh the page, even on the load testing machine)


fanf (approved):
Seems to work as expected (both in firefow and chromium)

fanf:
OK, merging this PR

PR #6142, opened 2025-01-23
Fixes #26200: Error at rudder start after an upgrade from 8.0 to 8.1.11

fanf:
https://issues.rudder.io/issues/26200

So, the change are rather small in code, but there a lot of files moving here and there. 
The main change is that I created two dedicated `Bootstrap.check()` sequence for early migration change regarding LDAP/PostgreSQL schema evolution. These two sequences are called as soon as the connection toward these base is up, and they are executed before making the connection service available to other services. 

The file dance is done because I reflected that structure in package naming. Now, unders `bootstrap.checks`, we have: 

- `earlyconfig.{db,ldap}` for the new checks belonging to the two new early sequence. I move here anything looking like a schema migration
- `endconfig` where I moved the old sub-package `action`, `consistency`, `migration`, `onetimeinit`.  

The same structure was needed also for the test packages and classes. 

Appart that, I update the bootstrap sequence class to get a name and a logger, so that the log for each sequence are different. 

And I needed to split a migration class in two to separate the DB schema evolution (put in `earlyconfig.db`) and other business migration (remaining in `endconfig`). 

Finally, I created a doc with ALL the init sequence high level logic in `rudder-bootstrap.md`

I tested that it corrects the linked bug by deleting my `globalscore` table and starting Rudder, which works now.


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/BootstrapChecks.scala
```diff
@@ -86,22 +98,25 @@ class SequentialImmediateBootStrapChecks(_checkActions: BootstrapChecks*) extend
...
-        s"[#${i}] ${check.description}"
-      } else {
-        s"${check.description}"
-      }
-      BootstrapLogger.logEffect.info(msg)
-      check.checks()
-      BootstrapLogger.logEffect.debug(
-        msg + s": OK in [${formatter.print(new Duration(System.currentTimeMillis - start).toPeriod)}] ms"
-      )
+  override def checks(): Unit = {
```
VinceMacBuche, line 104: I still think this should be a ZIO

VinceMacBuche:
I so think we mess initialization and checks in rudder config and it make ti very hard to understand when things are initialized. My guess would be that noting should be done in RudderConfig and they should only be ran separetely in Boot.scala. Maybe storing in different checks category like you do, but not running them in RudderConfig

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/BootstrapChecks.scala
fanf: it could, but that has nothing to do with the bug here. It's a perfect cleaning for next minor version

fanf:
I don't see how we could do that withtout rewriting most of our services. It would need, I think:
- that the instanciation does nothing, in particular does not use other service business methods (since they won't be init at that point)
- out of `RudderConfig`, that we then have "init()" methods for the services, all called in the correct order. 

So we would need to look by hand in what order things need to be init, and that's complicated at scale. 

We can have as a general rule: 
- anything that is linked to data migration goes to the base corresponding own sequence. This is needed because without that, other service may try to use the new schema on old data structure, leading to chaos. 
- anything that is needed for the service to be fully instanciated and callable by other service constructor, including if it has some internal consistency checks, goes in the constructor, 
- anything that looks like a hook or a post-instantiation needed step to "finish instantiation and have a working thing" is done in RudderConfig (if possible grouping them so that we have a place to check and add things... But that's not always possible for scoping reasons)

Then, remains all things that are linked to the service "starting to do work", like schedulers. Perhaps for these one, yes a dedicated step would be good (and we're reinventing spring, it's a bit sad). Starting to "do work" should be done as late as possible to speed to time for rudder to "boot". 
Perhaps bootchecks could go in such a latter phase too. 

It feels a big and difficult change for a patch release thought: we don't know preciselly if that will work, it's possible that we break things

fanf:
PR updated with a new commit

fanf:
Tests and spotless pass localy, it seems something is not OK with the CI.

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #6146, opened 2025-01-29
LDAPConnection code doesn't compile with scala 3 rules

mbaechler:
In scala 3, with -Werror, an IO can't have Object on the error channel unless it's done on purpose.
In this case, it seems more a programming error than a deliberate decision.


fanf:
Seems all good. I don't see anything that is not 8.1-targettable. I will open a ticket and retarget accordingly.

mbaechler:
merged on 8.1 / 8.2 / 8.3 and master

PR #6147, opened 2025-01-29
refactor: specs2 `and` is supposed to be used to chain two assertions…

mbaechler:
… on the same value


fanf:
Merged with https://github.com/Normation/rudder/pull/6221

PR #6150, opened 2025-01-29
refactor: avoid implicit discard of values in ZIO

mbaechler:
`UIO[Foo].catchAll(log).unit` resolves the value as Any whereas `UIO[Foo].unit.catchAll(log)` does not, we should use the latter.


mbaechler:
merged on 8.1 / 8.2 / 8.3 and master

PR #6154, opened 2025-01-30
Fixes #26249: Technique can be created with technique ID starting with a forbidden character

clarktsiory:
https://issues.rudder.io/issues/26249

Upon selection, the whole UI state of the draft is not validated (name and ID), however it should because it may have been left into an invalid state. We just need to refactor the validation of the UI state and apply it upon a new draft selection.

We no longer need to prevent the ID from starting with `_`, it is not validated on the API and is supposed to be prefixed for CFEngine.


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

VinceMacBuche (approved):

PR #6160, opened 2025-02-04
Fixes #26303: Migrate logback.xml to new fixed logfile 

VinceMacBuche:
https://issues.rudder.io/issues/26303

Parses logback.xml, check wether fixed logfile appender exist (check if it has an if element that has an appender element that has a STDOUT name), then add/remove elements that are needed in 8.3

Make a backup of the current logback file, replace with new logback file

Comment that are outside the first element are removed ... Maybe we can add them , but i don't they are this important


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):
LGTM, seems to work as expected !

fanf:
OK, merging this PR

PR #6164, opened 2025-02-06
Fixes #26317: rudder-package should create the license folder when not already there

amousset:
https://issues.rudder.io/issues/26317


clarktsiory (approved):
LGTM, it is also done [when installing](https://github.com/amousset/rudder/blob/d4165b812f66c048ad6ad868420b8a4618371a8d/relay/sources/rudder-package/src/license.rs#L74)

PR #6165, opened 2025-02-06
Fixes #26322: When several plugins are using the same menu, only the last one is displayed

VinceMacBuche:
https://issues.rudder.io/issues/26322

If two plugins are setting the same menu, data that were already set are replaced because we are resetting the menu with the one from the new plugin


fanf (changes requested):
OK, so the problem I used to have is that if the plugin is not `make` at least one time, it's not found in the map of plugins because the plugin ID is not present in the build metadata file. But we don't need late checking. 

Can you please remove the comment above the change ? It tells the opposite of what is done.

fanf:
OK, merging this PR

PR #6173, opened 2025-02-11
Fixes #26335: Port RestApiAccount api endpoint to zio-json 

fanf:
https://issues.rudder.io/issues/26335

WIP status at 2025-02-14

- The backend is working accordingly to what we want, API tests work
- I didn't clean-up things for now: in particular `RestApiAccounts` must be deleted (butI did clean-up things related to the old format)
- elm does not work, at least because elm side isn't able to decode rfc3339 datetime and is expected an other format for the date pickler. 
    - there will be some other inconsistencies, since the API input for update / regenerate token / create token have changed a bit
    - I don't know if there's other things not working since for now, it's blocked here. 

On the backend side, the most notable change are that we have different data structures for different action and return values. 
In particular: 
- create and update are different, since when creating we need to be able to optionnaly choose ID and generate or not a token, 
- there is a difference between action that can expose a token (create, regenerate) and the one that just query information to avoid displaying a token. There is also a difference between the `ClearTextSecret` and the `ApiToken` (hashed).

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
@@ -72,10 +72,10 @@ class RestApiAccounts(
 
   // used in ApiAccounts snippet to get the context path
   // of that service
-  val relativePath: List[String] = "secure" :: "apiaccounts" :: Nil
+  val relativePath: List[String] = "oldapiaccounttoremove" :: Nil
 
   serve {
-    case Get("secure" :: "apiaccounts" :: Nil, req) =>
+    case Get("oldapiaccounttoremove" :: Nil, req) =>
       implicit val prettify: Boolean = restExtractor
         .extractBoolean("prettify")(req)(identity)
         .getOrElse(Some(false))
@@ -119,7 +119,7 @@ class RestApiAccounts(
 
       })
 
-    case "secure" :: "apiaccounts" :: Nil JsonPut body -> req =>
+    case "oldapiaccounttoremove" :: Nil JsonPut body -> req =>
       implicit val prettify: Boolean = restExtractor
         .extractBoolean("prettify")(req)(identity)
         .getOrElse(Some(false))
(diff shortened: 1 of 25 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/RestApiAccounts.scala
```diff
@@ -72,10 +72,10 @@ class RestApiAccounts(
 
   // used in ApiAccounts snippet to get the context path
   // of that service
-  val relativePath: List[String] = "secure" :: "apiaccounts" :: Nil
```
fanf, line 75: this file need to be deleted once we have checked the different use cases

webapp/sources/utils/src/main/scala/com/normation/utils/DateFormaterService.scala
```diff
@@ -58,7 +60,7 @@ object DateFormaterService {
       ZonedDateTime.ofInstant(java.time.Instant.ofEpochMilli(d.getMillis), ZoneId.of(d.getZone.getID, ZoneId.SHORT_IDS))
   }
 
-  object json {
+  trait DateTimeCodecs {
```
fanf, line 63: make it a trait to ease composition with other codes

clarktsiory:
PR updated : fixed API tests (missing `accounts` object in JSON data)

clarktsiory (commented):
Some suggested changes that may benefit to our model, and Scaladoc that I will fix

clarktsiory:
PR updated to fix the Elm Accounts app, with the corresponding expiration policy model, and with some fixes : POST instead of PUT, dates in the REST API need to be ISO8601 now.
There would still need to be change to improve the Elm model (more union types : for enabled/disabled, and ACLs), but that would be in another PR

clarktsiory:
PR updated to fix :  
* Scaladoc,
* sorting of ACLs
* API endpoint description

clarktsiory:
PR rebased

clarktsiory:
PR rebased

clarktsiory:
PR rebased

clarktsiory:
PR rebased

clarktsiory (approved):
LGTM, some improvements in the model, and additional tests covering them in https://github.com/Normation/rudder/pull/6186
We would also want an OpenAPI documentation later

PR #6175, opened 2025-02-11
Fixes #26338: Prettify About page

RaphaelGauthier:
https://issues.rudder.io/issues/26338

- When there is an empty value, it is replaced by a grey text ‘No data available’, and there is no longer a copy-to-clipboard button in this case.

- The "Launch options" value can now be resized and scrolled, and cannot exceed a certain maximum height so that it doesn't take up too much vertical space

about-2


clarktsiory (changes requested):
Nice change ! :+1: 
There is only one little thing that has not been fixed : the copy to clipboard button is misaligned, I think there is a `align-items: center;` on the `.info-container .info` element  
Screenshot from 2025-02-11 15-33-15

clarktsiory (approved):

RaphaelGauthier:
OK, squash merging this PR

PR #6177, opened 2025-02-11
Fixes #26336: Improve user interaction and displayed information on plugins management page 

clarktsiory:
https://issues.rudder.io/issues/26336

In terms of features, we need to know information about which selected plugins can be installed/enabled/disabled/uninstalled.
To better split this, there are two modules : 
* one which contains all the logic of install/enable,etc. : `Action.elm`
* another which makes the selection of allowed plugins : `Select.elm`
which only expose the needed datatypes (most of them are opaque), the mainly used functions are :  
```mermaid
graph LR
  Action --findPluginActions--> Select --select--> DataTypes
```
This model allows to define the action that are allowed/disallowed, then to apply them for selected plugins, and tests have been added for the `select` method on all cases for allowing/disallowing an action on a plugin.

At last, this information is used in aggregated format (count, group by errors/success), and displayed both in the filter bar and in the confirmation modal.

---
In addition to this splitting, the main changes in the existing codebase consist of : 
* modelling a separate type for the plugin parsed from the API JSON, and mapping it to the plugin type used for display
* modelling the view in the UI as a union type that represents it better, essentially the `ViewPluginsList` case with all the plugins, actions, selections and filters
* using setters to enforce invariants when changing fields together (e.g. selection needs to be reset with a new plugins list), minimize copy-paste and enforce code style
* minimizing data transformation in `View.elm`, there is already fewer logic because the model has changed to be closer to the actual data being displayed (e.g. license errors)

Additional minor changes for the final result include refactoring and adding SCSS :
 
Peek 2025-02-25 10-44

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/elm.json
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/elm.json
@@ -30,6 +30,7 @@
             "elm-community/list-extra": "8.7.0",
             "elm-community/maybe-extra": "5.3.0",
             "elm-community/string-extra": "4.0.1",
+            "elm-explorations/test": "2.2.0",
             "hecrj/html-parser": "2.4.0",
             "isaacseymour/deprecated-time": "1.0.0",
             "jinjor/elm-debounce": "3.0.0",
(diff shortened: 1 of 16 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
special thanks to @mbaechler for the guidance ;)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

RaphaelGauthier (approved):

clarktsiory:
OK, squash merging this PR

PR #6186, opened 2025-02-19
Fixes #26390: API Accounts model needs to be consistent accross backend and frontend 

clarktsiory:
https://issues.rudder.io/issues/26390

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
@@ -1,0 +1,226 @@
+/*
+ *************************************************************************************
+ * Copyright 2025 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.rest.lift
+
+import com.normation.errors.*
+import com.normation.eventlog.ModificationId
+import com.normation.rudder.api.*
+import com.normation.rudder.apidata.ZioJsonExtractor
+import com.normation.rudder.domain.logger.ApiLoggerPure
+import com.normation.rudder.rest.*
+import com.normation.rudder.rest.ApiAccounts as API
+import com.normation.rudder.rest.data.*
+import com.normation.rudder.rest.implicits.*
+import com.normation.rudder.users.UserService
+import com.normation.utils.StringUuidGenerator
+import io.scalaland.chimney.syntax.*
+import net.liftweb.http.*
+import org.joda.time.DateTime
+import zio.syntax.*
+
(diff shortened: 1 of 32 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/utils/src/main/scala/com/normation/utils/DateFormaterService.scala
```diff
@@ -57,37 +58,49 @@ import zio.json.*
 
 object DateFormaterService {
 
-  implicit class JodaTimeToJava(d: DateTime) {
+  extension (self: DateTime) {
+    def toJavaInstant: Instant = Instant.ofEpochMilli(self.getMillis)
+
```
clarktsiory, line 63: taken from https://github.com/Normation/rudder/pull/6768/changes#diff-1639ef728935239f42f82409cc34e8608a284bed16f9c5e33c1d9892af47754b

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
```diff
@@ -245,18 +245,27 @@ class ApiAccountApiServiceV1(
 
   override def regenerateToken(id: ApiAccountId)(using qc: QueryContext): IOResult[ApiAccountDetails] = {
     for {
-      a    <- getPublicApiAccount(id)
-      pair <- mapper.updateToken(a)
-      _    <- writeApi.save(pair._1, ModificationId(uuidGen.newUuid), qc.actor)
-    } yield mapper.toDetailsWithSecret.tupled(pair)
+      a                <- getPublicApiAccount(id)
+      pair             <- mapper.updateToken(a)
+      (account, secret) = pair
```
fanf: you don't need that anymore in scala 3 (the version we use)
`(account, secret) <- getPublicApiAccount(id)`

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_accounts.yml
```diff
@@ -146,7 +146,7 @@ response:
             "status": "enabled",
             "creationDate": "2025-02-10T16:37:19Z",
             "expirationPolicy": "datetime",
-            "expirationDate":"2025-03-10T16:37:19Z",
+            "expirationDate":"2025-03-12T16:37:19Z",
```
fanf: This is extremely fishy. Why do we have a +2h hour here ? It looks like a timezone semantic change - perhaps it's what we want, but it needs to be double check and explain. 
At least, `creationDate + 1 month` is the previous value, not the new one.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
fanf: OK, it's more than fishy: 

image


image


We need to display the time zone in the table, and be consistent - it looks like we are not managing correctly user inputs and their transformation to instant.

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_accounts.yml
clarktsiory: It's not a timezone issue, it's rather a change in semantic, because : 
* "Joda" durations has the semantic of `1 month` = difference between 10th february 2025 and 10th march 2025
* there is no such duration of "months" in ZIO durations, so I put `30 days` (and february is the shortest month, so it reveals that change, with 2days delta : 12th march 2025)

:sweat_smile: it looks like a major change that we wouldn't want, that would be surprising for users.
I can look up for Java time duration instead of ZIO ones...

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_accounts.yml
clarktsiory: i managed to use `java.time.Period.ofMonths`, with the caveat that the `Instant` needs to be converted to zoned or local date time before and converted back after, since instants are dumber and do not know how to reason with months (notably february and leap years :grin:)

clarktsiory:
PR updated with a new commit : spotless was fixed

I will try to make the dates display in the UI have a timezone, and make the datepicker consistent with timezones (which is very hard), in a future PR

clarktsiory:
PR updated with a new commit : fixing a forgotten test

clarktsiory:
OK, squash merging this PR

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_accounts.yml
mbaechler: The rule of thumb is : Instant is not able to do arithmetic, you need a proper timezone to do so.

PR #6187, opened 2025-02-19
Fixes #26396: Test for instance ID is failing on Docker CI

clarktsiory:
https://issues.rudder.io/issues/26396

Just comment the test as it is not critical one.
Also the geteuid in policies can fail, we can remove that one.


amousset (approved):

clarktsiory:
OK, merging this PR

PR #6195, opened 2025-02-21
Fixes #26417: Keep line breaks in reports

RaphaelGauthier:
https://issues.rudder.io/issues/26417

- I've removed the unused css class **.ok** applied on `<td>` element
- I've added a `div.value-container` to "Value" and "Message" cells, on which I've changed the way line breaks are handled, and I've applied a maximum height to it (a scroll bar appears when the content is too long).

line-breaks


RaphaelGauthier:
Commit modified

amousset:
:+1: Would it be possible to also use a mono font when there are several lines?

RaphaelGauthier:
PR updated with a new commit

amousset (approved):

PR #6200, opened 2025-02-25
Fixes #25962: Authorize global parameter names that are not alpha numeric only 

VinceMacBuche:
https://issues.rudder.io/issues/25962

The idea is to bring make global properties (parameters) follow the same character restriction than group/node properties. 
Today, they have fewer chars allowed which means that we can create node or group properties that can't have a global default. 

This change doesn't affect any existing global properties because it only extends the set of allowed character, none of the existing ones are removed. 

The escaped string must be used in any cfengine-related code which want to refer to the parameter, like in https://github.com/Normation/rudder-techniques/pull/1868

:warning: We don't expose the escaped name to user in the UI or wherever easy, because actually we don't want user to refer directly to that parameter, we want them to use the `${node.properties[]}` syntax with the non-escaped name, and the other direct syntax will be removed in Rudder 9.0.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
@@ -219,6 +219,13 @@ final case class ParameterEntry(
     escapedValue:  String,
     agentType:     AgentType
 ) {
+
+  val escapedParameterName = parameterName.replaceAll("""[^\p{Alnum}_]""", "_")
+  // returns the escaped name of the parameter
+  def getEscapedParameterName(): String = {
+    escapedParameterName
+  }
+
   // returns the name of the parameter
   def getParameterName(): String = {
     parameterName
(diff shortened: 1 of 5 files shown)


VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/DataStructures.scala
```diff
@@ -219,6 +219,13 @@ final case class ParameterEntry(
     escapedValue:  String,
     agentType:     AgentType
 ) {
+
+  val escapedParameterName = parameterName.replaceAll("""[^\p{Alnum}_]""", "_")
+  // returns the escaped name of the parameter
+  def getEscapedParameterName(): String = {
```
fanf, line 225: this is never used ?

fanf: It's in https://github.com/Normation/rudder-techniques/pull/1868

VinceMacBuche: It was explained by the comment above that we need this function for string template

fanf:
OK, merging this PR

PR #6202, opened 2025-02-25
Fixes #26434: Make the rudder plugins scriptfiles logs to /var/log/rudder/rudder-package

Fdall:
https://issues.rudder.io/issues/26434


amousset:
Missing info about what we are logging:

* What package operation
* A timestamp

PR #6205, opened 2025-02-25
Fixes #26438: Parent produces errors on stderr and a global error status for the event

amousset:
https://issues.rudder.io/issues/26438

The problem here is that APT (actually dpkg) writes to stdout which breaks CFEngine's protocl. stdout is then closed, hence the pile of broken pipe errors.

Let's just silence apt, we already collect everything we need through other channels.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -962,6 +962,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "filedescriptor"
+version = "0.8.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

Fdall (approved):

amousset:
OK, squash merging this PR

PR #6206, opened 2025-02-26
fix: comparison will always return false due to type mismatch

mbaechler:


clarktsiory:

PR #6210, opened 2025-02-26
refactor: remove shapeless from the codebase as it's not supported in…

mbaechler:
… scala3


clarktsiory:

PR #6212, opened 2025-02-26
refactor: `CompliancePercent` precision is never read

mbaechler:


clarktsiory:

PR #6213, opened 2025-02-26
fix: some case classes were not supposed to be opaque types

mbaechler:


clarktsiory:

PR #6214, opened 2025-02-26
Fixes #26446: Every time webapp starts the logback file gets duplicate XML to configure fixed log file 

VinceMacBuche:
https://issues.rudder.io/issues/26446

The condition on attribute was checking if there was an stdout appender within a if. The condition should be to not have an stdout appender ...

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/endconfig/migration/TestMigrateLogbackXml.scala
+++ b/webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/endconfig/migration/TestMigrateLogbackXml.scala
@@ -194,6 +194,8 @@ class TestMigrateLogbackXml extends Specification with ContentMatchers with Afte
 
   override def afterAll(): Unit = tmpDir.delete()
 
+  sequential
+
   "Properly migrate xml logback file" >> {
     prettyPrinter.formatNodes(migration.migrateLogback(xml)) must beEqualTo(prettyPrinter.format(result))
   }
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/endconfig/migration/TestMigrateLogbackXml.scala
```diff
@@ -211,5 +212,19 @@ class TestMigrateLogbackXml extends Specification with ContentMatchers with Afte
 
     content must beEqualTo(previous_content)
   }
+  "may be ran twice without breaking the logback.xml file" >> {
+
+    val logback8_2_migrated = tmpDir / "logback-8.2-migrated.xml"
+    val previous_content    = logback8_2_migrated.contentAsString
+    val logback8_2          = tmpDir / "logback-8.2.xml"
+    System.setProperty(LOGBACK_CONFIGURATION_FILE, logback8_2.pathAsString)
+    sleep(1000)
```
VinceMacBuche, line 221: Sleep or else it does not like previous test and they write on same backup file

clarktsiory (commented):
LGTM, there is only a probably missing `sequential` in tests, and a unnamed boolean argument that was added, I don't understand what it does and Dotty would not be very happy...
It can be merged after changing that

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/endconfig/migration/TestMigrateLogbackXml.scala
clarktsiory: maybe it's a missing a `sequential` for specs2 to run them one after another ?

VinceMacBuche:
Commit modified

PR #6215, opened 2025-02-27
Fixes #26447: Policy backup and plugins pages should not be available as read-only admin

clarktsiory:
https://issues.rudder.io/issues/26447

The tab system in `Settings` page needs to be refactored, as it is now needed in the `Maintenance` page.
Also the CSS selectors are not similar, we need to be able to insert the tab content at a specific place, instead of just appending it, so we need to specify modes within a `TabUtils`


clarktsiory:
PR updated with a new commit

fanf (approved):
Now works as expected :+1:

clarktsiory:
OK, squash merging this PR

PR #6217, opened 2025-02-27
Fixes #26354: Adding a property column to nodes list causes lines to double height thus screen shows 2 times less nodes

RaphaelGauthier:
https://issues.rudder.io/issues/26354

Now the ‘provider’ label is in absolute position, so has no impact on the display of the property value. Here is the result :
properties-label


clarktsiory (approved):
LGTM !

PR #6218, opened 2025-02-27
Fixes #26330: Inconsistent compliance computation between Directive compliance per Node and Node compliance (directive compliance page is probably wrong)

VinceMacBuche:
https://issues.rudder.io/issues/26330


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):
Nice refactoring, looks like it works as intented. 
We could use `Sub <: ComponentCompliance` I think given our model, but I'm interested in looking how well that works in newer scala.

PR #6225, opened 2025-03-04
Fixes #26260: The "Edit columns" button in the nodes webpage is displayed three times

clarktsiory:
https://issues.rudder.io/issues/26260

Now with the two nodes tables in the same page, we need to select elements within the table wrapper, this consists in prepending the selector of the wrapper to all selectors (most of them are selector on HTML id but jQuery selectors are not strict on uniqueness)


clarktsiory:
PR rebased

RaphaelGauthier (approved):

PR #6226, opened 2025-03-05
Fixes #26478: We do need shapeless version to keep dependency tree correct

fanf:
https://issues.rudder.io/issues/26478


fanf:
It looks like we don't have any other lib than doobie using shapeless, so auto resolution works.

PR #6227, opened 2025-03-05
Fixes #26479: Bad "OnSuccess" delay in log

fanf:
https://issues.rudder.io/issues/26479

Use the actual duration and not the ZIO promise on it. 
Plus that log is important enough to be in `info`, not just debug


clarktsiory (changes requested):
The `d.toString` is not a human-readable format for a Duration, we should use `render`

fanf:
It's a scala one, not a Zio one

PR #6233, opened 2025-03-06
Fixes #26483: Error when trying to add a node property when “Change audit logs” are mandatory

clarktsiory:
https://issues.rudder.io/issues/26483

Fixing the `NodeApi` to take the `reason` in the JSON instead of reading request form params.
I left the validation of the reason length there, we will want to refactor it later, as it is being used also in the "Group categories API" in 8.2 (it has been migrated to zio-json in 8.3 so the check was lost)


clarktsiory:
Testing Github BOT

clarktsiory:
Commit modified

clarktsiory:
Commit modified

fanf (approved):

clarktsiory:
PR rebased

PR #6234, opened 2025-03-06
Fixes #26465: Rejected inventory with `dummy-node-id` when version missing version section in SOFTWARE/RUDDER

fanf:
https://issues.rudder.io/issues/26465

Several biggish change in that PR that I though would be simpler: 

*Enforce the fact that agent version is mandatory.*
It's the case since Rudder 7.0, can't work since Rudder 8.0 and the new node fact where the agent is mandatory and has a non optionnal version. 
- that lead to the discovery that A LOT of our test inventory files are not compliant with these rules, and so a lot of little change in them to add an agent version
- some tests where also in contradiction to those rules and were changed accordingly
- some files were testing compliance with Rudder 4.1 nova agent :rofl: 

A side effect of that is that with that change, we get the correct error message in the `/var/rudder/inventories/failed/xxxx.log` since it is the actual message of the parsing error, and not a message further in the process telling that since there is no version, the serialisation/query for saving is impossible. 

*Add an warning message when there is several agent version from software*

image
 
This doesn't change existing logic, the first one is still taken, but it lets users debugging things know why the version they in Rudder is not the version of the installed agent


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #6237, opened 2025-03-07
Fixes #26504: Score are not cleaned correctly at startup

VinceMacBuche:
https://issues.rudder.io/issues/26504


VinceMacBuche:
Commit modified

fanf (approved):
It looks like it does what it means. A test would have been top.

PR #6242, opened 2025-03-11
Fixes #26512: rudderc generate invalid method call when using methods that are not compatible with Linux 

VinceMacBuche:
https://issues.rudder.io/issues/26512


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

PR #6244, opened 2025-03-12
Fixes 26517: Adds diff to report in template module

m4rtinh4rt:


policies/module-types/template/tests/repport.rs
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: GPL-3.0-or-later
```
amousset, line 1: a single "p" is enough in the file name :smile:

policies/module-types/template/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+        } else {
+            rudder_debug!("Output file {output_file_d} does not exist");
+            false
+        };
+
+        let outcome = match (already_correct, mode) {
+            (true, _) => Outcome::success(),
+            (false, PolicyMode::Audit) => {
+                if already_present {
+                    bail!("Output file {output_file_d} is present but content is not up to date")
```
amousset, line 179: we want a diff here too. We should probably extract the diff itself in a small function.

policies/module-types/template/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+                } else {
+                    bail!("Output file {output_file_d} does not exist")
+                }
+            }
+            (false, PolicyMode::Enforce) => {
+                // Backup current file
+                if already_present {
+                    let backup_file = parameters
```
amousset, line 187: the backup should be extracted too

policies/module-types/template/src/lib.rs
```diff
@@ -0,0 +1,244 @@
...
+                    output,
+                    already_present,
+                ))
+            }
+        };
+        Ok(outcome)
+    }
+}
+
+pub fn repaired_report(
```
amousset, line 218: the numder of parameters / numer of line is not good, this is a sign it is probably not the right abstraction

policies/module-types/template/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+            }
+        };
+        Ok(outcome)
+    }
+}
+
+pub fn diff(old: String, new: String) -> String {
+    let diff = TextDiff::from_lines(&old, &new);
+    let mut unified = diff.unified_diff();
+    return unified.context_radius(10).header("old", "new").to_string();
```
amousset, line 222: I think we might want to use a smaller radius to keep reports small, maybe 2 or 3

amousset: And probably set a hard limit to prevent breaking the reporting stack (for example 10kB), and replace the diff by a message explaining why it is not shown.

policies/module-types/template/src/lib.rs
```diff
@@ -0,0 +1,243 @@
...
+}
+
+fn backup_file(output_file: &Path, parameters: &Parameters) -> Result<(), anyhow::Error> {
```
amousset, line 225: This function does not need the whole `Parameters` struct, it should probably only take the backip_dir as an `&Path`

PR #6247, opened 2025-03-13
Fixes #26515: Missing api account account creation diff element

fanf:
https://issues.rudder.io/issues/26515

- the case for creation time was missing, so adding it
- also changed a little bit UI so that we can see if a token exists or no. It will need more changes because we will want to be able to just delete a token, perhaps think about the look => https://issues.rudder.io/issues/26538
image


fanf:
OK, merging this PR

PR #6250, opened 2025-03-14
Fixes #26348: Timeout on directive save

clarktsiory:
https://issues.rudder.io/issues/26348

Optimizing the lookup for the events to reassign to a node in the directive tree : instead of searching every node to their corresponding events in `O(n)`, we can use a `Map` and use the DOM element ID as key when traversing it.

Querying for the DOM element by the ID is quite fast, and negligible even on the load-testing machine with a thousand directives.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/tree.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/tree.js
@@ -36,26 +36,21 @@
 */
 
 
-  var allEvents = [];
+const allEvents = new Map();
 
-  // Since #24062, we inlined on click events from our lift generated nodes,
-  // there nodes are transformed into a tree using jstree
-  // jstree does not keep the event that were bound to the previous nodes
-  // We need to reattach the events here
-  function attachInlinedEvents( id ) {
-    var child = $(id).find('[id^="lift-event-js-"]'); // all nested child elements with a 'lift-event-js-...' id
-    child.each(function () {
-      var nodeId = this.id;
-      var nodes = allEvents.find(function (x) { return x[0].id === nodeId });
-      if (nodes && nodes[1]) {
-        var event = nodes[1];
-        if (event) {
-          $(this).off("click");
-          $(this).on("click", event.handler);
-        }
-      }
-    });
+// Since #24062, we inlined on click events from our lift generated nodes,
+// there nodes are transformed into a tree using jstree
+// jstree does not keep the event that were bound to the previous nodes
+// We need to reattach the events here
+function attachInlinedEvents( id ) {
+  for (let [id, events] of allEvents) {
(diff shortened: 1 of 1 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/tree.js
```diff
@@ -634,13 +629,15 @@ function openTreeNodes(treeId, settingsStorageId, data){
...
   // Save node event handlers, in order to customize node by reassigning the click event handler
-  var allNodes = $('li.techniqueNode > a[id^="lift-event-js-"], li.techniqueNode > * > span[id^="lift-event-js-"], li.directiveNode > a[id^="lift-event-js-"], li.directiveNode > * > span[id^="lift-event-js-"]');
-  var events =  allNodes.map(function () {
-    return $._data(this, 'events')["click"]; // we need to intercept the click now, the object will change after the jstree init
-  }).get().map(function (x, i) { return [allNodes[i], x] });  // cannot use a node object as key, so keep a [node, events] array
-  allEvents.push(...events);
+  const allNodes = $('li.techniqueNode > a[id^="lift-event-js-"], li.techniqueNode > * > span[id^="lift-event-js-"], li.directiveNode > a[id^="lift-event-js-"], li.directiveNode > * > span[id^="lift-event-js-"]');
+  allNodes.each(function () {
+    const events = $._data(this, 'events')?.click;
+    if (events && events[0]) {
```
VinceMacBuche, line 639: Why checking first element of events ? is the goal to check that events is not empty ? If so why not "events.length !== 0" ?

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/tree.js
```diff
@@ -36,26 +36,21 @@
...
-          $(this).on("click", event.handler);
-        }
-      }
-    });
+// Since #24062, we inlined on click events from our lift generated nodes,
+// there nodes are transformed into a tree using jstree
+// jstree does not keep the event that were bound to the previous nodes
+// We need to reattach the events here
+function attachInlinedEvents( id ) {
+  for (let [id, events] of allEvents) {
```
VinceMacBuche, line 46: it could have been "Map.foreach", but it's ok

clarktsiory: 1. I just tried it and it is confusing, the first argument is the value and the second one is the key :upside_down_face:

clarktsiory:
PR updated with a new commit

fanf (approved):

fanf:
OK, squash merging this PR

PR #6260, opened 2025-03-20
Fixes #26538: Missing API account UI info/actions for token

fanf:
https://issues.rudder.io/issues/26538

*ACL encoding*

The main change is that we replaced ACL `{"path":"xxx", "verb":"get"}` by `{"path":"xxx", "actions":["get"]}`. 
That is what is already used in LDAP backend for account serialization, and it's easier/leaner for rest API. That simplify the mapping and remove one more lift-json occurence. 

In backend, we always group/sort ACL so that we avoid duplicates and API response is predictable. 

We also more directly map for json DTO to business object using chimney partial transform to represent the possibility to fail. 
To make them a bit more usable, I added a mapper from their result/error type to our `PureResult`/`IOResult` types. 

For the elm app, we change the ACL encoding for the communication with scala API, but we also keep the existing encoding because it's what is used in the `api-authorization` plugin, and I prefered to minimize changes. 

*Token API path*

I also changed the token rest command path so that in addition to `regenerate`, we can have `delete`, so that a use can remove a token from an API account to use it with OIDC for example. 
There isn't any UI using it for now, it might be added later next to the regenerate button. 

*API doc menu entries*

Also, correct some naming in API doc in tags leading to duplicated menu entries: 
image


fanf:
PR updated with a new commit

webapp/sources/pom.xml
```diff
@@ -926,6 +926,11 @@ limitations under the License.
       <artifactId>chimney_${scala-binary-version}</artifactId>
       <version>${chimney-version}</version>
     </dependency>
+    <dependency>
+      <groupId>io.scalaland</groupId>
+      <artifactId>chimney-cats_${scala-binary-version}</artifactId>
+      <version>${chimney-version}</version>
+    </dependency>
```
clarktsiory, line 933: am I missing something or is it not used in this PR ?
It seems to be useful for accumulating errors using the `Validated` before converting to our error datatype, instead of directly using the `Result`, as mentioned in the doc : https://chimney.readthedocs.io/en/stable/cookbook/#cats-instances

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonDecoder.elm
```diff
@@ -35,7 +35,7 @@ decodeAccount datePickerInfo =
         |> optional "token" (string |> andThen toToken) Token.ClearText
         |> optional "tokenGenerationDate" (maybe string) Nothing
         |> custom (decodeExpirationPolicy datePickerInfo)
-        |> optional "acl" (map Just (list <| decodeAcl)) Nothing
+        |> optional "acl" (map Just ((list <| decodeAcls) |> map (List.concatMap identity)) ) Nothing
```
clarktsiory, line 38: mix of |> and <| is not a recommended practice and the latter is redundant here
`concatMap identity` also corresponds to `concat` (alias flatten in Scala) : 
```suggestion
        |> optional "acl" (map Just (list decodeAcls |> map List.concat)) Nothing
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
```diff
@@ -215,6 +229,14 @@ class ApiAccountApiServiceV1(
     } yield mapper.toDetailsWithSecret.tupled(pair)
   }
 
+  override def deleteToken(id: ApiAccountId): IOResult[ApiAccountDetails] = {
+    for {
+      a <- readApi.getById(id).notOptional(s"API account with ID '${id.value}' was not found")
```
clarktsiory, line 234: there is something unclear with the `getById` which can also get the system account, we could pass the system account ID and that line would not fail. It would not leak the account, but we wouldn't want to save a new system account

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonEncoder.elm
```diff
@@ -55,19 +55,28 @@ encodeAccount { zone } account =
...
 
 
+-- this one is used to talk to the Rudder API
+-- we don't need to group ACL on the same path for different verbs, the scala part will do that for us.
 encodeAcl : AccessControl -> Value
 encodeAcl acl =
     object
         [ ( "path", string acl.path )
-        , ( "verb", string acl.verb )
+        , ( "actions", list string [acl.verb] ) -- in elm UI, we have only pair of (path, verb) and json, (path, [actions])
```
clarktsiory, line 64: I still have doubts about this API design : sure the API supports merging by `path` so we can always pass singletons like this, or a single path with multiple actions, but would have no ambiguity if the path is a key of the JSON object : 
```json
{
   "/api/account/token/": { "actions": ["GET","POST"] },
   "/api/nodes": { "actions": ["GET","POST"] },
   ...
}
```
But that would again change a lot

webapp/sources/api-doc/paths/usermanagement/activate.yml
```diff
@@ -67,7 +67,7 @@ put:
                 description: Details of the error
                 example: "Could not activate user 'johndoe'; cause was: Inconsistency: User 'johndoe' does not exist therefore cannot be activated"
   tags:
-    - 🧩 User Management
+    - User Management
```
clarktsiory, line 70:
```suggestion
    - User management
```

webapp/sources/api-doc/paths/usermanagement/coverage.yml
```diff
@@ -70,7 +70,7 @@ post:
                         items:
                           type: string
   tags:
-    - 🧩 User Management
+    - User Management
```
clarktsiory, line 73:
```suggestion
    - User management
```

webapp/sources/api-doc/paths/usermanagement/disable.yml
```diff
@@ -67,7 +67,7 @@ put:
                 description: Details of the error
                 example: "Could not disable user 'johndoe'; cause was: Inconsistency: User 'johndoe' does not exist therefore cannot be disabled"
   tags:
-    - 🧩 User Management
+    - User Management
```
clarktsiory, line 70:
```suggestion
    - User management
```

webapp/sources/api-doc/paths/usermanagement/activate.yml
clarktsiory: because this is case-sensitive, there are duplicate menus for both tags otherwise : 
image

clarktsiory (changes requested):
LGTM overall and it is a significant improvement in the API

I had some wonderings about the error handling, a few suggestions to fix the doc and to improve code readability

webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
```diff
@@ -365,6 +366,29 @@ object errors {
...
+
+    private def errorToString(e: Error): String = s"${e.path.asString}: ${e.message.asString}"
+
+    def toPureResult: PureResult[A] = {
+      in match {
+        case Result.Value(value)   => Right(value)
+        case Result.Errors(errors) =>
+          val err = errors.iterator.toList match {
+            case Nil       => // we know that case is impossible, but there isn't anything to change to NonEmptyList
+              Unexpected(s"A chimney NonEmptyErrorsChain was empty, please call developers")
```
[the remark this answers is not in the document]
fanf: Why so? If we want cats data structures, you can use `.asValidatedNec` directly, why adding one hop? That error won't happen, it's an internal lib with a clear message, it's the same level as using a `var` internally for me.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPEntityMapper.scala
```diff
@@ -1161,8 +1151,59 @@ class LDAPEntityMapper(
...
+  implicit val encoderJsonApiAcl: JsonEncoder[JsonApiAcl] = JsonEncoder.list[JsonApiAuthz].contramap(_.acl)
+
+  /*
+   * When we get JsonApiAcl, we always group/sort by path when transforming into business elements
+   */
+  implicit val transformJsonApiAcl: PartialTransformer[JsonApiAcl, List[ApiAclElement]] = {
+    PartialTransformer.apply[JsonApiAcl, List[ApiAclElement]] {
+      case JsonApiAcl(acl) =>
+        partial.Result
+          .traverse(acl.iterator, (x: JsonApiAuthz) => x.transformIntoPartial[ApiAclElement], true)
```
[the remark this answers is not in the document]
fanf: ok

webapp/sources/pom.xml
fanf: it was, and finally I reverted what I wanted to do. Still, since we are starting to use partial transformer, we are using cats more, it looked like it could be usefull and it's minuscule. I can remove it, I've no really strong opinion.

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonEncoder.elm
fanf: that kind of API are horrible to use in practice from a human, command line usage: it complexifies what the user must do to use it compared to predictable kv pairs. Also, it's not clear what happens when you duplicate keys in json, it's implementation dependant and not normalized. 
We can completly model with a map in the business side to ensure our own invariant.

webapp/sources/api-doc/paths/usermanagement/activate.yml
fanf: hum, it's not clear what the convention is. See for ex `Secret Management`
OK for going lower case, it's seems to be the majority

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
fanf: yes, whatever we do with delete, we should be able to delete the system account

webapp/sources/pom.xml
clarktsiory: okay, I am more than fine with leaving the dependency there :cat:

webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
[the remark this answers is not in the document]
clarktsiory: it's just that the API in the case of `Error` has the syntax sugar that seem to be what our own datatype need for instantiation, the one-liner here is :
```scala
      in.asValidatedNel.leftMap(e => Accumulated(e.map(err => Inconsistency(errorToString(err))))).toEither
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
clarktsiory: but in that case it will save a new API account in LDAP, without token but with the same ID as in the in-memory system API account..
It is unrelated to this PR but it should be noted for further changes

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonEncoder.elm
clarktsiory: it is what swagger uses in the JSON definition [https://petstore.swagger.io/v2/swagger.json](for example).
And in fact our [existing public endpoint](https://docs.rudder.io/api/v/20/#tag/API-Info/operation/apiGeneralInformations) to list available API endpoint is closer to that kind of model. Also it gets harder to decide on how to merge fields if we need more fields than `path` and `actions`.

Certainly we could later improve our business model.
And I am still fine with the current model of this PR, since it is only used for changing API access and does not look like it will need other knowledge.

webapp/sources/utils/src/main/scala/com/normation/ZioCommons.scala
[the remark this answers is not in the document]
fanf: ok

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
fanf: no, we should prevent also save from overriding existing accounts, and update to apply on anything but public api account

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonEncoder.elm
fanf: Yep, we did error in the past :) 
Still, it has nothing to do with the merge subject since, really, JSON is perfectly fine with duplicated keys, and the behavior is implementation dependent - ie, we can't rely on the fact that it's a json `{}` to avoid duplicates. That logic need to be purelly defined on our business side.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ApiAccountApi.scala
```diff
@@ -150,46 +163,57 @@ trait ApiAccountApiService {
...
-  }
-
-  override def getAccounts(): IOResult[List[ApiAccountDetails.Public]] = {
-    readApi.getAllStandardAccounts.map(_.toList.map(_.transformInto[ApiAccountDetails.Public]))
-  }
 
-  override def getAccount(id: ApiAccountId): IOResult[Option[ApiAccountDetails.Public]] = {
+  // we need to take care only for PublicApi account, and not of the system account in that API.
+  // We don't want to do anything with it: don't get it, don't update it, don't delete it.
+  // This will be better done once we have a clear topology of ApiAccount with subtypes
```
clarktsiory, line 180: maybe to be added as improvement for #6186

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/JsonDecoder.elm
```diff
@@ -35,7 +35,7 @@ decodeAccount datePickerInfo =
         |> optional "token" (string |> andThen toToken) Token.ClearText
         |> optional "tokenGenerationDate" (maybe string) Nothing
         |> custom (decodeExpirationPolicy datePickerInfo)
-        |> optional "acl" (map Just (list <| decodeAcl)) Nothing
+        |> optional "acl" (map Just (list decodeAcls |> map List.contat)) Nothing
```
clarktsiory, line 38:
```suggestion
        |> optional "acl" (map Just (list decodeAcls |> map List.concat)) Nothing
```

clarktsiory (changes requested):
There is a typo in the Elm `concat` function, and a suggested renaming (optional), it can be merged after fixing that !

fanf:
PR updated with a new commit

fanf:
OK, merging this PR

PR #6266, opened 2025-03-20
Fixes #26579: Name the old event logs page "change logs"

fanf:
https://issues.rudder.io/issues/26579


clarktsiory (approved):
LGTM, it renames everything as expected in the UI

We should keep in mind that there is a technical object, and display object now

fanf:
OK, merging this PR

PR #6272, opened 2025-03-21
Fixes #26584: License information not displayed on “About” page

clarktsiory:
https://issues.rudder.io/issues/26584

The parsing of the license always ended up as `Nothing` because in case of unlimited number of nodes, the API returns `null`.
The model needs to be changed to be the same as in the API

**EDIT:** re-add license `supportedVersion` for copying it to the clipboard as it has been deleted in #6273


clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

fanf (approved):

PR #6273, opened 2025-03-21
Fixes #26590: Make tabs accessible by URL

RaphaelGauthier:
https://issues.rudder.io/issues/26590

We can now access tabs directly using an anchor in the url!

- When pages load, if an anchor is used that matches the ID of a tab on the page, then that tab will automatically be opened.
- This anchor is updated when the user changes tab, meaning that when the page is reloaded, the user is redirected to the last open tab.
- If the anchor does not correspond to any tab on the page, then nothing happens.
- :information_source: There is one exception: the Nodes page. This page already uses the anchor to store search data, so the tab id is stored in the anchor's json object.
 - :warning: This navigation system is not yet functional in Elm applications (such as Rules, Groups, Techniques...).
 

**Example below :arrow_down: :** the URL contains the anchor `#reportsDatabase`, which corresponds to the id of the “Reports database” tab. The user can directly access this tab via the URL `/rudder/secure/administration/maintenance#reportsDatabase`, and if he refreshes the page he will automatically return to this tab.

nav-tab-url

**Next steps :**
- Store tab navigation in history
- Improve node tab names (example: *#node_reports* **⇒** *#tab_compliance*)
- Remove node compliance tab redirection logic that uses the URL parameter `?displayCompliance=True/False`.
- Remove the use of href in the navigation menu of the Hooks page to avoid conflicts
- In node compliance tables, the link 🖊️ should redirect to the compliance tab
- Add redirection `secure/nodeManager/searchNodes#...` ⇒ `secure/nodeManager/nodes#...`

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
@@ -1180,13 +1180,13 @@ function propertyFunction(value, inherited) { return function (nTd, sData, oData
 function callbackElement(oData, displayCompliance) {
   var elem = $("<a></a>");
   if("callback" in oData) {
-      elem.click(function(e) {
-        oData.callback(displayCompliance);
-        e.stopPropagation();
-      });
-      elem.attr("href","javascript://");
+    elem.click(function(e) {
+      oData.callback(displayCompliance);
+      e.stopPropagation();
+    });
   } else {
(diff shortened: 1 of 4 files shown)


RaphaelGauthier:
Commit modified

RaphaelGauthier:
Commit modified

RaphaelGauthier:
Commit modified

RaphaelGauthier:
Commit modified

clarktsiory (approved):
LGTM !
Nice refactoring from `var` to `let`/`const` !

clarktsiory:
OK, merging this PR

PR #6278, opened 2025-03-24
Fixes #26532: Menu color leads to unclarity of meaning

RaphaelGauthier:
https://issues.rudder.io/issues/26532

- I removed the highlight color from the active parent section, which was confusing because it was lighter than all the other menu items
- I've rotated the arrows indicating the open/closed status of the menus by 90°, to be consistent with the other arrows on the pages ( right : closed, down: open)

Here is the result 
menu-colors
:


clarktsiory (approved):
LGTM !
Nice refactoring CSS variable refactoring, and it works perfectly !

fanf (approved):
Lgtm, it seems to address know concerns, we can do other iterations latter on if there's still issues

PR #6274, opened 2025-03-21
Fixes #26545: No compliance on dashboard in Rudder 8.3

clarktsiory:
https://issues.rudder.io/issues/26545

The `RequestVar` needs to be accessed outside of a ZIO for-yield : this is a really surprising behavior :shrug:, maybe specific to the new ZIO version in 8.3  ?

PR #6279, opened 2025-03-24
Fixes #26599: Rudder 8.3 Beta 2 : Missing Score pie charts on dashboard

clarktsiory:
https://issues.rudder.io/issues/26599

The call `scoreService.getAll()` faces the same issue as in #6274 when it calls `nodeFactsRepo.getAll()` : it sees empty nodes when inside the for-yield, but when initialized early it sees the correct list of nodes

**EDIT:**
the `CurrentUser.queryContext` is a also an access of `RequestVar` which is impure state, it needs to be done early oustide of the for-comprehension


fanf (changes requested):
It shouldn't matter. We evaluate an IO in a for, it should work. Else there is an other bug somewhere.

clarktsiory:
Commit modified

fanf:
OK, merging this PR

PR #6280, opened 2025-03-24
Fixes #26605: CurrentUser.queryContext is null when used in a ZIO for

fanf:
https://issues.rudder.io/issues/26605

Purelly mechanic weaving of `(implicit qc: QueryContext)` up to the root of call tree. 

Two things: 
- we could change the signature of plugin dispatch method to inherit the current user of their parent. Not sure it worth it, and it will go in 9.x in any cases, 
- we must not add the implicit param in lift methods that are called from the html, because Lift won't know how to give it to them - that's mainly `dispatch` (which is inherited in any case), but there is some other public method in snippets (`def head` for ex). I hope I missed none.


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/components/ComponentDirectiveEditForm.html
```diff
@@ -1,18 +1,5 @@
 <xml:group>
 
-<component-itemdependencies>
```
fanf, line 3: this is dead code. Removing it for clarity.

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #6286, opened 2025-03-26
Fixes #26620: Renaming tab and parameter for licence info in setup page

ElaadF:
https://issues.rudder.io/issues/26620
It just a renaming of displayed info, variable naming etc was not done to prevent potential issue. 
Typically, in many places, we use `Account` for the `Licence Id` for variable naming or pattern matching, also used in API output, but I think that we don't want to change all of that, but let me know if you want to rename some object.


ElaadF:
Commit modified

fanf (approved):

fanf:
OK, merging this PR

PR #6287, opened 2025-03-26
Fixes #26625: Ubuntu 22.04 agent cannot be accepted on the server

fanf:
https://issues.rudder.io/issues/26625

`AcceptNode` wasn't a `DispatchSnippet` so we had a direct call from the HTML to scala method without indirection. I changed it to dispatch snippet, so that we can refactor without fearing to break thing without compiler error.

Also correct some unrelated typos


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

PR #6288, opened 2025-03-27
Fixes #26544: SearchNodes page in Rudder : Error 404

RaphaelGauthier:
https://issues.rudder.io/issues/26544


clarktsiory (commented):
Works as expected ! 
There is just some variables that could be changed to `const` instead of `let` (the benefit here is just not having to search anywhere because the variable is exactly what it is assigned, if there is naming conflict they could just be renamed for example)

RaphaelGauthier:
PR updated with a new commit

clarktsiory (approved):

PR #6289, opened 2025-03-27
Fixes #26630: Dubius vertical alignement in generic method

RaphaelGauthier:
https://issues.rudder.io/issues/26630

I've tried to fix vertical alignment issues as much as possible using flexbox properties, but some problems seem to be related to the browser font (for example, no specific css rule is applied to the “_A friendly name for this component_” input placeholder, in my firefox and chrome my text is well aligned, but the font seems different from the one on the ticket screenshots).
So I don't know if there's a universal way of fixing these problems...

Here is the result :
Capture d’écran du 2025-03-27 16-54-55


clarktsiory (approved):
LGTM, it solved the policy mode alignment issue for me, the badge are sometimes still misaligned but as you said it depends on the browser and font

PR #6293, opened 2025-03-27
Fixes #26644: Expose techniqueArchiver in RudderConfig and Mock objects

fanf:
https://issues.rudder.io/issues/26644

We need to expose `techniqueArchiver` for security benchmark technique management in `RudderConfig` API. 
Some services weren't correctly alphabetically sorted in the big declaration, so I put them back in place. 
 
Also, we need bulk operation for `TechniqueArchiver` 
- mass add a category and all children techniques, 
- delete a category and all children techniques. 

Plus, we need to parse technique. I didn't want to directly expose `TechniqueParser`, because it only knows about `metadata.xml` (which is what we want for now for benchmark), but in the long time, we would like to have something in common with the archive manager that can know what to do with yaml, json, or xml. So I added a `parseTechnique` in archiver too. 

And add `techniqueArchiver` in mock test object.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
@@ -38,8 +38,11 @@
 package com.normation.rudder.repository.xml
 
 import com.normation.NamedZioLogger
+import com.normation.cfclerk.domain.RootTechniqueCategoryId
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
Merging by hand because typo in ncf and error of webapp is JS timing taking too long.

fanf:
OK, squash merging this PR

PR #6285, opened 2025-03-25
Fixes #26604: Compliance by directive is broken

VinceMacBuche:
https://issues.rudder.io/issues/26604

PR #6294, opened 2025-03-28
Fixes #26646: Upmerge of compliance by directive change breaks 8.3

fanf:
https://issues.rudder.io/issues/26646

So, the change in API output are consistent with the compliance computation evolution from https://github.com/Normation/rudder/pull/6285 and the fact that in 8.3, we added a technique with blocks in test repository.


clarktsiory (approved):
LGTM

PR #6296, opened 2025-04-02
Fixes #26587: Some settings buttons are grey

RaphaelGauthier:
https://issues.rudder.io/issues/26587


fanf (approved):
OK, merging by hand because of unraleted problems

PR #6297, opened 2025-04-02
Fixes #26639: Global properties are unfolded in weird way 

RaphaelGauthier:
https://issues.rudder.io/issues/26639

The css rules that define the height of these table lines have been moved to the wrong scss file. Now it looks like before:
global-properties-result


fanf (approved):
Merging by hand because of error on unralted things

fanf:
OK, merging this PR

PR #6301, opened 2025-04-03
Fixes #26571: Headers does not scale on table size on search node tab

RaphaelGauthier:
https://issues.rudder.io/issues/26571

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
@@ -920,13 +920,16 @@ function initBsTabs(isJsonHash = false){
     triggerEl.addEventListener('click', function (event) {
       event.preventDefault();
       tabTrigger.show();
-
       const newHash = this.getAttribute("data-bs-target");
 
       if (isJsonHash) {
         updateHashString("tab",newHash);
       }else{
         history.replaceState(undefined, undefined, newHash);
+      }
+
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
```diff
@@ -920,14 +920,17 @@ function initBsTabs(isJsonHash = false){
...
 
       if (isJsonHash) {
         updateHashString("tab",newHash);
       }else{
         history.replaceState(undefined, undefined, newHash);
       }
+
+      if (window.location.pathname.includes("nodeManager/nodes")){
+        $("#nodes, #serverGrid").DataTable({"retrieve": true}).columns.adjust().draw();
+      }
```
clarktsiory, line 933: That leads to a duplicate logic for "is the URL the one for the nodes page", in the calling function `initAndCheckTabs`.

The `isJsonHash` already matches the same condition, we could either : 
* rename the `isJsonHash` to `isNodePage`, but the JSON hash could be used elsewhere, in the groups and directives pages
* add another param `adjustServerGrid = false` to be used as condition

clarktsiory (changes requested):

RaphaelGauthier:
PR updated with a new commit

clarktsiory (approved):
LGTM, it fixes the issues and works as expected !

RaphaelGauthier:
OK, merging this PR

PR #6302, opened 2025-04-03
Fixes #26675: Error in documentation /api/changeRequests response data

skaerg:
https://issues.rudder.io/issues/26675


skaerg:
- The list all change request response schema contains a data object that contains a rules array, when data should have been an array whose contents is that of the rules array
- The other request's response schema contained a rules field that should be named changeRequests instead

PR #6305, opened 2025-04-03
Fixes #26671: Confusing warning header on the plugin webpage

fanf:
https://issues.rudder.io/issues/26671

So, the main idea is to expose the token state and make the token optional in elm, to match backen semantic. The token state can be `undef`, `generatedV1` for old token, and `generatedV2` for new format. 

It prefred to enumerate all the state because if we do a v3, it's likely that we will need to change all the place and so I made them easy to find. 

That `tokenState` is transferred by API. 
Then, in elm, we just map that into the token account data structure, and I changed the check for the warning messge to look for token with a `GeneratedV1` token state. 
These token also get the `token-disabled` badge.

Finally, I added a check in the LDAP unserialisation method for the token version, so that we can disable `ApiAccount` right here, even if they were enabled on last save. 

I adapted the test to check that the `tokenState` is correct in API, and that new token are created with the correct state (they were not in tests :sweat: )

Resulting account management page with the warning: 
image

And after token regeneration and enabling again the account: 
image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/View.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/View.elm
@@ -2,11 +2,11 @@ module Accounts.View exposing (..)
 
 import Accounts.ApiCalls exposing (..)
 import Accounts.DataTypes exposing (..)
-import Accounts.DataTypes as Token exposing (..)
+import Accounts.DataTypes as TokenState exposing (..)
 import Accounts.ViewModals exposing (..)
 import Accounts.ViewUtils exposing (..)
 import Html exposing (..)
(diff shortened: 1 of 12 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts/View.elm
```diff
@@ -41,7 +41,7 @@ view model =
                                 [ if hasClearTextTokens then
                                     div [ class "alert alert-warning" ]
                                         [ i [ class "fa fa-exclamation-triangle" ] []
-                                        , text "You have API accounts with tokens generated on an pre-8.1 Rudder versions."
+                                        , text "You have API accounts with tokens generated on an pre-8.1 Rudder versions. "
```
fanf, line 44: missing space between `versions.` and `They`

fanf:
OK, merging this PR

PR #6306, opened 2025-04-03
Fixes #26683: Don't check for expiration nodes that just got new reports

fanf:
https://issues.rudder.io/issues/26683

Ignore nodes which just got new reports in expiration check. 
Also clean methods and event which were never called anywhere in code.


VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #6311, opened 2025-04-04
Fixes #26692: Remove semaphore and use a sliding queue for compliance invalidation

fanf:
https://issues.rudder.io/issues/26692

# Sliding Queue of size nb nodes * 10

Since we have a queue containing a `Seq[T]` of size 1, it's the same a having a queue of `T` of size N. 

In the previous case, the inner `seq` was unbounded, leading to memory exhaustion in worst case. 
I chose a bounded queue to avoid that. It's sliding, ie keeping the most recent events and dropping the older one when full. 

The use of the queue as intented allows to remove the semaphor used up of it, since it's the whole of a queue to be able to add things atomically. 
It also remove a `takeAll` + `merge two array` from the hot path, likely removing some useless translation between collection. 

Then, on the down side, we do the same thing as before, just `takeAll` (getting a seq of T) in place of `take` (getting a seq of T). 

The sizing of the queue is chosen to `10*number of nodes` (with a minimum of 1000 to avoid the case where there is only 1 root node, and then 2000 are added in the next hour). 

# Batch save of lastCompliance

I also added a batch size of the same value as elsewhere (500 by default) for the method which save compliance in `NodeLastCompliance` table, since having very long connection with a lot of data might lead to some exhaustion of resources. 

Based on top of https://github.com/Normation/rudder/pull/6306

Impact: 

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ComputeNodeStatusReportService.scala
@@ -45,6 +45,7 @@ import com.normation.rudder.domain.logger.TimingDebugLoggerPure
 import com.normation.rudder.domain.reports.*
 import com.normation.rudder.domain.reports.NodeStatusReport
 import com.normation.rudder.facts.nodes.ChangeContext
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

VinceMacBuche (approved):

VinceMacBuche:
OK, squash merging this PR

PR #6317, opened 2025-04-09
Fixes #26725: Reverting change in queue since it leads to CPU over use

fanf:
https://issues.rudder.io/issues/26725

The reason of the cpu load is that `queue.takeAll` is non blocking. Which is :exploding_head:


VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #6324, opened 2025-04-10
Fixes #26106: Sometimes when we click on an element (technique or rule), the whole page reloads

clarktsiory:
https://issues.rudder.io/issues/26106

Remove the `href` from links, because the browser could use it to reload the page, instead of running the `onClick` defined in Elm. 
We used the `href` to open in a new tab e.g. with a middle click. Now, as a replacement, we open a new tab using the `auxclick` event : https://developer.mozilla.org/en-US/docs/Web/API/Element/auxclick_event

This occurs in both rules and technique editor.

**EDIT:**
Simple fix : use [onClickPreventDefault](https://package.elm-lang.org/packages/elm-community/html-extra/3.4.0/Html-Events-Extra#onClickPreventDefault) instead of `onClick`

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueList.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueList.elm
@@ -5,6 +5,7 @@ import Either exposing (Either(..))
 import Html exposing (..)
 import Html.Attributes exposing (..)
 import Html.Events exposing (..)
+import Html.Events.Extra exposing (onClickPreventDefault)
 import List.Extra
 import Maybe.Extra
 import NaturalOrdering as N exposing (compare)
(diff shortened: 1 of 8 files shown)


RaphaelGauthier:
This isn't exactly what we want, as we lose the browser's native functions (copy link url, open in new tab, etc.) and moreover, when we open a new object in another tab, this object also opens in the active tab...

The solution seems simpler than that: we need to keep the href attribute, but set a preventDefault on click. This disables the default link opening, but keeps the browser's native functionality.

clarktsiory:
Commit modified

clarktsiory:
Commit modified

RaphaelGauthier:
It's almost perfect, I think there are just a few more cases to fix:
- `Rules/View.elm`, **l.44**
- `Groups/View.elm`, **l.37**
- `GroupRelatedRules/View.elm`, **l.33** and **82**

RaphaelGauthier (approved):

PR #6325, opened 2025-04-11
Fixes #26743: Performance issue on the rule pages

ncharles:
https://issues.rudder.io/issues/26743

This is a wip that traces some improvement (~15%) but it needs a more fundamental refactoring
Sets are *really* expensive, and should be only used when we are certainwe need them

I'm fairly convinced we could dramatically improve the situation, most notably by replacing the Set in the AggregatedStatusReport by Chunk


fanf:
Replaced by #6335

PR #6333, opened 2025-04-17
Fixes #25374: System group description is editable in the UI but cannot be saved

RaphaelGauthier:
https://issues.rudder.io/issues/25374

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
@@ -397,6 +397,7 @@ class NodeGroupForm(
                     </div>
                   </div>
     & "group-cfeclasses" #> NodeSeq.Empty
+    & "#longDescriptionFieldMarkdownContainer i" #> NodeSeq.Empty
     & "#longDescriptionField" #> (groupDescription.toForm_! ++ Script(
       JsRaw(
         s"""setupMarkdown(${Str(groupDescription.defaultValue).toJsCmd}, "longDescriptionField")"""


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
```diff
@@ -399,7 +399,8 @@ class NodeGroupForm(
     & "group-cfeclasses" #> NodeSeq.Empty
     & "#longDescriptionField" #> (groupDescription.toForm_! ++ Script(
       JsRaw(
-        s"""setupMarkdown(${Str(groupDescription.defaultValue).toJsCmd}, "longDescriptionField")"""
+        s"""setupMarkdown(${Str(groupDescription.defaultValue).toJsCmd}, "longDescriptionField")
+           |$$("#longDescriptionFieldMarkdownContainer > label > i").addClass("d-none")""".stripMargin
```
clarktsiory, line 403: should the `groupDescription.toForm_!` not be `groupDescription.readOnlyValue` instead ?

clarktsiory (commented):
Hiding the pencil icon using JS is fine.
But we should keep in mind that this bloats the page, we could instead simplify the server-rendered HTML by removing the elements with Lift (I have not tested this code) : 

```scala
& "#longDescriptionFieldMarkdownContainer i" #> NodeSeq.Empty
```

In fact, some other elements could also be removed since the markdown are not used, I found that in the [HTML template](https://github.com/Normation/rudder/blob/branches/rudder/8.2/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/components/NodeGroupForm.html#L77) : 
```scala
& "#longDescriptionFieldContainer" #> NodeSeq.Empty
```

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala
clarktsiory: Maybe not, since this is a markdown field, it may need to be rendered as a form with the script.
There is an alternative to hiding the icon, see the comment below

RaphaelGauthier:
PR rebased

RaphaelGauthier:
PR updated with a new commit

clarktsiory (approved):
LGTM, it works as expected : the icon is not rendered.
It seems some other containers cannot be removed because the markdown lib may need them to display the description

clarktsiory:
OK, squash merging this PR

PR #6334, opened 2025-04-17
Fixes #26767: 

clarktsiory:
https://issues.rudder.io/issues/26767

For the Api account, the token is already redacted, we need the same for the user password

* from the strict type, avoid invalid representation of user (empty hash) : 
   * we know what to do with empty (invalid ?) hashes : they are for external users, so mark it as specific type
   * we know valid hashes
   * we also know how to generate random ones 

**TODO:** the password seems to be used to `getPassword` (override Spring abstract method) and as `""` value when user credentials is to be overridden by a plugin


clarktsiory:
PR updated with a new commit

clarktsiory:
PR replaced by https://github.com/Normation/rudder/pull/6515

PR #6335, opened 2025-04-17
Fixes #26712: When directives are skipped, they are multiplied in the directive tab of the rule

fanf:
https://issues.rudder.io/issues/26712

So, this is a big, overdue clean-up of the skipped mess that organically growed during the years. 

It shoudl correct all `skipped` problem we got, including https://issues.rudder.io/issues/26712 and https://issues.rudder.io/issues/24843. 

It also lead to a MASSIVE performance improvent, that looks quadratic for number of node or number or rules, and lead to at least an order of magnitude quicker computation of compliance for rules. 

> Note: this PR is rather low risk since it only change the computation of skipped directives in API/UI. I thing that at worst, we get new problems with skipped directive. 

## Architecture: keep skipped directive in NodeStatusReport with an "overridden" info

The only idea is that we should really just keep the skipped directives in `NodeStatusReport` with a tag "I'm overridden by that rule`. It's really cheap to have the directive added at that point, and it's very cheap to derive what is skipped where if we have that information AT THE DIRECTIVE LEVEL. 

Onve we have that, since we do have real directive object (just with a marker), we: 
- don't have to try to add them latter on ,
- don't need to build complicated logic to retrieve what is overridden where. 

To do so, I chose to add a new attribute `overridden: Option[RuleId]` in `DirectiveStatusReport`. This has the benefits of being fully compatible with all the existing JSON for `NodeStatusReport` serialization and is simple. 

> Note: the saved `NodeStatusReport` in base will be false wrt to skipped directive. That will be corrected at the first generation, which will happen on restart. The only case I can think where it could lead a missing "skipped" info is for nodes in "keep compliance because the node is not responding". It looks OK. 

Its problem is that it's not very obvious that a directive is overridden with just that, and I explored two other options: 

- have two types, `DirectiveNodeStatus` and `OverriddenDirectiveNodeStatus` which is obvious and make possible JSON customization for the overridden part. But it would have lead to (even more) changes, and I feared breaking things a lot. 

- add a `PolicyMode`: `skipped`, since we latter have a `ComputedPolicyMode`. But it doesn't make sense in second though, since the policy mode is not computed in `NodeStatusReport`. It *could* make sense to do so, since we have all the information, and actually, it looks like something we want to do. But I believe it belongs to another refactoring where we move all the horrible logic in `ComplianceAPIService` into real business objects with the knowledge to map and remap compliance from a point of view of node, rule, directive, or group. This was a HUGE step I didn't want to take here. 

## Performance analysis

To understand what was happening, I used `async-profiler` in a dedicated autonomous test that can be run by hand in `RunComplianceComputation`. 

This is a really nice tool that helped me spot the big performance problem thanks to flamegraph. 

On a 5000 nodes / 300 rules tests, I got from 22 seconds to 2 seconds. On a 10k nodes / 500 rules, from several minutes to < 10 s. 

One big part of the graph just disappeared: 
image

## Consequences: 

- everything is simpler and just so much more logical
- `overrides: List[OverriddenPolicy]` disappear from `NodeStatusReport` and `RuleStatusReport`. 
- performance are EXTREMELLY improved. By order of magnitude. 

## Next/TODO in a futur PR

There is so much complicated logic in `complianceAPIService` and so little tests that there MUST be bugs there. 
The next big step is to clean that up: 
- isolated the logic for the different merge parts, which is the definition of business rules that need to be tested
- make the transformation pure
- call that new service from the existing one.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -64,7 +64,6 @@ import com.normation.rudder.domain.reports.ComplianceLevel
 import com.normation.rudder.domain.reports.CompliancePrecision
 import com.normation.rudder.domain.reports.ComponentStatusReport
 import com.normation.rudder.domain.reports.DirectiveStatusReport
-import com.normation.rudder.domain.reports.NodeStatusReport
 import com.normation.rudder.domain.reports.ValueStatusReport
 import com.normation.rudder.facts.nodes.CoreNodeFact
 import com.normation.rudder.facts.nodes.NodeFactRepository
@@ -81,7 +80,6 @@ import com.normation.rudder.rest.RestExtractorService
 import com.normation.rudder.rest.RestUtils.*
 import com.normation.rudder.rest.data.*
 import com.normation.rudder.services.reports.ReportingService
-import com.normation.rudder.services.reports.ReportingServiceUtils
 import com.normation.rudder.web.services.ComputePolicyMode
 import com.normation.rudder.web.services.ComputePolicyMode.ComputedPolicyMode
 import com.normation.zio.currentTimeMillis
@@ -699,20 +697,34 @@ class ComplianceAPIService(
 
                   // if rule cannot be found in "rules" it means the level prevent from returning rule details, so : no compliance
                   rules.find(_.id == ruleId).map { rule =>
-                    val nodeIds    = RoNodeGroupRepository
+                    val nodeIds = RoNodeGroupRepository
                       .getNodeIdsChunk(allGroups, rule.targets, nodeFacts.mapValues(_.rudderSettings.isPolicyServer))
                       .toSet
-                    val policyMode = getRulePolicyMode(
-                      rule,
-                      allDirectives,
-                      nodeIds.toSet,
-                      nodeFacts.mapValues(_.rudderSettings).toMap,
-                      globalPolicyMode
+
+                    def defaultMode                   = (
+                      None,
+                      getRulePolicyMode(
+                        rule,
+                        allDirectives,
+                        nodeIds,
+                        nodeFacts.mapValues(_.rudderSettings).toMap,
+                        globalPolicyMode
+                      )
                     )
+                    // if all directive on that rules are skipped, the rule is skipped too
+                    val (overrideDetails, policyMode) = if (ruleDirectiveReports.forall(_._2.overridden.nonEmpty)) {
+                      ruleDirectiveReports.collectFirst {
+                        case (_, DirectiveStatusReport(_, _, Some(rid), _)) =>
+                          val rname = rules.collectFirst { case r if r.id == rid => r.name }.getOrElse("unknown")
+                          (Some(SkippedDetails(rid, rname)), ComputePolicyMode.skippedBy(rid, rname))
+                      }.getOrElse(defaultMode)
+                    } else defaultMode
+
                     ByDirectiveByRuleCompliance(
                       ruleId,
                       rule.name,
                       ComplianceLevel.sum(componentsCompliance.map(_.compliance)),
+                      overrideDetails,
                       policyMode,
                       componentsDetails
                     )
@@ -749,6 +761,7 @@ class ComplianceAPIService(
     for {
       t1        <- currentTimeMillis
       allGroups <- nodeGroupRepo.getAllNodeIdsChunk()
+      allRules  <- rulesRepo.getAll()
       t2        <- currentTimeMillis
       _         <- TimingDebugLoggerPure.trace(s"getByRulesCompliance - nodeGroupRepo.getAllNodeIdsChunk in ${t2 - t1} ms")
 
@@ -778,10 +791,6 @@ class ComplianceAPIService(
       reportsByRule = reportsByNode.flatMap { case (_, status) => status.reports.flatMap(_._2.reports) }.groupBy(_.ruleId)
       t7            = System.currentTimeMillis()
       _            <- TimingDebugLoggerPure.trace(s"getByRulesCompliance - group reports by rules in ${t7 - t6} ms")
-
-      // make a map of directive overrides for each rule, to add to the directives of a rule
-      directivesOverrides <-
-        getDirectiveOverrides[ByRuleDirectiveCompliance](ruleObjects, reportsByNode, directives, _.toComplianceByRule(_))
 
       t8               <- currentTimeMillis
       _                <- TimingDebugLoggerPure.trace(s"getByRulesCompliance - get directive overrides and rules infos in ${t8 - t7} ms")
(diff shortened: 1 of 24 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RunCompianceComputation.scala
clarktsiory, line 1: Its name has a typo (it should have been `RunComplianceComputation`), 
but also the only test in this file is named `RunTestCompliance`, for discoverability should it not be `TestRunComplianceComputation` and the file name too ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -699,20 +697,34 @@ class ComplianceAPIService(
...
+                        nodeIds,
+                        nodeFacts.mapValues(_.rudderSettings).toMap,
+                        globalPolicyMode
+                      )
                     )
+                    // if all directive on that rules are skipped, the rule is skipped too
+                    val (overrideDetails, policyMode) = if (ruleDirectiveReports.forall(_._2.overridden.nonEmpty)) {
+                      ruleDirectiveReports.collectFirst {
+                        case (_, DirectiveStatusReport(_, _, Some(rid), _)) =>
+                          val rname = rules.collectFirst { case r if r.id == rid => r.name }.getOrElse("unknown")
```
clarktsiory, line 718: isn't that a `find` ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -699,20 +697,34 @@ class ComplianceAPIService(
...
+                          (Some(SkippedDetails(rid, rname)), ComputePolicyMode.skippedBy(rid, rname))
+                      }.getOrElse(defaultMode)
+                    } else defaultMode
```
clarktsiory, line 721: The `defaultMode` is only used in certain cases but is eagerly evaluated here.
Also for testability, maybe this business logic should be defined as a separate method, even in the companion object

clarktsiory: I see that `computeOverridingMode` does almost the same thing, so it could be a method next to that one

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -970,20 +1001,33 @@ class ComplianceAPIService(
...
+                      val (overrideDetails, policyMode) = if (nodeDirectives.forall(_._2.overridden.nonEmpty)) {
+                        nodeDirectives.collectFirst {
+                          case (_, DirectiveStatusReport(_, _, Some(rid), _)) =>
+                            val rname = rules.get(rid).map(_._1.name)
+                            (
+                              Some(SkippedDetails(rid, rname.getOrElse("unknown"))),
+                              ComputePolicyMode.skippedBy(rid, r.name)
+                            )
+                        }.getOrElse(defaultMode)
+                      } else defaultMode
```
clarktsiory, line 1023: Same as above, there may be unwanted computations because of defaultMode

webapp/sources/rudder/rudder-web/src/main/elm/sources/DirectiveCompliance/ViewUtils.elm
```diff
@@ -48,6 +48,7 @@ badgePolicyMode globalPolicyMode policyMode =
           This rule is applied on at least one node or directive that will <b style='color:#9bc832;'>enforce</b>
           one configuration, and at least one that will <b style='color:#3694d1;'>audit</b> them.
           """
+        "skipped"   -> "<div style='margin-bottom:5px;'>This rule is in <b style='color:#eda800;'>skipped</b> mode.</div>"
```
clarktsiory, line 51: Is the explanation of that "because all its directives are skipped" ?

clarktsiory (commented):
I had a few suggestions and questions, other than that the change looks okay and I will try to test it for real

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
fanf: `collectFirst` is faster than `find(..).map()`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
fanf: For the eager evaluation, yes, could be `def`

yes, see last part of the PR description about all the business logic that need to get elsewhere, but not in that one

webapp/sources/rudder/rudder-web/src/main/elm/sources/DirectiveCompliance/ViewUtils.elm
fanf: yes

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RunCompianceComputation.scala
fanf: it's not a test in the sense where the testing framework must ignore it. If we put `Test` at the beginning or end of the name, the test framework will try to Do Things, resulting in Not What We Want. 
The discoverability is discutable, because appart if you want to autonomously run that - and so, you're exactly looking for running computation on compliance -, it is rather useless. 

Actually, it is almost always a bad idea to that kind of dead-code in a project, because it tend to drift away from the code base and just add maintenance burden for no gain, because nobody reuse it. I wanted to let it, because it demonstrates interesting things that were complicated :tm: to set up right for the flamegraphes. 
I'm pretty sure in a couple years, we will just delete it without much more gain - but at least, for a time, the "how to autonomously get a flamegraphe with `async-agent` will exist".

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RunCompianceComputation.scala
```diff
@@ -0,0 +1,634 @@
...
+    // and default parameters. The flamegraph will be written in html in `file`.
+    // See https://github.com/async-profiler/async-profiler/blob/v4.0/src/arguments.cpp
+    // for more parameters. You can look for memory allocation, time spent in tree form, etc.
+    profiler.execute(s"start,event=cpu,tree,file=${file}")
+    println(s"computing compliance...")
+
+    // This is the
+    /////////////// -- ///////////////
+    val (t, x) = cs.getRulesCompliancePure(Some(1))(QueryContext.systemQC).timed.runNow
+    /////////////// -- ///////////////
```
clarktsiory, line 115: The comment is unfinished

fanf:
PR updated with a new commit

clarktsiory (changes requested):

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
clarktsiory: okay I also though at first glance of all the duplicate logic : we should really take some time to reorganize all that later :grin:

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
clarktsiory: true :+1:

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM, I tried on existing skipped directives and rules and saw no change in behavior.
The "skipped" tooltip badge will yet need to be fixed later

fanf:
OK, squash merging this PR

PR #6337, opened 2025-04-24
Fixes #26792: Update scala dependencies

VinceMacBuche:
https://issues.rudder.io/issues/26792

### DIFF
--- a/webapp/sources/pom.xml
+++ b/webapp/sources/pom.xml
@@ -406,45 +406,45 @@ limitations under the License.
          parser option - see https://github.com/scala/scala-xml/releases/tag/v2.1.0 -->
     <scala-xml-version>1.3.0</scala-xml-version>
     <lift-version>4.0.0-M1</lift-version>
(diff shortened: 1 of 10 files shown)


fanf:
there is a spotless problem, but I wonder if we should not clean-up the `ruddersysevent` insertion logic as a whole before that update, because we don't use it for a long time now.

fanf (changes requested):
Tests don't pass, plus perhaps we want to just remove some DB code first

VinceMacBuche:
Commit modified

VinceMacBuche:
insertReports is only used in tests, maybe it's not a problem to keep it here.

fanf (commented):
Puting the DB code in the test file in: https://github.com/Normation/rudder/pull/6352

VinceMacBuche:
PR rebased

VinceMacBuche:
PR rebased

PR #6346, opened 2025-04-30
Fixes #26847: A user that has only the "deployer_read" right does not have access to the settings API

skaerg:
https://issues.rudder.io/issues/26847


clarktsiory (approved):
LGTM, I assume that since `Validator.Read` has access to `enable_self_validation` and `enable_self_deployment`, `Deployer.Read` should too

clarktsiory:
OK, merging this PR

PR #6348, opened 2025-04-30
Fixes #26852: Rest API Error 500 when not authorized

VinceMacBuche:
https://issues.rudder.io/issues/26852


fanf (approved):
Nice

fanf:
I'm making the missing spotless. 

So the main idea is to add the `returnCode` in our error class for HTTP, and put the correct ones for each error case.

fanf:
Merging by hand because of error in relayd.

PR #6350, opened 2025-05-02
Fixes #26853: Migration from Box to ZIO : Refactoring of classes ChangeRequest and ModificationValidationPopup

skaerg:
https://issues.rudder.io/issues/26853

This is part of the plan to remove the dependency to the lift library (replacing the `Box` type with `ZIO`).

In particular, the change-validation plugin contained a call to `BoxToIO` in `com.normation.plugins.changevalidation.api.ChangeRequestApi`, which required that the return type of `com.normation.rudder.domain.workflows.ChangeRequest#change` be changed. Hence, the return type is now `PureResult[T_CHANGE]` (previously `Box[T_CHANGE]`).

This PR is linked to #6353 , which uses change, as well as https://github.com/Normation/rudder-plugins/pull/827 , which contains the corresponding change (replaces the call to `BoxToIO` with `PureResultToIO` in `ChangeRequestApi`), and removes most of the remaning instances of the `Box` type in change-validation.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
@@ -320,7 +320,7 @@ class CommitAndDeployChangeRequestServiceImpl(
       }.toBox
 
       for {
-        change <- directiveChanges.changes.change
+        change <- directiveChanges.changes.change.toBox
         diff   <- change.diff match {
                     case DeleteDirectiveDiff(tn, d)       =>
                       dependencyService
@@ -340,7 +340,7 @@ class CommitAndDeployChangeRequestServiceImpl(
 
     def doNodeGroupChange(change: NodeGroupChanges)(implicit qc: QueryContext): Box[TriggerDeploymentDiff] = {
       for {
-        change <- change.changes.change
+        change <- change.changes.change.toBox
         diff   <- change.diff match {
                     case DeleteNodeGroupDiff(n)   =>
                       dependencyService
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -443,7 +443,10 @@ class CommitAndDeployChangeRequestServiceImpl(
             case (_, _: AddGlobalParameterDiff)                               => false
             case _                                                            => true
           }
-        }).openOr(true)
+        }) match {
+          case Right(b) => b
+          case Left(_)  => true
+        }
```
clarktsiory, line 449: that' a `getOrElse(true)`

skaerg:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
```diff
@@ -608,7 +613,7 @@ class ModificationValidationPopup(
     }
 
     (for {
-      cr <- boxcr
+      cr <- purecr
       id <- workflowService
               .startWorkflow(cr)(
                 ChangeContext(
```
clarktsiory, line 619: that is not the usual effect type for the for-comprehension : we have `IOResult` as a result from `workflowService.startWorkflow(cr)`, so we likely need to do `purecr.toIO` instead

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/ModificationValidationPopup.scala
```diff
@@ -620,20 +625,16 @@ class ModificationValidationPopup(
...
-        onSuccessCallback(id)
-      case eb: EmptyBox =>
-        val e = (eb ?~! "Error when trying to save your modification")
-        logger.error(e.messageChain)
-        e.chain.foreach { ex =>
-          parentFormTracker.addFormError(error(ex.messageChain))
-          logger.error(s"Exception when trying to update a change request:", ex)
-        }
-        onFailureCallback(Text(e.messageChain))
+    }).chainError("Error when trying to save your modification") match {
```
clarktsiory, line 632: once we do `purecr.toIO`, we will have the usual `chainError(...).either.runNow`

clarktsiory (changes requested):
Almost there, most changes look ok, only some nitpicking

skaerg:
Commit modified

PR #6353, opened 2025-05-04
Fixes #26866: Port remaining lift-json into zio-json

fanf:
https://issues.rudder.io/issues/26866


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
I'm not able to unit-test the missing API, the whole thing is a mess and should be rewritten. It mixes IO with data, dynamic json construct, non standard API patterns, and grew organically without the needed tiddy-up. 

I tested by hand and added some more serialization test and put the lowest hanging fruit appart.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -203,20 +190,35 @@ class TechniqueApi(
...
+            case e: Throwable =>
+              Left(
+                Inconsistency(
+                  s"Parsing request parameter '${key}' as a boolean failed, current value is '${head}'. Error message is: '${e.getMessage}'."
+                )
+              )
+          }
+        case Some(list)        => Left(Inconsistency(s"${list.size} values defined for 'id' parameter, only one needs to be defined"))
+      }
+    }
```
clarktsiory, line 221: The previous implementation in `RestExtractorService` was also looking into the JSON, but here the HTTP method is DELETE, it looks like we should not support JSON body at all, so it is OK.

So, the only usage is for the `force` param, but it is not documented yet : https://docs.rudder.io/api/v/21/#tag/Techniques/operation/deleteTechnique, and should be documented as a query param

clarktsiory (approved):
LGTM ! Very nice change !

PR #6357, opened 2025-05-05
Fixes #26861: Migration from Box to ZIO : Refactor XmlUnserialisation 

skaerg:
https://issues.rudder.io/issues/26861

This is part of the plan to remove the dependency to the lift library from Rudder and Rudder plugins.
This PR migrates the following classes in XmlUnserialisation : 
- DirectiveUnserialisationImpl
- NodeGroupCategoryUnserialisationImpl
- NodeGroupUnserialisationImpl
- RuleUnserialisationImpl
- RuleCategoryUnserialisationImpl
- GlobalParameterUnserialisationImpl
- ChangeRequestChangesUnserialisationImpl

This change also includes a new `XmlUtils` object, which offers methods to look up a specific child node or attribute of a given XML node, and the error handling that comes with it. XmlUtils may be used in order to migrate the remaining classes in XmlUnserialisation

This PR is part of the migration from Box to ZIO in change-validation, and is linked to the following PR in the Rudder plugins repository : https://github.com/Normation/rudder-plugins/pull/828

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
@@ -37,6 +37,7 @@
 
 package com.normation.rudder.services.marshalling
 
+import cats.implicits.*
 import com.normation.GitVersion
 import com.normation.GitVersion.ParseRev
 import com.normation.box.*
@@ -45,6 +46,10 @@ import com.normation.cfclerk.domain.TechniqueName
 import com.normation.cfclerk.domain.TechniqueVersion
 import com.normation.cfclerk.services.TechniqueRepository
 import com.normation.cfclerk.xmlparsers.SectionSpecParser
+import com.normation.errors.AccumulateErrors
+import com.normation.errors.Inconsistency
+import com.normation.errors.OptionToPureResult
+import com.normation.errors.PureResult
 import com.normation.eventlog.EventActor
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.api.*
@@ -72,6 +77,7 @@ import com.normation.rudder.services.queries.CmdbQueryParser
 import com.normation.utils.Control.traverse
 import net.liftweb.common.*
 import net.liftweb.common.Box.*
+import net.liftweb.util.Helpers.tryo
 import org.apache.commons.text.StringEscapeUtils
 import org.joda.time.format.ISODateTimeFormat
 import scala.util.Failure as Catch
@@ -92,22 +98,23 @@ final case class XmlUnserializerImpl(
 
 class DirectiveUnserialisationImpl extends DirectiveUnserialisation {
 
-  override def parseSectionVal(xml: NodeSeq): Box[SectionVal] = {
-    def recValParseSection(elt: XNode): Box[(String, SectionVal)] = {
-      if (elt.label != "section") Failure("Bad XML, awaiting a <section> and get: " + elt)
+  override def parseSectionVal(xml: NodeSeq): PureResult[SectionVal] = {
+    def recValParseSection(elt: XNode): PureResult[(String, SectionVal)] = {
+      if (elt.label != "section") Left(Inconsistency(s"Bad XML, expected a <section> but got: ${elt}"))
       else {
         for {
-          name     <- (elt \ "@name").headOption ?~! ("Missing required attribute 'name' for <section>: " + elt)
+          name     <- (elt \ "@name").headOption.notOptionalPure(s"Missing required attribute 'name' for <section>: ${elt}")
           // Seq( (var name , var value ) )
-          vars     <- traverse(elt \ "var") { xmlVar =>
+          vars     <- (elt \ "var").accumulatePure { xmlVar =>
                         for {
-                          n <- (xmlVar \ "@name").headOption ?~! ("Missing required attribute 'name' for <var>: " + xmlVar)
+                          n <-
+                            (xmlVar \ "@name").headOption.notOptionalPure(s"Missing required attribute 'name' for <var>: ${xmlVar}")
                         } yield {
                           (n.text, xmlVar.text)
                         }
                       }
           // Seq ( SectionVal )
-          sections <- traverse(elt \ "section")(sectionXml => recValParseSection(sectionXml))
+          sections <- (elt \ "section").accumulatePure(sectionXml => recValParseSection(sectionXml))
         } yield {
           val s = sections.groupBy { case (n, s) => n }.map { case (n, seq) => (n, seq.map { case (_, section) => section }) }
           (name.text, SectionVal(s, vars.toMap))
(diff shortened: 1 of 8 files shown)


skaerg:
Commit modified

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -143,41 +152,51 @@ class DirectiveUnserialisationImpl extends DirectiveUnserialisation {
...
-                            tryo(s.text.toBoolean)
-                          ) ?~! ("Missing attribute 'isEnabled' in entry type directive : " + xml)
-      priority         <- (directive \ "priority").headOption.flatMap(s =>
-                            tryo(s.text.toInt)
-                          ) ?~! ("Missing or bad attribute 'priority' in entry type directive : " + xml)
-      isSystem         <- (directive \ "isSystem").headOption.flatMap(s =>
-                            tryo(s.text.toBoolean)
-                          ) ?~! ("Missing attribute 'isSystem' in entry type directive : " + xml)
-      policyMode        = (directive \ "policyMode").headOption.flatMap(s => PolicyMode.parse(s.text).toBox)
+      fileFormatOk     <- TestFileFormat(directive).toPureResult
```
clarktsiory, line 161: you could create a method that implements the check in another method in `TestFileFormat`. Also, by looking at the implementation, it is only used to do a check, so maybe :   

```scala
object TestFileFormat {
  def check(xml: Node, fileFormat: String = Constants.XML_CURRENT_FILE_FORMAT.toString): PureResult[Unit] = {
    if (xml.attribute("fileFormat").map(_.text) == Some(fileFormat)) Right(())
    else Left(Inconsistency("Bad fileFormat (expecting ${fileFormat}): ${xml}"))
  }
}

**Comment by skaerg** [comment]
PR updated with a new commit

**Comment by skaerg** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala [hunk 2]
```diff
@@ -213,7 +213,7 @@ class CommitAndDeployChangeRequestServiceImpl(
       def getCurrentValue(rule:  Rule) = roRuleRepository.get(rule.id).toBox
       def compareMethod(initial: Rule, current: Rule) = CheckDivergenceForMerge.compareRules(initial, current)
       def xmlSerialize(rule:     Rule): Box[Node] = Full(xmlSerializer.rule.serialise(rule))
-      def xmlUnserialize(xml:    Node) = xmlUnserializer.rule.unserialise(xml)
+      def xmlUnserialize(xml:    Node) = xmlUnserializer.rule.unserialise(xml).toBox
```
clarktsiory, line 216: The interfaces defined in `CommitAndDeployChangeRequestServiceImpl#isMergeableConfigurationChangeRequest` are all used for the feature of "change validation".
They should also be all migrated to an `IOResult` API : 

```scala
    trait CheckChanges[T] {
      // Logging function
      def failureMessage(elem:     T): String
      // Find Current value from initial value
      def getCurrentValue(initial: T): PureResult[T]
      // Function that validates that the change request is still valid
      def compareMethod(initial:   T, current: T): Boolean
      // How to serialize the data into XML
      def xmlSerialize(elem:       T): PureResult[Node]
      // How to unserialize XML into the datatype
      def xmlUnserialize(xml:      Node): PureResult[T]
      // Transform date from LDAP to XML then transform it back to data using the XML parser
      def normalizeData(elem: T): PureResult[T] = {
        ...
      }

      // Check if the change request is mergeable
      def check(initialState: Option[T]): IOResult[Boolean] = {
          ...
       }
```

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/marshalling/TestXmlUnserialisation.scala
```diff
@@ -133,12 +131,10 @@ class TestXmlUnserialisation extends Specification with BoxSpecMatcher {
...
       changeRequestChangesUnserialisation.unserialise(change) match {
-        case f: Failure =>
-          val msg = s"I wasn't expecting the failure: ${f.messageChain}"
-          f.rootExceptionCause.foreach(ex => ex.printStackTrace())
+        case Left(err) =>
+          val msg = s"I wasn't expecting the failure: ${err.fullMsg}"
           ko(msg)
-        case Empty => ko(s"Unexpected Empty!")
-        case Full(_) => ok("unserialization was a success")
+        case Right(_)  => ok("unserialization was a success")
```
clarktsiory, line 137: This could simply be `changeRequestChangesUnserialisation.unserialise(change) must beRight`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -874,16 +908,114 @@ class ApiAccountUnserialisationImpl extends ApiAccountUnserialisation {
...
+    (entry \ label).headOption
+      .map(f)
+      .notOptionalPure(missingChildErrMsg(label, entryType, entry))
+  }
+
+  def flatMapChildPure[A](
+      entry: XNode,
+      label: String,
+      f:     XNode => Option[A]
+  )(implicit entryType: XmlEntryType): PureResult[A] = {
```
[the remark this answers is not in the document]
clarktsiory: I even think that it would more fluent to read when always opting for `getChild`, even if there is a transformation on the child :
    * `getChildPure(node, "id")` -> `getChild(node, "id")`
    * `mapChildPure(node, "id", _.text)` -> `getChild(node, "id", _.text)`
    * `getChildPureRec(node, "id")` -> `getChildRec(node, "id")`
    * etc.
    * `flatMapChildPure(node, "id", _.text.toIntOption)` -> `getAndTransformChild(node, "id", _.text.toIntOption)`

`getChildPure` is a specialization of `mapChildPure` but without any mapping function, you can define two functions with the same name but different number of arguments, and the JVM will allow that thanks to method polymorphism.

I prefer using polymorphism because : 
* it is easier to find declinations in the IDE
* there are not too many method names in the API
* the intent to apply a transformation can always be specified with explicit argument names : `getChild(..., transform = _.text)`

Also, choosing "XML" semantics instead of "Scala map/flatMap" makes XML parsing more readable : get child, transform them (instead of map/flatMap)
Unfortunately to replace the `flatMap`, polymorphism does not allow us to have two methods having the same name with `f: XNode => A` (map) and `f: XNode => Option[A]`, since it does not support generics very well, so the compiler will be confused about `Function1[XNode, A]` and `Function1[XNode, Option[A]]`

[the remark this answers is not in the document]
clarktsiory: Also, a personal preference (that would change the PR a lot) : the implicit class syntax could be used here, for XML I find the object oriented syntax more suited that function calls.

I find `directive.getChild("id", _.text)` more fluent than `mapChild(directive, "id", _.text)` (yes, I am still in the object-oriented vibe...)

We could of course expose both functions and syntax by simply adding : 
```scala
implicit class XmlUtilsSyntax(node: XNode) {
  def getChild[A](
      label: String,
      f:     XNode => A
  )(implicit entryType: XmlEntryType): PureResult[A] = ...
  /* etc. for each method */
}

File: webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala [hunk 5]
```diff
@@ -622,145 +623,178 @@ class ChangeRequestChangesUnserialisationImpl(
...
+            changeRule   <- getChildPureTwo(rule, "firstChange", "change")
+            actor        <- mapChildPureRec(changeRule, "actor", actor => EventActor(actor.text))
+            date         <- mapChildPureRec(
+                              changeRule,
+                              "date",
+                              date => ISODateTimeFormat.dateTimeParser.parseDateTime(date.text)
+                            )
+            reason        = (changeRule \\ "reason").headOption.map(_.text)
+            diff         <- flatMapChildPureRec(changeRule, "diff", _.attribute("action").headOption.map(_.text))
+            diffRule     <- getChildPureRec(changeRule, "directive")
```
clarktsiory, line 668: its a rather a `rule`

clarktsiory (changes requested):
It is great having such a migration, it covers already a lot.

There were some other methods in `CommitAndDeployChangeRequestServiceImpl` that still could be migrated, as propagated changes (to avoid `toBox`).

Also the XmlUtils API is nice, I have some suggestions to make it easier to use.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -268,36 +269,39 @@ class CommitAndDeployChangeRequestServiceImpl(
...
-            false
-        }
+        }).chainError("Can not merge change request")
+          .fold(
+            err => {
+              logger.debug(err.fullMsg)
+              false
+            },
+            _ => true
+          )
```
clarktsiory, line 292: this could be a `.tapError(err => logger.debug(err.fullMsg))` if the `logger` was pure.
By looking at the logger, it could be migrated as a pure one, there are some usages which will need to be refactored in this class

clarktsiory (changes requested):

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
skaerg: Should I rename and refactor the existing `ChangeRequestLogger` so that it is a pure logger instead ? or only add a `ChangeRequestLoggerPure` object alongside the existing logger (e.g. similarly to `ChangeValidationLogger` and `ChangeValidationLoggerPure` where both options are available) ?

clarktsiory: You can add a new `ChangeRequestLoggerPure` that extends `NamedZioLogger` instead, you do not need to remove the existing one (we often have both the impure and pure logger)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -874,16 +892,139 @@ class ApiAccountUnserialisationImpl extends ApiAccountUnserialisation {
...
+    s"Missing or bad child '${label}' in sequence and sub-sequences of entry type ${entryType} : ${entry}"
+  }
+
+  private def missingAttributeErrMsg(attributeName: String, entryType: XmlEntryType, entry: XNode): String = {
+    s"Missing attribute '${attributeName}' in entry type ${entryType} : ${entry}"
+  }
+
+  def checkEntry(entry: XNode)(implicit entryType: XmlEntryType): PureResult[XNode] = {
+    if (entry.label == entryType.value) Right(entry)
+    else Left(Inconsistency(s"Entry type is not a ${XML_TAG_NODE_GROUP}: ${entry}"))
```
clarktsiory, line 941:
```suggestion
    else Left(Inconsistency(s"Entry type is not a ${entryType.value}: ${entry}"))
```

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -129,7 +130,7 @@ class CommitAndDeployChangeRequestServiceImpl(
       (config, trigger) <- changeRequest match {
                              case config: ConfigurationChangeRequest =>
                                if (!isMergeableConfigurationChangeRequest(config)) {
-                                 ChangeRequestLogger.info(
+                                 logger.info(
```
clarktsiory, line 133: right above this, there is some code that can be put in the for-comprehension instead : 
```scala
workflowEnabled().either.runNow match {
      case Right(b) => if (b) logger.info(s"Saving and deploying change request ${changeRequest.id.value}")
      case _        => ()
}
```

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -115,21 +116,18 @@ class CommitAndDeployChangeRequestServiceImpl(
...
     implicit val qc: QueryContext = cc.toQuery
 
-    workflowEnabled().either.runNow match {
-      case Right(b) => if (b) logger.info(s"Saving and deploying change request ${changeRequest.id.value}")
-      case _        => ()
-    }
-
     for {
+      workflowEnabled   <- workflowEnabled().toBox
+      _                  = if (workflowEnabled) logger.info(s"Saving and deploying change request ${changeRequest.id.value}")
```
clarktsiory, line 126: the logger is pure, so this will just skip the log. It needs to be `<-`

clarktsiory (changes requested):

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
clarktsiory: the for-comprehension likely needs to be for `IOResult`, and not `Box`, the `toIO` at the end should be avoided

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/CommitAndDeployChangeRequestService.scala
```diff
@@ -115,21 +116,19 @@ class CommitAndDeployChangeRequestServiceImpl(
...
 
-    workflowEnabled().either.runNow match {
-      case Right(b) => if (b) logger.info(s"Saving and deploying change request ${changeRequest.id.value}")
-      case _        => ()
-    }
-
     for {
+      _                 <- workflowEnabled().toBox.map(workflowEnabled => {
+                             if (workflowEnabled) logger.info(s"Saving and deploying change request ${changeRequest.id.value}")
+                           })
```
clarktsiory, line 127: This will still not do the log, the pure logger needs to be evaluated (the result of this expression is `Box[IOResult[Unit]]`).

The whole for-comprehension needs to be a `IOResult` one, there are nested `toBox` and `toIO` and we don't want that. In fact the `toIO` and the end needs to be removed (you will likely need to do `saveConfigurationChangeRequest(...).toIO`)

clarktsiory (changes requested):

skaerg:
PR updated with a new commit

PR #6359, opened 2025-05-06
Fixes #26878: Remove default root description 

VinceMacBuche:
https://issues.rudder.io/issues/26878

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/action/RemoveDefaultRootDescription.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/action/RemoveDefaultRootDescription.scala
@@ -1,0 +1,70 @@
+/*
+ *************************************************************************************
+ * Copyright 2025 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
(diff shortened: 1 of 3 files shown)


clarktsiory (approved):
LGTM

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/action/RemoveDefaultRootDescription.scala
```diff
@@ -0,0 +1,76 @@
...
+import com.normation.zio.*
+import zio.syntax.ToZio
+
+class RemoveDefaultRootDescription(nodeFactRepository: NodeFactRepository) extends BootstrapChecks {
+
+  override val description = "Remove Default node documentation "
+
+  override def checks(): Unit = {
+    (for {
+      root <- nodeFactRepository.get(NodeId("root"))(QueryContext.systemQC).notOptional("no root node in inventory")
```
fanf, line 61: there's a `ROOT_POLICY_SERVER_ID` in place of using `NodeId("root")`. It's better to never use that kind of string directly in code.

fanf:
I'm not sure about that one. OK, the description isn't good, but won't it be better to actually add the missing "edit" action in UI, remove the default description, and let user do what they want with the existing one (ie, no force delete ?)

fanf:
OK.
The update of description by API and display was demanded by users managing the node lifecycle mainly by api, it is relevant without corresponding ui for editing.

VinceMacBuche:
Commit modified

PR #6362, opened 2025-05-07
Fixes from scala3 migration

mbaechler:
This is another set of fixes required in scala3 that happens to also be valid in scala2.
Merging them in the scala2 codebase will make the switch to scala3 simpler.


fanf:
Super seeded by https://github.com/Normation/rudder/pull/6363

PR #6364, opened 2025-05-09
fix: simplify xml parsing syntax that is not supported in scala3 and …

mbaechler:
…fix acl parsing

`case <property>{p @ _*}</property>` syntax is not working in scala3 any longer.
While rewriting the 2 usages in the codebase, I figured the authz parsing is broken.
So this is the test and the fix (s/isEmpty/nonEmpty).


fanf:
Super-seeded by https://github.com/Normation/rudder/pull/6366 (merged in 8.2)

PR #6367, opened 2025-05-09
refactor: ensure EventActor is serialized as expected and make the co…

mbaechler:
…de scala3 compatible


fanf:
Backported in Rudder 8.2 in https://github.com/Normation/rudder/pull/6369

PR #6371, opened 2025-05-14
Fixes #26900: Score service does not really clean unmanaged score, only remove them in cache but they are still in database

VinceMacBuche:
https://issues.rudder.io/issues/26900

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
@@ -49,6 +49,7 @@ import doobie.implicits.*
 import doobie.implicits.toSqlInterpolator
 import doobie.postgres.implicits.pgEnumString
 import doobie.util.invariant.InvalidEnum
+import doobie.util.update.Update
 import zio.Ref
 import zio.interop.catz.*
 import zio.json.ast.Json
@@ -57,10 +58,10 @@ trait ScoreRepository {
 
   def getAll(): IOResult[Map[NodeId, List[Score]]]
   def getAllOneScore(scoreId: String): IOResult[Map[NodeId, Score]]
-  def getScore(nodeId:        NodeId, scoreId: Option[String]): IOResult[List[Score]]
-  def getOneScore(nodeId:     NodeId, scoreId: String):         IOResult[Score]
-  def saveScore(nodeId:       NodeId, score:   Score):          IOResult[Unit]
-  def deleteScore(nodeId:     NodeId, scoreId: Option[String]): IOResult[Unit]
+  def getScore(nodeId:        NodeId, scoreId:      Option[String]): IOResult[List[Score]]
+  def getOneScore(nodeId:     NodeId, scoreId:      String):         IOResult[Score]
+  def saveScore(nodeId:       NodeId, score:        Score):          IOResult[Unit]
+  def deleteScore(nodeId:     Seq[NodeId], scoreId: Option[String]): IOResult[Unit]
 
 }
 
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -158,9 +158,10 @@ class ScoreServiceImpl(
   }
   def cleanScore(name: String)(implicit qc: QueryContext): IOResult[Unit] = {
     for {
-      _         <- cache.update(_.map { case (id, gscore) => (id, gscore.copy(details = gscore.details.filterNot(_.scoreId == name))) })
+      newGlobal         <- cache.updateAndGet(_.map { case (id, gscore) => (id, gscore.copy(details = gscore.details.filterNot(_.scoreId == name))) })
+      _         <- ZIO.foreach(newGlobal){ case (id, gscore) => globalScoreRepository.save(id, gscore)}
       newScores <- scoreCache.updateAndGet(_.map { case (id, scores) => (id, scores.filterNot(_.scoreId == name)) })
-      _         <- update(newScores)
```
VinceMacBuche, line 163: I forgot that update keeps value for score that are not present in "newScore", and thought it would clean missing score.
I replaced by adding two queries to remove score directly

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -158,9 +162,16 @@ class ScoreServiceImpl(
...
-      _         <- cache.update(_.map { case (id, gscore) => (id, gscore.copy(details = gscore.details.filterNot(_.scoreId == name))) })
+      _         <- ScoreLoggerPure.info(s"Cleaning score ${name}")
+      // Remove from score caches
+      newGlobal <- cache.updateAndGet(_.map {
+                     case (id, gscore) => (id, gscore.copy(details = gscore.details.filterNot(_.scoreId == name)))
+                   })
       newScores <- scoreCache.updateAndGet(_.map { case (id, scores) => (id, scores.filterNot(_.scoreId == name)) })
-      _         <- update(newScores)
+      // Remove from databases
+      _         <- ZIO.foreach(newScores.keySet) { case nodeId => scoreRepository.deleteScore(nodeId, Some(name)) }
```
fanf, line 172: I think we want to have a repo method with signature like `def deleteScore(name: ScoreId, nodes: Seq[NodeId]): IOResult[???]` so that we can do only one sql request with a prepared statement and avoid high I/O when there is a lot of nodes involved.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -158,9 +162,16 @@ class ScoreServiceImpl(
...
+      _         <- ZIO.foreach(newGlobal) { case (id, gscore) => globalScoreRepository.save(id, gscore) }
```
fanf, line 173: Same here

fanf (changes requested):
I think we must create batch-aware SQL methods to avoid killing postgresql under the number of request once the number of nodes will become big. 
Else the code/logic seems ok. 
Having tests to demonstrate the cases in sql would help a lot.

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreRepository.scala
```diff
@@ -180,11 +184,13 @@ class ScoreRepositoryImpl(doobie: Doobie) extends ScoreRepository {
...
 
-  override def deleteScore(nodeId: NodeId, scoreId: Option[String]): IOResult[Unit] = {
-    val whereNode = Some(fr"nodeId = ${nodeId.value}")
+  override def deleteScore(nodeId: Seq[NodeId], scoreId: Option[String]): IOResult[Unit] = {
+    val whereNode = Some(fr"nodeId = ?")
     val whereName = scoreId.map(n => fr"scoreId = ${n}")
     val where     = Fragments.whereAndOpt(whereNode, whereName)
-    val q         = sql"delete from scoredetails " ++ where
-    transactIOResult(s"error when deleting score for node ${nodeId.value}")(xa => q.update.run.transact(xa).unit)
+    val q         = Update[NodeId]((sql"delete from scoredetails " ++ where).toString)
```
fanf: the `toString` is strange

fanf: So, `toString` does not give you the sql, but the string `Fragment(the sql here)`

fanf: You need ```    val whereNode = Some(fr"nodeId = ?")
    val whereName = scoreId.map(n => Fragment.const0(s"scoreId = '${n}'")) // const0 since we don't want it to become '?'
    val where     = Fragments.whereAndOpt(whereNode, whereName)
    val sql       = (sql"delete from scoredetails " ++ where).update.sql
    val q         = Update[NodeId](sql)
    transactIOResult(s"error when deleting score for nodes ${nodeId.map(_.value).mkString(",")}")(xa =>
      q.updateMany(nodeId).transact(xa).unit
    )
```

fanf (changes requested):
The SQL is not correct and lead to an error at startup if there is score to delete.

VinceMacBuche:
Commit modified

PR #6374, opened 2025-05-15
Fixes #26832: Groups with special target has clone button but cannot be cloned

clarktsiory:
https://issues.rudder.io/issues/26832


fanf (approved):
LGTM

PR #6375, opened 2025-05-15
Fixes #26779: Archive “download as zip” creates backup with wrong name (but right contents)

fanf:
https://issues.rudder.io/issues/26779

Bad `hh` in place of  `HH`. Plus normalize toward iso and utc

image


clarktsiory (approved):
LGTM

PR #6376, opened 2025-05-15
Fixes #26747: Number of “nodes in audit mode” wrong on "About" page

fanf:
https://issues.rudder.io/issues/26747

We were just not displaying node in mixed mode info.


amousset:
what does "mixed mode" means in a node context?

amousset:
My understanding is this should match the policy mode computed from the node details override and global policy mode, and i don't see how it could be "mixed".

fanf:
It means that the node has both enforced and audit policies applied on it. It can happen when you have a default global mode locally overridden for a given directive (typically: default == `enforce`, one directive in `audit`)

fanf:
it's the same aggregation logic than for `rules`, but for the other container.

clarktsiory (approved):

PR #6382, opened 2025-05-19
Fixes #26921: Improve drag'n drop ergonomics in the techniques editor

RaphaelGauthier:
https://issues.rudder.io/issues/26921

Drag'n drop has undergone some changes:
- when a method is dragged, it remains visible (but greyed out)
- The appearance of drop zones no longer has an impact on the size of the list, so items remain in place when dragged and dropped.
- the appearance of the drop zones has therefore changed: it is no longer a zone delimited by a dotted border, but a separator that indicates the drop zones. This separator is highlighted when it becomes the target for the drop.
- The margins between methods and blocks have been reduced

I've also moved the info icon at the top right of the method to the method name badge, and removes the "method" label to free up space (not visible in the gif) : 
image

Thanks to these changes, it's now easier to move an element, and harder to do so by mistake

Here is a preview :
dragndrop-improved


RaphaelGauthier:
PR updated with a new commit

VinceMacBuche (approved):

PR #6386, opened 2025-05-19
Fixes #26925: 

clarktsiory:
https://issues.rudder.io/issues/26925


clarktsiory:
PR updated with a new commit

fanf:
I would prefer that you make `UserAccount.Write` and `UserAccount.Edit` exactly the same and only use `Write` rights everywhere. We want to merge the two. And so modify `UserApi.UpdateApiToken` if necessary

clarktsiory:
PR updated with a new commit

fanf (approved):
OK, we will deal with the `edit` thing later

PR #6388, opened 2025-05-21
Fixes #26936: 

fanf:
https://issues.rudder.io/issues/26936

Move safe `SAXParserFactory` into its own object in `utils`, and use it everywhere. 
Also, create an utility method for technique category parsing to avoid having user not going through the safe parser. 

Also change the returned message of API to avoid letting too much go out (remove stack trace of system error in particular, likely to get displayed when a vulnerability is probed)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/LiftApiDispatcher.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/LiftApiDispatcher.scala
@@ -278,7 +278,7 @@ class LiftHandler(
 
 /*
  * some type machinery to be able to have an implementation for old RuleApi up to V13
- * and an other for API with zio_json in V14 (and choose processing based on that)
+ * and another for API with zio_json in V14 (and choose processing based on that)
  */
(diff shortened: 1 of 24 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/LiftApiDispatcher.scala
```diff
@@ -263,6 +263,7 @@ class LiftHandler(
       case ApiError.BadParam(x, _)   => "Bad parameters for request: " + x
       case ApiError.Authz(x, _)      => "Authorization error: " + x
     }
+    println("*********** here")
```
amousset, line 266: this looks like a fun debugging session

fanf: It wasn't even here

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM !

PR #6390, opened 2025-05-22
Fixes #26939: User management UI does not display when a user info is a complex JSON object

clarktsiory:
https://issues.rudder.io/issues/26939

Using the `Decoder.dict Decoder.string` can result in a silent error when the JSON has non-string values.
It happens with the `otherInfo` field which is additional user information that can be updated from the API, as JSON.

In the UI, complex JSON objects (int, objects) are not supposed to be edited, and parsed as JSON, but string only.
So, we decode the `otherInfo` field as any JSON object instead, and split the key-values into ones that are editable as string, and others that are read-only JSON : 
 
Screenshot from 2025-05-22 16-04-53

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/UserManagement.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/UserManagement.elm
@@ -7,7 +7,7 @@ import Http exposing (..)
 import String exposing (isEmpty)
 
 import UserManagement.ApiCalls exposing (activateUser, addUser, disableUser, getRoleConf, getUsersConf, postReloadConf, updateUser, updateUserInfo)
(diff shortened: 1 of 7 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/UserManagement.elm
```diff
@@ -216,35 +216,35 @@ update msg model =
...
+        UserInfoName name ->
             let
                 userForm = model.userForm
                 userInfoForm = userForm.userInfoForm
-                newUserInfoForm = { userInfoForm | email = email }
+                newUserInfoForm = { userInfoForm | name = name }
                 newUserForm = { userForm | userInfoForm = newUserInfoForm, isValidInput = ValidInputs }
             in
                 ({model | userForm = newUserForm}, Cmd.none)
-        UserInfoFields fields ->
```
clarktsiory, line 243: unused Msg

clarktsiory:
PR updated with a new commit

RaphaelGauthier (approved):

PR #6393, opened 2025-05-23
Fixes #26944: 

fanf:
https://issues.rudder.io/issues/26944

We were using `filename` as the file name, and not a path, which it is really. So just use the real filename. 

Also changed for archive API, even if there not risk their since the file is never written, but if we change our way of doing things in the futur, I prefer to have that checked.


clarktsiory (approved):
LGTM, so sometimes we use `File(...).name`, and sometimes we explicitly check canonical path within a known workspace like the configuration-repo

PR #6398, opened 2025-05-26
Fixes #26957: 

fanf:
https://issues.rudder.io/issues/26957

Remove path trasversal in `TechniqueApi` creation. There's two place, in resources and in technique itself. 

That PR also rework a bit the `FileUtils` sanitization API: 
- change signature from `IOResult[Either[OutsideBaseDir, File]]` to `IOResult[File]` because we really don't want to treat an `OutsideBaseDir` as a potential ok result in the nominal case, and so it simplifies usage for nominal code,
- ass a `checkSanitizedIsIn(baseDir, path)` to add more flexibility in some case (it's just exposing what was internal before)


fanf:
OK, merging this PR

PR #6399, opened 2025-05-27
Fixes #26970: For compat with older XML eventlog, we need to also manage API action

fanf:
https://issues.rudder.io/issues/26970

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
@@ -793,6 +793,13 @@ class ApiAccountUnserialisationImpl extends ApiAccountUnserialisation {
     // mess up with authz
     import cats.implicits.*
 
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -793,6 +793,13 @@ class ApiAccountUnserialisationImpl extends ApiAccountUnserialisation {
     // mess up with authz
     import cats.implicits.*
 
+    def parse(s: String): Box[List[HttpAction]] = {
+      (s.split(",").map(_.trim).toList.filter(_.nonEmpty).traverse(HttpAction.parse)) match {
+        case Left(e)  => Failure(e)
+        case Right(x) => Full(x)
+      }
```
clarktsiory, line 800: for the sake of the already functional DSL :grin: (with `com.normation.errors.*` imported) :
```suggestion
      s.split(",").map(_.trim).toList.filter(_.nonEmpty).traverse(HttpAction.parse).toBox
```

clarktsiory (approved):

PR #6403, opened 2025-05-27
Fixes #26979: 

clarktsiory:
https://issues.rudder.io/issues/26979

Change the authz of the rollback endpoint in event logs (like https://github.com/Normation/rudder/pull/6401, upmerge is needed)


clarktsiory:
OK, merging this PR

PR #6407, opened 2025-05-28
Fixes #26996: Add argon2id support for local hash

amousset:
https://issues.rudder.io/issues/26996

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
@@ -82,11 +82,12 @@ sealed abstract class PasswordEncoderType(override val entryName: String) extend
 }
 
 object PasswordEncoderType extends Enum[PasswordEncoderType] {
-  final case object MD5    extends PasswordEncoderType("MD5")
-  final case object SHA1   extends PasswordEncoderType("SHA-1")
-  final case object SHA256 extends PasswordEncoderType("SHA-256")
-  final case object SHA512 extends PasswordEncoderType("SHA-512")
-  final case object BCRYPT extends PasswordEncoderType("BCRYPT")
+  final case object MD5      extends PasswordEncoderType("MD5")
+  final case object SHA1     extends PasswordEncoderType("SHA-1")
+  final case object SHA256   extends PasswordEncoderType("SHA-256")
+  final case object SHA512   extends PasswordEncoderType("SHA-512")
+  final case object BCRYPT   extends PasswordEncoderType("BCRYPT")
+  final case object ARGON2ID extends PasswordEncoderType("ARGON2ID")
 
   // Default value, same as in RudderPasswordEncoder
   val DEFAULT: PasswordEncoderType = BCRYPT
@@ -102,14 +103,15 @@ object PasswordEncoderType extends Enum[PasswordEncoderType] {
 
 }
 
-class PasswordEncoderDispatcher(bcryptCost: Int) {
+class PasswordEncoderDispatcher(bcryptCost: Int, argon2Memory: Int, argon2Parallelism: Int, argon2Iterations: Int) {
   def dispatch(encoderType: PasswordEncoderType): PasswordEncoder = {
     encoderType match {
-      case PasswordEncoderType.MD5    => RudderPasswordEncoder.MD5
-      case PasswordEncoderType.SHA1   => RudderPasswordEncoder.SHA1
-      case PasswordEncoderType.SHA256 => RudderPasswordEncoder.SHA256
-      case PasswordEncoderType.SHA512 => RudderPasswordEncoder.SHA512
-      case PasswordEncoderType.BCRYPT => RudderPasswordEncoder.BCRYPT(bcryptCost)
+      case PasswordEncoderType.MD5      => RudderPasswordEncoder.MD5
+      case PasswordEncoderType.SHA1     => RudderPasswordEncoder.SHA1
+      case PasswordEncoderType.SHA256   => RudderPasswordEncoder.SHA256
+      case PasswordEncoderType.SHA512   => RudderPasswordEncoder.SHA512
+      case PasswordEncoderType.BCRYPT   => RudderPasswordEncoder.BCRYPT(bcryptCost)
+      case PasswordEncoderType.ARGON2ID => RudderPasswordEncoder.ARGON2ID(argon2Memory, argon2Parallelism, argon2Iterations)
     }
   }
 
@@ -124,9 +126,12 @@ class PasswordEncoderDispatcher(bcryptCost: Int) {
 
 object RudderPasswordEncoder {
 
+  import org.apache.commons.codec.binary.Base64
+  import org.bouncycastle.crypto.generators.Argon2BytesGenerator
   import org.bouncycastle.crypto.generators.OpenBSDBCrypt
+  import org.bouncycastle.crypto.params.Argon2Parameters
 
-  val random = new SecureRandom()
+  private val random = new SecureRandom()
 
   // see https://stackoverflow.com/a/44227131
   // produce a random hexa string of 32 chars
@@ -154,7 +159,7 @@ object RudderPasswordEncoder {
 
     def fromPasswordEncoderType(encoderType: PasswordEncoderType): SecurityLevel = {
       encoderType match {
-        case PasswordEncoderType.BCRYPT                                                                                   => Modern
+        case PasswordEncoderType.BCRYPT | PasswordEncoderType.ARGON2ID                                                    => Modern
         case PasswordEncoderType.MD5 | PasswordEncoderType.SHA1 | PasswordEncoderType.SHA256 | PasswordEncoderType.SHA512 =>
           Legacy
       }
@@ -176,12 +181,12 @@ object RudderPasswordEncoder {
   }
 
   // One pass, non-salted hashes. Unsuitable for password storage.
-  val MD5               = new DigestEncoder("MD5")
-  val SHA1              = new DigestEncoder("SHA-1")
-  val SHA256            = new DigestEncoder("SHA-256")
-  val SHA512            = new DigestEncoder("SHA-512")
+  val MD5                                                      = new DigestEncoder("MD5")
+  val SHA1                                                     = new DigestEncoder("SHA-1")
+  val SHA256                                                   = new DigestEncoder("SHA-256")
+  val SHA512                                                   = new DigestEncoder("SHA-512")
   // Proper password hash functions
-  def BCRYPT(cost: Int) = new PasswordEncoder() {
+  def BCRYPT(cost: Int)                                        = new PasswordEncoder() {
     override def encode(rawPassword: CharSequence):                           String  = {
       val salt: Array[Byte] = new Array(16)
       random.nextBytes(salt)
@@ -200,20 +205,69 @@ object RudderPasswordEncoder {
       }
     }
   }
+  def ARGON2ID(memory: Int, parallelism: Int, iterations: Int) = new PasswordEncoder() {
+    override def encode(rawPassword: CharSequence):                           String  = {
+      /// 32 bytes
+      val hashSize = 32
+      /// 16 bytes = 128 bits, the recommended value ("sufficient for all applications")
+      val saltSize = 16
+
+      val argon2Version = Argon2Parameters.ARGON2_VERSION_13
+
+      val salt: Array[Byte] = new Array(saltSize)
+      random.nextBytes(salt)
+
+      val params = new Argon2Parameters.Builder(Argon2Parameters.ARGON2_id)
+        .withVersion(argon2Version)
+        .withIterations(iterations)
+        .withMemoryAsKB(memory)
+        .withParallelism(parallelism)
+        .withSalt(salt)
+        .build()
+
+      // Compute the hash
+      val generate = new Argon2BytesGenerator
+      generate.init(params)
+      val hash: Array[Byte] = new Array(hashSize)
+      generate.generateBytes(rawPassword.toString.toCharArray, hash)
+
+      val base64Salt = Base64.encodeBase64URLSafe(salt)
+      val base64Hash = Base64.encodeBase64URLSafe(hash)
+
+      // We store hashes with a shadow-like format
+      val shadowHash =
+        s"$$argon2id$$v=$argon2Version$$m=$memory,t=$iterations,p=$parallelism$$$base64Salt$$$base64Hash"
+
+      shadowHash
+    }
+    override def matches(rawPassword: CharSequence, encodedPassword: String): Boolean = {
+      try {
+
+        OpenBSDBCrypt.checkPassword(encodedPassword, rawPassword.toString.toCharArray)
+      } catch {
+        case e: Exception =>
+          ApplicationLoggerPure.Auth.debug(s"Invalid password format: ${e.getMessage}")
+          false
+      }
+    }
+  }
   // Default algorithm to use for hashing passwords
-  val DEFAULT           = BCRYPT
+  val DEFAULT                                                  = ARGON2ID
 
   def getFromEncoded(encodedPassword: String, securityLevel: SecurityLevel): Either[String, PasswordEncoderType] = {
     // Two cases: modern = /etc/shadow-like or legacy = plain hashes
 
     // https://github.com/bcgit/bc-java/blob/5b1360854d85fd27b75720015be68f9e172db013/core/src/main/java/org/bouncycastle/crypto/generators/OpenBSDBCrypt.java#L41C1-L48C1
     // Allow types: 2, 2x, 2y, 2a, 2b
-    val bcryptRegex = "^\\$2[abxy]?\\$.+".r
+    val bcryptRegex    = "^\\$2[abxy]?\\$.+".r
+    val argon2idPrefix = "$argon2id$"
 
     val hexaRegex = "^[a-fA-F0-9]+$"
 
     if (encodedPassword.matches(bcryptRegex.regex)) {
       Right(PasswordEncoderType.BCRYPT)
+    } else if (encodedPassword.startsWith(argon2idPrefix)) {
+      Right(PasswordEncoderType.ARGON2ID)
     } else if (securityLevel == SecurityLevel.Legacy && encodedPassword.matches(hexaRegex)) {
       // Guess based on hexadecimal string length
       encodedPassword.length match {
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -323,7 +323,12 @@ class AppConfigAuth extends ApplicationContextAware {
   }
 
   @Bean def passwordEncoderDispatcher: PasswordEncoderDispatcher = {
-    new PasswordEncoderDispatcher(RudderConfig.RUDDER_BCRYPT_COST)
+    new PasswordEncoderDispatcher(
+      RudderConfig.RUDDER_BCRYPT_COST,
+      RudderConfig.RUDDER_ARGON2_MEMORY,
+      RudderConfig.RUDDER_ARGON2_PARALLELISM,
+      RudderConfig.RUDDER_ARGON2_ITERATIONS
+    )
   }
 
   @Bean def checkUsersFile: CheckUsersFile = new CheckUsersFile(rudderUserListProvider)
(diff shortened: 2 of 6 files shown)


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -200,20 +297,90 @@ object RudderPasswordEncoder {
       }
     }
   }
+  def ARGON2ID(memory: Int, parallelism: Int, iterations: Int) = new PasswordEncoder() {
```
fanf: I think you should have a `private class Argon2idPasswordEncoder(memory: Int, parallelism: Int, iterations: Int) extends PasswordEncoder { ....`  because IIRC, this is not creating the same thing, and the thing that is created is not what we want (I *think* it has far fewer static (in the JVM sense) part that what we would like).

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -102,18 +134,79 @@ object PasswordEncoderType extends Enum[PasswordEncoderType] {
...
+// Utility class for password hash storage with a shadow-like format.
+// There is no real standard, following https://argon2.online/
+case class Argon2IDHashString(
+    version:     Int,
+    memory:      Int,
+    iterations:  Int,
+    parallelism: Int,
+    // use Chunk to compare by value
+    salt:        Chunk[Byte],
+    hash:        Chunk[Byte]
```
fanf: it looks like there is 3 kinds of data in that case class (with different lifecycles):
- `version`, `memory`, `iteration`, `parallelism` looks like stable, reusable accross hash parameters,
- `salt` is to be defined for one case of hash, 
- `hash` is partly given by external means, so it's different. 

Perhaps the first two are the same, with just the first being linked to default value. 

We see that hash is different, because it's of no use in the `presentedHash` computation below, appart for its size - which is also a generic parameter. 

So I would split it like that: 
```
case class Argon2IDHashStringParam(
    version:     Int,
    memory:      Int,
    iterations:  Int,
    parallelism: Int,
    // use Chunk to compare by value
    salt:        Chunk[Byte],
    hashSite:        Int
}
case class Argon2IDHashString(params: Argon2IDHashStringParam, hash: Chunk[Byte])
```

So that we can share `Argon2IDHashStringParam` when comparing, without the `hash`.

If you need to deep update values in embeded case classes, you can use `quicklens`, for example:

```
import com.softwarmill.quicklens.*

val hs = Argon2IDHashString(param, hash)
val updated = hs.modify(_.params.salt).setTo(xxx)
```

It also allows to create an helper function for param builder, so you are sure you always use the same: 
```
def buildParam(param: Argon2IDHashStringParam) = new Argon2Parameters.Builder(param.hashType)
        .withVersion(param.version)
        .withIterations(param.iterations)
        .withMemoryAsKB(param.memory)
        .withParallelism(param.parallelism)
        .withSalt(param.salt)
        .build()
```
(also, that show that `salt` is in the same logical grouping, so we have 2 kinds of data, not 3)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -102,18 +134,79 @@ object PasswordEncoderType extends Enum[PasswordEncoderType] {
...
+) {
+  def toShadowString = {
+    val encoder     = Base64.getEncoder.withoutPadding
+    val encodedSalt = encoder.encodeToString(salt.toArray)
+    val encodedHash = encoder.encodeToString(hash.toArray)
+    s"$$argon2id$$v=$version$$m=$memory,t=$iterations,p=$parallelism$$$encodedSalt$$$encodedHash"
+  }
+}
```
fanf: We prefer to put these kind of serialization in the corresponding companion object. It even can be called with a `.toShadowString` as if it was defined like here  (see below)

The idea to do so is dual:
- first, it's better to group serilisation/parsing in the same place because often, when you change one you need to change the other, 
- second, we prefer to have case classes as dumb as possible, just the data. It's easier to mentally parse "case class class is for data, logic is elsewhere, likely in the companion or a related service that use the data". It's also the translation of an aspect of FP where you transform immutable data in a flow of function defined out of the data, more than you update an object thanks to method in it.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -102,18 +134,79 @@ object PasswordEncoderType extends Enum[PasswordEncoderType] {
...
+  def apply(version: Int, memory: Int, iterations: Int, parallelism: Int, salt: Array[Byte], hash: Array[Byte]) = {
+    new Argon2IDHashString(
+      version = version,
+      memory = memory,
+      iterations = iterations,
+      parallelism = parallelism,
+      salt = Chunk.fromArray(salt),
+      hash = Chunk.fromArray(hash)
+    )
+  }
```
fanf: so here, you could have: 
```
implicit class ToShadowString(hashString: Argon2IDHashString) {
  def toShadowString: String =    {
    val encoder     = Base64.getEncoder.withoutPadding
    val encodedSalt = encoder.encodeToString(salt.toArray)
    val encodedHash = encoder.encodeToString(hash.toArray)
    s"$$argon2id$$v=$version$$m=$memory,t=$iterations,p=$parallelism$$$encodedSalt$$$encodedHash"
  }
}
```
And then call `val hs = Argon2IDHashString(...) ; hs.toShadowString`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -124,9 +217,11 @@ class PasswordEncoderDispatcher(bcryptCost: Int) {
 
 object RudderPasswordEncoder {
 
+  import org.bouncycastle.crypto.generators.Argon2BytesGenerator
   import org.bouncycastle.crypto.generators.OpenBSDBCrypt
+  import org.bouncycastle.crypto.params.Argon2Parameters
 
-  val random = new SecureRandom()
+  private val secureRandom = new SecureRandom()
```
fanf, line 146: :+1:

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -82,14 +84,44 @@ sealed abstract class PasswordEncoderType(override val entryName: String) extend
...
+    val defaultIterations  = 3
+    val defaultParallelism = 1
+
+    // 32 bytes
+    // Used by most implementations.
+    val hashSize = 32
+
+    // 16 bytes = 128 bits
+    // The recommended value in the RFC, deemed "sufficient for all applications" in the specs.
+    val saltSize = 16
```
fanf: these parameters should be in a `Argon2IDHashStringParam` (see below)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -200,20 +297,90 @@ object RudderPasswordEncoder {
...
+              .withMemoryAsKB(stored.memory)
+              .withParallelism(stored.parallelism)
+              .withSalt(stored.salt.toArray)
+              .build()
+            val generate     = new Argon2BytesGenerator
+            generate.init(storedParams)
+            val presentedHash: Array[Byte] = new Array(stored.hash.size)
+            generate.generateBytes(rawPassword.toString.toCharArray, presentedHash)
+            Right(presentedHash)
+          }
```
fanf: from a scala idiomatic way, we prefer to create an helper function that will encapsulate things that can throw exceptions:

```
def computeHash(rawPassword: String, params: Argon2IDHashStringParam): Either[String, Array[Byte]) = {
try {
  val storedParams = buildParam(params)
  val generate     = new Argon2BytesGenerator
  generate.init(storedParams)
  val presentedHash: Array[Byte] = new Array(params.hashSize)
  generate.generateBytes(rawPassword.toString.toCharArray, presentedHash)
  Right(presentedHash)
} catch {
        case e: Exception =>
          Left(s"Error while computing Argon2 hash: ${e.getMessage}")
      }
}
```

And if `MessageDigest` doesn't throw, you can put it the yield directly (else, same as above), so you have a short for loop: 

```
    override def matches(rawPassword: CharSequence, encodedPassword: String): Boolean = {
      val result = for {
          stored              <- Argon2IDHashString.parseShadowString(encodedPassword)
          // Compute the hash
          presentedHash <- computeHash(rawPassword, stored.params)
        } yield  MessageDigest.isEqual(stored.hash.toArray, presentedHash)

     result match { .... }
  }
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -200,20 +297,90 @@ object RudderPasswordEncoder {
...
+          // Actual comparison
+          res           <- Right(MessageDigest.isEqual(stored.hash.toArray, presentedHash))
+        } yield res
+
+        result match {
+          case Left(e)  =>
+            ApplicationLoggerPure.Auth.warn(s"Error while computing Argon2 hash: ${e}")
```
fanf: `ApplicationLoggerPure` is to use in a `zio` for loop, else it's effect free (does nothing). You need `ApplicationLoggerPure.Auth.logEffect.warn`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -102,18 +134,79 @@ object PasswordEncoderType extends Enum[PasswordEncoderType] {
...
+      memory = memory,
+      iterations = iterations,
+      parallelism = parallelism,
+      salt = Chunk.fromArray(salt),
+      hash = Chunk.fromArray(hash)
+    )
+  }
+
+  def parseShadowString(hashString: String): Either[String, Argon2IDHashString] = {
+    val pattern = """\$argon2id\$v=(\d+)\$m=(\d+),t=(\d+),p=(\d+)\$([^$]+)\$([^$]+)""".r
```
clarktsiory: the regex pattern is compiled every time a hash is parsed which can be very costly.
The pattern should be put in the companion object instead, if it is not needed elsewhere it could be private : 
```scala
object Argon2IDHashString {
  private val pattern = """\$argon2id\$v=(\d+)\$m=(\d+),t=(\d+),p=(\d+)\$([^$]+)\$([^$]+)""".r
  ...
}

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala [hunk 10]
```diff
@@ -176,15 +271,17 @@ object RudderPasswordEncoder {
...
-  val SHA256            = new DigestEncoder("SHA-256")
-  val SHA512            = new DigestEncoder("SHA-512")
+  val MD5                                                      = new DigestEncoder("MD5")
+  val SHA1                                                     = new DigestEncoder("SHA-1")
+  val SHA256                                                   = new DigestEncoder("SHA-256")
+  val SHA512                                                   = new DigestEncoder("SHA-512")
   // Proper password hash functions
-  def BCRYPT(cost: Int) = new PasswordEncoder() {
+  def BCRYPT(cost: Int)                                        = new PasswordEncoder() {
+    val defaultCost = 12
```
clarktsiory: is this actually used ? there seems to be a duplicate definition : `PasswordEncoder.BCRYPT.defaultCost`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
clarktsiory: another consideration : to me it seems like the configuration is so critical that we can't afford to confuse all the integer parameters : `version`, `memory`, `iteration`, `parallelism`

I think we should use [opaque types](https://docs.scala-lang.org/scala3/book/types-opaque-types.html#opaque-types) (at least if we don't want scala 2-incompatible code yet, then use value classes : `case class ArgonVersion(value: Int) extends AnyVal`)

Then we can also use all those types in all our API, so that is a lot safer, and we would have something like the following structure:
```scala
case class Argon2IDHashStringParam(
    version: ArgonVersion,
    memory: ArgonMemory,
    iterations: ArgonIterations,
    ...
```

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -731,14 +731,64 @@ object RudderParsedProperties {
...
+  }
+
+  val RUDDER_ARGON2_ITERATIONS: Int = {
+    val defaultValue = PasswordEncoderType.ARGON2ID.defaultIterations
+    try {
+      config.getInt("rudder.argon2.iterations")
+    } catch {
+      case ex: ConfigException =>
+        ApplicationLogger.debug(
+          "Property 'rudder.argon2.iterations' is absent or empty in rudder.configFile. Default iterations to $defaultValue."
```
clarktsiory:
```suggestion
          s"Property 'rudder.argon2.iterations' is absent or empty in rudder.configFile. Default iterations to $defaultValue."
```

clarktsiory (changes requested):
the change is great ! I only have some additional suggestions

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -322,7 +322,14 @@ class AppConfigAuth extends ApplicationContextAware {
   }
 
   @Bean def passwordEncoderDispatcher: PasswordEncoderDispatcher = {
-    new PasswordEncoderDispatcher(RudderConfig.RUDDER_BCRYPT_COST)
+    new PasswordEncoderDispatcher(
```
amousset, line 325: I did not find how to easily reuse the one from `RudderConfig`

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/Argon2.scala
```diff
@@ -0,0 +1,208 @@
...
+
+// All parameters required to create the encoder.
+case class Argon2EncoderParams(
+    variant:     Int,
+    version:     Int,
+    memory:      Argon2Memory,
+    iterations:  Argon2Iterations,
+    parallelism: Argon2Parallelism,
+    hashSize:    Int,
+    saltSize:    Int
```
fanf: Why not giving a type to all things?
I totally see how we could mixup the remaining ones too

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -82,14 +83,47 @@ sealed abstract class PasswordEncoderType(override val entryName: String) extend
...
+
+    /// Argon2 hash type. Use "id", the recommended variant in most cases.
+    val hashType = Argon2Parameters.ARGON2_id
+
+    /// Latest version
+    val defaultVersion = Argon2Parameters.ARGON2_VERSION_13
+
+    // Defaults for user-controlled settings
+    // 19 MiB, from OWASP recommendations
+    val minimumMemory      = 19922944
```
fanf: Should be an Argon2Memory
Same for others

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/PasswordEncoderTypeTest.scala
```diff
@@ -0,0 +1,127 @@
+/*
+ *************************************************************************************
+ * Copyright 2018 Normation SAS
+ *************************************************************************************
```
fanf, line 4: The date is incorrect

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -200,20 +237,41 @@ object RudderPasswordEncoder {
...
+    }
+    override def matches(rawPassword: CharSequence, encodedPassword: String): Boolean = {
+      Argon2Hash.checkPassword(rawPassword.toString.getBytes, encodedPassword) match {
+        case Left(e)                  =>
+          ApplicationLoggerPure.Auth.logEffect.warn(s"Error while checking Argon2 hash: $e")
+          false
+        case Right(isPasswordCorrect) => isPasswordCorrect
+      }
+    }
+  }
```
fanf, line 240: neat. Much better.

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/PasswordEncoderTypeTest.scala
```diff
@@ -0,0 +1,127 @@
...
+
+      test("recognize bcrypt a") {
+        assertTrue(
+          RudderPasswordEncoder
+            .getFromEncoded("$2a$12$mgAVHJ2/312Q.hdWT0EzjOZHrGicXV/2K.1CLsnM3gOYqi5twcwtW", modern)
+            .is(_.right) == PasswordEncoderType.BCRYPT
+        )
+      }
+      test("recognize bcrypt y") {
+        assertTrue(
```
fanf: `assertTrue` is a broad machinery that I like, but @clarktsiory would say that it's better to `assert` for these case: 

```
     import zio.test.Assertion.*
     assert(
        RudderPasswordEncoder
          .getFromEncoded("$2a$12$mgAVHJ2/312Q.hdWT0EzjOZHrGicXV/2K.1CLsnM3gOYqi5twcwtW", modern)
     )(isRight(equalTo(PasswordEncoderType.BCRYPT))
```

And Clark is not wrong, it helps for type safety and is a bit less magicky

amousset: What is the difference wrt type safety?

fanf: the unit assertion are more directed (`isRight` checks that we have an either, etc) while the magic macro of `assertTrue` will try to merge type much harder, and often succeed even when you would have prefer not.

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/Argon2.scala
```diff
@@ -0,0 +1,267 @@
...
+// All parameters required to create the encoder.
+case class Argon2EncoderParams(
+    variant:     Argon2Variant,
+    version:     Argon2Version,
+    memory:      Argon2Memory,
+    iterations:  Argon2Iterations,
+    parallelism: Argon2Parallelism,
+    hashSize:    Argon2HashSize,
+    saltSize:    Argon2SaltSize
+)
```
fanf, line 149: :rocket:

fanf:
LGTM, appart the magnitude factor in the memory parameter in argon algo

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

PR #6413, opened 2025-06-02
Fixes #27010: 

clarktsiory:
https://issues.rudder.io/issues/27010

Being more strict about typing and validation in the JS code : validate the value of the radio button into `before` or `after`, or `null` (do nothing in that case)


clarktsiory:
Commit modified

VinceMacBuche (approved):

PR #6415, opened 2025-06-02
Fixes #27012: Scala3 - reorganize imports, clean unused values

fanf:
https://issues.rudder.io/issues/27012

- remove unused import, especially the ones where `*` is already present, and sometimes because the var is not used,
- remove unused vars, esp. in for loop in the lhs, 
- remove unreachable case in pattern matching (we only have `Certificate` now)
- add the serialization for doobie for EventActor,
- remove `final` in case objects


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
OK, merging this PR

PR #6419, opened 2025-06-03
Fixes #27026: Workspace directory is missing when saving a technique

fanf:
https://issues.rudder.io/issues/27026

An optionnal action has become mandatory for no reasons.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
@@ -427,12 +427,11 @@ class TechniqueApi(
         _            <- ZIO
                           .whenZIO(IOResult.attempt(workspaceDir.exists)) {
                             IOResult.attempt("Error when moving resource file from workspace to final destination") {
-                              finalDir.createDirectoryIfNotExists(true)
+                              finalDir.createDirectoryIfNotExists(createParents = true)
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -427,12 +427,11 @@ class TechniqueApi(
         _            <- ZIO
                           .whenZIO(IOResult.attempt(workspaceDir.exists)) {
                             IOResult.attempt("Error when moving resource file from workspace to final destination") {
-                              finalDir.createDirectoryIfNotExists(true)
+                              finalDir.createDirectoryIfNotExists(createParents = true)
                               workspaceDir.moveTo(finalDir)(File.CopyOptions.apply(true))
                               workspaceDir.parent.parent.delete()
                             }
                           }
-                          .notOptional(s"Error: the workspace directory for techniques '${workspaceDir}' is missing")
```
fanf, line 435: We used to have: 

```
      IOResult.attempt("Error when moving resource file from workspace to final destination")(if (workspaceDir.exists) {
        finalDir.createDirectoryIfNotExists(true)
        workspaceDir.moveTo(finalDir)(File.CopyOptions.apply(true))
        workspaceDir.parent.parent.delete()
        "ok"
      } else {
        "ok"
      })
```

Before https://github.com/Normation/rudder/pull/6398

PR #6423, opened 2025-06-04
Fixes #27035: ReportType json serialization in base is incorrect

fanf:
https://issues.rudder.io/issues/27035

To make the change transparent for both Scala 2 and 3, I switched `ReportType` to `Enumeratum` and parse by hand. It has also a fallback to the a decoder able to deserialize the old syntaxt for transparent migration. 

It also updates ZIO lib and ZIO json to last version, because there was a ton of serialization bugs corrected in between.


clarktsiory (approved):
LGTM

clarktsiory (approved):
LGTM, that is really a recurring issue from deriving Json codecs :see_no_evil:

fanf:
OK, merging this PR

PR #6424, opened 2025-06-05
Fixes #27153: Add type and clean-up campaign code

fanf:
https://issues.rudder.io/issues/27153

Some cleaning-up in campaign code: 

- use standard layout for `CampaignEventState` (`case object/class` in `object CampaignEventState`)
- create enums for query parameters in place of string: `CampaignSortOrder`, `CampaignSortDirection`
- factor out default value for `CampaignSortOrder` in pattern matching 
- use the actual `CampaignEventState` in place of its string value in query
- remove `MainCampaignService` from `CampaignHandler` which is not used and create a dependency loop


fanf:
Replaced by https://github.com/Normation/rudder/pull/6476

PR #6425, opened 2025-06-05
Fixes #27005: License setup should update the list of plugins

clarktsiory:
https://issues.rudder.io/issues/27005

When updating settings in [the public API](https://docs.rudder.io/api/v/21/#tag/Plugins/operation/updateSettings) (the endpoint needs `Administration.Write` rights), the rudder package index needs an attempt to be updated. In case of an error the, like a 401 if credentials are wrong, the response will directly contain the error (the setting file is still updated) 

EDIT: did the cleaning of the APIs, there were some unsuitable JSON encoders, I added a new one that prevents from leaking the password by never encoding it, and added tests

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestPluginInfo.scala
+++ b/webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestPluginInfo.scala
@@ -37,6 +37,11 @@
 
 package com.normation.rudder.rest
 
+import com.normation.JsonSpecMatcher
+import com.normation.errors.IOResult
+import com.normation.errors.RudderError
+import com.normation.plugins.*
+import com.normation.plugins.settings.*
 import com.normation.rudder.api.ApiVersion
 import com.normation.rudder.rest.data.JsonGlobalPluginLimits
 import com.normation.rudder.rest.data.JsonPluginDetails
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestPluginInfo.scala
```diff
@@ -140,7 +140,7 @@ class TestRestPluginInfo extends Specification {
     )
   )
 
-  val pluginApi = new PluginApi(null, null, pluginInfo.succeed)
+  val pluginApi = new PluginApi(null, null, null, pluginInfo.succeed)
```
fanf: this looks like an anti-pattern. We should not have nulls in our tests. At least, it show that that part is not tested, but even without that, a constructor that works with 3/4 of its parameter set to null points to a bad splitting of responsibility. 

I don't have a good general answer, but at least for that one: 
- can you add a test for that API ? 
- can you add something that makes checkable that the effect was called ? (perhaps it's something like the date of last update returned in the api result, or a mock instance for `updateIndex` that check that it's called, or whatever that let us know it happened)

fanf (changes requested):
I would like that such a change be reflected in at least one test somewhere

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/TestRestPluginInfo.scala
clarktsiory: I remove the first `null` which is a `restExtractorService` which was just unused.
For the remaining part, I added a test and cleaned the API, added a new JSON structure which does not encode passwords to replace existing ones (and remove extra unused encoder/decoder definitions)

fanf (approved):
Very nice

clarktsiory:
OK, squash merging this PR

PR #6429, opened 2025-06-05
Fixes #26711: Rule detail compliance has no indication that it is loading

RaphaelGauthier:
https://issues.rudder.io/issues/26711

A skeleton loading animation now appears in the compliance tabs during compliance loading :
loading-compliance-table


RaphaelGauthier:
Commit modified

clarktsiory (approved):
LGTM, the diff becomes very hard to read since the `if` and `let` and `case` that are nested are.
We could refactor more into functions to prevent that in the future

PR #6430, opened 2025-06-05
Fixes #26912: Creating technique with resources in folders can break technique library

clarktsiory:
https://issues.rudder.io/issues/26912

I could not identify the source of the corrupted state (I could not even boot my server anymore), nor apply a workaround (I even reset my configuration repo git history), but to avoid any inconsistency in the immutable map, just   skip non-existing keys


fanf (approved):
OK, let's do that. We will need to rebuild the whole technique editor backend to make it much more resilient, it's very brittle for now.

PR #6432, opened 2025-06-05
Fixes #27040: Inventory role allows to get system compliance and technical logs 

clarktsiory:
https://issues.rudder.io/issues/27040

User-based compliance are accessible under `compliance_read`, and can be found in system logs : a user with "inventory" role should not be able to see them. **EDIT: in fact users without `node_read`**
For now, until we know how to specifically handle system policy types in the system logs, that are visible in some parts of the UI, we should hide those parts for such user.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/reports_server.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/reports_server.html
@@ -5,19 +5,27 @@
   <lastreportgrid-intro></lastreportgrid-intro>
 
 
-  <button id="AllLogButton" class="btn btn-primary">Show logs <i class="fa fa-table"></i></button>
+  <lift:authz role="node_read">
+    <lift:authz role="compliance_read">
+      <button id="AllLogButton" class="btn btn-primary">Show logs <i class="fa fa-table"></i></button>
+    </lift:authz>
+  </lift:authz>
 
   <lastreportgrid-grid></lastreportgrid-grid>
 
   <div class="clearfix"></div>
 
-  <div id="logRun"  class="space-top" style="display:none;">
-    <h3 class="page-title space-top">All logs for current run</h3>
-    <button id="AllLogHideButton" class="btn btn-primary">Hide logs</button>
-    <table id="complianceLogsGrid" cellspacing="0">
-    </table>
-  </div>
-  <lastreportgrid-missing></lastreportgrid-missing>
-  <lastreportgrid-unexpected></lastreportgrid-unexpected>
+  <lift:authz role="node_read">
+    <lift:authz role="compliance_read">
+      <div id="logRun"  class="space-top" style="display:none;">
+        <h3 class="page-title space-top">All logs for current run</h3>
+        <button id="AllLogHideButton" class="btn btn-primary">Hide logs</button>
+        <table id="complianceLogsGrid" cellspacing="0">
+        </table>
+      </div>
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/server/server_details.html
```diff
@@ -151,7 +151,9 @@ <h3>Groups containing this node <span class="badge" id="nbGroups"></span></h3>
               </div>
 
               <div id="node_logs" class="tab-pane">
-                <div id="logsDetails"></div>
+                <lift:Authz.forbiddenrole role="inventory" unless="compliance_read">
+                  <div id="logsDetails"></div>
+                </lift:Authz.forbiddenrole>
```
clarktsiory: this is for the "Technical logs" tab

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/reports_server.html
```diff
@@ -5,19 +5,24 @@
...
+    <div id="logRun"  class="space-top" style="display:none;">
+      <h3 class="page-title space-top">All logs for current run</h3>
+      <button id="AllLogHideButton" class="btn btn-primary">Hide logs</button>
+      <table id="complianceLogsGrid" cellspacing="0">
+      </table>
+    </div>
+    <lastreportgrid-missing></lastreportgrid-missing>
+    <lastreportgrid-unexpected></lastreportgrid-unexpected>
+  </lift:Authz.forbiddenrole>
+
```
clarktsiory: This is for the "System status" tab.
The `lastreportgrid-grid` element is the one for the system compliance, so it is left as is

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/reports_server.html
```diff
@@ -5,19 +5,24 @@
   <lastreportgrid-intro></lastreportgrid-intro>
 
 
-  <button id="AllLogButton" class="btn btn-primary">Show logs <i class="fa fa-table"></i></button>
+  <lift:Authz.forbiddenrole role="inventory" unless="compliance_read">
```
fanf: I'm not sure why you need a "forbiddenrole" - plus, it's a dangerous way to think about rights, because it doesn't compose. 

Why not just specifying that that bit need `compliance_read` ? Then, you can see it only if you have both `inventory` - actually `node_read` (because of page right access) and also `compliance_read` ? 

What we may reach is the need to be able to express "or", I'm not sure we can now - and is possible with `<lift:authz role="node_write"><lift:authz role="compliance_read">protected, need both node_write and compliance_read</lift:authz></lift:authz>` for

fanf (changes requested):
I really dont' want to have "unless" semantic in right

webapp/sources/rudder/rudder-web/src/main/webapp/templates-hidden/reports_server.html
clarktsiory: right, I just literally thought that "inventory" needed to be considered differently, but `node_read && compliance_read` seems to be the right restriction :+1:

clarktsiory:
PR updated with a new commit

clarktsiory:
Commit modified

clarktsiory:
PR rebased

fanf (approved):

PR #6436, opened 2025-06-09
Fixes #27051: Event logs rollback does not do any action

clarktsiory:
https://issues.rudder.io/issues/27051

the previous condition was always true, the condition needs to be checked the other way around

EDIT: there is some compilation issue in SCSS because of a missing import


amousset (approved):

clarktsiory:
OK, merging this PR

PR #6439, opened 2025-06-11
Fixes #27061: Rudder server depends on cf-promises being in old path

peckpeck:
https://issues.rudder.io/issues/27061


amousset:
I hope this is not a conffile

PR #6444, opened 2025-06-11
Fixes #27067: Error when parsing date for last user loggin awaited format is EEE MMM dd HH:mm

fanf:
https://issues.rudder.io/issues/27067

Nothing fancy: 

- extract the parser into the companion object, 
- make it have a list of possible formats, take the first that works. We assume that we are in english format for the bios. 
- formats don't need to be a parameter of the class, it's a given by fusion and never changed in 15 years. 

Add tests, and a 9.0 inventory that has the alternative format


webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionInventoryParser.scala
```diff
@@ -57,25 +57,50 @@ import scala.xml.*
...
+  private val biosParsers: List[DateTimeFormatter] = biosFormats.map(toFusionDateParsers)
+
+  private val loggedUserFormats: List[String]            = List("EEE MMM dd HH:mm", "EEE MMM dd HH:mm:ss")
+  private val loggedUserParsers: List[DateTimeFormatter] = loggedUserFormats.map(toFusionDateParsers)
+
+  private def toFusionDateParsers(format: String): DateTimeFormatter = {
+    DateTimeFormat.forPattern(format).withLocale(Locale.ENGLISH).withZoneUTC()
+  }
+
+  private def parseDateTime(s: String, parsers: List[DateTimeFormatter], formats: List[String]): Either[String, DateTime] = {
```
clarktsiory, line 72: it seems that `parsers = format.map(toFusionDateParsers)` in all cases, should it not be simplified to defining variables only for `formats` ?

webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionInventoryParser.scala
```diff
@@ -57,25 +57,50 @@ import scala.xml.*
...
+    def parse(p: DateTimeFormatter, s: String): Option[DateTime] = {
+      try {
+        Some(p.parseDateTime(s))
+      } catch {
+        case _: IllegalArgumentException => None
+      }
+    }
+
+    parsers
+      .foldLeft(Option.empty[DateTime]) { case (dt, p) => dt.orElse(parse(p, s)) }
```
clarktsiory, line 82: seems like a `reduceK` in cats :smile_cat:

webapp/sources/ldap-inventory/inventory-fusion/src/main/scala/com/normation/inventory/provisioning/fusion/FusionInventoryParser.scala
fanf: I don't want to redefine the parser at each call, it's rather heavy

PR #6445, opened 2025-06-11
fix: Tests should work regardless of the JVM timezone

mbaechler:


fanf:
Replaced by https://github.com/Normation/rudder/pull/6452, that implementation was causing `==` problem because timezone was not homogeneous in all instances of `DateTime`

PR #6447, opened 2025-06-11
Scala 3 migration

mbaechler:


fanf:
Merged in branche 9.0 - we are Scala 3 !

PR #6448, opened 2025-06-12
Fixes #27081: Update to last version of Scala 3

fanf:
https://issues.rudder.io/issues/27081

For plugins, we need: 
- public: https://github.com/Normation/rudder-plugins/pull/878
- private : https://github.com/Normation/rudder-plugins-private/pull/1123

Below: description of the main change

### Mandatory `using` for explicit implicits

If you pass an implicit parameter explicitly, it is now mandatory to use the keyword `using`. 
Actually it's just a warning in 3.7 but at some point, we will need to pay that tax, it could be when we officially switch to scala 3. 

### No more trailing _ for function application in paramters

Just use the name, without postponing a `_` behind. 

### `_ => XXX` in signature becomes `? => XXX`

The underscore is replaced by `?` when use in a type signature

### Fastparse change impact

See: 
https://stackoverflow.com/questions/77966105/how-to-test-fastparse-parsers-in-a-separate-class

We must change the way we call fastparse. 

### `A with B` is replaced by `A & B` in signature

All in the tible.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
@@ -241,7 +241,7 @@ trait AnalyseInterpolation[T, I <: GenericInterpolationContext[T]] {
       }
     }
 
-    build _
+    build
   }
 
   /*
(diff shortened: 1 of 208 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala
```diff
@@ -497,7 +497,7 @@ object PropertyParser {
   import fastparse.NoWhitespace.*
 
   def parse(value: String): PureResult[List[Token]] = {
-    (fastparse.parse(value, all(_)): @unchecked) match {
+    fastparse.parse(value, { case given P[?] => all }) match {
```
clarktsiory, line 500: interesting

clarktsiory (approved):
👏 we had a lot of explicit calls of implicits
hurray, we have caught up now ! 🎉

fanf:
OK, squash merging this PR

PR #6451, opened 2025-06-12
Fixes #26976: Rudder webapp migration create table ScoreDetails with message constraint

fanf:
https://issues.rudder.io/issues/26976

Add two migration, each for removing the non empty contstraint on message: one for `GlobalScore`, the other for `ScoreDetails`

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
@@ -438,7 +438,7 @@ Create table GlobalScore (
 , details jsonb  NOT NULL
 );
 
-Create table scoreDetails (
+Create table ScoreDetails (
   nodeId  text NOT NULL
 , scoreId text NOT NULL
 , score   score NOT NULL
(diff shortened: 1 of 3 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -438,7 +438,7 @@ Create table GlobalScore (
 , details jsonb  NOT NULL
 );
 
-Create table scoreDetails (
+Create table ScoreDetails (
```
fanf, line 441: for consistency with `GlobalScore` and migration

VinceMacBuche (approved):

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
VinceMacBuche: postgres stores every tables as lowercase (and ignore cases) so you can do whatever you want ! fine to me

PR #6453, opened 2025-06-12
Fixes #27086: Change validation method to find steps need to know if user has rights

clarktsiory:
https://issues.rudder.io/issues/27086

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
@@ -226,15 +226,19 @@ trait WorkflowService {
   def stepsValue: List[WorkflowNodeId]
 
   def findNextSteps(
-      currentUserRights: Seq[String],
-      currentStep:       WorkflowNodeId,
-      isCreator:         Boolean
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
```diff
@@ -226,15 +226,19 @@ trait WorkflowService {
...
 
   def findNextSteps(
-      currentUserRights: Seq[String],
-      currentStep:       WorkflowNodeId,
-      isCreator:         Boolean
+      currentUserRights:       Seq[String],
+      currentStep:             WorkflowNodeId,
+      isCreator:               Boolean,
+      hasValidatorWriteRights: Boolean,
+      hasDeployerWriteRights:  Boolean
```
fanf: It looks  like user rights should be in QC and with a real type, not string

clarktsiory:
PR rebased

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
PR rebased

PR #6435, opened 2025-06-06
Fixes #27047: Export technical logs table into CSV

clarktsiory:
https://issues.rudder.io/issues/27047

Continuation of https://github.com/Normation/rudder/pull/6420

PR #6456, opened 2025-06-13
Fixes #27096: Export change logs table into CSV

clarktsiory:
https://issues.rudder.io/issues/27096

The DOM is the same than in https://github.com/Normation/rudder/pull/6435


clarktsiory:
Commit modified

PR #6460, opened 2025-06-17
Fixes #27103: Update front-end dependencies

clarktsiory:
https://issues.rudder.io/issues/27103

We now need to include ESM modules as the newer JS libraries (here [_jsondiffpatch_](https://github.com/benjamine/jsondiffpatch/releases)) are switching paradigm, or we could not upgrade (this is also the case for _chartjs_, but hopefully it still ships UMD dists). So there is some change in the gulpfile, we need to include the modules tree : there are javascripts imports and a single javascript entrypoint file is no longer sufficient to embark a library.

For _bootstrap_, the dependency upgrade is not breaking, there are simply some warnings that can be ignored from gulp, as discussed [here](https://github.com/twbs/bootstrap/issues/40962#issuecomment-2427637501), and it requires pinning the version of _sass_. Also the _gulp-sass_ plugin API exposes a synchronous version which is [reported to be more efficient](https://www.npmjs.com/package/gulp-sass#usage).

There is some dependency vulnerability which needs to be fixed with an [override](https://docs.npmjs.com/cli/v11/configuring-npm/package-json#overrides) because the package has not an up-to-date npm release but only a fix in [the github source](https://github.com/gulp-sourcemaps/identity-map/pull/12) without any release (we should watch for upcoming versions of this package)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/webapp/secure/configurationManager/changeLogs.html
+++ b/webapp/sources/rudder/rudder-web/src/main/webapp/secure/configurationManager/changeLogs.html
@@ -3,8 +3,7 @@
   <head_merge>
     <title>Rudder - Change logs</title>
     <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/style/rudder/rudder-changelogs.css" media="screen">
-    <script type="text/javascript" data-lift="with-cached-resource" src="/javascript/libs/jsondiffpatch.umd.slim.js"></script>
-    <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/style/libs/html.css" />
+    <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/javascript/libs/jsondiffpatch/lib/formatters/styles/html.css" />
   </head_merge>
   <div class="rudder-template">
     <div class="one-col">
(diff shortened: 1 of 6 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-web/src/main/webapp/secure/configurationManager/changeLogs.html
```diff
@@ -3,8 +3,7 @@
   <head_merge>
     <title>Rudder - Change logs</title>
     <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/style/rudder/rudder-changelogs.css" media="screen">
-    <script type="text/javascript" data-lift="with-cached-resource" src="/javascript/libs/jsondiffpatch.umd.slim.js"></script>
-    <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/style/libs/html.css" />
+    <link type="text/css" rel="stylesheet" data-lift="with-cached-resource" href="/javascript/libs/jsondiffpatch/lib/formatters/styles/html.css" />
```
clarktsiory, line 6: the diff is not working for the moment, see https://github.com/Normation/rudder/pull/6466 to make the ESM module work

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

RaphaelGauthier (approved):

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

clarktsiory:
PR rebased

clarktsiory:
OK, merging this PR

PR #6467, opened 2025-06-19
Fixes #27131: Iterator items that have exactly the same values are considered to be a single object

RaphaelGauthier:
https://issues.rudder.io/issues/27131

We need to compare the index of the modified item, not its value.


clarktsiory (approved):
LGTM

PR #6468, opened 2025-06-19
Fixes #27090: Bad logger used for authentication error

fanf:
https://issues.rudder.io/issues/27090

- only use one logger
- call the logger statically to know what is uses where. 

The rule is: 
- if it's about authencation module configuration, it's just `application` logger, 
- if it's about an actual authentication attempt, it's `application.authentication`. 

There was the same error for REST authentication failure. 

Now:
image

image


clarktsiory (approved):
LGTM

PR #6473, opened 2025-06-22
Fixes #27147: Enable fatal warning and disable variable initialization check

fanf:
https://issues.rudder.io/issues/27147

Enable back `-Werror`. 

I needed to disable some checks: 

- disable safe initiationlization check. There is too much work to do, and that work is hard, long, and not correlated with the switch to scala 3
- remove an existantial warning on `BoxTrait`, because we can't do anything about it, 

And correct two other: 

- `ExecutionBatch`: actually, all `LastRunAvailable` are also `ExpectedConfigAvailable`. That makes little sense for a run with a deleted exepected config, so I'm not sure what happens here, but the encoding say so and scalac3 saw it - there is an `Unexpected` case
- `Query` lexer/parser : there was `lift-json` objects used as intermediary, that we really don't need anymore. Just use direct `zio-json` decoding, simplify everything. Some tests were irrelevant (checking for non empty string while the soundess is done later on)
- some sub-component encoding was relying on uncheck runtime erasure, so I changed it a bit. 

Now, everything compiles without warning :tada: 

Also, I changed `spotless` from `verify` (of very late phase, just before installing) to `validate` (a very early phase, before compilation).

### DIFF
--- a/webapp/sources/pom.xml
+++ b/webapp/sources/pom.xml
@@ -166,9 +166,10 @@ limitations under the License.
             <arg>-language:higherKinds</arg>         <!-- Allow higher-kinded types -->
             <arg>-language:implicitConversions</arg> <!-- Allow definition of implicit functions called views -->
             <arg>-Xmax-inlines</arg><arg>100</arg>
-            <arg>-Ysafe-init</arg>                   <!--  Wrap field accessors to throw an exception on uninitialized access. -->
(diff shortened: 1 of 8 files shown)


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM ! :rocket:

PR #6475, opened 2025-06-23
Fixes #27154: Create CampaignEventsStateHistory table

fanf:
https://issues.rudder.io/issues/27154

Needed by https://github.com/Normation/rudder-plugins-private/pull/1071

This PR allows to split the campaign event table into two parts: 
- the old one, `CampaignEvents`, represents the current state of the campaign, 
- the new one, `CampaignEventsStateHistory`, represents the log of states that the campaign went through. 

The migration adds new states to prepare for the future : `deleted` (not sure if needed), `failure` (to differentiate a user-skipped event from a failed one), and the pre/post hooks.
The migration also change the event that were `skipped` but with an error message to `failure` state. 

The other big change is that now, `state` is only an (enum) type, and not the full JSON with added information. This added information is now stored in `CampaignEventsStateHistory.data` column (jsonb). 
This was needed to be able to prepare the fact that for future `pre/post-hooks` events, we will want to be able to see what happened even if the campaign current state is not the one with details. 

So, that make for a bit of a complicated migration process for existing table in 4 steps.
 Of course for them, only the current state can be added in history. 
The misgration process is (well, I hope) unit tested in `TestMigrateTableCampaignEvents`.

Then, other changes are mainly impact of that. I tried to make as little changes as possible in external APIs, so the added tests for the campaign API. Nothing changed here, even if backend storage is totaly different. 

The business representation changed a bit, but not that much once I finally found a satisfying encoding: 
- there is a new enum(eratum) for the state kind: `CampaignEventStateType`
- it's used in another sealed hierachy, which is NOT an enum, `CampaignEventState`. This data structure is the one derived for API and matches encoding in 8.3
- in the DB part, I used intermediate data structure to match table column for insert, and a broader one for select that is join between the two
- then there is dedicated data structure for each kind of details and their json encoder/decoder (which are just `derives` thanks to scala 3)

I chose to not have the state kind in the `data` column, only the actual optionnal details, so that `finished`/etc has not details, and for `skipped` there is no duplication of the state part, only a `reason`. 
This is nice when working in sql and avoid having to think about inconsistencies between columns. 
On the other hand, it means that I need to test each decoder one after the other to find the correct one. 

And that's in that PR. 
The next one will be about actually adding pre/post-hooks state in the workflow engine and corresponding execution.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
@@ -41,12 +41,12 @@ import com.normation.GitVersion
 import com.normation.GitVersion.Revision
 import com.normation.NamedZioLogger
 import enumeratum.*
+import io.scalaland.chimney.*
+import java.time.Instant
 import java.time.ZoneId
 import org.joda.time.DateTime
 import org.joda.time.DateTimeZone
-import zio.json.jsonDiscriminator
-import zio.json.jsonField
-import zio.json.jsonHint
+import zio.json.*
 
 case class CampaignParsingInfo(
     campaignType: CampaignType,
@@ -93,14 +93,18 @@ case class CampaignId(value: String, rev: Revision = GitVersion.DEFAULT_REV) {
 
   def withDefaultRev: CampaignId = this.copy(rev = GitVersion.DEFAULT_REV)
 }
-object CampaignId                                                            {
-  // parse a directiveId which was serialize by "id.serialize"
+
+object CampaignId {
+
+  // parse an id which was serialized by "id.serialize"
   def parse(s: String): Either[String, CampaignId] = {
     GitVersion.parseUidRev(s).map {
       case (id, rev) =>
         CampaignId(id, rev)
     }
   }
+
+  implicit val codecCampaignId: JsonCodec[CampaignId] = JsonCodec.string.transformOrFail(CampaignId.parse, _.serialize)
 }
 
 sealed trait CampaignStatusValue extends EnumEntry {
@@ -253,6 +257,38 @@ trait CampaignDetails
 
 case class CampaignType(value: String)
 
+object CampaignType {
+  implicit val campaignTypeEncoder: JsonEncoder[CampaignType] = JsonEncoder[String].contramap(_.value)
+  implicit val decoderCampaignType: JsonDecoder[CampaignType] = JsonDecoder.string.map(CampaignType.apply)
+}
+
+case class CampaignEventId(value: String)
+
+object CampaignEventId {
+  implicit val campaignEventIdDecoder: JsonDecoder[CampaignEventId] = JsonDecoder[String].map(CampaignEventId.apply)
+  implicit val campaignEventIdEncoder: JsonEncoder[CampaignEventId] = JsonEncoder[String].contramap(_.value)
+}
+
+/*
+ * Our different kind of states and their stable names (need compat between rudder version).
+ */
+sealed abstract class CampaignEventStateType(override val entryName: String) extends EnumEntry
+
+object CampaignEventStateType extends Enum[CampaignEventStateType] {
+  case object Scheduled extends CampaignEventStateType("scheduled")
+  case object Running   extends CampaignEventStateType("running")
+  case object Finished  extends CampaignEventStateType("finished")
+  case object Skipped   extends CampaignEventStateType("skipped")
+  case object Deleted   extends CampaignEventStateType("deleted")
+  case object Failure   extends CampaignEventStateType("failure")
+
+  override def values: IndexedSeq[CampaignEventStateType] = findValues
+
+  implicit val encoder: JsonEncoder[CampaignEventStateType] = JsonEncoder[String].contramap(_.entryName)
+  implicit val decoder: JsonDecoder[CampaignEventStateType] =
+    JsonDecoder[String].mapOrFail(withNameInsensitiveEither(_).left.map(_.getMessage))
+}
+
 case class CampaignEvent(
     id:           CampaignEventId,
     campaignId:   CampaignId,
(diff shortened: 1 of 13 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/earlyconfig/db/TestMigrateTableCampaignEvents.scala
```diff
@@ -0,0 +1,427 @@
...
+        )
+      )
+    }
+  }
+
+  "Repository on new table" should {
+    org.slf4j.LoggerFactory
+      .getLogger("sql")
+      .asInstanceOf[ch.qos.logback.classic.Logger]
+      .setLevel(ch.qos.logback.classic.Level.TRACE)
```
clarktsiory: its for debug ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
```diff
@@ -253,6 +257,33 @@ trait CampaignDetails
...
+  case object Running   extends CampaignEventStateType("running")
+  case object Finished  extends CampaignEventStateType("finished")
+  case object Skipped   extends CampaignEventStateType("skipped")
+  case object Deleted   extends CampaignEventStateType("deleted")
+  case object Failure   extends CampaignEventStateType("failure")
+
+  override def values: IndexedSeq[CampaignEventStateType] = findValues
+
+  implicit val codecCampaignEventState: JsonCodec[CampaignEventStateType] = DeriveJsonCodec.gen
+}
```
clarktsiory: I find the definition of the `codec` confusing : when deriving a codec, does it use the `entryName` ? If so, couldn't it use the `derives` syntax then ?

Also, why not use the `EnumEntry.Lowercase` to be 100% sure to specialize the entryname to the "lowercase" version of the enum case ? (even if we hardcode `"scheduled"`, `"running"`, etc. for grep)

clarktsiory (changes requested):
LGTM overall, only some question, and one debug statement left

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
fanf: For `derives` : I decided to add it late in the dev, and that part was already OK. I will look to use it. 
I don't think the codec uses the entry name. I'm not sure there is a way to tell `zio-json` "use that field value to differentiate between sealed member names". You can individually specify the name with `@jsonHint`. You can tell it to use a field in place of the `"scheduled": { "..." }` construct. But I didn't find a way to say "use that attribute value for jsonhint value"

fanf: I'm not sure where you want me to add `EnumEntry.Lowercase` ? It's not compatible with specifying `entryName`, since it override the field and we do, too, with the abstract class that should be a parametrized trait.

clarktsiory: Okay, why not encode/decode using the "defaults" for enumeratum Enums then, i.e. :
```scala
implicit val encoder: JsonEncoder[MyEnum] = JsonEncoder[String].contramap(_.entryName)
implicit val decoder: JsonDecoeder[MyEnum] = JsonDecoder[String].mapOrFail(withNameInsensitiveEither(_))
```
? (I wonder why we don't have a generic encoder for any type that extends `EnumEntry`, since we do that many times, a circe library that integrates with enumeratum to do exactly that even exists, but not with zio-json)

clarktsiory: Oh yes here the `Lowercase` is supposed to override the `entryName` but I thought that the argument of the abstract class was not an `entryName`, we were used to defining a `name` argument before using enumeratum.

So the `entryName` mandatory argument looks like it "should be" lowercase, but it's not strictly, nothing prevents later from adding another case with an uppercase argument, plus if it's conventionally lowercase, it is repetitive. If it is for grep-ing, I personally don't see the benefit here since I end up grep-ing those words without case-sensitivity (I agree for other cases with kebab-case or camelCase, but I think just having `Lowercase`/`Uncapitalised` is fine)

fanf: yep, clearly it should be generics.
For the why not: I think I just oscillated too many times with too many encoding to loop back to that.

fanf: (the main problem/strangeness is for the `CampaignEventState` hierarchy)

fanf: it's not only for greping. It's for stability accross refactoring and avoiding the "oups, I changed the name of the scala case object and broke APIs". That could be handled by unit testing each case for exact form, but 1/ we're nowhere near that in unit testing coverage, and 2/ it's the kind of unit testing that is extremelly fastidious, easy to overlook when adding a new case, and easy to overlook when we do a change ("ok, the name changed in code, it changes there too"). 

If you DO write the entry name explicitly, it is a STRONG signal that the naming was a deliberate choice with IMPACTS. 
Each time you automate somethings (with a marker, a `derives`, etc), you make change on those things more automated too, and shut-down you brain and let the autopilot change code. Which is even more relevant in a world with agentic-thinks everywhere. So yes, it's willing choice.

fanf: the rule of thumb is: if there's no external API using the name you chose, whatever you want. If there is, make it very visible, and as fastidious as the impact is to change the name. If it's an external API, put the name at several places cross-referencing and without automated way to change them all easily.

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/earlyconfig/db/TestMigrateTableCampaignEvents.scala
clarktsiory: it's still left there, is it necessary for the tests output or is it for local debug purpose ? (
in the latter case it can just be commented)

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/test/scala/bootstrap/liftweb/checks/earlyconfig/db/TestMigrateTableCampaignEvents.scala
fanf: no, I just haven't deal with that one yet

clarktsiory:
OK, squash merging this PR

PR #6483, opened 2025-06-27
Fixes #27128: Drop support for legacy password hash algorithms

amousset:
https://issues.rudder.io/issues/27128


fanf (approved):
Nice, I really like that PR :)

PR #6484, opened 2025-06-30
Fixes #27173:  Add and Remove/deprecate API for Rudder 9.0 - version 22

fanf:
https://issues.rudder.io/issues/27173

- move currently supported API versions from `RudderConfig` to a dedicated object `SupportedApiVersions` in `rudder-core`
- use these value by default in tests, so that we can check that we correctly support the last version (without doing anything)
- remove deprecated versions < 21 (ie, keep Rudder 8.3 and up)
- remove code used for compatibility in older, deleted versions (old node format, old allowed networks, old system info, etc)
- add documentation for API version 21 (Rudder 8.3) and 22 (Rudder 9.0)


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM, nice code deletion !

PR #6485, opened 2025-07-01
Fixes #27174: CA list is not initialized at installation

peckpeck:
https://issues.rudder.io/issues/27174


peckpeck:
OK, merging this PR

PR #6490, opened 2025-07-03
Fixes #27145: Creating global parameter with change-validation enabled leads to 404

skaerg:
https://issues.rudder.io/issues/27145


skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
```diff
@@ -135,8 +135,14 @@ class LinkUtil(
...
-  def redirectToChangeRequestLink(id: ChangeRequestId): JsCmd =
-    RedirectTo(baseChangeRequestLink(id))
+  def redirectToChangeRequestLink(id: ChangeRequestId, contextPath: String = S.contextPath): JsCmd = {
+    if (S.contextPath == "") {
+      RedirectTo(s"${contextPath}${baseChangeRequestLink(id)}")
+    } else {
+      RedirectTo(baseChangeRequestLink(id))
+    }
+
+  }
```
clarktsiory: I don't see why there is a check on the default value `S.contextPath` : when we pass an explicit contextPath then it means that we want to always use it, and not only when there is a issue leading to an empty one...
So it could be simplified to `RedirectTo(changeRequestLink(id, contextPath))` 

And I'm not even sure that `S.contextPath == ""` is disallowed, in a server with only a single webapp

clarktsiory (changes requested):
LGTM on the IOResult refactoring, only the link util require some change

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
skaerg: alas, `RedirectTo` automatically adds the context path as a prefix when `S.contextPath` is set , i.e. calling `RedirectTo(changeRequestLink(id))` when `S.contextPath` is set would return something that starts with `rudder/rudder/...`
I agree that this solution is not the most elegant ; could there be another way to do this ?

clarktsiory: So, to keep the API of the link utils look predictable, I suggest to redefine another `RedirectTo` that skips the check on the context path, since the issue is that the current one adds the context path wrt some logic, and because we don't want that logic when we pass an explicit `contextPath`.

It would then end up into : 
```scala
  if (S.contextPath != contextPath) {
    RedirectToStrict(s"${contextPath}${baseChangeRequestLink(id)}")
  } else {
    RedirectTo(baseChangeRequestLink(id))
  }
```
if I'm not mistaken

skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/model/JsInitContextLinkUtil.scala
```diff
@@ -132,11 +132,16 @@ class LinkUtil(
...
 
-  def redirectToChangeRequestLink(id: ChangeRequestId): JsCmd =
-    RedirectTo(baseChangeRequestLink(id))
+  def redirectToChangeRequestLink(id: ChangeRequestId, contextPath: String = S.contextPath): JsCmd = {
+    RedirectToFullPath(changeRequestLink(id, contextPath))
+  }
+
+  case class RedirectToFullPath(where: String) extends JsCmd {
+    override def toJsCmd: String = s"window.location = \"${where}\" ;"
+  }
```
clarktsiory: if used in with other IDs which are mostly strings, this has some XSS vulnerability because the location is not escaped.
So it should keep the same security guards as in the Lift code with `encJs`

clarktsiory (changes requested):

skaerg:
Commit modified

clarktsiory (approved):
LGTM, this is likely to happen in other places due to progress in migrating away from Lift

clarktsiory:
OK, merging this PR

PR #6491, opened 2025-07-03
Fixes #27201: Adding linux GM for the template module

m4rtinh4rt:
https://issues.rudder.io/issues/27201


m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

policies/lib/tree/30_generic_methods/file_from_template_options.cf
```diff
@@ -0,0 +1,105 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File from template
+# @description Build a file from a template (MiniJinja, Jinja2, mustache)
```
Fdall:
```suggestion
# @description Build a file from a template (MiniJinja, Jinja2, Mustache)
```

policies/lib/tree/30_generic_methods/file_from_template_options.cf
```diff
@@ -0,0 +1,105 @@
...
+# @documentation Coming soon.
+#
+# @parameter            destination Destination file (absolute path on the target node)
+# @parameter_constraint destination "allow_empty_string" : false
+# @parameter            engine Template engine (defaults to minijinja)
+# @parameter_constraint engine "allow_empty_string" : true
+# @parameter_constraint engine "select" : [ "", "minijinja", "jinja2", "mustache" ]
+# @parameter            data Data provided to the template
```
Fdall:
```suggestion
# @parameter            data Data provided to the template, defaults to the global context if empty
```

policies/lib/tree/30_generic_methods/file_from_template_options.cf
```diff
@@ -0,0 +1,105 @@
...
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+#####################################################################################
+
+# @name File from template
+# @description Build a file from a template (MiniJinja, Jinja2, mustache)
+# @documentation Coming soon.
```
Fdall, line 21: Will be needed before release

policies/lib/tree/30_generic_methods/file_from_template_options.cf
```diff
@@ -0,0 +1,105 @@
...
+# @parameter_constraint template_string "allow_empty_string" : true
+# @parameter            template_path  Source file containing a template to be rendered (absolute path on the target node)
+# @parameter_constraint template_path "allow_empty_string" : true
+# @parameter            show_content Show the file content in the report (defaults to true)
+# @parameter_constraint show_content "allow_empty_string" : true
+# @parameter_constraint show_content "select" : [ "", "true", "false" ]
+#
+# @class_prefix file_from_template_options
+# @class_parameter destination
+# @agent_support = ["cfengine-community"]
```
Fdall:
```suggestion
# @agent_support = ["cfengine-community", "dsc"]
```

policies/lib/tree/30_generic_methods/file_from_template_options.cf
```diff
@@ -0,0 +1,105 @@
...
+   # "datastate" string => string_mustache("{{%-top-}}"),
+   #			unless => isvariable($(this.promiser));
+   "datastate_path" string => "/var/rudder/tmp/datastate-${this.promiser_pid}.json";
+
+  classes:
+      "pass3" expression => "pass2";
+      "pass2" expression => "pass1";
+      "pass1" expression => "any";
+
+  files:
```
Fdall, line 88: The removal of the datastate file is missing. See https://github.com/Normation/rudder/blob/11ef4db06f575f104b91690682abb30a5a607f3d/policies/lib/tree/30_generic_methods/file_from_template_jinja2.cf#L291

policies/module-types/template/src/lib.rs
```diff
@@ -216,15 +219,8 @@ impl ModuleType0 for Template {
...
-            parameters.template_path.is_some(),
-            parameters.template_src.is_some(),
-        ) {
-            (true, true) => bail!("Only one of 'template_path' and 'template_src' can be provided"),
-            (false, false) => {
-                bail!("Need one of 'template_path' and 'template_src'")
-            }
-            _ => (),
+        if let (None, None) = (parameters.template_path, parameters.template_src) {
+            bail!("Need one of 'template_path' and 'template_src'")
```
Fdall: You removed the error when both of them are used at the same time

policies/module-types/template/src/lib.rs
```diff
@@ -236,12 +232,27 @@ impl ModuleType0 for Template {
         let output_file = &p.path;
         let output_file_d = output_file.display();
 
-        let output = match p.engine {
-            Engine::Mustache => {
-                Engine::mustache(p.template_path.as_deref(), p.template_src, p.data)?
+        let data = match (p.data.clone(), p.datastate_path) {
+            (Value::String(s), _) if !s.is_empty() => p.data,
```
Fdall: Why are the first 4 cases needed ?

policies/module-types/template/src/lib.rs
m4rtinh4rt: Both `template_path` and `template_src` are passed to the module. The previous validation logic was incorrect; there is a [condition](https://github.com/Normation/rudder/blob/002d14458d3334ea4658cb45b76e637ec93067fd/policies/module-types/template/src/lib.rs#L156) on the content to determine which one to use.


```
/// Source template path
#[serde(skip_serializing_if = "Option::is_none")]
template_path: Option<PathBuf>,
/// Inlined source template
#[serde(skip_serializing_if = "Option::is_none")]
template_src: Option<String>
```


The current serde serialization used does not return None if empty; None is returned only if the variable is not passed to the module. Therefore, a condition to check if the content is empty is needed, or a change to handle that in the serialization

Fdall: I feel like we still need an explicit error message for the end user when both options are used and not empty

policies/module-types/template/src/lib.rs
m4rtinh4rt: Depending on what is entered in the technique editor, the string sent to the module can be serialized into different JSON types, since the serialization does not return None if empty (same as the previous comment). Checks are required to handle this.

I am going to look into handling this case directly in the serialization to fix that and the issue mentioned in the previous comment.

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

policies/module-types/template/src/lib.rs
```diff
@@ -274,21 +274,21 @@ impl ModuleType0 for Template {
...
+                            datastate_path.to_string_lossy()
+                        )
+                    })?;
+                    serde_json::from_str(&data)?
+                } else {
+                    bail!("Could not get datastate file")
+                }
             }
-            _ => bail!("Could not get datastate file"),
+            v => v,
```
Fdall, line 291: ?

m4rtinh4rt: The other Serde values cannot be empty. The emptiness check is here to determine if an empty string is passed to the module. If so, we attempt to load the datastate.

PR #6493, opened 2025-07-03
Fixes #27200: Add a global warning and alert banner in user management with unsafe hashes

clarktsiory:
https://issues.rudder.io/issues/27200


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/UserManagementApi.scala
```diff
@@ -719,3 +718,48 @@ object UserManagementApiImpl {
...
+ * The internal API
+ */
+sealed trait UserManagementInternalApi extends EnumEntry with EndpointSchema with InternalApi with SortIndex
+
+object UserManagementInternalApi extends Enum[UserManagementInternalApi] with ApiModuleProvider[UserManagementInternalApi] {
+
+  final case object SafeHashes extends UserManagementInternalApi with ZeroParam with StartsAtVersion20 {
+    val z              = implicitly[Line].value
+    val description    = "Get the status of password encoder being used"
+    val (action, path) = GET / "usermanagementinternal" / "safe-hashes"
```
fanf: We don't have api with that case, even if internal please stick to conventions

amousset:
I think the message is not actionnable enough. We need to point to some docs containing more precise instruction, like https://www.notion.so/rudderio/Drop-des-hash-insecure-pour-les-users-Rudder-21e8750cc48380d89513cbcd2592c982?source=copy_link#2228750cc4838087a471f9dd534db24f

It could be part of the upgrade notes.

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

PR #6496, opened 2025-07-03
Fixes #27175: Plugins get silently disabled when number of licensed nodes exceeded.

clarktsiory:
https://issues.rudder.io/issues/27175

* the plugin status displayed in the Plugins page was only taking into account the status from rudder-package, not from the webapp which does license runtime check :
  * use the webapp plugin definition to return the correct `status` and details with the `statusMessage` field
  * display the `statusMessage` message as an error in Elm, it behaves just like when license is expired/missing
* refactor : 
  * some duplicate code in the snippet with the plugins icon warning
  * the Elm callouts and handling of license status to avoid code repetition/duplication

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginData.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/plugins/PluginData.scala
@@ -65,11 +65,12 @@ object PluginId {
   }
 }
 
-final case class Licensee(value: String)      extends AnyVal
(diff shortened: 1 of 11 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM, works as expected!

clarktsiory:
OK, squash merging this PR

PR #6497, opened 2025-07-03
Fixes #27180: Node compliance persistence fails when all reports from sysevents are cleaned

fanf:
https://issues.rudder.io/issues/27180

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
@@ -160,20 +160,34 @@ class NodeStatusReportRepositoryImpl(
                          // if we are asked to keep compliance, start by checking we do have one
                          case RunAnalysisKind.KeepLastCompliance =>
                            m.get(id) match {
-                             case Some(e) =>
-                               e.runInfo.kind match { // here we need to check for all expiring status from `def outDatedCompliance`
+                             case Some(cached) =>
+                               cached.runInfo.kind match { // here we need to check for all expiring status from `def outDatedCompliance`
                                  case RunAnalysisKind.ComputeCompliance | RunAnalysisKind.Pending =>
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
```diff
@@ -160,20 +160,34 @@ class NodeStatusReportRepositoryImpl(
...
                                    // If the saved node is pending, it means that we are in a case where the node never saved
                                    // anything else than pending since it was first processed, so the node never had a computed
                                    // compliance (it was a new report, directly in pending). Let it blue.
                                    // In both case: keep existing compliance, change run info to avoid loop.
-                                   Some((id, e.modify(_.runInfo).setTo(report.runInfo)))
+
+                                   // we need to copy new report information BUT NOT the one regarding "last run",
+                                   // which were lost if we reach that point
+                                   val runInfo = report.runInfo.copyLastRunInfo(cached.runInfo)
+                                   Some((id, cached.modify(_.runInfo).setTo(runInfo)))
```
ncharles, line 175: i'm unsure about this change: I don't see when the received report actually gets to the cache - the runInfo is overriden by the cached.runInfo

fanf: it was already in the cache. The case for "not in the cache" is the last one

fanf: ah yes, sorry I was thinking of the line below. That part is a false-start, and was not deleted when the part below was added

fanf: Ok, no it is actually correct. I'm adding a test to be assured of that.

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #6500, opened 2025-07-07
Fixes #27213: Allow to distribute Linux policies in tar.gz

fanf:
https://issues.rudder.io/issues/27213

So, the idea is to have a PoC with a `policy-generation-node-finished` hook that will do the archive/compression. 

The script: 
- create an archive in `/var/rudder/shate/${nodeid}/rules/policies.tar.bz2`
- the archive is the content of `/var/rudder/shate/${nodeid}/rules/cfengine-community` with the base directory renamed to `policies_next` 
-  `/var/rudder/shate/${nodeid}/rules/cfengine-community/` is not deleted (so that old nodes can still use that)

The benefits are : 

- it's extermelly quick to reach an E2E working solution and not block the whole feature any longer. 
- it's will become symetric to what is done for Windows agent, so can go forward on the next steps (performance, node generation rework) for both at the same time
- it's easier to tweak for non scala developers for the PoC duration (even for 9.0)

The cons are: 

- well, performance wise, it's Not Good : we can't implement an all-in-memory solution to policy generation with that
- it adds a dependency to `bzip2` (or other compression algorithm). 
- it's harder to precisely define what nodes get the new scheme. Here, it's all-or-none solution. We can disable it globally by making the hook non executable, but we can't defined by-group, or (not easily) by-node preferences. 

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-finished/91-compress-linux-policies
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-finished/91-compress-linux-policies
@@ -1,0 +1,40 @@
+#!/bin/sh
+
+# Hooks parameter are passed by environment variable:
+#
+# - RUDDER_GENERATION_DATETIME: generation datetime: ISO-8601 YYYY-MM-ddTHH:mm:ss.sssZ date/time that identify that policy generation start
+# - RUDDER_NODEID             : the nodeId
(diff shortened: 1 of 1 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
See #6505 for the error in comment for `RUDDER_POLICIES_DIRECTORY`

fanf:
OK, squash merging this PR

PR #6507, opened 2025-07-08
Fixes #27234: Upmerge of user API definition breaks with Scala 3

clarktsiory:
https://issues.rudder.io/issues/27234

This is true :
```scala
[ERROR] 743 |  final case object SafeHashes extends UserManagementInternalApi with ZeroParam with StartsAtVersion20 {
[ERROR]     |  ^^^^^
[INFO]     |  Modifier final is redundant for this definition
```


clarktsiory:
PR updated with a new commit : 

There are also unused imports

clarktsiory:
OK, squash merging this PR

PR #6510, opened 2025-07-08
Fixes #27245: Inventory user sees technical logs table

clarktsiory:
https://issues.rudder.io/issues/27245

Completely hide the tab when it's not relevant. From the data that the table contains, it requires `node_read` and `compliance_read` permissions (so it's the same as in https://github.com/Normation/rudder/pull/6432)


fanf (approved):
Ok

PR #6515, opened 2025-07-10
Fixes #26767: 

clarktsiory:
https://issues.rudder.io/issues/26767

For the Api account, the token is already redacted, we need the same for the user password : add a dedicated structure.

There are some sub-types, and we can avoid invalid representations (empty password, and secret unhashed password was used in the same `String` field)

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/User.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/User.scala
@@ -45,6 +45,8 @@ import com.normation.rudder.api.ApiAccount
 import com.normation.rudder.api.ApiAuthorization
 import com.normation.rudder.facts.nodes.NodeSecurityContext
 import com.normation.rudder.facts.nodes.QueryContext
+import com.normation.rudder.users.UserPassword.HashedUserPassword
+import com.normation.rudder.users.UserPassword.RandomHexaPassword
 import java.util.Collection
 import net.liftweb.http.RequestVar
 import org.springframework.security.core.GrantedAuthority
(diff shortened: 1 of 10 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf:
CI fails with something that seems legit: 
```
[2025-07-11T10:15:41.353Z] [ERROR]   RudderUserDetailsTest.an account without password field get a random 32 chars pass » TestFailed ✗ 1 was equal to 1
```

clarktsiory:
PR updated with a new commit : indeed the test did not pass because the password we get is purposefully a random one, I added a specific subtype for that

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/User.scala
```diff
@@ -122,9 +124,11 @@ case class RudderUserDetail(
   }
 
   override val (getUsername, getPassword, getAuthorities) = account match {
-    case RudderAccount.User(login, password) => (login, password, RudderAuthType.User.grantedAuthorities)
+    case RudderAccount.User(login, h: HashedUserPassword) => (login, h.exposeValue(), RudderAuthType.User.grantedAuthorities)
+    case RudderAccount.User(login, r: RandomHexaPassword) => (login, r.randomValue, RudderAuthType.User.grantedAuthorities)
+    case RudderAccount.User(login, _)                     => (login, "", RudderAuthType.User.grantedAuthorities)
```
clarktsiory, line 129: so, this makes me doubt : should we generate a random one ? it seems to match the case of "externally provided user".. 
from the comment explaining why we generate a random one, it seems like we want to avoid an unknown/empty password to match by accident, so could we actually generate a random one for the remaining cases (`UnknownPassword` and `SecretUserPassword`) ?

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit : for some reason the rebase was not done correctly, now it should...

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #6521, opened 2025-07-16
Fixes #27296: POLICY_SERVER_KEY_HASH must not happen a semicolon when empty

fanf:
https://issues.rudder.io/issues/27296

So, we were always adding the ";" event when `POLICY_SERVER_KEY_HASH` was empty. I corrected that by checking for the special empty case, and added a test for that - we were only testing the case were it was non empty.


fanf:
PR updated with a new commit

peckpeck (approved):

peckpeck:
OK, squash merging this PR

PR #6525, opened 2025-07-17
Fixes #27211: Cannot remove all rights from a user who has a single role, from the User management GUI

clarktsiory:
https://issues.rudder.io/issues/27211


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserManagement.scala
```diff
@@ -453,18 +457,27 @@ final case class JsonStatus(
 final case class JsonUserFormData(
     username:    String,
     password:    String,
-    permissions: List[String],
+    permissions: Option[List[String]],
```
fanf, line 460: I would much, much more prefer a sealed classe with either norights or a non empty list of non empty rights. But I'm not sure it earns us much, since the semantic parsing of rights is not done when json is decoded, but latter. So perhaps it's better to just keep the logic there.

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
```diff
@@ -397,6 +397,281 @@ response:
...
+  code: 200
+  content: >-
+    {
+      "action" : "updateUser",
+      "result" : "success",
+      "data" : {
+        "updatedUser" : {
+          "username" : "user1",
+          "password" : "",
+          "permissions" : []
```
fanf, line 421: this is not what we want. It's a lie, the user permission is `[user]`

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
```diff
@@ -397,6 +397,281 @@ response:
...
```
fanf, line 643: we can't be sure it's OK, since it always returned what it was given in input

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
```diff
@@ -397,6 +397,281 @@ response:
...
+        }
+      }
+    }
+---
+description: Update a user with no_rights
+method: POST
```
fanf, line 787: shouldn't be get now ? It looks like a duplicate of the previous one ?

fanf (changes requested):
There is something fishy with the returned value in upated user when perms are not updated

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
clarktsiory: yes, it is there because the field is required in the response, so now when not updating permissions we need to fetch them and look in the file, but we forget the content of the file during the update, so we need another way to do that, I put it in another issue : https://issues.rudder.io/issues/27310

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
clarktsiory: it's not a duplicate but an actual update, resulting in "empty permissions" for the user, all the permissions are removed for the user

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserManagement.scala
clarktsiory: You mean, inside of the `Option` ? The `Option` is necessary and the point of this PR : it's an "optional update"/"patch it or leave it", but when patching, we are not parsing the list of permissions as "Rudder rights". They are actually `String` values that are put in the file as is (I guess, because with custom roles there could be some "invalid" rights at some point that will be valid later)

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
fanf: ok, but can you add "get" requests between each case so that we DO SEE the change in the user ? Until https://issues.rudder.io/issues/27310 there is no way to know what happened server side, so the test is not very informative

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserManagement.scala
fanf: yes, inside the option, I understand the option part is for "optional update". 
I'm a bit wary with the idea that we are just accepting random strings there, but ok, they can't be fully parsed rights since there is plugins and etc. Perhaps we should at least check they are "well formed" for rights, to at least avoid having obvious injection. I don't see that (perhaps I'm missing it)

clarktsiory: We are expressly adding non-parsable roles/authz because there is no check for "custom roles" or "plugin roles", as for authz. The format of authz is `xx_(read|edit|write)`, but we don't have any format for roles so I don't see any way to avoid this.

And if we don't accept unknown strings, the behavior of file reload may be really strange, especially when editing custom roles, it may lead to users having "disappearing roles" on refresh because the roles have been refactored in some order for example

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
clarktsiory: done in https://github.com/Normation/rudder/pull/6525/commits/ea21b8bb4cbb0593e6ee019737bcaa3d3e12d3c7

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserManagement.scala
fanf: yes, something to enhance / make stricter in future versions

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_usermanagement.yml
```diff
@@ -397,6 +397,419 @@ response:
...
+          "username" : "user1",
+          "password" : "",
+          "permissions" : []
+        }
+      }
+    }
+---
+description: Get users after resetting a user permission
+method: GET
+url: /api/latest/usermanagement/users
```
fanf, line 650: can you only `GET /api/latest/usermanagement/update/user1` So that there's less jitter around in future PR?

clarktsiory: so uuuh... there is no endpoint to get a single user :sweat_smile:

fanf: AH

fanf: well, it will do then :)

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #6527, opened 2025-07-18
Fixes #27119: CSP headers for pages without scripts are always set with static nonce

clarktsiory:
https://issues.rudder.io/issues/27119

Keep the `script-src: 'nonce-<randomnonce>' 'script-dynamic'` in pages where the nonce has been generated to be put in script tags (`<script nonce="<therandomnonce>/>`),

But in other pages (e.g. served files, like `css`, `svg`, ...) we don't have script tags and the nonce has not been generated so we need `script-src: 'none'` (it's [the safest policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/script-src#none))


fanf (approved):
ok, complies with what we chose to do

PR #6534, opened 2025-07-22
Fixes #27329: Move the graph creation functions from homepage.js to another js file.

RaphaelGauthier:
https://issues.rudder.io/issues/27329

This PR makes it possible to use graph initialisation functions on all Rudder pages (including plugins) without having to use the homepage.js file (which, as its name suggests, is now specific to the homepage).


RaphaelGauthier:
PR replaced by https://github.com/Normation/rudder/pull/6536

PR #6541, opened 2025-07-25
Fixes #27364: Add a generic json codec for enumeratum

fanf:
https://issues.rudder.io/issues/27364

An enumeratum json codec that can be used with:

```
import enumeratum.*
import zio.json.other.*

  sealed trait A extends EnumEntry

  object A extends Enum[A] with EnumCodec[A] {
    case object CaseOne extends A
    case object casetwo extends A

    override def values: IndexedSeq[A] = findValues
  }
```

It comes with two variants that differ only in the parsing: 
-  `EnumCodecCaseSensitive[A]` : expects exactly `enumEntry` value
- `EnumCodecCaseInsensitive[A]`: pase case insensively

The default is case insensitive parsing, given by the alias `EnumCodec[A]`

### DIFF
--- a/webapp/sources/utils/src/main/scala/zio/json/other/package.scala
+++ b/webapp/sources/utils/src/main/scala/zio/json/other/package.scala
@@ -1,0 +1,42 @@
+package zio.json
+
+import _root_.enumeratum.*
+
+package object enumeratum {
+
+  def buildDecoder[A <: EnumEntry](cn: Enum[A], parse: String => Either[NoSuchMember[A], A]): JsonDecoder[A] = {
+    JsonDecoder.string.mapOrFail { s =>
+      parse(s).left.map { e =>
+        s"Error decoding enum '${cn.getClass.getSimpleName
+            .replaceAll("\\$", "")}': ${e.getMessage} ; values: ${cn.values.mkString("'", "','", "'")}"
+      }
+    }
+  }
+
+  /*
+   * Exact codec for EnumEntry
+   * It is isomorphic to enum `entryName`, case sensitive
+   */
+  trait EnumCodecCaseSensitive[A <: EnumEntry] {
+    this: Enum[A] =>
+
+    implicit val decoderEnumEntry: JsonDecoder[A] = buildDecoder(this, this.withNameEither)
+    implicit val encoderEnumEntry: JsonEncoder[A] = JsonEncoder.string.contramap(_.entryName)
+  }
+
+  /*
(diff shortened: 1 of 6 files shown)


webapp/sources/utils/src/test/scala/zio/json/other/EnumCodecTest.scala
```diff
@@ -0,0 +1,138 @@
...
+import org.junit.runner.RunWith
+import zio.*
+import zio.json.*
+import zio.test.*
+import zio.test.Assertion.*
+import zio.test.junit.ZTestJUnitRunner
+
+/*
+ * It's absolutely mandatory to define that out of ZIOSpecDefault
+ * because macro used in it breaks in cryptic way enumeratum "findValues" macro
```
clarktsiory, line 51: 🤯

webapp/sources/utils/src/main/scala/zio/json/other/package.scala
```diff
@@ -0,0 +1,42 @@
+package zio.json
+
+import enumeratum.*
+
+package object other {
```
clarktsiory: why `other` ? to me it is not very intuitive, with ZIO there is the `interop` namespace.
Is the idea the same : to have a single namespace for integration with "external" libraries ?

Another solution would be a namespace for each library, e.g. `zio.json.enumeratum`, wouldn't it be the prefered solution ?

fanf: I agree `other` is not a very good name and I used it as a placeholder for discussion.

We can't contribute easily to existing namespaces, it creates infinite shadowing problems. Typically, I would have love to have it under `zio.json` directly, since it's just one more encoder. But it causes mayhem 

We can use `zio.json.enumeratum`, but it means we need one import for each of our own codec.

clarktsiory: I think the `zio.json.enumeratum` is the most suited way, it also seems to be same naming convention in the ZIO ecosystem, e.g. for zio-config in https://zio.dev/zio-config/integrations#enumeratum it's `zio.config.enumeratum`

For the import, I don't see any other way if we can't have it directly under `zio.json`

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM but CI is failing because of a missing spotless

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #6543, opened 2025-07-28
Fixes #27365: Allow per-hook-kind logger

fanf:
https://issues.rudder.io/issues/27365

The main problem to adress is the cost of multiplying loggers classes in the case of per-campaign (or other ephemeral objects) loggers. 
We could add a cache for the "hook" loggers, which is quite simple, but it duplicates the work of logback that already does exactly that. 

So the chosen solution is to build a 0-cost abstraction in top of `SLF4J` logger to add our own methods. This is possible thanks to scala 3 `opaque` type + `extension methods`. 

I also removed `ZioLogger` to replace it with the new opaque `RudderLogger` because now, ZIO framework has its own logging framework, which blur things. It wasn't use anywhere, so that is a cheap enhancement (`NamedZioLogger` is used everywhere).

Once we have that in place, we can: 

- have one dedicated logger for each hook types, ie for each directory under `/opt/rudder/etc/hooks.d`. That gives us the immediate quick-win of being able to debug one hook sets without being poluted by all other unrelated hooks
- document in default logback.xml how to use them.

### DIFF
--- a/webapp/sources/.scalafmt.conf
+++ b/webapp/sources/.scalafmt.conf
@@ -2,12 +2,14 @@ version = "3.7.17"
 maxColumn = 130
 runner.debug = true
 runner.dialect = scala213Source3
+// see https://github.com/scalameta/scalameta/blob/main/scalameta/dialects/shared/src/main/scala/scala/meta/Dialect.scala
+// for all supported dialectOverride
 runner.dialectOverride.allowExportClause = true
 runner.dialectOverride.allowOpaqueTypes = true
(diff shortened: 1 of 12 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM, finally extension methods at their best !

fanf:
OK, squash merging this PR

PR #6549, opened 2025-07-31
Fixes #27369: Nodes API payload for agentKey has the wrong JSON format

fanf:
https://issues.rudder.io/issues/27369


clarktsiory (approved):
LGTM !

PR #6552, opened 2025-08-01
Fixes #27264: Random error after node-to-relay is applied and other dynamic group and node accepted by API problems

fanf:
https://issues.rudder.io/issues/27264


fanf:
replaced by https://github.com/Normation/rudder/pull/6608

PR #6553, opened 2025-08-01
Fixes #27370: Dashboard will not display charts when switching between bechmarks

clarktsiory:
https://issues.rudder.io/issues/27370

* We need to `destroy()` existing charts present within the given DOM element (this resolves the issue when switching between benchmarks)
* We also need to wait for the element to be present, e.g. in https://github.com/Normation/rudder-plugins-private/pull/1070 there is some Elm code which initializes the chart even if the `view` is not fully rendered yet (this resolves the issue when switching between tabs)


RaphaelGauthier:
LGTM !

clarktsiory:
OK, merging this PR

PR #6555, opened 2025-08-02
Fixes #27392: Refactor campaign service so that it's testable

fanf:
https://issues.rudder.io/issues/27392


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -314,7 +314,7 @@ ALTER TABLE statusupdate set (autovacuum_vacuum_threshold = 0);
  *************************************************************************************
  */
 
-CREATE TYPE campaignEventState AS enum ('scheduled', 'prehooks', 'running', 'posthooks', 'finished', 'skipped', 'deleted', 'failure');
+CREATE TYPE campaignEventState AS enum ('scheduled', 'pre-hooks', 'running', 'post-hooks', 'finished', 'skipped', 'deleted', 'failure');
 
 CREATE TABLE CampaignEvents (
   campaignId   text
```
fanf: I think that on table `CampaignEventsStateHistory`, we : 
- don't want end date
- that start date is the date of save
- that save date is part of the primary key

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignDateScheduler.scala
```diff
@@ -0,0 +1,187 @@
...
+import org.joda.time.*
+
+/*
+ * This object contains the logic to compute campaign event dates.
+ * It takes care of time zone and other date shiny things.
+ */
+object CampaignDateScheduler {
+
+  implicit class DateTimeTzOps(date: DateTime) {
+    def adjustScheduleTimeZone(optTimeZone: Option[ScheduleTimeZone])(implicit defaultTz: DateTimeZone): DateTime = {
```
clarktsiory, line 52: it could be an `extension` method now in Scala 3

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignEventRepository.scala
```diff
@@ -65,6 +67,11 @@ trait CampaignEventRepository {
   def saveCampaignEvent(event: CampaignEvent): IOResult[Unit]
 
   def numberOfEventsByCampaign(campaignId: CampaignId): IOResult[Int]
+
+  /*
+   * A function that
+   */
```
clarktsiory: the comment is unfinished

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
```diff
@@ -154,18 +154,18 @@ class FsCampaignHooksService(
...
-    val dir        = rootDir + "/" + name + "/" + hookType.entryName
+    val loggerName = FsCampaignHooksRepository.CampaignDirName + "." + c.info.id.serialize + "." + hookType.entryName
+    val dir        = rootDir + "/" + FsCampaignHooksRepository.CampaignDirName + "/" + c.info.id.serialize + "/" + hookType.entryName
 
     (
       for {
+        -         <- PureHooksLogger.For(loggerName).debug(s"Running ${hookType.entryName} for campaign '${c.info.id.serialize}'")
         systemEnv <- IOResult.attempt(java.lang.System.getenv.asScala.toSeq).map(seq => HookEnvPairs.build(seq*))
         hooks     <- RunHooks.getHooksPure(dir, HOOKS_IGNORE_SUFFIXES)
         res       <- for {
```
clarktsiory, line 174: the `dir` is a String, it could as well be refactored to a `File` since the `RunHooks.getHooksPure` method manipulates a file.

And to avoid path traversal or nasty thing with files, I think `dir` should be checked against the base campaign directory (e.g. with `sanitzePath`), since the campaign id is not escaped

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
```diff
@@ -37,582 +37,258 @@
...
+   * Main event processing queue processing logic:
+   * here, we process each event in queue in its own fiber:
+   * - take one element
+   * - fork a fiber to process it (wait for the correct time, do something, etc)
+   */
+  def loop(): UIO[Unit] = {
+    for {
+      c <- queue.take
+      _ <- handle(c, DateTime.now(DateTimeZone.UTC)).forkDaemon
+    } yield ()
```
clarktsiory, line 297: okay, it's an UIO but in case the `handle` is in error, the error is forgotten, shouldn't there be some handling of business errors and defects errors here in the whole loop (including the only failable `handle: IOResult`) ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignEventRepository.scala
```diff
@@ -349,39 +348,56 @@ class CampaignEventRepositoryImpl(doobie: Doobie, campaignSerializer: CampaignSe
...
       beforeDate:   Option[DateTime] = None
-  ): IOResult[Unit] = {
+  ): ConnectionIO[RuntimeFlags] = {
 
     import _root_.cats.syntax.list.*
     val eventIdQuery      = id.map(c => fr"eventId = ${c.value}")
     val campaignIdQuery   = campaignId.map(c => fr"campaignId = ${c.value}")
     val campaignTypeQuery = campaignType.map(c => fr"campaignType = ${c.value}")
-    val stateQuery        = states.toNel.map(s => Fragments.in(fr"e.state", s))
+    val stateQuery        = states.toNel.map(s => Fragments.in(fr"state", s))
```
clarktsiory: it's too bad we don't have doobie query type-checking tests to catch this

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
```diff
@@ -0,0 +1,493 @@
...
+          } yield EventOrchestration.Queue(event.id)
+        } else {
+          CampaignLogger.debug(
+            s"${event.id.value} should be treated by ${event.campaignType.value} handler for Running state"
+          ) *> serviceHandleEvent(campaign, event)
+        }
+
+      case TPostHooks =>
+        for {
+          res <- effects.runPreHooks(campaign, event)
```
clarktsiory: the effect should then be `runPostHooks`

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/campaign/CampaignOrchestrationLogicTest.scala
```diff
@@ -0,0 +1,278 @@
...
+                     ZIO.unit
+                     // If the test is TRunning, since the logic is to wait for the running window to
+                     // close, we fork the process, then advance the clock
+                   } else if (event.state.value == TRunning) {
+                     for {
+                       t     <- currentTime(TimeUnit.MILLISECONDS)
+                       adjust = event.end.plusMillis(5).plus(bufferDelayAroundRun).getMillis - t
+                       _     <- o.handle(id, d, bufferDelayAroundRun, bufferDelayAroundRun).fork
+                       _     <- TestClock.adjust(Duration(adjust, TimeUnit.MILLISECONDS))
+                       _     <- IOResult.attempt(Thread.sleep(200)) // for TRunning etc to happen
```
clarktsiory: Is the `Thread.sleep` necessary ? does it not block the whole thread, why not `ZIO.sleep` which to my eyes is suited to the test semantics here ?

clarktsiory (changes requested):
Nice refactoring to a more testable event-based system !
There are a few things to be addressed, and some suggestions

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
fanf: I looked back, and the standard process always save two times: 
- one time when we enter into the phase - it's the normal logic pour that workflow, save are done between the previous and the current step, with current step initial value
- one more time at the end, to actually save the hooks result. 

So I just change the insert to do an update if the couple (step, eventid) already exists. It updates end date and data, which should give us the first and last update time, plus the interesting data.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
fanf: the error is logged/processed in handle, and if the last level fail, zio will print an error (and then we can add more management code for that case) - but perhaps I didn't understood ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignDateScheduler.scala
fanf: I just extracted that part from the other file, I'm not sure introducing more divergence is good

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
clarktsiory: Okay, I didn't look at the `handle` implementation, but only it's signature, and it's not an `UIO` but an `IOResult`, so to me it doesn't catch all errors. I think it should at least be an `UIO` by adding a `catchAll` then.

For defect error (actually unexpected ones), I assume it will not interrupt the loop since we have `forkDaemon`, but two things that suggest to me that error handling is still needed here in a loop : 
* the defects will throw with "fiber error", not with a campaign logger error log
* in the future if we remove the `forkDaemon` for any reason, it falls into the case of "endless loops" for which we would really like to catch all defects (see https://github.com/Normation/rudder/pull/5837)

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
clarktsiory: okay !

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -210,7 +233,47 @@ object RunHooks {
...
+      envVariables,
+      HookExecutionHistory.DoNotKeep,
+      globalWarnAfter,
+      unitWarnAfter,
+      unitKillAfter
+    ).map { case (c, _, d) => (c, d) }
+  }
 
+  /**
+     * Runs a list of hooks. Each hook is run sequencially (so that
```
clarktsiory:
```suggestion
     * Runs a list of hooks. Each hook is run sequentially (so that
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
fanf: I don't think it belongs to that PR. 
(and no urgency: for now, all hooks configuration are static/by configuration. This is the first example of one taking a variable, and that variable is not user-inputed. So yes for a by construction correction, but not in that branch/PR)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
fanf: The test should have caught that :thinking:

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/campaign/CampaignOrchestrationLogicTest.scala
fanf: I changed to waiting for the fiber end, which can block in case of bugs but the test suite has a timeout, so it should be ok

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
fanf: OK, change the test to actually check for a static thing, so there's no doubt.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
fanf: The fork is part of the design, we can't remove it without changing all the logic, see the comment on tha class.
The loop is an UIO, since forking the process does not fail. 

The forked process is handled, catchAll and logged, and if it fails, it looks like it's a logic error that should draw a stack trace so that we (the dev) get notified about it. I can avoid giving the error in the existing catchAll, but I think it will just make the bug harder to debug in that case, since we will also need the user to copy the log before the stack trace.

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
fanf: ok, let me make it clearer and document each bit

clarktsiory: okay so if it is part of the design then only documenting that the loop does not fail should be enough, indeed error is logged in the handling method

clarktsiory: (and by design, it's mainly because the `q.take` is supposed to be run in another fiber than the `q.offer` which is the `handle` ?)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
clarktsiory: okay, it could be done in another PR then, it's not urgent

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
fanf: in that kind of queue-based process, we don't really care who/where offer is done. What must not fail is the `take -> handle -> loop` part, else it brokes the engine whick can stop running, and we don't have any other way to start it than rebooting. 
That part is addressed either in being sure that no error at all can happen in the handling part - including ZIO runtime error, or jvm errors regarding threading that could be worked-around -, or by delegating the handling in an other thread for processing and letting it die if it wants. 

The other common problem, which is completely different and must be addressed in other way, is the risk for infinite (fast) looping without progress. 
For now, we don't have tooling to avoid it in current design. The standard answer is to design a datastructure for what are authorized next step for any step, so that you have a truth table for a finite automata of the process. 
In general, every case is easy appart the one where you loop to the same time "after some time" to see if the universe progressed when you were sleeping. 
This can be addressed with a two-step design, where after sleep you go to a circuit breaker in charge of looking for how long it was since last time you came here and if any progress was made, and chose to loop or fail based on that. 

I didn't want to do that in that PR, I wanted to keep the drift from the original PR smallish (conceptually, at least). But it should completly be done. Having a workflow engine without finite automata is a sure path toward complicated bugs and unforseen edge cases that will be hard to deal with.

fanf:
OK, merging this PR

fanf:
OK, merging this PR

PR #6556, opened 2025-08-02
Fixes #26883: Add campaign hooks in the campaign workflow engine

fanf:
https://issues.rudder.io/issues/26883

Create the `CampaignHooksService` and the corresponding step in campaign workflow and persistence in base. 
There is no unit tests, they will be added after https://github.com/Normation/rudder/pull/6555 make them easier to code. 

So, the main thing is the addition of: 

## `CampaignHooksService` and its implementation

That adds a service that run campaign related hooks, which are located on FS along with other Rudder hooks. The difference is that for these ones, each campaign can have different hooks and so, under the `.../hook.d/campaign/` directoty, there is one directoty for each campaign with its subdirectories for `pre-hooks` and `post-hooks`. 

- `CampaignHooksRepository` allows to create and delete the relevant directories when a campaign is created/deleted. 
- `CampaignHooksService`: the hook execution by itself is not interesting: it's the same than for all other hooks, with the following environment variables (all obvious I think): 
```
"CAMPAIGN_ID"
"CAMPAIGN_NAME"
"CAMPAIGN_TYPE"
"CAMPAIGN_EVENT_ID"
"CAMPAIGN_EVENT_NAME"
```
- `RunHooks` was updated to be able to add a variant of `asyncExec` that keep the whole logs of all run hooks, not only the last (failledà one. That will allow to store them in the corresponding campaign event and for forensic. 

## Campaign logic

The main things is to add the too new state and their type. I normalized their name to `pre-hooks` and `post-hooks` (with a '-') to be  consistent on FS and base and JSON. 

The logic is to call the hook service when we are in the corresponding event state. 

## Refactoring

Some light refactoring was done:
- pattern matching on the even state types, 
- addind a `T` to differentiate between event state and event state type. 
- adding the hook name as a recorded data in the `HookReturnCode` so that we have the command name with the logs,
- hide `CampaignArchiver` in `CampaignRepositoryImpl`, it's an implementation details that should not be top-level

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
@@ -1,0 +1,197 @@
+/*
+ *************************************************************************************
+ * Copyright 2025 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.campaigns
+
+import better.files.File
+import com.normation.errors.*
+import com.normation.rudder.hooks.*
+import com.normation.zio.*
+import scala.jdk.CollectionConverters.*
+import zio.*
+import zio.syntax.*
+
+/*
+ * This service is in charge of managing campaign hooks.
+ */
+trait CampaignHooksService {
+
+  /*
+   * Run hooks that are configured to run before the campaign time slot is reached.
+   * Hooks results are collected to for trace.
+   */
+  def runPreHooks(c: Campaign, e: CampaignEvent): IOResult[HookResults]
+
+  /*
+   * Run hooks that are configured to run before the campaign time slot is reached.
+   * Hooks results are collected to for trace.
+   */
+  def runPostHooks(c: Campaign, e: CampaignEvent): IOResult[HookResults]
+}
+
+object NoopCampaignHooksService extends CampaignHooksService {
+  override def runPreHooks(c: Campaign, e: CampaignEvent): IOResult[HookResults] = {
+    HookResults(List(HookResult("pre-hook", 0, "", "", ""))).succeed
+  }
+
+  override def runPostHooks(c: Campaign, e: CampaignEvent): IOResult[HookResults] = {
+    HookResults(List(HookResult("post-hook", 0, "", "", ""))).succeed
+  }
+}
+
+/*
+ * Hooks are by-campaign, so we need some help to create the overall structure when
+ * a new campaign is created, and delete the whole thing when the campaign is deleted.
+ */
+trait CampaignHooksRepository {
+  /*
+   * Create the pre-/post-hook sub-directories specific to that campaign
+   */
+  def initHooks(cid: CampaignId): IOResult[Unit]
+
+  /*
+   * Delete hook directory related to that campaign.
+   */
+  def deleteHooks(cid: CampaignId): IOResult[Unit]
+}
+
+object NoopCampaignHooksRepository extends CampaignHooksRepository {
+  override def initHooks(cid:   CampaignId): IOResult[Unit] = ZIO.unit
+  override def deleteHooks(cid: CampaignId): IOResult[Unit] = ZIO.unit
+}
+
+/*
+ *  The standard layout for hooks is:
+ *  - HOOKS_D/campaigns
+ *     |- <campaign-uuid>
+ *     |     |- pre-hooks
+ *     |     |    ` scripts
+ *     |     `- post-hooks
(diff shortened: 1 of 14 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
```diff
@@ -276,12 +289,14 @@ object CampaignEventId {
...
-  case object Deleted   extends CampaignEventStateType("deleted")
-  case object Failure   extends CampaignEventStateType("failure")
+  case object TScheduled extends CampaignEventStateType("scheduled")
+  case object TPreHooks  extends CampaignEventStateType("pre-hooks")
+  case object TRunning   extends CampaignEventStateType("running")
+  case object TPostHooks extends CampaignEventStateType("post-hooks")
+  case object TFinished  extends CampaignEventStateType("finished")
+  case object TSkipped   extends CampaignEventStateType("skipped")
+  case object TDeleted   extends CampaignEventStateType("deleted")
+  case object TFailure   extends CampaignEventStateType("failure")
```
clarktsiory, line 299: I find the `T` quite inconvenient for readability, to resolve ambiguity, I would personally prefer using full paths e.g. `CampaignEventStateType.Running`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -291,7 +354,15 @@ object RunHooks {
...
             } yield {
-              c
+              history match {
+                case HookExecutionHistory.DoNotKeep => (c, Nil)
+                case HookExecutionHistory.Keep      =>
+                  val logs = historyList match {
+                    case Noop :: Nil => x :: Nil
+                    case l           => x :: l
+                  }
+                  (c, logs)
```
clarktsiory, line 364: Is that logic the one that is tested for the behavior of keeping history ?

I don't really understand the inner pattern match here, I fear the `Noop :: Nil` case, it means the `Noop` first element should specifically be replaced by the next history ?
If so, then, why not everything starting with a `Noop` ?

To have a name, perhaps another type that better represents the specific `Noop :: Nil` case could be introduced  ? As I understand, it would be either the empty list (initializer of the fold), or an existing list, or the case where "there is a first noop result that should be replaced", so maybe with Scala 3 :

```scala
type FoldHistoryList = Initializer | FirstNoopResult | ::[HookReturnCode] 
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
```diff
@@ -0,0 +1,197 @@
...
+ *     |          ` scripts
+ *     ...
+ */
+class FsCampaignHooksRepository(HOOKS_D: String) extends CampaignHooksRepository {
+  val campaignRootHooksDir: File = File(HOOKS_D) / FsCampaignHooksRepository.CampaignDirName
+
+  private def getDir(cid: CampaignId): File = {
+    // perhaps we will want per-revision hook directory some day?
+    campaignRootHooksDir / cid.serialize
+  }
```
clarktsiory, line 113: risk of path traversal for that if the `CampaignId` is not escaped

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
```diff
@@ -0,0 +1,197 @@
...
+                                       ),
+                                       systemEnv,
+                                       HookExecutionHistory.Keep,
+                                       1.minutes // warn if a hook took more than a minute
+                                     )
+                       timeHooks1 <- currentTimeMillis
+                       _          <-
+                         PureHooksLogger.For(loggerName).trace(s"Inventory received hooks ran in ${timeHooks1 - timeHooks0} ms")
+                     } yield res
+      } yield HookResults((res._1 :: res._2).map(HookResult.fromCode))
```
clarktsiory, line 191: in scala 3 you can keep my eyes away from the horrendous 🙈   
```suggestion
      } yield HookResults((res(0) :: res(1)).map(HookResult.fromCode))
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
fanf: In practice, it's unbelievabely unconvenient. This type is used in a lot of pattern matching, list, etc. 
I changed to that as a relief

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
fanf: Yeah, but no, because for the reader, it's not a tuple (well typed) anymore, but some seq, which should never be accessed by index without care. I would just prefer to have filter correctly implemented in for loop

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
fanf: I can put the T in the end to so if it helps wrt readability

clarktsiory: OK, so it's at the end I would even prefer a full `Type` : `ScheduledType` :smile:

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
fanf: Noop means that there's no operation, it can't be produced as a result, or even if it is, it means "there's no result", like a `None`. I can just filter out `Noop` in resulting list, it will be clearer

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
clarktsiory: (a solution is to use the `sanitizePath` to check that the directory is still in the campaigns repo base dir)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
clarktsiory: okay, indeed deconstructing the tuple in the for comprehension is the most elegant solution (we are awaiting an higher scala 3 version for that ?)

fanf: Yes, it's on 3.4.0 (https://github.com/scala/scala3/pull/18842 - I think). It will be a revolution. 20 years awaiting for that pain to be corrected

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/DataTypes.scala
fanf: can do that too

fanf:
OK, merging this PR

PR #6561, opened 2025-08-06
Fixes #27412: Using Cons for big lists can blow the compiler stack and prevent scala project compilation entirely

mbaechler:
https://issues.rudder.io/issues/27412


clarktsiory:
OK, merging this PR

PR #6568, opened 2025-08-14
Fixes #27450: Fix the rudder_info! macro to make it usable in audit mode

m4rtinh4rt:
https://issues.rudder.io/issues/27450

CFEngine offers multiple [log levels](https://docs.cfengine.com/docs/3.26/reference-promise-types.html#log_priority). The `info` level previously used in our macro is intended to be used only in [enforce mode](https://github.com/cfengine/core/blob/37e62318b14c9b581fea20b0f02b069b758223ac/libpromises/mod_custom.c#L1039). The `rudder_info!` macro will now use the `notice` priority level to retain the same semantics while avoiding CFEngine errors when running in audit mode.


amousset:
Could you add a more complete explanation in the issue description?

amousset (approved):

PR #6570, opened 2025-08-21
Fixes #22595: Move & update AuthorizationType into rudder-core

clarktsiory:
https://issues.rudder.io/issues/22595

Move some authorization & user-related structures to rudder-core.
Some chimney transformers needed to be refactored and cleaned (to avoid making the visibility of secret user passwords field public)

It is needed to be able to use the `AuthenticatedUser` in the right scope for https://github.com/Normation/rudder/pull/6453 

**EDIT:** added a commit to refactor the `AuthenticatedUser` to be an actually authenticated one (a trait of `RudderUserDetail` : 99459a1d8ad14d5fc3bbab33d0abf6915ce9ed57
We already do a preliminary check for autorizations, if the check is passed then an `AuthzToken` is passed to API implementation : it makes sense that it contains the fully authenticated and authorized user now.
Something remains unchecked and not at the right place : in Lift snippets, a `QueryContext(actor, nodePerms)` is still needed, but no prior authorization check is made so the actor is left `"unknown"`

---
N.B. : this has some ~trivial impact on plugins : https://github.com/Normation/rudder-plugins/pull/879

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/User.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/User.scala
@@ -39,6 +39,7 @@ package com.normation.rudder.users
 
 import com.normation.eventlog.EventActor
 import com.normation.rudder.AuthorizationType
+import com.normation.rudder.Rights
 import com.normation.rudder.api.ApiAccount
 import com.normation.rudder.api.ApiAuthorization
 import com.normation.rudder.facts.nodes.NodeSecurityContext
@@ -70,29 +71,26 @@ object RudderAccount {
 }
 
 /**
- * An authenticated user with the relevant authentication information. That structure
- * will be kept in session (or for API, in the request processing).
+ * An authenticated user with the relevant authentication information within the API.
+ * That structure will be kept in session (or for API, in the request processing).
  */
 trait AuthenticatedUser {
-  def user: Option[RudderAccount.User]
+  def account: RudderAccount
 
-  def account: Option[ApiAccount]
+  def name:  String = account match {
+    case RudderAccount.User(login, _) => login
+    case RudderAccount.Api(api)       => api.name.value
+  }
+  def login: String = name
+
+  def authz:       Rights
+  def apiAuthz:    ApiAuthorization
+  def nodePerms:   NodeSecurityContext
+  implicit def qc: QueryContext = {
+    QueryContext(EventActor(name), nodePerms)
+  }
 
   def checkRights(auth: AuthorizationType): Boolean
-
-  def getApiAuthz: ApiAuthorization
-
-  final def actor: EventActor = EventActor((user, account) match {
-    case (Some(user), _)    => user.login
-    case (_, Some(account)) => account.name.value
-    case (None, None)       => "unknown"
-  })
-
-  def nodePerms: NodeSecurityContext
-
-  def queryContext: QueryContext = {
-    QueryContext(actor, nodePerms)
-  }
 }
 
 sealed trait UserPassword {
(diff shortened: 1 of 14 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR rebased

clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/User.scala
```diff
@@ -38,17 +38,170 @@ package com.normation.rudder.users
...
+}
+
+/**
+ * An authenticated user with the relevant authentication information. That structure
+ * will be kept in session (or for API, in the request processing).
+ */
+trait AuthenticatedUser {
+  def user: Option[RudderAccount.User]
+
+  def account: Option[ApiAccount]
```
fanf: I'm very uneased by that two fields. I know it's "just" a move from the other place, but we should have been more disciplined in the review introducing that. 
 
First, non technical, but for comprehension: field naming. We have a `RudderAccound.User` that is a `user`, and a `ApiAccount` that is an account. It really should be an `api` for consistency, both are accounts. Then, I understand `api` doesn't hold what we want it to mean here, so we have to use something else, likely: `userAccount` and `apiAccount`. 

Then: does it really makes sense to have an authenticated user without an user acount nor an api account ? Which leads to: what important information do we need from an authenticated account that these field provides and is not generalisable ? 

If what we need is some kind of authentication origin, that we trace like that, then we should have a dedicated structure for that. 
If what we need is some kind of common property (actor, at least), then it should be isolated at such - likely in the in the constructor of `AuthenticatedUser`. 
And having an `unknown` actor is almost certainly NOT what we want for an **authenticated** user. 
We should not be able to represent unwanted state. 

So, that can go into an other PR, but it can't remind like that in 9.0. Even if there is clear reason for why we should allow theses very, very strange state, they should be isolated and special cased (typically if we have a "authenticated by 'generic unknown' user, then it should be a special case, with final rights, so that it doesn't happen by chance)

clarktsiory: I agree that the naming is not consistent, in fact `user` is only used internally, so it could be `private`.

The API is based on the one of a `RequestVar[Option[RudderUserDetail]]`, because in the lifecycle of the `CurrentUser` variable, the user can be not yet authenticated, since it is a global mutable variable I don't see a way out of that.
I think it makes sense to have "empty"-ish fields : have everything as `Option`s, or `checkRight` always returning `false`. So it's already the case for some fields, but the "unknown" actor seems to be here for convenience.

Since the usage is to have the `QueryContext(actor, nodePerms)`, the API structure `AuthzToken { val qc: QueryContext }` seems to be more appropriate to get an "unknown" actor from an optional one, see https://github.com/Normation/rudder/pull/6453.

Yes, clearly the case for "unknown" should actually be a sub-case of `QueryContext`, which itself should be an enum of :
* there is "no actor", so `canSee` is false and hence there are no permissions
* there is an actor with node permissions

clarktsiory:
PR updated with a new commit

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserService.scala
```diff
@@ -38,8 +38,8 @@
 package com.normation.rudder.users
 
 /**
- * A minimalistic definition of a service that give access to currently logged user .
+ * A minimalistic definition of a service that give access to currently logged user, if any.
  */
 trait UserService {
-  def getCurrentUser: AuthenticatedUser
+  def getCurrentUser: Option[AuthenticatedUser]
```
fanf, line 44: This change impacts plugins `api-authorization` and `change-validation` tests and should be reported there

fanf (approved):

clarktsiory:
OK, merging this PR

PR #6572, opened 2025-08-21
Fixes #27384: Missleading format for parameter category in API

ElaadF:
https://issues.rudder.io/issues/27384


webapp/sources/rudder/rudder-rest/src/test/resources/api/api_rules.yml
```diff
@@ -443,7 +443,7 @@ headers:
   - "Content-Type: application/json"
 body: >-
   {
-    "category": "category1"
+    "categoryId": "category1"
```
fanf: You also need to test with the alias that `category` is still working

clarktsiory:
OK, squash merging this PR

PR #6574, opened 2025-08-26
Fixes #27486: Add includeSystem parameter to filter system groups in API

clarktsiory:
https://issues.rudder.io/issues/27486

Reusing the existing parameter extractor (used in the directives tree).
The filtering is also very similar, see https://github.com/Normation/rudder/blob/cd64fe5017654f3a988080d0c3fb2d6cc5bbc915/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/DirectiveApi.scala#L280-L288

### DIFF
--- a/webapp/sources/api-doc/paths/groups/tree.yml
+++ b/webapp/sources/api-doc/paths/groups/tree.yml
@@ -4,6 +4,15 @@ get:
   summary: Get groups tree
   description: Get all available groups and their categories in a tree
   operationId: GetGroupTree
+  parameters:
+    - name: "includeSystem"
+      in: query
+      required: false
+      schema:
+        type: boolean
+        default: true
+      example: false
+      description: Whether to include system categories and groups in results.
   responses:
     "200":
       description: Groups information
@@ -67,6 +76,9 @@ get:
                             - root
                           dynamic: true
                           enabled: true
+                          properties: []
+                          target: "group:hasPolicyServer-root"
+                          system: true
                     - id: 38dd2107-a73b-45fb-916d-e110312abb87
                       name: production groups
                       description: ""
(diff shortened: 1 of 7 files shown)


webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
```diff
@@ -579,6 +579,7 @@ response:
         ]
       }
     }
+---
```
clarktsiory, line 582: it's funny : the test above got ignored because the `---` were missing

clarktsiory:
PR updated with a new commit : public documentation update (also fixed directives tree documentation)

webapp/sources/api-doc/paths/groups/tree.yml
```diff
@@ -4,6 +4,15 @@ get:
   summary: Get groups tree
   description: Get all available groups and their categories in a tree
   operationId: GetGroupTree
+  parameters:
+    - name: "includeSystem"
+      in: query
+      required: false
+      schema:
+        type: boolean
+        default: true
```
fanf, line 13: humpf, so bad the default is not the same than in directive. Can we do an other PR for Rudder 9.0 that change the default value to "false" as it is in directiveTree for the API version corresponding to Rudder 9.0 (and keep the other default for previous API versions) ? If we don't do it now, we will forget and live with an inconsistancy. 

The external effect is that user using that API won't see anymore system groups (ok), and we will need to correct where we call it in rules and groups at least I think.

clarktsiory: OK that's right, I took note to do that : https://issues.rudder.io/issues/27498

clarktsiory: I checked it and changed in 9.0 with these PRs (it was only used in Rules, and system-updates) : 
* https://github.com/Normation/rudder/pull/6579
* https://github.com/Normation/rudder-plugins-private/pull/1166

clarktsiory:
OK, squash merging this PR

PR #6576, opened 2025-08-28
Fixes #27428: Missing migration for existing directives with the bad select input identifier

VinceMacBuche:
https://issues.rudder.io/issues/27428

The migration check technique with a select parameter, then get all technique based on this directive, then check if the directives has a parameter using the name of a technique parameter (instead of an id) and if the name is used, remove it and add a parameter with the id of the technique parameter with previous value

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/migration/MigrateDirectiveWithSelectInputBroken.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/migration/MigrateDirectiveWithSelectInputBroken.scala
@@ -1,0 +1,138 @@
+/*
+ *************************************************************************************
+ * Copyright 2025 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+
+ *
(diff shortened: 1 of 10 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/migration/MigrateDirectiveWithSelectInputBroken.scala
```diff
@@ -0,0 +1,118 @@
...
+                                  case None        => directive
+                                  case Some(value) =>
+                                    directive.copy(parameters =
+                                      directive.parameters.removed(param.name).updated(param.id.value.toUpperCase, value)
+                                    )
+                                }
+                            }
+                          })
+                        }
+                        saved               <- ZIO.foreach(updatedDirectives)(d => {
```
fanf, line 100: `saved` is not used, it can be `_`
And so, `ZIO.foreach` can be `ZIO.foreachDiscard`. Same for to layer  `directives`

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/migration/MigrateDirectiveWithSelectInputBroken.scala
```diff
@@ -0,0 +1,118 @@
...
+                                                 writeDirectives.saveDirective(
+                                                   atId,
+                                                   d,
+                                                   modificationId,
+                                                   RudderEventActor,
+                                                   Some(s"Updating invalid parameters in directive ${d.id.serialize}")
+                                                 )
+                                               })
+
```
fanf: I would like to have a log that looks like a check (ie, don't tell user it migrates if there is nothing to migrate), and if migration did happen, then list UUIDs of impacted directive. 
Something lile: 

````
.... Checking if some directive have a select paramenter needing migration
... Migrated N directives: UUID1, UUID2, etc
````

Without that, it is quite difficult to know is any directives were migrated, so it can lead to harder than needed debug session. 
We can't rely on event log: I thought I could find them back by searching on the save message, but it does not work (which is a pitty).

fanf (commented):
Good ! Works as expected, but I would still like a better log bootstrap log to help pinpoint what directives were impacted if something goes wrong.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/migration/MigrateDirectiveWithSelectInputBroken.scala
VinceMacBuche: I use the saved as commentary of what is done. it has more meaning than _

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/checks/endconfig/migration/MigrateDirectiveWithSelectInputBroken.scala
VinceMacBuche: I hope this log line will be helpful to someone, I'm not sure a Logger for a migration for a case that may not really exists needed this much !! but it is done

VinceMacBuche:
PR rebased

VinceMacBuche:
OK, merging this PR

PR #6577, opened 2025-08-28
Fixes #26813: No manual when a filter for methods in the technique editor right panel

RaphaelGauthier:
https://issues.rudder.io/issues/26813

Now, when the user wants to view the documentation for a method that does not match the filters, it still appears in the list of methods.


clarktsiory (approved):
LGTM, works as mentionned

PR #6578, opened 2025-08-28
Fixes #26718: When the JSON property is invalid in global properties, the error is nasty

clarktsiory:
https://issues.rudder.io/issues/26718

The error : 
* no longer has a stack trace
* has a the generic action message fixed ("create"/"update" parameter)
* still has the parsing error message

Result :
image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateOrUpdateGlobalParameterPopup.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateOrUpdateGlobalParameterPopup.scala
@@ -129,7 +129,11 @@ class CreateOrUpdateGlobalParameterPopup(
       // in case of string, we need to force parse as string
       v <- if (jsonRequired) GenericProperty.parseValue(value) else Right(value.toConfigValue)
       _ <- if (jsonRequired && v.valueType() == ConfigValueType.STRING) {
-             Left(Inconsistency("JSON check is enabled, but the value format is invalid."))
+             Left(
+               Inconsistency(
+                 "JSON check is enabled, but the value appears to be a String. Please select the String format instead."
+               )
+             )
            } else Right(())
     } yield {
       v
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateOrUpdateGlobalParameterPopup.scala
```diff
@@ -127,9 +130,20 @@ class CreateOrUpdateGlobalParameterPopup(
     import GenericProperty.*
     for {
       // in case of string, we need to force parse as string
-      v <- if (jsonRequired) GenericProperty.parseValue(value) else Right(value.toConfigValue)
+      v <- if (jsonRequired) {
+             GenericProperty.parseValue(value).adaptErr {
+               // a system error displays full trace, but for an invalid JSON we only want a message
+               case SystemError(_, ex: ConfigException.Parse) => Inconsistency(s"The JSON value is invalid: ${ex.getMessage}")
```
fanf: That feels fragile. Unpacking system error for the kind of exception will break if we change the underlying (opaque, here) parsing framework.
Why not handling that in parseValue, with a meaningful ".msg" ?

fanf (changes requested):
Isn't there a better (more respectful of system threshold) way of doing what you need for the parse exception?

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateOrUpdateGlobalParameterPopup.scala
clarktsiory: :white_check_mark: I changed the `parseValue` to format the error without a stack trace, and adapted tests (to be more precise on the error-side)

fanf (approved):
Much better !

clarktsiory:
OK, squash merging this PR

PR #6580, opened 2025-09-01
Fixes #27124: Use ESM modules in rudder-web

clarktsiory:
https://issues.rudder.io/issues/27124

With the [modular JS system](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules), neither of the usual `Script(OnLoad(...))` in Scala code and `<script type="text/javascript" src="..."/>` work on ESM dependencies, in this case [jsondiffpatch](https://github.com/benjamine/jsondiffpatch/releases/tag/v0.6.0) (see https://github.com/Normation/rudder/pull/6460).

To have an equivalent : 
* we need to add a `type="module"` to create the element : `<script type="module" src="...">` 
* but the module file needs to contain "imports"

:arrow_right: create such an element with a `ScriptModule(OnLoad(...), imports = "import ...")` in Scala code.
This is done by refactoring the JS code used for change logs into a dedicated module which will imported

## ESM modules caveats
We need to be careful about `import` and `export` : 
* the dependency has been put in the `/<context-path>/<static-assets-base-uri>/javascript/libs/jsondiffpatch` static assets directory (see https://github.com/Normation/rudder/pull/6460) BUT when we want to import it within a JS module, we cannot directly reference the path so we need **relative imports**
* we need to also be careful to **export** all functions and values we use


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/snippet/WithNonce.scala
```diff
@@ -36,6 +36,19 @@ object WithNonce extends StatefulSnippet {
...
   }
 
+  def moduleWithNonce(base: Node, imports: String = ""): Node = {
+    base match {
+      case e: Elem =>
+        val child = scala.xml.Text(imports) ++ e.child
+        e.copy(
+          attributes = MetaData.update(base.attributes, base.scope, new UnprefixedAttribute("type", "module", Null)),
+          child = child
+          // TODO: create children for imports :  = e.child.text
```
fanf: why it that TODO remaining ? Either it needs to be done, or it is not a todo

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/StaticResourceRewrite.scala
```diff
@@ -0,0 +1,98 @@
...
+          val size        = conn.getContentLength.toLong
+          val in          = conn.getInputStream
+          StreamingResponse(in, () => in.close(), size = size, headers(contentType), cookies = Nil, code = 200)
+        }
+      }
+  }
+}
+
+object StaticResourceRewrite {
+  // the resource directory we want to server that way
```
fanf: s/to server/to serve/

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/EventListDisplayer.scala
```diff
@@ -136,21 +138,27 @@ class EventListDisplayer(repos: EventLogRepository) extends Loggable {
...
        changeTimezone(new Date('${jsDateWithTimeZone}'), '${currentTimezone}'),
        '${currentTimezone}',
        ${hoursBeforeNow}
      );
      createEventLogTable('${gridName}',[], '${S.contextPath}', refreshEventLogs)
      refreshEventLogs();
-    """)))) // JsRaw ok, escaped
+    """)),
+        imports =
+          s"import { createEventLogTable } from '${staticResourceRewrite.resourceUrl("/javascript/rudder/changeLogs.js")}'"
```
fanf: I don't get from where that comes. Did you forget to add the `changeLogs.js` file?

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/snippet/WithNonce.scala
clarktsiory: oups it was kept from a previous draft commit but actually this method is no longer relevant, it has now been removed

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/EventListDisplayer.scala
clarktsiory: :see_no_evil: yes it's better with the file added to git : https://github.com/Normation/rudder/pull/6580/files#diff-65dddd04c125123e71ce44285210fc9d06cb81841c1d2fa0dec3657e732e0fdc

RaphaelGauthier:
LGTM !

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/lift/JsCommands.scala
```diff
@@ -0,0 +1,22 @@
+package com.normation.rudder.web.lift
+
+import net.liftweb.http.js.JsCmd
+import scala.xml.Node
+import scala.xml.Unparsed
+
+object JsCommands {
+
+  // copied and adapted from Lift JsCommands code
```
fanf: Shouldn't imports be a list of paths that we check from path traversal and we rebuild the correct relative path from a directory? Having .. In imports is unsetling

clarktsiory: The usage is : 
```scala
ScriptModule(
   """
   ...
   createEventLogTable('${gridName}',[], '${S.contextPath}', refreshEventLogs)
     refreshEventLogs();
    """
    , imports =
      s"import { createEventLogTable } from '${staticResourceRewrite.resourceUrl("/javascript/rudder/changeLogs.js")}'"
)
```

We are not assuming that imports have a given format since the JS syntax can vary a lot, and the URL of the resource can be either relative or absolute. Mostly from Scala code, it would be absolute (the `resourceUrl` knows the whole context path + assets URI). It could be relative when called from within a JS module. 

The thing that we can enforce is that imports are put on top of the `<script/>`

fanf: Well, we don't have to accept all of javascript syntax - it's even a sure way toward too much complexity and vulnerabilities. 
We can restrict the cases to what we do want to use in the limited context of Rudder and our import. 
You can have variation for absolute or relative path, something like: 

```
ScriptModule(....).withStaticImport("createEventLogTable", "changeLogs.js").toHtml
```
And:
```
case class StaticImport(component: String, path: String)
case class ScriptModule(script: JsCmd, staticImports: List[StaticImport], relativeImport: List[RelativeImport]) {
  def withStaticImport(name, path) = { //check path traversal, throw security exception if ..
    this.modify(staticImports).using(StaticImport(name, path) :: _)
  }
  def toHtml = Unparsed("""....
}
object ScriptModule {
  def apply(js:JsCmd) = ScriptModule(js, Nil, Nil)
     
}

```

If you don't do something like that, the first time we will want to do dynamic module loading based on "things", we will give path as computed things from parameter and one of us, one time will forget to check for path traversal.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/lift/JsCommands.scala
clarktsiory: I changed it to incorporate this suggestion (the result is interesting since it makes use of opaque types to avoid type blindness on "string" and the "path").
I added tests for the security-related issues (external import and path traversal), and also the HTML output.
It looks friendlier and less error-prone now !

fanf (approved):
Ah ! This is now really good :+1:

PR #6588, opened 2025-09-08
Fixes #27456: Inherited properties API change in parent 

VinceMacBuche:
https://issues.rudder.io/issues/27456

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeProperties/DataTypes.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeProperties/DataTypes.elm
@@ -50,13 +50,13 @@ type alias Property =
 
 type alias HierarchyStatus =
   { hasChildTypeConflicts : Bool
-  , fullHierarchy : List ParentProperty
+  , fullHierarchy : ParentProperty
   , errorMessage : Maybe String
   }
 
 type alias ParentGlobalProperty = { valueType : String }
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeProperties/DataTypes.elm
```diff
@@ -170,17 +170,28 @@ valueTypeToValueFormat valueType =
...
+                       Nothing -> []
+          in
+            if prop.id /= model.nodeId then prop.valueType :: parent else parent
+        ParentNode prop ->
+          let
+            parent = case prop.parent of
+                       Just par -> getOtherHierarchyValueType par
+                       Nothing -> []
+          in
+            if prop.id /= model.nodeId then prop.valueType :: parent else parent
```
clarktsiory, line 190: it's the kind of repetition we could avoid by using generics, it's okay here with 2 small blocks of code though : 
```elm 
let
  getLeafHierarchy { parent, id, valueType } =
    let
      parent = case parent of
                       Just par -> getOtherHierarchyValueType par
                       Nothing -> []
    in
    if id /= model.nodeId then valueType :: parent else parent

  getOtherHierarchyValueType : ParentProperty -> List String
  getOtherHierarchyValueType p =
    case p of
      ParentGlobal prop -> [ prop.valueType ]
      ParentGroup prop -> getLeafHierarchy prop
      ParentNode prop -> getLeafHierarchy prop

  ...
```

PR #6599, opened 2025-09-12
Fixes #27559: Use CA in reports signature check

amousset:
https://issues.rudder.io/issues/27559


amousset:
PR updated with a new commit

PR #6607, opened 2025-09-18
Fixes #26637: System info API changed format in v21 and needs new documentation

P4uline:
https://issues.rudder.io/issues/26637


webapp/sources/api-doc/paths/system/info.yml
```diff
@@ -32,25 +32,159 @@ get:
...
-                        format: X.Y
-                      full-version:
-                        type: string
-                        example: 6.0.4
+                        example: "8.3.4-SNAPSHOT"
                         format: X.Y.Z
-                      build-time:
+                      buildTime:
                         type: string
+                        example: "2025-09-07T01:31:07Z"
```
P4uline, line 52: Is `buildTime` property really a build time ? The documentation describes a date in a string. If it's a time, then the type could be a number representing a time in milliseconds for instance. In any cases it shouldn't represent a date

If it's a date, then we should create a ticket to rename `buildTime` to `buildDate`

P4uline:
PR rebased

webapp/sources/api-doc/paths/system/info.yml
fanf: it's a date, but it's also a public API so we can't change it without some care. 
Plus, it's generally accepted that `build time/timestamp` can be whatever represent a specific instant. Rust, jenkins, etc use rfc3339 for that, too

fanf (approved):
Seems OK to me and compile fine.

PR #6611, opened 2025-09-22
Fixes #27574: Post-hooks for campaigns should be executed even even if pre-hooks are in failure

fanf:
https://issues.rudder.io/issues/27574

This commit handle one annoyance and a semantic change for campaign hooks:
- add clearer logs in case of warning or error with the corresponding log level and faulty script
- we always go to `post-hooks`, even if `pre-hooks` failed, so that we let user a chance to correct state changed in `pre-hooks`

# Clearer logs

Now, when a campaign hook fails or warn, it has a corresponding line in webapp logs: 

```
2025-10-03 22:40:58+0200 ERROR hooks.campaigns.1bb22fc5-f41c-46fb-a44b-54b68b8254f2.pre-hooks - Campaign 'test campaign' pre-hooks returned code '24' in 'test-hook.sh'
```

This needed some adaptation: 
- also keep hook names in the history from `RunHooks.asyncRunHistory`
- post process campaign hook history execution to log to correct level

# Workflow change: always go to `post-hook`

The idea here is that is we do something in `pre-hooks`, we may want to undo it in `post-hooks`. 
For that, we must always go to `post-hooks` - but we need to more info:
- one to let the user know that we came to that post-hooks because something failed and the campaign wasn't exec
- one to let the workflow know that after these post-hooks, the final campaign state must be "failure" even if all post hooks succeded. 

And actually, it's the same bit of information: we just need to set a `next-state` value in post-hook before we go to them: 
- if it's from a failure in `pre-hook`, the `next-state` is `failure` ,
- if it's from the running state, the `next-state` is `finished`. 
And we let the user know in the hook with a new environnment variable: `CAMPAIGN_NEXT_STATE`. 

From a code point of view, I had to: 
- change `CampaignEventState.PostHooks` to store `nextState: CampaignEventStateType` 
- change the workflow to direct to `post-hooks` in place of `failure` after pre hooks, seting `nextState`
- change serialisation to save the `nextState` in base for `post-hooks`
And add a unit test to demonstrate the behavior

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
@@ -142,11 +142,19 @@ class FsCampaignHooksService(
 ) extends CampaignHooksService {
 
   override def runPreHooks(c: Campaign, e: CampaignEvent): IOResult[HookResults] = {
-    runHooks(HOOKS_D, HOOKS_IGNORE_SUFFIXES, c, e, CampaignHookTypes.CampaignPreHooks)
+    runHooks(HOOKS_D, HOOKS_IGNORE_SUFFIXES, c, e, CampaignHookTypes.CampaignPreHooks, None)
   }
 
   override def runPostHooks(c: Campaign, e: CampaignEvent): IOResult[HookResults] = {
-    runHooks(HOOKS_D, HOOKS_IGNORE_SUFFIXES, c, e, CampaignHookTypes.CampaignPostHooks)
+    // we should have a post-hook state, use it
+    e.state match {
+      case CampaignEventState.PostHooks(nextState, _) =>
+        runHooks(HOOKS_D, HOOKS_IGNORE_SUFFIXES, c, e, CampaignHookTypes.CampaignPostHooks, Some(nextState))
+      case _                                          =>
+        Inconsistency(
+          s"Can not run post-hooks for campaign event '${e.id.value}' which has state '${e.state.value.entryName}'"
+        ).fail
+    }
   }
 
   /*
@@ -160,7 +168,8 @@ class FsCampaignHooksService(
       ignoreSuffixes: List[String],
       c:              Campaign,
       e:              CampaignEvent,
-      hookType:       CampaignHookTypes // pre or post
+      hookType:       CampaignHookTypes, // pre or post
+      nextState:      Option[CampaignEventStateType]
   ): IOResult[HookResults] = {
 
     val loggerName = FsCampaignHooksRepository.CampaignDirName + "." + c.info.id.serialize + "." + hookType.entryName
@@ -176,22 +185,42 @@ class FsCampaignHooksService(
                        res        <- RunHooks.asyncRunHistory(
                                        loggerName,
                                        hooks,
-                                       HookEnvPairs.build(
-                                         ("CAMPAIGN_ID", c.info.id.serialize),
-                                         ("CAMPAIGN_NAME", c.info.name),
-                                         ("CAMPAIGN_TYPE", c.campaignType.value),
-                                         ("CAMPAIGN_EVENT_ID", e.id.value),
-                                         ("CAMPAIGN_EVENT_NAME", e.name)
-                                       ),
+                                       HookEnvPairs
+                                         .build(
+                                           ("CAMPAIGN_ID", c.info.id.serialize),
+                                           ("CAMPAIGN_NAME", c.info.name),
+                                           ("CAMPAIGN_TYPE", c.campaignType.value),
+                                           ("CAMPAIGN_EVENT_ID", e.id.value),
+                                           ("CAMPAIGN_EVENT_NAME", e.name)
+                                         )
+                                         .optAdd("CAMPAIGN_NEXT_STATE", nextState.map(_.entryName)),
                                        systemEnv,
                                        HookExecutionHistory.Keep,
                                        1.minutes // warn if a hook took more than a minute
                                      )
                        timeHooks1 <- currentTimeMillis
                        _          <-
-                         PureHooksLogger.For(loggerName).trace(s"Campaign pre-hooks ran in ${timeHooks1 - timeHooks0} ms")
+                         PureHooksLogger.For(loggerName).trace(s"Campaign ${hookType.entryName} ran in ${timeHooks1 - timeHooks0} ms")
+                       _          <- ZIO.foreachDiscard(res._2) { // log at the corresponding level campaign hook that did a warning/error.
+                                       case (hookName, code) =>
+                                         (code match {
+                                           case _: HookReturnCode.Warning =>
+                                             PureHooksLogger
+                                               .For(loggerName)
+                                               .warn(
+                                                 s"Campaign '${c.info.name}' ${hookType.entryName} returned code '${code.code}' in '${hookName}'"
+                                               )
+                                           case _: HookReturnCode.Error   =>
+                                             PureHooksLogger
+                                               .For(loggerName)
+                                               .error(
+                                                 s"Campaign '${c.info.name}' ${hookType.entryName} returned code '${code.code}' in '${hookName}'"
+                                               )
+                                           case _ => ZIO.unit
+                                         })
+                                     }
                      } yield res
-      } yield HookResults((res._1 :: res._2).map(HookResult.fromCode))
+      } yield HookResults((res._1 :: res._2.map(_._2)).map(HookResult.fromCode))
     ).catchAll { err =>
       PureHooksLogger.For(loggerName).error(err.fullMsg) *>
       HookResults(List(HookResult(loggerName, 1, "", err.fullMsg, "System error when executing hook"))).succeed
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
@@ -41,8 +41,10 @@ import cats.implicits.*
 import com.normation.errors.*
 import com.normation.rudder.campaigns.CampaignEventState.*
 import com.normation.rudder.campaigns.CampaignEventStateType.*
+import com.normation.rudder.hooks.HookReturnCode
 import com.normation.utils.DateFormaterService
 import com.normation.utils.StringUuidGenerator
+import com.softwaremill.quicklens.*
 import org.joda.time.{Duration as JTDuration, *}
 import zio.*
 import zio.syntax.*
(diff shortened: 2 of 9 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -141,6 +148,10 @@ sealed trait HookReturnCode {
 }
 
 object HookReturnCode {
+
+  def isError(code:   Int): Boolean = code < 0 || code > 0 && code < 32
+  def isWarning(code: Int): Boolean = code >= 32 && code < 64
```
fanf, line 153: This will go well

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -283,7 +294,7 @@ object RunHooks {
                PureHooksLogger.For(logIdentifier).trace(s"  -> stdout : ${result.stdout}") *>
                PureHooksLogger.For(logIdentifier).trace(s"  -> stderr : ${result.stderr}")
              }
-        _ <- ZIO.when(result.code >= 32 && result.code <= 64) { // warning
+        _ <- ZIO.when(HookReturnCode.isWarning(result.code)) { // warning
```
fanf, line 297: This is a change that will certainly lead to no problem at all, it's really not the kind of change that lead to error, at all. 
Please double check definition of  `isWarning`.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -306,9 +317,9 @@ object RunHooks {
         // using script error because we do have an error code and it can help in some case
 
         ScriptError(path, result.code, result.stdout, result.stderr, msg)
-      } else if (result.code >= 1 && result.code <= 31) { // error
+      } else if (HookReturnCode.isError(result.code)) { // error
         ScriptError(path, result.code, result.stdout, result.stderr, msg)
-      } else if (result.code >= 32 && result.code <= 64) { // warning
+      } else if (HookReturnCode.isWarning(result.code)) { // warning
```
fanf, line 322: This is a change that will certainly lead to no problem at all, it's really not the kind of change that lead to error, at all. 
Please double check definition of  `isError` and `isWarning`.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -323,44 +334,48 @@ object RunHooks {
...
-                   }
-              _ <- f.interrupt
-              c  = translateReturnCode(path, r)
-              _ <- logReturnCode(c)
-            } yield {
-              history match {
-                case HookExecutionHistory.DoNotKeep => (c, Nil)
-                // Noop are not real hook execution, just filter them out from the result
-                case HookExecutionHistory.Keep      => (c, (x :: historyList).filterNot(_ == Noop))
+    val runAllSeq: IOResult[(HookReturnCode, List[(String, HookReturnCode)])] = {
```
fanf, line 340: add the `String` which is the command name in the history of returned code

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunHooks.scala
```diff
@@ -323,44 +334,48 @@ object RunHooks {
...
+                         PureHooksLogger.LongExecLogger.error(msg) *> Unexpected(msg).fail
+                     }
+                _ <- f.interrupt
+                c  = translateReturnCode(path, r)
+                _ <- logReturnCode(c)
+              } yield {
+                history match {
+                  case HookExecutionHistory.DoNotKeep => (c, Nil)
+                  // Noop are not real hook execution, just filter them out from the result
+                  case HookExecutionHistory.Keep      => (c, ((nextHookName, c) :: historyList).filterNot(_._2 == Noop))
```
fanf, line 374: we now keep the last code also as head of the history, because we need the name. 
So we could just have no first returned value and use a non empty list with some default value and check the head, but I think we will want to have different global value and last one at some point - perhaps in case of warning for example.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/campaign/CampaignOrchestrationLogicTest.scala
```diff
@@ -262,16 +301,48 @@ class CampaignOrchestrationLogicTest extends ZIOSpecDefault {
...
+          test("all state must have been saved, and we skip from pre-hook to post-hook") {
+            import com.normation.rudder.campaigns.CampaignEventState.*
+
+            assert(t.reverse.map(_.state))(
+              equalTo(
+                List(
+                  Scheduled,
+                  PreHooks(HookResults(Nil)),
+                  PreHooks(HookResults(HookResult(e.id.value, 1, "pre-hooks", "", "") :: Nil)),
+                  PostHooks(FailureType, HookResults(Nil)),
```
fanf, line 336: look, directly from pre-hooks to post-hooks.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
```diff
@@ -265,18 +272,26 @@ class CampaignOrchestrationLogic(effects: CampaignOrchestrationEffects) {
...
-          val nextState =
-            res.results.collectFirst { case r if isError(r) => Failure("post-hooks were in error", r.stderr) }.getOrElse(Finished)
-          EventOrchestration.SaveThenUpdateAndQueue(event.copy(state = PostHooks(res)), nextState)
+          val nextState = {
+            val optError = res.results.collectFirst { case r if isError(r) => Failure("post-hooks were in error", r.stderr) }
+            (optError, state.nextState) match {
+              case (None, FailureType)     => Failure("pre-hooks were in error and post-hooks completed successfully", "")
+              case (None, s)               => getDefault(s)
+              case (Some(f1), FailureType) => Failure("pre-hooks and post-hooks were in error. Last error:", f1.cause)
+              case (Some(f1), _)           => f1
```
fanf, line 286: this logic is just here to put something relevant in the failure message. We don't keep the whole error, so we don't know what hook errored at that point. 
In the futur, we will just have the whole history in the campaign event UI and the user will be able to see what broke.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
```diff
@@ -176,22 +185,42 @@ class FsCampaignHooksService(
...
+                                         )
+                                         .optAdd("CAMPAIGN_NEXT_STATE", nextState.map(_.entryName)),
                                        systemEnv,
                                        HookExecutionHistory.Keep,
                                        1.minutes // warn if a hook took more than a minute
                                      )
                        timeHooks1 <- currentTimeMillis
                        _          <-
-                         PureHooksLogger.For(loggerName).trace(s"Campaign pre-hooks ran in ${timeHooks1 - timeHooks0} ms")
+                         PureHooksLogger.For(loggerName).trace(s"Campaign ${hookType.entryName} ran in ${timeHooks1 - timeHooks0} ms")
```
fanf, line 203: all that foreach just for better logs :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignHooksService.scala
```diff
@@ -176,22 +185,42 @@ class FsCampaignHooksService(
...
-                                       ),
+                                       HookEnvPairs
+                                         .build(
+                                           ("CAMPAIGN_ID", c.info.id.serialize),
+                                           ("CAMPAIGN_NAME", c.info.name),
+                                           ("CAMPAIGN_TYPE", c.campaignType.value),
+                                           ("CAMPAIGN_EVENT_ID", e.id.value),
+                                           ("CAMPAIGN_EVENT_NAME", e.name)
+                                         )
+                                         .optAdd("CAMPAIGN_NEXT_STATE", nextState.map(_.entryName)),
```
fanf, line 196: the new parameter, only defined in `post-hooks`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
```diff
@@ -265,18 +272,26 @@ class CampaignOrchestrationLogic(effects: CampaignOrchestrationEffects) {
         }
 
       case PostHooksType =>
+        val state = event.state.asInstanceOf[PostHooks]
```
clarktsiory, line 275: could this not be avoided, by matching on the `event.state: CampaignEventState` instead ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
```diff
@@ -265,18 +272,26 @@ class CampaignOrchestrationLogic(effects: CampaignOrchestrationEffects) {
...
           res <- effects.runPostHooks(campaign, event)
         } yield {
           // hooks are special: we don't call per-campaign type handler and we need to first save before change state/update
-          val nextState =
-            res.results.collectFirst { case r if isError(r) => Failure("post-hooks were in error", r.stderr) }.getOrElse(Finished)
-          EventOrchestration.SaveThenUpdateAndQueue(event.copy(state = PostHooks(res)), nextState)
+          val nextState = {
+            val optError = res.results.collectFirst { case r if isError(r) => Failure("post-hooks were in error", r.stderr) }
+            (optError, state.nextState) match {
+              case (None, FailureType)     => Failure("pre-hooks were in error and post-hooks completed successfully", "")
```
clarktsiory, line 283: does the `FailureType` not mean that the post-hook completed with a "failure", instead of "successfully" ?

VinceMacBuche: no optError is None then it mean that post-hooks ran successfully, but we have nextState in Failure type, meaning that there was an error in pre-hooks

VinceMacBuche: The case you talk about is treated two lines below

fanf: 👍

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignOrchestrationLogic.scala
fanf: Perhaps, but most 8f the time it's far too powerful for the workflow engine, those state's coverture becomes very hard to check.
But I agree that having that is a smell in the encoding that exposes a mismatch. I chose ro keep it because it's very obviously Not Good and claims it needs to be refactored and I didn't want to massively extends the risks with a big refactoring so late.

fanf:
OK, squash merging this PR

PR #6621, opened 2025-09-29
Fixes #27604: Update onboarding documentations

P4uline:
https://issues.rudder.io/issues/27604


contributing/webapp.md
```diff
@@ -51,62 +82,80 @@ This is a **straightforward guide** aims to setup all tools and settings on a Li
...
-mv elm /usr/local/bin/
-```
 
-```bash
-wget -qO elm-0.19.1.gz https://github.com/elm/compiler/releases/download/0.19.1/binary-for-linux-64-bit.gz
-gzip -d elm-0.19.1.gz
-chmod +x elm-0.19.1
-mv elm-0.19.1 /usr/local/bin/
-```
+Follow the elm installer tutorial https://github.com/elm/compiler/blob/master/installers/linux/README.md
```
fanf: "The goal is to have elm 19.1 installed on your machine. If you know how to do that, do it like you want. If not and your distribution provides that version, just go for that. Else follow ...."

P4uline:
PR updated with a new commit

contributing/webapp.md
amousset: I dont't think we need/want a global Elm cli installed. We should use the version handled by npm instead (like we do for scalac/maven)

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

contributing/webapp.md
P4uline: Indeed `elm` is now managed by `npm` as a dependency of the webapp, see `package.json` : 
```
devDependencies": {
    ...
    "elm": "^0.19.1-6",
```
So I just removed completely the `elm` paragraph.

fanf: we need a stable-path local install of elm for intellij plugin

fanf: hum, we can deal with that in relation with the elm plugin installation, once we have one

P4uline:
OK, merging this PR

PR #6634, opened 2025-10-02
Fixes #27644: Rudder 9.0 Beta 2 : Error message when deleting technique in editor

fanf:
https://issues.rudder.io/issues/27644


webapp/sources/api-doc/paths/techniques/id-version.yml
```diff
@@ -91,19 +91,21 @@ delete:
...
-                    properties:
-                      id:
-                        type: string
-                        description: Deleted technique ID
-                        example: foo
-                      version:
-                        type: string
-                        description: Deleted technique version
-                        example: 1.0
+                    type: array
```
fanf, line 94: added array in return value

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -235,7 +235,16 @@ class TechniqueApi(
           force <- extractBoolean("force")(req).map(_.getOrElse(false)).toIO
           _     <- techniqueWriter.deleteTechnique(techniqueInfo._1, techniqueInfo._2, force, modId, authzToken.qc)
         } yield {
-          Json.Obj(("id", Json.Str(techniqueInfo._1)), ("version" -> Json.Str(techniqueInfo._2)))
+          // here, up to 8.3 we used to just return: `"data": { "techniques": { "id":...}}`
```
fanf, line 238: this is the part where we build by hand the return `data` object since we want to have an array or not depending of API version

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ApiCalls.elm
```diff
@@ -152,7 +152,7 @@ deleteTechnique  technique model =
         , headers = [header "X-Requested-With" "XMLHttpRequest"]
         , url     = getUrl model "techniques/" ++ technique.id.value ++ "/" ++ technique.version
         , body    = emptyBody
-        , expect  = Detailed.expectJson DeleteTechnique ( Json.Decode.at ["data", "techniques" ] ( decodeDeleteTechniqueResponse ))
+        , expect  = Detailed.expectJson DeleteTechnique ( Json.Decode.at ["data", "techniques" ] (list decodeDeleteTechniqueResponse ) |> headList)
```
fanf, line 155: and this is the elm part that change the decoding to know that there is an array, and we only want the (only one) item in it

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompiler.scala
```diff
@@ -414,7 +414,11 @@ class RuddercTechniqueCompiler(
                    }
       // clean-up generated files
       _         <- ZIO.foreach(TechniqueFiles.Generated.all) { name =>
-                     IOResult.attempt((gitDir / getTechniqueRelativePath(technique) / name).delete(true))
+                     val d = gitDir / getTechniqueRelativePath(technique) / name
```
P4uline: I would inline the values `d` and `x` as they're not used further.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
```diff
@@ -331,25 +348,79 @@ class MockTechniques(configurationRepositoryRoot: File, mockGit: MockGitConfigRe
...
-        "",
-        ""
-      ).succeed
+  // a false technique compiler that return a value depending of the name of the
+  // technique in parameter (known by the name of the parent path in technique.path)
+  // - if it contains "technique_with_error", then return 42 + error message
+  // - else returns 0
+  // ALSO: it saves the metadata of techniques before write in `gitMock.abstractRoot / ${technique_name}_metadata_save`
+  // for investigation like in TestMigrateJsonTechniquesToYaml
+  val rudderc: RuddercService = new RuddercService {
```
P4uline, line 357: This code looks unused, if so I would simply remove it so we would get rid of a big block to maintain. Same for the other changes in the file.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -235,7 +235,16 @@ class TechniqueApi(
...
+          // in place of the normalized `"data": { "techniques": [ {"id":...}]}
+          // so we need to correct before API version 21
+          // see https://issues.rudder.io/issues/27644
+          val res     = Json.Obj(("id", Json.Str(techniqueInfo._1)), ("version" -> Json.Str(techniqueInfo._2)))
+          val content = {
+            if (version.value < 22) res
+            else Json.Arr(Chunk(res))
+          }
+          Json.Obj(("techniques", content))
```
P4uline, line 247: Here I would perform a little refactoring to get rid of the if/else statement in favor of a pattern matching, just to avoid imperative code style. Because it doesn't get along with functional programming and it could make future changes difficult.

Something like:
```
val techniques = Json.Obj(("id", Json.Str(id)), "version" -> Json.Str(v))
techniqueInfo match {
  case (id, v) if version.value < 22 => Json.Obj(("techniques", techniques))
  case (id, v) => Json.Obj(("techniques", Json.Arr(Chunk(techniques))))
}
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
```diff
@@ -235,7 +235,16 @@ class TechniqueApi(
           force <- extractBoolean("force")(req).map(_.getOrElse(false)).toIO
           _     <- techniqueWriter.deleteTechnique(techniqueInfo._1, techniqueInfo._2, force, modId, authzToken.qc)
         } yield {
-          Json.Obj(("id", Json.Str(techniqueInfo._1)), ("version" -> Json.Str(techniqueInfo._2)))
+          // here, up to 8.3 we used to just return: `"data": { "techniques": { "id":...}}`
+          // in place of the normalized `"data": { "techniques": [ {"id":...}]}
+          // so we need to correct before API version 21
+          // see https://issues.rudder.io/issues/27644
+          val res     = Json.Obj(("id", Json.Str(techniqueInfo._1)), ("version" -> Json.Str(techniqueInfo._2)))
```
P4uline, line 242: Just a general remark to that I would avoid tuple type as much as I can, because of the lake of readability of `._1` and `._2`. It make the code review difficult. But let's keep as is, refactoring the type `techniqueInfo` is out of the scope of the ticket.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/TechniqueCompiler.scala
fanf: ok - it was for debugging and can now be inlined

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/MockServices.scala
fanf: it is used just below, line 390:
```
  val techniqueCompiler = new RuddercTechniqueCompiler(
    rudderc,
    _.path,
    mockGit.configurationRepositoryRoot.pathAsString
  )```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
fanf: that make more duplicated code and less obvious what is the part that differs. I tend to prefer to scope as small as possible the delta, so that it's clear why there is an `if`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/TechniqueApi.scala
fanf: it won't be a problem in a near futur since scala 3.7, tupled can be named, see: https://www.scala-lang.org/api/current/docs/other-new-features/named-tuples.html
We are just waiting for the tooling to converge, and to have a bit more of code in scala 3.7+. Then, we will be able to migrate all these `_1`. 

Moreover, tuple types have a real difference in several places: 
- it makes clearer that such a thing isn't a domain related data structure, but is here just here for technicalities (like here, or more canonically for DB results),
- some performance aspects: tuple cost nothing wrt to class size and runtime path optimization,
- it's actually what the parameter is, like in that case : we have pairs of things.

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #6712, opened 2025-11-12
Fixes #27873: Rudder package should always clean the /opt/rudder/share/plugins/<plugin id> folder before unpacking a plugin

Fdall:
https://issues.rudder.io/issues/27873

### DIFF
--- a/relay/sources/rudder-package/src/archive.rs
+++ b/relay/sources/rudder-package/src/archive.rs
@@ -11,6 +11,7 @@ use std::{
 
 use anyhow::{Context, Ok, Result, anyhow, bail};
 use ar::Archive;
+use regex::Regex;
 use serde::{Deserialize, Serialize};
 use tracing::{debug, info};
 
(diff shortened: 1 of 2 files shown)


Fdall:
PR updated with a new commit

relay/sources/rudder-package/src/archive.rs
```diff
@@ -154,6 +156,22 @@ impl Rpkg {
...
+        let safe_name_pattern = Regex::new(r"^[a-zA-Z0-9\-_]+$")?;
+        if safe_name_pattern.is_match(self.metadata.short_name()) {
+            bail!("The metadata short name contains illegal characters");
+        }
+        Ok(PathBuf::from(PACKAGE_CONTENT_DEFAULT_FOLDER).join(self.metadata.short_name()))
+    }
+
+    fn get_default_package_scripts_extract_folder(&self) -> Result<PathBuf> {
+        let safe_name_pattern = Regex::new(r"^[a-zA-Z0-9\-_]+$")?;
+        if safe_name_pattern.is_match(&self.metadata.name) {
```
amousset, line 169: we should probably ensure safe names in the metadata constructors, likely with a dedicated type enforcing these constraints

amousset: A bit like we do for `Condition` in rudderc (with only a stirng inside obviously), with specific Deserializer

relay/sources/rudder-package/src/archive.rs
```diff
@@ -213,11 +231,18 @@ impl Rpkg {
             db.uninstall(&self.metadata.name, false, webapp)?;
```
amousset, line 231: shouldn't the cleanup be done by the uninstall step instead?

PR #6983, opened 2026-03-07
Fixes #28503: Migrate Node hash repo to zio-jon

fanf:
https://issues.rudder.io/issues/28503

This might be better targeted to 9.2 - keeping in draft for now. 

The core of that PR is to remove `lift-json` from `NodeConfigurationCacheRepository.scala`. 

That core manages the hashes of node's policy information to know if they are up to date of nore a new policy generation. 
The saved hashes used to be in LDAP, but are in a file (`/var/rudder/policy-generation-info/node-configuration-hashes.json`) since Rudder 6.1. So first, we can clean-up the LDAP related code. 

Next, we switch from an ad-hoc json serialization to a standard one, adding `case classes` that will be automatically derived into json codec. 

The chosen design is to place these case classes in the companion objects of the related business classes, with a restricted visibility : they are not used anywhere else, and placing them here is more idiomatic than placing them in the repository implementation, and it avoid fighting against scala for `given` resolution (no `import` needed). 

So, for each class, we have: 
- the business class (for ex `NodeConfigurationHash`), 
- in the companion object : 
  - a `protected[nodeconfig]` DTO with a `Json` suffix (for ex: `NodeConfigurationHashJson`)
  - chimney transformers between the two, in the corresponding companion object so that the `given` is correctly looked-up, 
  - the JSON codec for the DTO. 

In addition, I moved `NodeId`, `RuleId` and `DirectiveId` JSON codecs to their companion objects to continue to make that design more systematic.

*Format impact on `node-configuration-hashes.json`*

The main functionnal change is that there is now many more spaces and lines in `node-configuration-hashes.json`. This is because the previous encoding chose to have one line per node exactly. 
The idea was that is was simpler for an human looking in the file that way, but the reality is not very convincing. 
The lines are VERY long, since each policy is >80 chars, and a node generally have tens of them. 
So the new encoding, which is `.toJsonPretty`, put each policy on one line. 
If anybody wants to look into that file, it will be easier for a given node to see the list of policies and their hashes.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
@@ -43,12 +43,6 @@ import com.normation.cfclerk.domain.TechniqueVersion
 import com.normation.cfclerk.domain.Variable
 import com.normation.errors.*
 import com.normation.inventory.domain.NodeId
-import com.normation.ldap.sdk.LDAPConnectionProvider
-import com.normation.ldap.sdk.LDAPEntry
-import com.normation.ldap.sdk.RwLDAPConnection
-import com.normation.rudder.domain.RudderDit
-import com.normation.rudder.domain.RudderLDAPConstants.A_NODE_CONFIG
-import com.normation.rudder.domain.RudderLDAPConstants.OC_NODES_CONFIG
 import com.normation.rudder.domain.logger.ApplicationLoggerPure
 import com.normation.rudder.domain.logger.PolicyGenerationLoggerPure
 import com.normation.rudder.domain.policies.DirectiveId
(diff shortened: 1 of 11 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
```diff
@@ -147,94 +130,61 @@ final case class NodeConfigurationHash(
...
-          hash.parameterHash,
-          hash.nodeContextHash
-        )
-      ))
-      ~ ("p" -> JArray(
-        hash.policyHash.toList.map(p => {
-          JArray(
-            List(p.draftId.ruleId.serialize, p.draftId.directiveId.serialize, p.draftId.techniqueVersion.serialize, p.cacheValue)
+  final private[nodeconfig] case class NodeConfigurationHashJson(
+      i: (NodeId, Instant, Int, Int, Int),
```
clarktsiory, line 146: I didn't know tuples were automatically encoded as json lists...

clarktsiory (approved):
LGTM !

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
fanf: Yes, a surprise for me too, I was more enclined to believe they would be some kind of map (but it doesn't work for more than 2 dimensions)

fanf:
OK, merging this PR

PR #6985, opened 2026-03-09
Fixes #28507: Add ADR for Handling IO encoding in Rust modules

m4rtinh4rt:
https://issues.rudder.io/issues/28507

### DIFF
--- a/adr/system/28504-io-encoding-in-rust-modules.md
+++ b/adr/system/28504-io-encoding-in-rust-modules.md
@@ -1,36 +1,39 @@
-# IO encoding in Rust modules
+# Disk I/O encoding in Rust modules
 
-- Status: draft
+- Status: accepted
 - Deciders: AMO, FDA, MHA
 - Date: 2026/03/09
 
-## Context 
+## Context
 
 When reading from a file, all Rust modules expect the data to be valid UTF-8.
-This requirement comes both from the language itself—since a Rust String is
-expected to contain valid UTF-8—and from the serialization library used, which
-also expects UTF-8 input. As a result, errors occur on Windows when the modules
-encounter files encoded in UTF-16 with a
-[BOM](https://en.wikipedia.org/wiki/Byte_order_mark). Additionaly Powershell
+This requirement comes from the language itself, as a Rust String is expected
+to contain valid UTF-8, and from the serialization library used, which
+also expects UTF-8 input.
+
+As a result, errors occur on Windows when the modules encounter files encoded
+in UTF-16 with a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark).
+Additionaly Powershell
 [requires](https://docs.microsoft.com/en-us/windows/desktop/intl/using-byte-order-marks)
 a BOM added at the beginning of all files when using UTF-8 encoding.
 
 ## Decision
 
-- Introduce the new `rudder_read_file` function.
+Introduce two new functions for interacting with files.
+
+- `rudder_read_file`
   - This function behaves like `std::fs::read_to_string`, with encoding detection and conversion added.
   - It supports file encoded as UTF-8, UTF-8 with BOM and UTF-16 BOM LE.
-  - Returns a valid UTF-8 string or an error if the file cannot be decoded.
+  - It returns a valid UTF-8 string or an error if the file cannot be decoded.
 
-- Introduce the new `rudder_write_file` function.
-  - Takes as parameter the file path, the data string, and an enum specifying the target encoding.
-  - All writes to disk should be atomic.
-  - On Windows, files written by this function must include a BOM:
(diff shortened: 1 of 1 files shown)


m4rtinh4rt:
PR updated with a new commit

adr/system/28504-io-encoding-in-rust-modules.md
```diff
@@ -0,0 +1,39 @@
...
+to contain valid UTF-8, and from the serialization library used, which
+also expects UTF-8 input.
+
+As a result, errors occur on Windows when the modules encounter files encoded
+in UTF-16 with a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark).
+Additionaly Powershell
+[requires](https://docs.microsoft.com/en-us/windows/desktop/intl/using-byte-order-marks)
+a BOM added at the beginning of all files when using UTF-8 encoding.
+
+## Decision
```
amousset, line 25: This should first describe the general principles we chose to follow before explaining the implementation.

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

adr/system/28504-io-encoding-in-rust-modules.md
```diff
@@ -0,0 +1,41 @@
...
+  - Takes as parameter the file path, the data string in UTF-8, and an enum specifying the target encoding.
+  - It supports writing files as UTF-8, UTF-8 with BOM and UTF-16 BOM LE.
+  - All writes to disk are atomic.
+  - On Windows, files written by this function include a BOM:
+    - UTF-8 content, a three-byte BOM: `EF BB BF`
+    - UTF-16 LE content, a two-byte BOM: `FF FE`
+
+## Consequences
+
+The new functions must be used when performing I/O operations on files.
```
amousset: only on Windows or everywhere?

adr/system/28504-io-encoding-in-rust-modules.md
```diff
@@ -0,0 +1,41 @@
+# Disk I/O encoding in Rust modules
+
+- Status: accepted
+- Deciders: AMO, FDA, MHA
+- Date: 2026/03/09
+
+## Context
```
amousset, line 7: could you add a note about current BOM usage (rudderc & windows agent at least)

adr/system/28504-io-encoding-in-rust-modules.md
```diff
@@ -0,0 +1,41 @@
...
+- `encoding::unicode_file_to_string`
+  - This function behaves like `std::fs::read_to_string`, with encoding detection and conversion added.
+  - It supports file encoded as UTF-8, UTF-8 with BOM and UTF-16 BOM LE.
+  - It returns a valid UTF-8 string or an error if the file cannot be decoded.
+
+- `encoding::write_file`
+  - Takes as parameter the file path, the data string in UTF-8, and an enum specifying the target encoding.
+  - It supports writing files as UTF-8, UTF-8 with BOM and UTF-16 BOM LE.
+  - All writes to disk are atomic.
+  - On Windows, files written by this function include a BOM:
```
amousset, line 40: why only on Windows? What would be the behavior on Linux?

adr/system/28504-io-encoding-in-rust-modules.md
m4rtinh4rt: I think we want to use those functions everywhere to abstract the OS-specific logic (BOM vs. non-BOM) so we don’t have to repeat it in every module.

adr/system/28504-io-encoding-in-rust-modules.md
m4rtinh4rt: The current code in rudderc adds an UTF-8 BOM only for the Windows backend.

```Rust
impl Backend for Windows {
    fn generate(
        &self,
        technique: Technique,
        resources: &Path,
        _standalone: bool,
    ) -> Result<String> {
        // Powershell requires a BOM added at the beginning of all files when using UTF8 encoding
        // See https://docs.microsoft.com/en-us/windows/desktop/intl/using-byte-order-marks
        // Bom for UTF-8 content, three bytes: EF BB BF https://en.wikipedia.org/wiki/Byte_order_mark
        const UTF8_BOM: &[u8; 3] = &[0xef, 0xbb, 0xbf];
        let mut with_bom = String::from_utf8(UTF8_BOM.to_vec()).unwrap();
        with_bom.push_str(&Self::technique(technique, resources)?);
        Ok(with_bom)
    }
}

```

It is not needed on Linux and [could potentially break things](https://unix.stackexchange.com/a/414426).

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

PR #6987, opened 2026-03-09
Fixes #28509: Port quicksearch to zio-json

fanf:
https://issues.rudder.io/issues/28509

# Commit 1: First part of the work: port to "new" API framework. 
This is a bit tedious because we once again it the "platform is limited to 254 params", so I tried a new approach where I user an intermediate object to avoid building a list because of the `lazy val`. 

Everything else is rather automatic, and we remove the direct use of `userService`, which is good!

# Commit 2: port to zio-json

Nothing special, the usual workflow: 
- create case classes/objects to describe the data structure, 
- use our standard utilities for response generation. 

The only difference compared to other API is that I needed to define some special constructors (the `CompanionObject.from` methods) to minimze the changes, because some computation is done on the inputs compared to what is returned in JSON.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
@@ -1668,6 +1668,26 @@ object ArchiveApi       extends Enum[ArchiveApi] with ApiModuleProvider[ArchiveA
   def values = findValues
 }
 
+sealed trait QuickSearchApi extends EnumEntry with EndpointSchema with InternalApi with SortIndex {
+  override def dataContainer: Option[String] = None
+}
+object QuickSearchApi       extends Enum[QuickSearchApi] with ApiModuleProvider[QuickSearchApi]   {
+
+  case object CompleteTagsValue extends QuickSearchApi with ZeroParam with StartsAtVersion23 with SortIndex {
+    val z: Int = implicitly[Line].value
+    val description    = "Do a quicksearch"
+    val (action, path) = GET / "quicksearch"
+    val authz: List[AuthorizationType] = {
+      import com.normation.rudder.AuthorizationType.*
+      Group.Read :: Rule.Read :: Directive.Read :: Technique.Read :: Parameter.Read :: Configuration.Read :: Node.Read :: Nil
+    }
+  }
+
+  def endpoints: List[QuickSearchApi] = values.toList.sortBy(_.z)
+
+  def values = findValues
+}
+
 /*
  * All API.
  */
@@ -1687,6 +1707,7 @@ object AllApi {
     InfoApi.endpoints :::
     HookApi.endpoints :::
     ApiAccounts.endpoints :::
+    QuickSearchApi.endpoints :::
     // UserApi is not declared here, it will be contributed by plugin
     Nil
   }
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -1668,6 +1668,23 @@ object ArchiveApi       extends Enum[ArchiveApi] with ApiModuleProvider[ArchiveA
...
+sealed trait QuickSearchApi extends EnumEntry with EndpointSchema with InternalApi with SortIndex {
+  override def dataContainer: Option[String] = None
+}
+object QuickSearchApi       extends Enum[QuickSearchApi] with ApiModuleProvider[QuickSearchApi]   {
+
+  case object CompleteTagsValue extends QuickSearchApi with ZeroParam with StartsAtVersion23 with SortIndex {
+    val z: Int = implicitly[Line].value
+    val description    = "Do a quicksearch"
+    val (action, path) = GET / "quicksearch"
+    val authz: List[AuthorizationType] = AuthorizationType.Configuration.Read :: AuthorizationType.Node.Read :: Nil
```
clarktsiory: Shouldn't it be extended to all the authz we check within the `filter` implementation of the endpoint ? 
```scala
    val nodeOK       = u.checkRights(AuthorizationType.Node.Read)
    val groupOK      = u.checkRights(AuthorizationType.Group.Read)
    val ruleOK       = u.checkRights(AuthorizationType.Configuration.Read) || u.checkRights(AuthorizationType.Rule.Read)
    val directiveOK  = u.checkRights(AuthorizationType.Configuration.Read) || u.checkRights(AuthorizationType.Directive.Read)
    val techniqueOK  = u.checkRights(AuthorizationType.Technique.Read)
    val parametersOK =
      u.checkRights(AuthorizationType.Configuration.Read) || u.checkRights(AuthorizationType.Parameter.Read)
```
➡️ it would be missing `Group.Read`, `Rule.Read`, `Directive.Read`, `Technique.Read` and `Parameter.Read`

clarktsiory (changes requested):
LGTM apart from the authz definition for the endpoint

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
fanf: Hum. Configuration authz is strange, because its comment says `// configure param, techniques, directives, rules, groups` so it should actuall be a super set of `param`, `technique`, `directive`, `rule`, `group`. 
I think we have something to clean-up here, likely remove `AuthorizationType.Configuration` which is not a real atomic type, and use a compount authorization for that. 
But in the meantime, the correction looks more like adding `|| u.checkRights(AuthorizationType.Configuration.Read)` for technique and group too. 
And perhaps use the `configurationKind` check in endpoint.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
fanf: done (just adding the missing authz)

clarktsiory (approved):

PR #7024, opened 2026-03-24
Fixes #28571: Have an alias for dataTables classes

clarktsiory:
https://issues.rudder.io/issues/28571

Ths likely needs to be retargeted in 9.2 : we should either rename "dataTable" completely, or we should not be wondering whether the CSS is used/taking risk of not applying an alias

See https://github.com/Normation/rudder/pull/7273


clarktsiory:
Commit modified

clarktsiory:
Commit modified

clarktsiory:
PR replaced by https://github.com/Normation/rudder/pull/7273

PR #7048, opened 2026-04-02
Fixes #28681: Buttons in tree headers do not auto-wrap when the screen becomes too small

RaphaelGauthier:
https://issues.rudder.io/issues/28681

This PR globally fixes the line-break issue in the sidebar headers:

Capture d’écran du 2026-04-02 15-46-09

----

Capture d’écran du 2026-04-02 15-45-04

----

Capture d’écran du 2026-04-02 15-44-42

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueList.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueList.elm
@@ -89,7 +89,7 @@ techniqueList model techniques =
   in
     div [ class "template-sidebar sidebar-left col-techniques", onClick OpenTechniques ] [
       div [ class "sidebar-header"] [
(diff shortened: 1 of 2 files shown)


skaerg (commented):
Above the techniques tree, the "Techniques" title wraps on itself; is this unavoidable ? Also, the label that shows the number of user-created techniques on the right (the small grey label that contains a "0" ) gets more deformed as the window becomes smaller 
Image

Similarly, the system-updates "Add campaign" button still gets partially hidden even though it auto-wraps to the next line, and the "System update campaigns" title auto-wraps in the middle of the words, but maybe these effects are unavoidable as well ?
Image

RaphaelGauthier:
@skaerg I think this is fine, as this issue only occurs when the window is extremely small (~ <700px). To fix this, we should completely overhaul the design of the sidebars at this resolution (e.g. making them collapsible?)

skaerg (approved):

PR #7051, opened 2026-04-02
Fixes #28683: Change logs pagination cannot be set to "All"

skaerg:
https://issues.rudder.io/issues/28683

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
@@ -2065,6 +2065,7 @@ function createEventLogTable(gridId, data, contextPath, refresh, serverTimezone)
           }
 
       }
+    , "lengthMenu": [ [10, 25, 50, 100, 500, 1000], [10, 25, 50, 100, 500, 1000] ]
     , "sDom": '<"dataTables_wrapper_top newFilter d-flex"f<"d-flex ms-auto my-auto" B <"dataTables_refresh ms-2" r>>'+
       '>t<"dataTables_wrapper_bottom"lip>'
     , "buttons" : [ csvButtonConfig("change_logs") ],


P4uline (changes requested):
Getting all logs will lead to memory issues

skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -233,23 +233,30 @@ class EventLogService(
...
+      } else {
+        (None, s"all logs since log #${start}")
+      }
+    }
 
     (for {
       events      <-
         repo
-          .getEventLogByCriteria(queryCriteria, Some(length), Nil, from)
-          .chainError(s"Error when trying fetch eventlogs from database for page ${(start / length) + 1}")
```
skaerg, line 250: i opted to modify this error message as `${(start / length) + 1}` does not give an accurate view of how many logs are retrieved, and because it could also lead to a division by 0

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -233,23 +233,30 @@ class EventLogService(
...
       // we also need the orderBy here because it should come before the offset
       criteria.getOrElse(fr"1=1") ++ orderBy ++ offset
     )
+    val (lengthOpt, selectedLogs) = {
+      if (length >= 0) {
+        (Some(length), s"logs #${start} to #${start + length}")
+      } else {
+        (None, s"all logs since log #${start}")
+      }
+    }
```
clarktsiory: I would recomment failing (e.g. with a 400 Bad Request and nice debugging message) when the provided length is not valid i.e. `0 < length < MAX_LEN` where the maximum could be the same as the max one for pagination (1000) because the API is internal and dedicated for that.

This would also make our API more reliable to changes (evolution of max length) as it can be directly tested in the UI, and in unit tests, and it would also make it more RESTful.

Silently ignoring invalid parameters could make things harder to debug in the future

P4uline: I would have opened a new thread on slack to discuss this, there are a couple of angles to consider:
- decoupling layers made in 9.1
- desing of API like you mentionned, an expressive message would be a great help indeed :+1: 
- is this business rule valid in which context

In the context of getting the logs of the user, i.e. in the change logs (or in the activity square's dashboard in 9.1), I am pretty sure we don't to get more than 1000 logs. 
Is it the case in other contexts ?  Putting the limit to 1000 seems very arbitrary... Are we 100% sure it will never need to cross that limit ? 
I am not sure so I would like a discussion on slack to gather more opinions on it.

Another point: the service is refactored in 9.1, there is no user service in 9.0, but this is where we should implement this business rule in 9.1.

clarktsiory: Indeed @P4uline the API has changed in 9.1, that is a good catch.
Thank you for pointing all of these aspects of the change. Since we have enforced a service that encapsulates business rules in 9.1, we need to watch out the considerations of changing the code in the different branches.


### My proposals for 9.0 and 9.1 
Because of the divergence in the two branches, and our policy of addressing potential security risks in maintained branches, I have these suggestions : 
* **in 9.0** (this PR) : avoid the security issue by enforcing a hard limit, and as I pointed out, the `limit` shouldn't be an unsanitized value from the frontend, and we should prioritize security before the concerns of having a clean architectural solution, at least in that branch   
  * since it would a radically different solution from 9.1 and is only to be maintained for a few more months, I would just enforce the constraints at the same place for `0 < limit` and `limit < MAX_LEN`
* **in 9.1** (it would be better in a dedicated PR, rather than an upmerge) : we will be able to enforce a clean business logic
  * @skaerg could make a PR that reflects the same change, but at the right place
  * @P4uline since the separation of concerns that was added in 9.1, could you point out the right place for the change in 9.1 ? (my take is that, we could be using a dedicated _opaque_ type for the `limit` field in the business object, since we don't have _refined_ types) 

### What is the hard limit ?
Regarding the choice of `MAX_LEN`, the solution is simple because it is an internal API, and we know exactly where this is used : 
  * in the event logs, the max has always been 1000
  * in the dashboard, the max number of elements being returned is 20
Assuming in both cases that it would be unlikely more than 1000, I would set the max limit to `1000`, and change it only if the business constraints evolve

skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -246,12 +246,16 @@ class EventLogService(
...
+        if (start < 0) Inconsistency("Start log id cannot be inferior to zero").fail
+        (Some(length), s"logs #${start} to #${start + length}").succeed
+      events                    <-
         repo
-          .getEventLogByCriteria(queryCriteria, Some(length), Nil, from)
-          .chainError(s"Error when trying fetch eventlogs from database for page ${(start / length) + 1}")
-      totalRecord <- repo.getEventLogCount(None)
-      totalFilter <- repo.getEventLogCount(criteria, from)
+          .getEventLogByCriteria(queryCriteria, lengthOpt, Nil, from)
+          .chainError(s"Error when trying fetch event logs from database for ${selectedLogs}")
```
clarktsiory: As discussed (in previous comments and IRL), the change in the backend code would be complicated to upmerge, and is a different matter since it is about having a "nicer API".

Could you simply remove the backend change, and report it in a new "enhancement" issue, and :
* make a PR in 9.0 that does that, plus also adds an upper bound for the `length
* make the same PR in 9.1, taking the changed code into account
?

clarktsiory (changes requested):
There is that + I don't know why the CI would report an error like : 
```scala
[ERROR] Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.43.0:check (default-cli) on project rudder-rest: Unable to check file /srv/jenkins/workspace/rudder_PR-7051/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala: <input>:252: error: [dialect scala213source3 [with overrides]] <- expected but . found

[ERROR]         (Some(length), s"logs #${start} to #${start + length}").succeed
[ERROR]                                                                ^
```

skaerg:
PR updated with a new commit

clarktsiory (approved):

P4uline (approved):

clarktsiory:
OK, squash merging this PR

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
skaerg: https://issues.rudder.io/issues/28754

PR #7052, opened 2026-04-02
Fixes #28651: Ignored node compliance are seens as no report

fanf:
https://issues.rudder.io/issues/28651

So, the correction is easy, but finding the best place to put it was not obvious. 
The idea is that a node in `Ignored` state should not have any `NodeStatusReport`. It's absolutly necessary to have that computation done in only one place so that everything is consistent. 

The idea is: somewhere appropriated, for all disabled nodes, replace whatever `NodeStatusReport` the node could have to a `NodeStatusReport` with type `ReportsDisabledInInterval` and zero compliance reports in it. 

I first thought to put the change in `FindNewNodeStatusReports` (either in `findUncomputedNodeStatusReports()` since it would make sense to just avoid getting reports from DB and computing things into it, or in `buildNodeStatusReports` which is the computation part, and used in all path where we do compute compliance). 
But it was hard to test, and it's just a pur computation (a simple if/then/else, even), and it didn't seem to fit that part of the code which does a lot of I/O. 

So I looked where it would better fit without hurting the general design and separation of concerns, and `ExecutionBatch.scala` is the perfect fit: it's pure `NodeStatusReport` computation. 

Then, we need to force cleaning-up compliance cache when the node becomes ignored, because since it removes it from policy generation (it's the goal), we won't have new expected reports, and so we won't have update until the next node run or until we overflow the run frequency. 

For that, I used a `CoreNodeFactHook`, which is really such a good design/power tool. Actuall, there was already a hook for that purpose, so I just extended it. 

There is a new info log for that, it looks like it's interesting enought from ops debug pow: 

```
2026-04-02 16:38:39+0000 INFO  nodes - Node 'node1.rudder.local' [e25a8aec-1e1c-4c5d-871e-617867b86060] state changed from 'enabled' to 'ignored': remove that node from compliance and expected report cache to recompute them
```

Remaning things are just cleaning up (typo, bad naming, old commented code...). 

And ignored node appears with compliance "N/A" which seems ok. 

image

image

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactChangeEventCallback.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactChangeEventCallback.scala
@@ -66,7 +66,7 @@ import com.normation.rudder.services.reports.InvalidateCache
 import com.normation.utils.DateFormaterService
 import com.normation.utils.StringUuidGenerator
 import java.time.Instant
-import zio.ZIO
+import zio.*
 
 /*
  * This file store callbacks for node events.
@@ -171,7 +171,10 @@ class ScoreUpdateOnNodeFactChange(scoreServiceManager: ScoreServiceManager, scor
 }
 
 /*
- * Callback related to cache invalidation when a node changes
+ * Callback related to cache invalidation when a node changes.
+ * It is:
+ *  - when a node is accepted or deleted,
+ *  - when its NodeState changes
  */
 class CacheInvalidateNodeFactEventCallback(
     cacheExpectedReports:   InvalidateCache[CacheExpectedReportAction],
@@ -182,6 +185,25 @@ class CacheInvalidateNodeFactEventCallback(
   import com.normation.rudder.services.reports.CacheExpectedReportAction.*
 
   override def name: String = "node-fact-cec: invalidate caches"
+
+  private def clearCaches(nodeId: NodeId): IOResult[Unit] = {
+    val a = CacheExpectedReportAction.RemoveNodeInCache(nodeId)
+    for {
+      _ <-
+        cacheNodeStatusReports
+          .invalidateWithAction(Seq((nodeId, CacheComplianceQueueAction.ExpectedReportAction(a))))
+          .catchAll(err => {
+            NodeLoggerPure.Delete
+              .error(s"Error when removing node ${nodeId.value} from node configuration cache: ${err.fullMsg}")
+          })
+      _ <- cacheExpectedReports
+             .invalidateWithAction(Seq((nodeId, a)))
+             .catchAll(err => {
+               NodeLoggerPure.Delete
+                 .error(s"Error when removing node ${nodeId.value} from compliance cache: ${err.fullMsg}")
+             })
+    } yield ()
+  }
 
   override def run(change: NodeFactChangeEventCC): IOResult[Unit] = {
     change.event match {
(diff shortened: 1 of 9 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -935,240 +928,251 @@ object ExecutionBatch extends Loggable {
...
       // and perhaps a configId
-
-      runInfo: RunAndConfigInfo, // reports we get on the last know run
+      nodeState: NodeState,
+      runInfo:   RunAndConfigInfo, // reports we get on the last know run
 
       agentExecutionReports: Seq[Reports]
   ): NodeStatusReport = {
+    // early return, no need to continue
+    if (!nodeState.isEnabled) {
```
fanf, line 939: this is the actual correction of the problem. In that case, I create a `NoReportInInterval` or `NoRunNoExpectedReport` based on what information I (still) have.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactChangeEventCallback.scala
```diff
@@ -202,24 +226,21 @@ class CacheInvalidateNodeFactEventCallback(
...
-                NodeLoggerPure.Delete
-                  .error(s"Error when removing node ${node.id.value} from node configuration cache: ${err.fullMsg}")
-              })
-          _ <- cacheExpectedReports
-                 .invalidateWithAction(Seq((node.id, a)))
-                 .catchAll(err =>
-                   NodeLoggerPure.Delete.error(s"Error when removing node ${node.id.value} from compliance cache: ${err.fullMsg}")
-                 )
-        } yield ()
+      case NodeFactChangeEvent.Updated(oldNode, newNode, attrs)        =>
```
fanf, line 227: this is the node repo hook extension to clean up caches. It was almost the same code as for when we delete a node, so I factored-it out.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactChangeEventCallback.scala
```diff
@@ -183,6 +187,26 @@ class CacheInvalidateNodeFactEventCallback(
 
   override def name: String = "node-fact-cec: invalidate caches"
 
+  private def clearCaches(nodeId: NodeId, log: UIO[Unit]): IOResult[Unit] = {
+    val a = CacheExpectedReportAction.RemoveNodeInCache(nodeId)
+    for {
+      _ <- log
```
clarktsiory: to have the `log` outside, at the call site of the `clearCaches` function would be easier to understand for me (`Logger.info(...) *> clearCaches(args)` instead of `clearCaches(args..., Logger.info(...))`)

clarktsiory (commented):
LGTM, just a remark for the readability of logs

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactChangeEventCallback.scala
fanf: ok

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactChangeEventCallback.scala
```diff
@@ -66,6 +66,7 @@ import com.normation.rudder.services.reports.InvalidateCache
 import com.normation.utils.DateFormaterService
 import com.normation.utils.StringUuidGenerator
 import java.time.Instant
+import zio.UIO
```
clarktsiory:
```suggestion

```

clarktsiory (changes requested):
OK, but looks like there is an unused import

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #7053, opened 2026-04-02
Fixes #28682: We should have consistant behaviour when creating group with empty criteria

fanf:
https://issues.rudder.io/issues/28682

We weren't taking appart the case where there is 0 criteria, and since we are *filtering* things, we weren't filtering anything at all. 
Since the "save" without criteria actually create a query with 0 criteria, we were falling in that case.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
@@ -300,9 +300,16 @@ class NodeFactQueryProcessor(
       subgroupLines = subGroupMatcher(sortedLines.subGroup, groupRepo)
       ldapLines     = ldapMatcher(sortedLines.ldap, query)
       nodeLines     = nodeFactMatcher(sortedLines.coreNodeFact)
-      lineResult   <- ZIO.foldLeft(globalLines ++ subgroupLines ++ ldapLines ++ nodeLines)(group.neutral: NodeFactMatcher) {
-                        case (matcher, line) =>
-                          line.map(group.compose(matcher, _))
+      allMatcher    = globalLines ++ subgroupLines ++ ldapLines ++ nodeLines
+      lineResult   <- if (allMatcher.isEmpty) {
+                        // take care of the empty set case, which breaks our inner set law, by adding a never matching
+                        // matcher. With that, other latter transformation like `inversion` will work.
+                        AbsorbingMatcher("no criteria provided", false).succeed
+                      } else {
+                        ZIO.foldLeft(allMatcher)(group.neutral: NodeFactMatcher) {
+                          case (matcher, line) =>
+                            line.map(group.compose(matcher, _))
+                        }
                       }
     } yield {
       // inverse now if needed, because we don't want to return root if not asked *even* when inverse is present
--- a/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/queries/TestNodeFactQueryProcessor.scala
+++ b/webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/queries/TestNodeFactQueryProcessor.scala
@@ -200,7 +200,7 @@ class TestNodeFactQueryProcessor {
     val q2 = TestQuery(
       "q2",
       parser("""
-      {  "select":"node", "composition":"and", "where":[
+      { "select":"node", "composition":"and", "where":[
         { "objectType":"node"   , "attribute":"nodeId"  , "comparator":"eq", "value":"node1" },
         { "objectType":"node"   , "attribute":"nodeId"  , "comparator":"eq", "value":"node5" }
       ] }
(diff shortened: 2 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
```diff
@@ -300,9 +300,15 @@ class NodeFactQueryProcessor(
...
+      allMatcher    = globalLines ++ subgroupLines ++ ldapLines ++ nodeLines
+      lineResult   <- if (allMatcher.isEmpty) {
+                        // add a never matching matcher so that other transformation works
+                        AbsorbingMatcher("no criteria provided", false).succeed
+                      } else {
+                        ZIO.foldLeft(allMatcher)(group.neutral: NodeFactMatcher) {
+                          case (matcher, line) =>
+                            line.map(group.compose(matcher, _))
+                        }
                       }
```
clarktsiory, line 313: This looks like a composition of all lines using the `group`, which has group laws on the `compose`, `neutral` and `zero` elements.
And here the `AbsorbingMatcher` looks exactly like the `zero` element, except that the value is `true` or `false` depending on the case  : 
* `false` when the composition is `and` (because an _AND_ on "no elements" defaults to `false`)
* `true` when the composition is `or` (because an _OR_ on "no elements" defaults to `true`)

So, in the case of an `OR` composition with no criteria, should we make an exception to that, by always defaulting to `false` and returning no elements, like in the proposed change ?

I would personally prefer to keep consistency with the laws that we have defined, and encapsulate the whole `if (empty) group.zero else ZIO.foldLeft(...)(group.neutral)(group.compose(...))` within a newly defined group operator

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/queries/TestNodeFactQueryProcessor.scala
```diff
@@ -200,15 +200,35 @@ class TestNodeFactQueryProcessor {
...
+    val q4 = TestQuery(
+      "q4",
+      parser("""
+      { "select":"node", "composition":"and", "transform":"invert", "where":[
+      ] }
+      """).openOrThrowException("For tests"),
+      s
+    )
+
+    testQueries(q0 :: q1 :: q2 :: q3 :: q4 :: Nil, doInternalQueryTest = false)
```
clarktsiory: it could be reassuring to have queries and test cases for `or` too (see the previous comment)

clarktsiory (commented):
That was a nasty error, and I now understand why this happens : we forgot to check for "zero" and use it when using our _group._
 The behavior is quite consistent with the group laws, but I'm simply not sure if we should enforce laws to return an actual "zero", or a hardcoded "empty query"

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
fanf: I think you are confusing the absorbing element for the monoid operation with the empty set. 
The absorbing element in a set wrt to `or` is the whole set. 
Absorbing elements are not part of group definition, the binary relation just need a neutral element. `group.zero` is not something I know about, I think it's just an abuse for the neutral element in the addition law on integer. 

So yes, we need a special case, and no, I don't think we break any consistency with the law. 
I can still maybe use the `group.zero` we define, because I think we just have a special treatment for it latter on, but it algorithmic and has nothing to do with set operation.

clarktsiory: `group.zero` is the "stop elements", it's used to optimize operations : 

```
true AND true AND false AND true AND ...
                  ^^^^^ 
                  stop evaluations here
```
So, indeed we may want to stop early even with `OR`, I agree that it can be handled at that level...

I was just realizing we missed something when effectively using the API of the group to translate from the initial set, but anyway there is `ZIO` in-between, so I don't think the translation has any further simplification.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
fanf: I know what the group zero is, but it's not the problem here. It *could* work because I think we do a `if(matcher == group.zero) then early out`, but it doesn't have any sens wrt what the matcher is as a matcher in the case of `or`.

fanf: nope, doesn't work

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/queries/NodeFactQueryProcessor.scala
fanf: I added comment to try to make it clearer

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/services/queries/TestNodeFactQueryProcessor.scala
fanf: added the relevant cases

clarktsiory (approved):
OK !

fanf:
OK, squash merging this PR

PR #7054, opened 2026-04-02
Fixes #28592: When we create a Node with a PUT, the properties are not taken into account

fanf:
https://issues.rudder.io/issues/28592

*SO*. We had a bit forgotten to migrate the node creation with API to `NodeFactRepository`. The tests were good them. So use that code in place of the direct LDAP queries, that can't work anymore. 

With all these implicit and cc.toQuery, the upmerge will be funny.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
@@ -44,27 +44,20 @@ import com.normation.errors.*
 import com.normation.eventlog.ModificationId
 import com.normation.inventory.domain.*
 import com.normation.inventory.domain.NodeId
-import com.normation.inventory.ldap.core.InventoryDit
-import com.normation.inventory.ldap.core.UUID_ENTRY
-import com.normation.ldap.sdk.LDAPConnectionProvider
-import com.normation.ldap.sdk.RwLDAPConnection
 import com.normation.rudder.api.ApiVersion
 import com.normation.rudder.apidata.DefaultDetailLevel
 import com.normation.rudder.apidata.JsonQueryObjects.*
 import com.normation.rudder.apidata.JsonResponseObjects.*
 import com.normation.rudder.apidata.NodeDetailLevel
 import com.normation.rudder.apidata.RenderInheritedProperties
-import com.normation.rudder.apidata.RestDataSerializer
 import com.normation.rudder.apidata.ZioJsonExtractor
 import com.normation.rudder.apidata.implicits.*
 import com.normation.rudder.config.ReasonBehavior
 import com.normation.rudder.config.UserPropertyService
-import com.normation.rudder.domain.NodeDit
 import com.normation.rudder.domain.logger.NodeLogger
 import com.normation.rudder.domain.logger.NodeLoggerPure
 import com.normation.rudder.domain.logger.TimingDebugLoggerPure
 import com.normation.rudder.domain.nodes.Node
-import com.normation.rudder.domain.nodes.NodeState
 import com.normation.rudder.domain.policies.GlobalPolicyMode
 import com.normation.rudder.domain.properties.CompareProperties
 import com.normation.rudder.domain.properties.FailedNodePropertyHierarchy
@@ -85,10 +78,8 @@ import com.normation.rudder.facts.nodes.SelectFacts
 import com.normation.rudder.facts.nodes.SelectNodeStatus
 import com.normation.rudder.properties.NodePropertiesService
 import com.normation.rudder.properties.PropertiesRepository
-import com.normation.rudder.reports.ReportingConfiguration
 import com.normation.rudder.reports.execution.AgentRunWithNodeConfig
 import com.normation.rudder.reports.execution.RoReportsExecutionRepository
-import com.normation.rudder.repository.ldap.LDAPEntityMapper
 import com.normation.rudder.rest.ApiModuleProvider
 import com.normation.rudder.rest.ApiPath
 import com.normation.rudder.rest.AuthzToken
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -2195,17 +2195,11 @@ object RudderConfigInit {
...
             new NodeApiService(
-              rwLdap,
               nodeFactRepository,
               propertiesRepository,
               roAgentRunsRepository,
-              ldapEntityMapper,
               stringUuidGenerator,
-              nodeDit,
-              pendingNodesDit,
-              acceptedNodesDit,
```
fanf, line 2208: much better

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RestTestSetUp.scala
```diff
@@ -875,43 +869,15 @@ class RestTestSetUp {
...
-        n <- mockNodes.nodeFactRepo.get(id).notOptional(s"Can not merge node: missing")
-        n2 = CoreNodeFact.updateNode(n, mergeNodeSetup(n.toNode, setup))
-        _ <- mockNodes.nodeFactRepo.save(n2)(testCC)
-      } yield {
-        n.id
-      }).mapError(err => CreationError.OnSaveInventory(err.fullMsg))
+    override def saveRudderNode(id: NodeId, setup: NodeSetup)(implicit cc: ChangeContext): IO[Creation.CreationError, NodeId] = {
+      import com.softwaremill.quicklens.*
+      // we need to have a stable date for tests
+      val refDate = DateTime.parse("2021-01-30T00:20:00Z")
```
clarktsiory, line 875: this joda `DateTime` will likely break the upmerge too

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
```diff
@@ -865,36 +850,19 @@ class NodeApiService(
...
-    (for {
-      con <- ldapConnection
-      _   <- inventoryExists(con, nodeId)
-    } yield ()).mapError(err => CreationError.OnSaveInventory(s"Error during node ID check: ${err.fullMsg}"))
+  def checkUuid(nodeId: NodeId)(implicit qc: QueryContext): IO[CreationError, Unit] = {
+    nodeFactRepository
+      .get(nodeId)
+      .map(_.nonEmpty)
+      .mapError(err => CreationError.OnSaveInventory(s"Error during node ID check: ${err.fullMsg}"))
+      .unit
```
clarktsiory: I'm not sure about that implementation, shouldn't there be an explicit rejection, instead of `.unit` ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
```diff
@@ -948,38 +918,16 @@ class NodeApiService(
...
-      // try t get node
-      entry   <- ldap.get(nodeDit.NODES.NODE.dn(id.value), NodeInfoService.nodeInfoAttributes*)
-      current <- entry match {
-                   case Some(x) => ldapEntityMapper.entryToNode(x).toIO
-                   case None    => default().succeed
-                 }
-      merged   = mergeNodeSetup(current, setup)
-      // we ony want to touch things that were asked by the user
-      nSaved  <- ldap.save(ldapEntityMapper.nodeToEntry(merged))
+      n <- nodeFactRepository.get(id).notOptional(s"Can not merge node: missing")
```
clarktsiory, line 927: are we not expected to save a non-existing node ? (I wonder why the default was there in the first place)

clarktsiory (changes requested):
there is simply an implementation that looks to not check anything, and a question about change in implementation behavior

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
fanf: hum. I think the API is tested for that case, but I will check

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
fanf: at that time of the process, the first "save inventory" was done and so we should have the node in repo. 
The template was here because once upon a time, inventories and nodes were different things. So you could have an inventory saved, without the "rudder node" aspect of the node existing in LDAP - so the template part. 
Now, this is all correctly managed transparently by `NodeFactRepository`, and so it's much, much simpler. We could even merge the two steps into one (save inventory / save rudder node),  but it was a bit too much of a change for a 8.3 patch release.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/scala/com/normation/rudder/rest/RestTestSetUp.scala
fanf: Yep. Nice upmerge to follow.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/NodeApi.scala
fanf: Rewrote it to make it clearer and added a test.

clarktsiory (approved):
LGTM now

fanf:
OK, squash merging this PR

PR #7055, opened 2026-04-03
Fixes #28670: Wrong message for node deletion in activity 

clarktsiory:
https://issues.rudder.io/issues/28670

A logic flaw : we don't display the link for deleted node (since they no longer exist), but the "deleted" word was forgotten


clarktsiory:
Commit modified

P4uline (approved):

PR #7057, opened 2026-04-03
Fixes #28669: A "compliance" hover appears on top of the node lists in group page

clarktsiory:
https://issues.rudder.io/issues/28669

The [fixedHeader](https://datatables.net/extensions/fixedheader/) plugin is actually unused, and is the cause of the problem of "shadow headers" in nodes table. It can simply be removed as it was only declared in the nodes table.


fanf (approved):
It is clearly broken for a long time, so there's no reason to keep it. Less code is better.

fanf:
We will keep it in 9.0 to avoid surprising side effects, but we know in case of security problem, we should be able to remove it in version anterior to the 9.1 upgrade too.

PR #7060, opened 2026-04-08
Fixes #28665: Cannot export node list if there pagination has more elements

clarktsiory:
https://issues.rudder.io/issues/28665

We were hacking around with DataTables, where we should have used the [render](https://datatables.net/reference/option/columns.render) option : 
* when the `type` of render is the `exportCsv`, we need to pass raw JSON `data`, and format the JSON
* otherwise it is for display (the previous hack was around jQuery to modify the DOM, but we can simply pass the html to the render function now)

As for the format of the JSON for the CSV export : it was not consistent, as it could had line endings, and additional "inherited" text, because the displayed value contained that. Now the value is the JSON of the property and nothing else

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
@@ -84,18 +84,6 @@ const csvButtonConfig = (filename, additionalCls) => ({
   text: 'Export',
   exportOptions: {
     orthogonal: 'exportCsv',
-    format: {
-          body: function (html, row, col, node) {
-              // begin with default formatting
-              const isString = value => typeof value === 'string' || value instanceof String;
-              html = $.fn.DataTable.Buttons.stripData( html, null );
-              // N/A and curly braces are properties, get innerText ...
-              if (isString(html) && (html === "N/A" || html.startsWith("{") )) {
-                html = node.innerText
-              }
-              return html;
-          }
-    },
     customizeData: function (data) {
       // export compliance percent
       const complianceColumnIdx = data.header.findIndex(s => s.toLowerCase() === "compliance")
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -84,18 +84,6 @@ const csvButtonConfig = (filename, additionalCls) => ({
...
-              // begin with default formatting
-              const isString = value => typeof value === 'string' || value instanceof String;
-              html = $.fn.DataTable.Buttons.stripData( html, null );
-              // N/A and curly braces are properties, get innerText ...
-              if (isString(html) && (html === "N/A" || html.startsWith("{") )) {
-                html = node.innerText
-              }
-              return html;
-          }
-    },
```
clarktsiory, line 98: this was the hack that "parsed" the displayed HTML, but at this level we have no knowledge of the raw _data_, only the _html_, which was `undefined` when paginated elements were not displayed :upside_down_face:

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -1302,23 +1288,24 @@ function createNodeTable(gridId, nodeIds, refresh, scores) {
...
+        return {
+          "data": function (row, type) {
+            if (type === 'set') {
+              return;
+            } else if (type === 'sort') {
+              const prop = inherited ? row.inheritedProperties[value] : row.properties[value]
+              return JSON.stringify(prop);
+            }
+            const prop = inherited ? row.inheritedProperties[value] : row.properties[value]
+            return prop;
```
clarktsiory, line 1300: this makes the _data_ ready for consumption to the `render` function, which will dispatch : 
* rendering for CSV export : format the JSON
* rendering for display : create HTML, tooltips, etc. from the JSON

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -1183,34 +1171,32 @@ function scoreFunction(value) { return function (nTd, sData, oData, iRow, iCol)
...
-  } else {
-    var text = property.value;
-    if (typeof property === "object") {
-      text = JSON.stringify(property.value, undefined, 2)
+function propertyFunction(name) {
+  return function (data, type, row, meta) {
+    if (data === undefined) {
+      return null  // will use defaultContent
+    } else if (type === 'exportCsv') {
+      return JSON.stringify(data.value)
```
clarktsiory, line 1179: this is a significant change from the previous behavior : JSON objects would be formatted in compact fashion e.g. `{"a":1,"b":[1,2,3]}`

Previous behavior had prettified JSON (as in the display), but also an additional "inherited" text at start, which made the CSV column type hard to guess

clarktsiory: we can discuss this change, we could manually add the `"inherited"` text e.g. 
```
inherited
{
  "a": "1",
  "b": [1,2,3}
}
```
if needed

VinceMacBuche (approved):

PR #7061, opened 2026-04-08
Fixes #28608: API account tooltips are broken when table is sorted or refreshed

clarktsiory:
https://issues.rudder.io/issues/28608

### EDIT : New understanding of tooltips
* the only attribute that seems to be universally working is `data-bs-original-title`, instead of `title`/`data-bs-title`
* `removeBsTooltips` is only a hack for removing "stale" tooltips in some specific UX concerns (_"the tooltip is frozen"_)
  * it leads to memory leaks : when using tooltips in Elm apps and e.g. filtering items, the DOM changes, but event handlers of tooltips are still there
  * it is used in this PR because a "button click" when opening a modal leads to a frozen one
  
* we need a new function to properly [dispose](https://getbootstrap.com/docs/5.0/components/tooltips/#dispose) of tooltips, it's called `disposeBsTooltips` :
  * it needs to be always followed by an initialization of tooltips, because when "sorting elements", the DOM elements would swap
  * `resetTooltips` is a single Port that guarantees that we have "dispose", then "init", and it is used in this PR upon the update of "table filters"

* to maximize performance, we could restrict tooltips actions to a selector of "tooltips within the Elm app" : just add a `<div id="app-container">` and pass to the "dispose" and "init" functions
 
---

Adding a new function to dispose of tooltips completely : `removeBsTooltips` is  to not work in this context because of how Elm swaps the DOM elements, by leaving tooltips at their previous location (which made the bug appear when : sorting the table, filtering, and adding and API account).

I could not get rid of them :
* the previous one always remains with a `data-bs-original-title` :
image

* sometimes it's the `title` that would remain, even on a `-` :  
image

I tried removing every attributes (including the `title`) upon calling `dispose()` on the tooltip, and this solves the issue, but would make all tooltips not work globally when applying some filters on the table :upside_down_face:, so I didn't risk doing that

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts.elm
@@ -100,15 +100,21 @@ update msg model =
                 tenants =
                     editAccount |> Maybe.andThen .selectedTenants |> Maybe.withDefault []
             in
-            ( { model | ui = { ui | modalState = modalState }, editAccount = editAccount }, Cmd.batch [ shareAcl (encodeTokenAcl tokenId acl), focusAccountTenants (encodeAccountTenants tokenId tenants) ] )
+            ( { model | ui = { ui | modalState = modalState }, editAccount = editAccount }, Cmd.batch [ shareAcl (encodeTokenAcl tokenId acl), focusAccountTenants (encodeAccountTenants tokenId tenants), clearTooltips () ] )
 
 
         UpdateFilters filters ->
             let
                 ui =
                     model.ui
+                cmd =
+                    if filters /= ui.filters then
+                        resetTooltips ()
+
+                    else
+                        Cmd.none
             in
-            ( { model | ui = { ui | filters = filters } }, Cmd.none )
+            ( { model | ui = { ui | filters = filters } }, cmd )
 
         GetAccountsResult res ->
             case res of
@@ -123,7 +129,7 @@ update msg model =
 
                     in
                     ( { model | accounts = accounts, ui = { modelUi | loadingAccounts = False, pluginAclInit = True, pluginTenantsInit = True } }
-                    , Cmd.batch [ initTooltips "" ]
+                    , resetTooltips ()
                     )
 
                 Err err ->
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
@@ -906,23 +906,34 @@ function logout(cb){
 }
 
 // BOOTSTRAP 5
-function initBsTooltips(){
-  var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
+const bsTooltipsRemovalListener = () => removeBsTooltips()
+
+function initBsTooltips(selector = null){
+  const sel = selector ?? '[data-bs-toggle="tooltip"]'
+  var tooltipTriggerList = [].slice.call(document.querySelectorAll(sel));
   return tooltipTriggerList.map(function (tooltipTriggerEl) {
     let dataTrigger = $(tooltipTriggerEl).attr('data-bs-trigger');
     let trigger = dataTrigger === undefined ? 'hover' : dataTrigger;
-    tooltipTriggerEl.addEventListener('hide.bs.tooltip', () => {
-      removeBsTooltips();
-    });
-    tooltipTriggerEl.addEventListener('show.bs.tooltip', () => {
-      removeBsTooltips()
-    });
+    tooltipTriggerEl.addEventListener('hide.bs.tooltip', bsTooltipsRemovalListener);
+    tooltipTriggerEl.addEventListener('show.bs.tooltip', bsTooltipsRemovalListener);
     return new bootstrap.Tooltip(tooltipTriggerEl,{container : "body", html : true, trigger : trigger});
   });
 }
 
 function removeBsTooltips(){
   document.querySelectorAll(".tooltip").forEach(e => e.remove());
+}
+
+// remove attributes, handlers left by tooltips: sometimes (with Elm replacing the DOM elements) they are still present
+function disposeBsTooltips(selector) {
+  removeBsTooltips()
+  document
+    .querySelectorAll(selector)
+    .forEach(tooltip => {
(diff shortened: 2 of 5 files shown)


webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
```diff
@@ -906,24 +906,34 @@ function logout(cb){
...
+// remove attributes, handlers left by tooltips: sometimes (with Elm replacing the DOM elements) they are still present
+function disposeBsTooltips() {
+  document
+    .querySelectorAll('[data-bs-toggle="tooltip"],[data-bs-original-title]')
+    .forEach(tooltip => {
+      bootstrap.Tooltip.getInstance(tooltip)?.dispose()
+      tooltip.removeAttribute("data-bs-original-title")
+      tooltip.removeEventListener('hide.bs.tooltip', bsTooltipsRemovalListener)
+      tooltip.removeEventListener('show.bs.tooltip', bsTooltipsRemovalListener)
+    })
```
clarktsiory, line 936: this new function does the "unmounting" of the tooltip from the DOM element (at least in theory, from [the doc](https://getbootstrap.com/docs/5.0/components/tooltips/#dispose))

But as explained in the description, several attributes still remain... it won't clean up correctly

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
```diff
@@ -906,24 +906,34 @@ function logout(cb){
...
     let dataTrigger = $(tooltipTriggerEl).attr('data-bs-trigger');
     let trigger = dataTrigger === undefined ? 'hover' : dataTrigger;
-    tooltipTriggerEl.addEventListener('hide.bs.tooltip', () => {
-      removeBsTooltips();
-    });
-    tooltipTriggerEl.addEventListener('show.bs.tooltip', () => {
-      removeBsTooltips()
-    });
+    tooltipTriggerEl.addEventListener('hide.bs.tooltip', bsTooltipsRemovalListener);
+    tooltipTriggerEl.addEventListener('show.bs.tooltip', bsTooltipsRemovalListener);
```
clarktsiory, line 918: adding a stable reference to the event listener avoids duplicating the event listener function (it appeared several times when I looked in the inspector)

Also, this allows calling `removeEventListener(..., registeredListener)`

RaphaelGauthier:
I've tried different things that improve the result, here is my conclusion:

- Adding a **clearTooltips** port was a good idea, but `disposeBsTooltips()` does not work as expected, `removeBsTooltip()` is more effectice as it actually destroy the generated `.tooltip` elements. So we better use this one in the `subscribe()` function.

- As explained in the Bootstrap Tooltip documentation: _"Feel free to use either title or data-bs-title in your HTML. When title is used, Popper will replace it automatically with data-bs-title when the element is rendered."_. It looks like this creates another delay before the tooltip initialisation, so it is better to use ‘**data-bs-title**’ directly rather than ‘**title**’ attribute in order to avoid this extra delay

- Finally, the timeout we use before each tooltip initialisation is one of the main causes of the problem.
Its duration is purely arbitrary and may not correspond at all to the time required before initialising the tooltips.
The best solution I have found is to wait until the elements containing the tooltips are present in the DOM, by using the [MutationObserver api](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver).

So instead of:
```
app.ports.initTooltips.subscribe(function (msg) {
  setTimeout(function () {
    initBsTooltips();
  }, 600);
});
```
We should do something like:
```
const tooltipSelector = '.dataTable [data-bs-toggle="tooltip"]';
app.ports.initTooltips.subscribe(function () {
  waitForElement(tooltipSelector).then((el) => {
    initBsTooltips();
  });
});
```

The last remaining issue (I hope :crossed_fingers:) is the resolution delay of the MutationObserver's promise, which takes few millisecond to complete. But I can’t see how to improve on that.

RaphaelGauthier:
Last but not least, we should replace the attribute "title" with "data-bs-original-title"... This makes the generated .tooltip element actually disappeared :skull:

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts.elm
```diff
@@ -107,8 +107,14 @@ update msg model =
             let
                 ui =
                     model.ui
+                cmd =
+                    if filters /= ui.filters then
+                        Cmd.batch [ clearTooltips (), initTooltips () ]
```
RaphaelGauthier: Cmd.batch is not garanteed to be sequential (clear and init sequentially)

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Accounts.elm
clarktsiory: I found no easy Elm solution for this (other than a rather [complex library](https://package.elm-lang.org/packages/lobanov/elm-taskport/latest/)) : the solution I opted for was to have a single `resetTooltips` port.

Tooltips are always reset anyway... (and sometimes we want the actual "removeBsTooltips" for frozen ones)

clarktsiory:
PR updated with a new commit : forgot the `initBsTooltips` optimization

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
```diff
@@ -906,24 +906,35 @@ function logout(cb){
 }
 
 // BOOTSTRAP 5
-function initBsTooltips(){
-  var tooltipTriggerList = [].slice.call(document.querySelectorAll('[data-bs-toggle="tooltip"]'));
+const bsTooltipsRemovalListener = () => removeBsTooltips()
+
+function initBsTooltips(selector = null){
+  const sel = select ?? '[data-bs-toggle="tooltip"]'
```
RaphaelGauthier: `select ??` must be `selector` here!

clarktsiory:
PR updated with a new commit : fix the `initBsTooltips` selector optimization

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder.js
clarktsiory: that's a nice catch :see_no_evil:
it was a last minute addition so I didn't pay attention to the typo !

RaphaelGauthier (approved):

clarktsiory:
OK, squash merging this PR

PR #7063, opened 2026-04-09
Fixes #28654: Notification error on user without admin_read rights on node page

clarktsiory:
https://issues.rudder.io/issues/28654

The previous implementation was getting 3 settings values from `/api/settings` (i.e. by getting all settings values), but that endpoint would return a `500 Server error` error without admin rights.

The solution is to fetch each individual settings `/api/settings/enable_change_request`, etc., which will return `403 Forbidden`.
We need to fetch 3 of them into a single `type alias ChangeMessageSettings`, so I followed the idea of [this blog post](https://korban.net/posts/elm/2019-02-15-combining-http-requests-with-task-in-elm/) to do that : 
* use the Elm `Task` module and compose the Http results using the [`map3`](https://package.elm-lang.org/packages/elm/core/latest/Task#map3) method (without that, there would have been 3 messages and so much additional noise)
* handle 403 error once again with the `onError` method

### Benefits of using Task
* We are short-circuiting on the first error (because of how `onError` is used), which would have been a pain to handle in the update function otherwise
* All the logic are local and composable, otherwise there would have been many indirection levels between the "updates", "api calls" and "messages"

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Agentpolicymode/JsonDecoder.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Agentpolicymode/JsonDecoder.elm
@@ -35,16 +35,19 @@ decodeSaveNodeDetails =
   at [ "data" ] decodeNodePolicyMode
 
 
-decodeGetChangeMessageSettings : Decoder ChangeMessageSettings
-decodeGetChangeMessageSettings =
-  at ["data", "settings" ] decodeChangeMessageSettings
+-- Same as in Rules
+decodeGetEnableChangeMsg : Decoder Bool
+decodeGetEnableChangeMsg =
+  at ["data", "settings", "enable_change_message" ] bool
 
-decodeChangeMessageSettings : Decoder ChangeMessageSettings
-decodeChangeMessageSettings =
-  succeed ChangeMessageSettings
-    |> required "enable_change_message"    Json.Decode.bool
-    |> required "mandatory_change_message" Json.Decode.bool
-    |> required "change_message_prompt"    Json.Decode.string
+decodeGetMandatoryMsg : Decoder Bool
+decodeGetMandatoryMsg =
+  at ["data", "settings", "mandatory_change_message" ] bool
+
+decodeGetMsgPrompt : Decoder String
+decodeGetMsgPrompt =
+  at ["data", "settings", "change_message_prompt" ] string
+
 
 
 decodeGetGlobalSettings : Decoder Settings
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Agentpolicymode/ApiCalls.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Agentpolicymode/ApiCalls.elm
@@ -3,6 +3,7 @@ module Agentpolicymode.ApiCalls exposing (..)
 import Http exposing (..)
 import Http.Detailed as Detailed
 import Json.Decode exposing (Decoder)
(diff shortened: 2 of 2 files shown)


clarktsiory:
Commit modified

fanf:
Here, you are doing three requests to get the values, which is extremely inefficient. 

You could do a specialized private endpoint, exposing exactly what the UI need in a purposed-made format, in only one round, with the rights matching the one of the page. In addition, that endpoint would be able seamlessly evolve with the client-side needs, even in patch release. 

I understand it needs a backend addition, but twisting public APIs to get what we want in wastful way is not a sustainable path (more coupling with public APIs making it harder to make at the same time good public API and things that are useful for the pages, more latency everywhere, etc).

I'm not sure we already have such endpoints, perhaps we should have them in a dedicated API sub-path, like `pages` or whatever.

clarktsiory:
The app that is being changed here loads its data asynchronously : the settings are only used when displaying a "modal",  when saving the policy mode setting (thus when the user clicks on the "Save" button).
So making those API requests would not directly be visible to users.

I even think that, in the principle of minimizing load on the server, we should only fetch those settings upon the click on the "Save" button, but that would require having a loader, @RaphaelGauthier do you think it should be part of this change ?


@fanf for the case of efficiency and latency, this appears more efficient than the previous implementation which was fetching ALL settings (which appears as a huge load on the server, I have identified [a potential cause](https://issues.rudder.io/issues/28701)). I simply found the settings API to be annoying to build upon : the architecture of it makes it fragile when defining autorizations, especially when fetching "multiple settings" at once, since every individual setting is bound to 1 or more API authorizations.

I agree that having a dedicated endpoint for that could be a solution, but that is one more endpoint to maintain within an already fragile part, which is meant to be redesigned and has heavier maintenance cost than an Elm app evolving with a stable public API.

I'm willing to do the change anyway, since it proves to be useful for the Rules app, that also requires those same settings. @RaphaelGauthier would you agree with that change ?

webapp/sources/rudder/rudder-web/src/main/elm/sources/Agentpolicymode/JsonDecoder.elm
```diff
@@ -35,16 +35,19 @@ decodeSaveNodeDetails =
...
+  at ["data", "settings", "enable_change_message" ] bool
+
+decodeGetMandatoryMsg : Decoder Bool
+decodeGetMandatoryMsg =
+  at ["data", "settings", "mandatory_change_message" ] bool
+
+decodeGetMsgPrompt : Decoder String
+decodeGetMsgPrompt =
+  at ["data", "settings", "change_message_prompt" ] string
+
```
RaphaelGauthier, line 51: @clarktsiory To avoid duplicating code, shouldn’t we move the ChangeRequest module out of the Rules folder and import it directly so we can use its functions?

clarktsiory: Yes, we could create a shared module for Decoders/Api calls related to : 
```elm
type alias ChangeMessageSettings =
  { enableChangeMessage    : Bool
  , mandatoryChangeMessage : Bool
  , changeMessagePrompt    : String
  }
```

But that does not resolve whether it should a new API call/combined calls, nor if we should "delay" the API call to when the user clicks on a "Save" button...

fanf:
Let's merge like that, and we will resolve architecture problem in a minor release.

fanf (approved):
LGTM

clarktsiory:
OK, merging this PR

PR #7065, opened 2026-04-09
Fixes #28611: When no rules are applied, the message in the compliance tab of a node is missleading

clarktsiory:
https://issues.rudder.io/issues/28611

image

I needed to add [`rounded-0`](https://getbootstrap.com/docs/5.3/utilities/borders/#sizes) because somehow there is a default radius to the blue background : 
image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeCompliance/View.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeCompliance/View.elm
@@ -70,7 +70,7 @@ view model =
               , tbody []
                 ( if List.length childs <= 0 then
                   [ tr[]
-                    [ td[class "empty", colspan 2][i [class"fa fa-exclamation-triangle"][], text "There is no compliance for this directive."] ]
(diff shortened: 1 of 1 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

skaerg (commented):
functionally the fix looks good to me, though i am wondering if it is consistent with existing displays of the same kind : e.g. in the system-updates plugin, the table of nodes that are targeted by an update campaign is displayed as a warning if there are no nodes :

Image

if a given group has no recorded compliance info, the compliance table also displays a warning (instead of an "info") : 

Image

in essence, i am wondering whether we have a clear distinction between tables that should display a warning vs. tables that should display an info message when they are empty
 
do you have an opinion @RaphaelGauthier @clarktsiory ? is there already an established distinction for the cases where a warning vs. an info message should be displayed ?

clarktsiory:
There are several warnings in the UI when elements are unused and not only in tables : for example when a directive is not used in any rule (which also means it won't have any compliance).

That is a legitimate question @skaerg, as for "nodes", the reported issue can apply to "ignored" nodes or freshly added node : nodes are independent from configurations so they could deserve special treatment, contrary to "stale" configuration elements.

skaerg (approved):

PR #7066, opened 2026-04-09
Fixes #28695: Remove column type in change logs table

clarktsiory:
https://issues.rudder.io/issues/28695

Removing the field from API, and the column from DataTables initialization.

The translation added in https://github.com/Normation/rudder/pull/7002 is kept, as it's still used for the rollback details, but we should likely remove it too in another PR


clarktsiory:
Commit modified

clarktsiory:
Commit modified

fanf (approved):
LGTM

clarktsiory:
OK, merging this PR

PR #7068, opened 2026-04-13
Fixes #28708: Clean-up dead code in compliance serialization

fanf:
https://issues.rudder.io/issues/28708

OK, it's removing `zio-json` code and not `lift-json` one... But dead code is dead code.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/db/Doobie.scala
```diff
@@ -289,20 +289,6 @@ object Doobie {
...
-
-  implicit val AggregatedStatusReportComposite: Write[AggregatedStatusReport] = {
-    import NodeStatusReportSerialization.*
-    Write[String].contramap(_.toCompactJson)
-  }
-
-  implicit val SetRuleNodeStatusReportComposite: Write[Set[RuleNodeStatusReport]] = {
-    import NodeStatusReportSerialization.*
-    Write[String].contramap(ruleNodeStatusReportToJson)
-  }
```
fanf, line 305: I think it used to be used in the detailed compliance table, which are not used anymore - we only have node last compliance, and it doesn't use that.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -512,15 +512,6 @@ object JsonCompliance {
   }
 
   implicit class JsonbyDirectiveCompliance(val directive: ByDirectiveCompliance) extends AnyVal {
-    def toJsonV6: JObject = (
```
fanf, line 515: API version *6*. Wow.

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_compliance.yml
```diff
@@ -456,6 +456,223 @@ response:
       }
     }
 ---
+description: Get a node group's global compliance for group G1 (simple example, limited to 2 levels, int precision)
```
fanf, line 459: the precision test was missing

fanf:
Replaced by https://github.com/Normation/rudder/pull/7071

PR #7074, opened 2026-04-16
Fixes #28723: Property value textarea is cropped while editing multiline string or JSON

RaphaelGauthier:
https://issues.rudder.io/issues/28723

Now when a property is being edited, the maximum line height has been significantly increased, and the user can scroll and resize the textarea. I also move the css of the table in the right file, and cleaned it up a bit.

https://github.com/user-attachments/assets/2899d3d5-c139-4eab-a6b5-bf9b5e8dbf3d


ElaadF (approved):

RaphaelGauthier:
OK, merging this PR

PR #7076, opened 2026-04-16
Fixes #28729: ADR core domain rules in service

P4uline:
https://issues.rudder.io/issues/28729


P4uline:
PR updated with a new commit

mbaechler (commented):
I think I get the intention of this ADR but I would like to point your attention to the fact the "business rule" concept is not explicitly defined.

My understanding is that "business rule" is basically the same thing as "some logic", which mean that if you take a decision with some code, it's always some kind of business rule.

Given that definition, business rules are needed in every layer, and it doesn't help with application design.

DDD solves that by saying "each component should handle a specific domain", for example: your REST API layer's domain is "REST API" and is allowed to handle logic as long as it's about REST API and not something else, Service layer is allowed to handle logic related to your core business (for example inventory) but not logic about storage or serialization, etc.

So I suggest to define what is the exact responsibility of each component/layer and what kind of logic it can handle.

Finally, I would say that an Elm app without at least some logic duplicated from your backend service layer would be quite poor UX, I would personally allow some code domain logic here.

P4uline:
> I think I get the intention of this ADR but I would like to point your attention to the fact the "business rule" concept is not explicitly defined.
> 
> My understanding is that "business rule" is basically the same thing as "some logic", which mean that if you take a decision with some code, it's always some kind of business rule.
> 
> Given that definition, business rules are needed in every layer, and it doesn't help with application design.
> 

To me business rules and some logic are different things. Some logic is generic, and when I talked about business rules I thought about the core domain of Rudder : audit and/or patch system configurations and/or system security management on environments, workstations, servers.

> DDD solves that by saying "each component should handle a specific domain", for example: your REST API layer's domain is "REST API" and is allowed to handle logic as long as it's about REST API and not something else, Service layer is allowed to handle logic related to your core business (for example inventory) but not logic about storage or serialization, etc.
> 
Agree

> So I suggest to define what is the exact responsibility of each component/layer and what kind of logic it can handle.
> 
I updated the PR with more context. I changed `business`  to `core domain` as it's more accurate.

> Finally, I would say that an Elm app without at least some logic duplicated from your backend service layer would be quite poor UX, I would personally allow some code domain logic here.

I'm not sure to be agree because I don't see any use cases, but I would be glad to get examples or details if you have something specific in mind.

Also I've been thinking how we can stick more on DDD principles so there will be probably others ADR incoming.

mbaechler:
> > I think I get the intention of this ADR but I would like to point your attention to the fact the "business rule" concept is not explicitly defined.
> > My understanding is that "business rule" is basically the same thing as "some logic", which mean that if you take a decision with some code, it's always some kind of business rule.
> > Given that definition, business rules are needed in every layer, and it doesn't help with application design.
> 
> To me business rules and some logic are different things. Some logic is generic, and when I talked about business rules I thought about the core domain of Rudder : audit and/or patch system configurations and/or system security management on environments, workstations, servers.

In the context of this ADR, I would say it's a perfectly valid definition

> [...]

> > Finally, I would say that an Elm app without at least some logic duplicated from your backend service layer would be quite poor UX, I would personally allow some code domain logic here.
> 
> I'm not sure to be agree because I don't see any use cases, but I would be glad to get examples or details if you have something specific in mind.
 
Let's say, to create a a REST resource, you have to respect a specific set of constraints.
You don't really want the frontend to let you specify an invalid input and let the backend tell you what's wrong.
You usually want the frontend to guide you in a way that makes the risk of sending an invalid input lower.
Now let's think about a very complex set of constraint: the frontend will have to encode some part of the business logic to provide the right UX.
Does it make sense?

> [...]

mbaechler (commented):
I have two more questions:

1. Usually, you want to implement some of the business rules in your domain model (the case classes that you work with in services) to enforce some invariants. Is it a design we want to use? If yes, could you specify where they belong?

2. Some business rules are best implemented in repositories (i.e. unique constraint). Should we take that into account in this ADR?

PR #7078, opened 2026-04-16
Fixes #28733: Port RestCompletion to zio-json

fanf:
https://issues.rudder.io/issues/28733

Minimalistic migration. I didn't even create an Endpoint.


clarktsiory (approved):
LGTM, indeed it can be migrated without changing from the "old internal api"

PR #7079, opened 2026-04-17
Fixes #28735: We should define report_id based classes even in no report case

amousset:
https://issues.rudder.io/issues/28735


Fdall (approved):

amousset:
OK, merging this PR

PR #7081, opened 2026-04-17
Fixes #28734: Migrate system api to zio-json

fanf:
https://issues.rudder.io/issues/28734

This migration is ad-hoc with some normalization: 
- I added tests, but for a lot of them it's not possible since they are returning binary results or are machine dependant (like healthchecks). Also, all the archive are similar, I only added one test for all. 
- I use schema everywhere. I tried to keep doc consistant, but I may have missed some - It might lead to some change in action naming, but that's really not important. 
- A lot of methods where using a kind of `Either[String, (String, Json)]` pattern then used to builed a json object or error. I kept that, adding the minimum of DTO : I kep String and List for very simple cases.


fanf:
PR updated with a new commit

clarktsiory (approved):
Nice !

clarktsiory (commented):
There are tests which are no longer passing in `com.normation.rudder.rest.SystemApiTest`

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #7082, opened 2026-04-17
Fixes #28738: Migrate several rudder-web js param to zio-json

fanf:
https://issues.rudder.io/issues/28738

Always a bit the same logic: 
- create a trivial `case class` with the shape of the parameter that `derives JsonDecoder`
- use that in place of lift json decoding, simplify 

I tried to change as little things as possible in these migration (so, not even old school formater).

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -64,6 +64,7 @@ import com.normation.rudder.rest.ApiModuleProvider
 import com.normation.rudder.rest.AuthorizationMappingListEndpoint
 import com.normation.rudder.rest.EndpointSchema
 import com.normation.rudder.rest.InfoApi as InfoApiDef
+import com.normation.rudder.rest.RudderJsonResponse.LiftJsonResponse
 import com.normation.rudder.rest.data.JsonGlobalPluginLimits
 import com.normation.rudder.rest.data.JsonPluginDetails
 import com.normation.rudder.rest.data.JsonPluginsDetails
@@ -487,13 +488,11 @@ class Boot extends Loggable {
           && req.json.isEmpty) =>
         () => {
           Full(
-            JsonResponse(
-              net.liftweb.json.parse(
-                """{"result": "error"
-                  |, "errorDetails": "The request has a JSON content type but the body is not valid JSON"}""".stripMargin
-              ),
+            LiftJsonResponse[Map[String, String]](
+              Map("result" -> "error", "errorDetails" -> "The request has a JSON content type but the body is not valid JSON"),
+              false,
               412
-            )
+            ).toResponse
           )
         }
     }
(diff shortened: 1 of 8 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -487,13 +488,11 @@ class Boot extends Loggable {
...
-              net.liftweb.json.parse(
-                """{"result": "error"
-                  |, "errorDetails": "The request has a JSON content type but the body is not valid JSON"}""".stripMargin
-              ),
+            LiftJsonResponse[Map[String, String]](
+              Map("result" -> "error", "errorDetails" -> "The request has a JSON content type but the body is not valid JSON"),
+              false,
               412
-            )
+            ).toResponse
```
clarktsiory, line 495: I wonder why it's not a "400 Bad Body" error, since it's about the body being empty

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/AgentScheduleEditForm.scala
```diff
@@ -97,40 +97,36 @@ class AgentScheduleEditForm(
...
-    import net.liftweb.json.*
-    val json = parse(s)
+    final case class JsonSchedule(
+        overrides:   Option[Boolean],
+        interval:    Int,
+        startHour:   Int,
+        startMinute: Int,
+        splayHour:   Int,
+        splayMinute: Int
+    ) derives JsonDecoder
```
clarktsiory: shouldn't the class be "outside" of the method ? I'm not sure of what impact on derivation this has...

clarktsiory (commented):
There is one where the case class does not seem in the right context

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
fanf: I didn't want to change it, and "precondition failed" is somehow ok. We can deal with that in an other PR dedicated to that specific change.

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/AgentScheduleEditForm.scala
fanf: it could, not sure either. In any case it is safe to put it outside with a `private` visibility

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
clarktsiory: okay !

fanf:
PR updated with a new commit

clarktsiory (approved):

fanf:
OK, squash merging this PR

PR #7084, opened 2026-04-17
Fixes #28739: Migrate RuleGrid and JsDataTable

fanf:
https://issues.rudder.io/issues/28739

Needed for https://github.com/Normation/rudder-plugins/pull/948

This one was a bit more complicated because I needed to remove all the homegrown kind of JsonEncore framework we add to transform it into real `JsonEncoder`. 

The main idea is that `JsTableData` encoder is just here to require `JsTableLine` encoders. 

Then, for each kind of `JsTableLine`, the same strategy is used: 
- create a private `JsonXXX` class that mimics the old json object that was provided by `override def json(freshName: () => String): JsObj` 
- (we can notice that `freshName` isn't used anywhere anymore, so I removed it, else it would have been one more implicit parameter to deal with)
- then, create a transformer from the `XXX` to the `JsonXXX` that basically do what was on the `json()` method, 
- delete everything from the `XXX` case class. 

One noticable thing: `RuleLine` json has some JS callback in it. Since it's not real valid json, I needed to add a `JsRaw` class with a `JsonEncoder` that just output the callback string.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
@@ -58,12 +58,13 @@ import com.normation.rudder.web.services.JsTableLine
 import com.normation.utils.Control.traverse
 import com.normation.utils.SimpleStatus
 import com.normation.zio.*
+import io.scalaland.chimney.*
+import io.scalaland.chimney.syntax.*
 import net.liftweb.common.*
 import net.liftweb.http.*
 import net.liftweb.http.js.*
 import net.liftweb.http.js.JE.*
 import net.liftweb.http.js.JsCmds.*
-import net.liftweb.json.*
 import net.liftweb.util.Helpers.*
 import org.apache.commons.text.StringEscapeUtils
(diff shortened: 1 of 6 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/RuleGrid.scala
```diff
@@ -789,42 +793,53 @@ final case class RuleLine(
...
+  // the callbacks are pure JS, which should not be authorized in pur JSON. It's historical.
+  final private case class JsRaw(value: String)
+  private object JsRaw {
 
-    val callbackField = callback.map(cb => ("callback" -> cb))
+    given JsonEncoder[JsRaw] = new JsonEncoder[JsRaw] {
+      override def unsafeEncode(a: JsRaw, indent: Option[Int], out: Write): Unit = {
+        out.write(a.value)
+      }
+    }
```
clarktsiory, line 815: I don't see the difference from an `opaque type`, which would avoid using the internal `Write` type from zio-json

clarktsiory (approved):
LGTM, comments can be addressed later

fanf:
OK, merging this PR

PR #7085, opened 2026-04-18
Fixes #28741: Port SharedFileApi to zio-json

fanf:
https://issues.rudder.io/issues/28741

This one was a bit more complicated because it is not idiomatic at all: 
- it's a direc liftjson API, 
- the API processing, the parameter parsing, the logic, everything was mixed-up, 
- it uses direct `json` AST exploration. 

# Refactoring a bit the code structure

I splitted the file in three parts: 
- the first one (`SharedFilesAPI`) is the same API declaration, it only deals with the wiring into lift API 
- the second (`SharedFilePostQueries`) is the query processing step that analyses parameters and call the actual action on file, 
- the last (`SharedFileApiProcessing`) actually do the requested processing (like delete file, move it, etc). 

The 2 and 3 are objects, they don't need external services. 

# Json processing

Since that file uses direct AST exploration for parameters in the flow in a dynamic way and I didn't want to change too much the logic to add parameters case classes, I kept the parsing of an AST (in zio-json) and using `jsonCursor` to mimic previous AST exploration. 
The resulting code is almost identical.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
@@ -42,6 +42,10 @@ import cats.syntax.apply.*
 import com.normation.box.*
 import com.normation.errors.*
 import com.normation.rudder.rest.OldInternalApiAuthz
+import com.normation.rudder.rest.RudderJsonRequest.*
+import com.normation.rudder.rest.RudderJsonResponse
+import com.normation.rudder.rest.internal.SharedFileApiProcessing.*
+import com.normation.rudder.rest.internal.SharedFilePostQueries.*
 import com.normation.rudder.users.UserService
 import com.normation.utils.DateFormaterService.toJodaDateTime
 import com.normation.utils.FileUtils.*
@@ -55,21 +59,17 @@ import net.liftweb.common.EmptyBox
 import net.liftweb.common.Failure
 import net.liftweb.common.Full
 import net.liftweb.common.Loggable
-import net.liftweb.http.JsonResponse
 import net.liftweb.http.LiftResponse
 import net.liftweb.http.Req
 import net.liftweb.http.StreamingResponse
 import net.liftweb.http.rest.RestHelper
-import net.liftweb.json.JsonAST.JArray
-import net.liftweb.json.JsonAST.JField
-import net.liftweb.json.JsonAST.JNull
-import net.liftweb.json.JsonAST.JObject
-import net.liftweb.json.JsonAST.JString
-import net.liftweb.json.JsonAST.JValue
 import org.apache.commons.fileupload2.core.FileUploadSizeException
 import scala.jdk.CollectionConverters.*
 import scala.util.Try
 import zio.ZIO
+import zio.json.*
+import zio.json.ast.*
+import zio.json.ast.Json.*
 import zio.syntax.*
 
 object SharedFilesAPI {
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
```diff
@@ -79,186 +80,59 @@ object SharedFilesAPI {
   val MAX_FILE_SIZE: Int = 8388608
 }
 
+@jsonExplicitNull
+final case class JStatusResponse(success: Boolean, error: Option[String]) derives JsonEncoder
+
+final private case class JResponseResult[A](result: A) derives JsonEncoder
```
clarktsiory, line 85: I didn't know we could `derive JsonEncoder` without having a `[A: JsonEncoder]` 🤯

fanf: the constraint is not used at that level, it's forwarded at the call site

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
```diff
@@ -79,186 +80,59 @@ object SharedFilesAPI {
...
+    size:   Long,
+    `type`: String,
+    date:   String,
+    rights: String
+) derives JsonEncoder
+
+private object JFileInfo {
+
+  given Transformer[File, JFileInfo] = (file: File) => {
+    val date = Files.getLastModifiedTime(file.path, File.LinkOptions.noFollow*).toInstant.toJodaDateTime
```
clarktsiory: I'm not sure about the "impure" transformer, which gets file attributes during transformation...
It's a small portion of code and only used once within a `IOResult.attempt`, so it's nitpicking, it can always be improved as a good practical example in another PR.

The pattern that I remember using in that case was to passing the data, obtained from impure calls, something like 

```scala
  given (using lastModified: Instant, perms: String) Transformer[File, JFileInfo] = (file: File) => {
    val date = lastModified.toInstant.toJodaDateTime
     ...
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
```diff
@@ -79,186 +80,59 @@ object SharedFilesAPI {
...
-        } else {
-          Unexpected(s"File '${file.name}' is not a regular file").fail
-        }
-      } else {
-        Unexpected(s"File '${file.name}' does not exist").fail
-      }
-    }
-  }
-  def directoryContent(directory: File):               IOResult[LiftResponse] = {
+  given JsonEncoder[File] = JsonEncoder[JFileInfo].contramap(_.transformInto[JFileInfo])
```
clarktsiory: this looks like it's unused (hopefully, since the transformer is not a pure one)

```suggestion
```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
```diff
@@ -322,121 +196,72 @@ class SharedFilesAPI(
                 }
 
               case _ => {
-                (req.json match {
+                val action = JsonCursor.field("action").isString
+
+                (req.body.flatMap(x => new String(x, StandardCharsets.UTF_8).fromJson[Json].toBox) match {
```
clarktsiory: there exists an util to parse the body as json properly :
```suggestion
                import com.normation.rudder.rest.RudderJsonRequest.*
                (req.fromJson[Json].toBox) match {
```

clarktsiory (commented):
Looks mostly good, a given encoder which adds some impurity could be removed.
Moreover, there is util function that simplifies the parsing of the body as Json

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
fanf: You don't change the impurity at all here, each call to `File` method is impure. If you want to go that way, you basically need to build each parameter of the `JFileInfo` before transforming it to `JFileInfo`, which is mostly useless.

fanf: Actually, `file.isDirectory` should also be guarded, `file.permissionsAsString` too

fanf:
PR updated with a new commit

clarktsiory (approved):

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/SharedFilesAPI.scala
clarktsiory: :+1: the latest solution that does not make use chimney is likely the best one :grin:

fanf:
OK, squash merging this PR

PR #7088, opened 2026-04-21
Fixes #28748: Migrate CampaignApiTest and last bit in rudder-rest to zio-jons

fanf:
https://issues.rudder.io/issues/28748

Need https://github.com/Normation/rudder/pull/7084 and https://github.com/Normation/rudder/pull/7085 to be merged before. 

Last bit of migration in `rudder-rest` with some cleaning. 

The only novel part is that I needed to switch `encoder` implicit parameter of `LiftJsonResponse` to public so that it can be used in the response in tests.

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/ApiAuthorization.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/ApiAuthorization.scala
@@ -54,7 +54,6 @@ import com.normation.rudder.users.RudderAccount
 import com.normation.rudder.users.UserService
 import net.liftweb.common.*
 import net.liftweb.http.LiftResponse
-import net.liftweb.json.JsonDSL.*
 
 /*
  * This trait allows to check for authorisation on a given boundedendpoint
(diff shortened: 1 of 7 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/ApiAuthorization.scala
```diff
@@ -325,7 +324,7 @@ object OldInternalApiAuthz {
           )
         }
         ApiLogger.warn(failure.messageChain)
-        RestUtils.toJsonError(None, failure.messageChain, ForbiddenError)
+        RudderJsonResponse.generic.forbiddenError(false, failure.messageChain)
```
clarktsiory: There is no known user, so it looks like it's a `401 Unauthorized` 
```suggestion
        RudderJsonResponse.generic.unauthorizedError(false, failure.messageChain)
```

clarktsiory (commented):
There is only one suggestion to change a HTTP semantic that is not consistent

Thanks to all your contributions, we finally see the end of road ! GG !

fanf:
PR updated with a new commit

fanf:
OK, merging this PR

PR #7089, opened 2026-04-22
Fixes #28764: Policy generation could fail with runtime reflection error

clarktsiory:
https://issues.rudder.io/issues/28764

The usage of lift-json makes it always produce an exception that breaks policy generation.
That blocking piece of code should be removed as we don't have diffing alternative, and it was an old (< Rudder 5) issue that has not surfaced until we test it now


fanf (approved):
LGTM. 
That's a good way to port to zio-json :)

PR #7090, opened 2026-04-22
Fixes #28759: Migrate Properties/Property hierarchy to zio-json

fanf:
https://issues.rudder.io/issues/28759

Last bit of `net.liftweb.json` in rudder. Needed by https://github.com/Normation/rudder-plugins/pull/949 and https://github.com/Normation/rudder-plugins-private/pull/1363

The first commit is a massive refactoring of `JsonResponseObject` to change the `implicit val JsonEncoder` defined in dedicated objects into `derives JsonEncoder` directly on `case class`es and `given`. 

There is only one interesting fact in that bit (appart the massive simplification): I needed to keep one `implicit val` to avoid imposing an `import ...given` in plugins. 

Remaingin part is just cleaning and transforming lift AST to `zio.json.ast` and adding spaces in tests. 

I normalized the HTML view of property hierarchy to always have the same naming shape (kind of level, quote around oject name, use property and not parameter, etc). No migration needed, it's purely human oriented info.


fanf:
PR updated with a new commit

fanf:
PR rebased

fanf:
PR rebased

clarktsiory (approved):
LGTM !

fanf:
OK, merging this PR

PR #7092, opened 2026-04-23
Fixes #28779: Add some space between buttons

P4uline:
https://issues.rudder.io/issues/28779

image (3)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Plugins/View.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Plugins/View.elm
@@ -233,7 +233,7 @@ radioButton groupName selected toMsg index ( labelText, value ) =
 
 displayActionButtons : PluginsViewModel -> List (Html Msg)
 displayActionButtons { selected, installAction, uninstallAction, enableAction, disableAction } =
-    [ button [ class "btn btn-primary me-2", onClick (CallApi updateIndex) ] [ i [ class "fa fa-refresh me-1" ] [], text "Refresh plugins" ]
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/Plugins/View.elm
```diff
@@ -134,7 +134,7 @@ actionInstallUpgradeButton x =
             [ disabled (Maybe.Extra.isNothing install && Maybe.Extra.isNothing upgrade) ]
     in
     button
-        ([ class "btn btn-default"
+        ([ class "btn btn-default me-2"
```
clarktsiory: This makes the structure harder to understand, because we put `me-2` locally in this function, knowing that the caller has a preceding element also with `me-2` AND a following element that is the last element !
_Illustration:_
```elm
div "plugins-actions-buttons"
[ button "me-2"              -- first button
, actionInstallUpgradeButton  -- applies me-2
, button ""                    -- last button
]
```


After some discussion, the best solution would be to add the class `gap-2` in addition to `plugins-actions-buttons`, the class of the parent container where `actionInstallUpgradeButton` is called !

RaphaelGauthier: Instead of adding a `me-2` class for each button, we should remove these class for all buttons and add  .gap-2 class on their container `div.plugins-actions-buttons` l.356 :
`div [ class "plugins-actions-buttons" ]` => `div [ class "plugins-actions-buttons gap-2" ]`

P4uline:
PR updated with a new commit

clarktsiory (approved):

PR #5362, opened 2024-01-29
Fixes #24099: Add a select type for technique parameter

VinceMacBuche:
https://issues.rudder.io/issues/24099

This feature adds constraints to technique parameter to generate different metadata.xml with support for different inputs. Everything is already ok rudderc side. we juste need to add it into rudder api and also in technique editor  UI

PR #7095, opened 2026-04-24
Fixes #28763: Use a select input for parameters that require one rather than a text input

RaphaelGauthier:
https://issues.rudder.io/issues/28763

For parameters where the values are limited to a list of possible options, the `<input type="text">` tag has been replaced with a `<select>` tag, in order to prevent errors and make selection easier.

Bonus:
- The "string" label next tho the parameter name has been removed because he has no purposes.
- I removed some unused css

### Before:
select-param-before

### After:
https://github.com/user-attachments/assets/ee9728f1-430b-42ae-a514-40efbd152b8d

----------
### EDIT:

:warning: **Actually, we can't use a `<select>` here because it prevents the user from using variables (technique parameters, iterators, etc.)**. So instead, we now use a `<datalist>` combined with a  `<input type="text">`!
This allows the user to select one of the suggested values directly, but also to use variables freely.

This is what it looks like (datalist display may vary from one browser to another):
Capture d’écran du 2026-04-29 10-50-55

So now, there is no more error when user try to use a variable (for now we don't check if the variable exists, we only check the syntax):
param-check-2

Alos, the error message has been updated:
param-check-1

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewMethod.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewMethod.elm
@@ -84,64 +84,62 @@ showParam model call state methodParam params =
           [ stopPropagationOn "mousedown" (Json.Decode.succeed (DisableDragDrop, True))
           , onFocus DisableDragDrop
           , readonly (not model.hasWriteRights)
+          , class "form-control"
           , id ("param-" ++ methodParam.name.value)
+          , value displayedValue
           , onInput (MethodCallParameterModified call methodParam.name)
+          -- to deactivate plugin "Grammarly" or "Language Tool" from
+          -- adding HTML that make disapear textarea (see  https://issues.rudder.io/issues/21172)
+          , attribute "data-gramm" "false"
+          , attribute "data-gramm_editor" "false"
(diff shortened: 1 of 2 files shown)


RaphaelGauthier:
I think there is a better solution, by using a type for the paramaters value.
It already exsists, it has been introduced a long time ago in [that PR](https://github.com/Normation/rudder/pull/5362) which add the corresponding constraint to the parameter. But it hasn't been used since:
```
type ParameterType = StringParameter | SelectParameter (List String)
```

But I'd need some help with that. What do you think @clarktsiory ?

RaphaelGauthier:
Commit modified

RaphaelGauthier:
We have to use `<datalist>` paired to a `<input type="text">` instead of a `<select>`, as user may want to use a technique parameter, an iterator or a variable...

RaphaelGauthier:
PR rebased

RaphaelGauthier:
PR updated with a new commit

RaphaelGauthier:
PR updated with a new commit

RaphaelGauthier:
PR rebased

RaphaelGauthier:
PR replaced by https://github.com/Normation/rudder/pull/7120

PR #7099, opened 2026-04-24
Fixes #28769: Migrate to Lift 4.0.0-RC1 in scala 3

fanf:
https://issues.rudder.io/issues/28769

Switch to Lift 4.0.0-RC1 and other remaning lib in scala 2.13. 
With that, we are now full scala3 without any dependencies in Scala 2.13. 

The main impacts are: 
- switch to `scala-xml` 2.4.0. Specs 4 was compiled with 1.3.1 but it looks like it works, and specs 5 is out of reach. In any case, it's test-side only, and they pass everywhere, included in plugins. 

- it looks like some `LiftRules` function behavior as changed, like `liftContextRelativePath` which is not interpreted anymore. It might be due to change in imports. Other similar case may happen.

# Impact of `scala-xml 2.4.0`

- comments are kepts in the doc, so in several places we have to filter out `#REM` labeled node. It might be needed elsewhere, 
- the order of attributes when written changed again, and it's still not the orginal one. 
Long term solution is to avoid XML, or use good old Java XML libs. 

# Removing scala-compiler 

`lift-json` used to need it, but not json4s (https://github.com/json4s/json4s/issues/108) so we don't need it anymore until it's cleaned-up in lift side.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
@@ -148,7 +148,7 @@ object Boot {
               replaceCSPRestrictionDirectives("object-src", "'none'")(_)
             )
             .pipe(
-              _ :+ ("base-uri" -> "'none'") :+ ("report-uri" -> s"${S.contextPath}/${LiftRules.liftContextRelativePath}/content-security-policy-report")
+              _ :+ ("base-uri" -> "'none'") :+ ("report-uri" -> s"${S.contextPath}/${LiftRules.liftContextRelativePath().mkString("/")}/content-security-policy-report")
             )
         )
         val newCspHeaders = csp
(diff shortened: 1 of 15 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/Boot.scala
```diff
@@ -133,6 +133,7 @@ object Boot {
       */
     private def addCspHeaders(allHeaders: List[(String, String)]): List[(String, String)] = {
       if (WithDisabledCSP.isDisabled) { // no headers to override
+        println(s"****** case no header: ${allHeaders}")
```
mbaechler: I guess it's a leftover?

fanf: Yes :)

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM ! 🥳

PR #7058, opened 2026-04-03
Enhance Elm TEA usage

mbaechler:
https://issues.rudder.io/issues/28794

Following discussion with Pauline, I decided to demonstrate in a PR some changes that make the code more idiomatic to Elm developers and allow to tests much more the logic of it.

PR #7101, opened 2026-04-24
Fixes #28794: cover an elm app with tests including the update function

mbaechler:
https://issues.rudder.io/issues/28794

Initial explanations are located here : https://github.com/Normation/rudder/pull/7058


clarktsiory (commented):
LGTM but commits should be squashed before merging

clarktsiory (approved):

PR #7102, opened 2026-04-24
Fixes #28805: Only share the cargo sources caches between jobs

amousset:
https://issues.rudder.io/issues/28805

* only share registry/cache which should be safe
* remove cargo-vet cache which has been unused for some time


amousset:
PR updated with a new commit

PR #7110, opened 2026-04-27
Fixes #28802: Export Rule compliance by directive to CSV

skaerg:
https://issues.rudder.io/issues/28802

This PR aims to add an "export to CSV" button for rules when sorted by directive.

image

commit history : 
1. define new datatypes and extend existing `GetRuleId` of the Compliance Api so that it can return the complianc of a given rule by directive in csv format, and add a button that calls this endpoint in the Elm app
2. use opaque types instead of strings for new CSV datatypes
3. refactoring of CSV datatypes + return the `GetRuleId` endpoint to the way it was before and create a new endpoint to return the compliance of a given rule to CSV : `GetNodeGroupComplianceTargetId`
4. move the new endpoint to the Rule Api
5. refactoring of util functions for compliance endpoints
6. refactoring of util functions for compliance endpoints
7. modify the format of the JSON response of the new endpoint so that it only returns the `byDirective` sort, and add a test in `api_rules.yml`
8. add a test `api_rules.yml` for the CSV response of the new endpoint

### DIFF
--- a/webapp/sources/.scalafmt.conf
+++ b/webapp/sources/.scalafmt.conf
@@ -12,6 +12,7 @@ runner.dialectOverride.allowGivenUsing = true
 runner.dialectOverride.allowOpaqueTypes = true
 runner.dialectOverride.allowTraitParameters = true
 runner.dialectOverride.allowTypeLambdas = true
+runner.dialectOverride.allowDependentFunctionTypes = true
 
 rewrite.scala3.convertToNewSyntax = true
 
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
@@ -1052,6 +1052,14 @@ object RuleApi       extends Enum[RuleApi] with ApiModuleProvider[RuleApi]      
     val authz:                  List[AuthorizationType] = AuthorizationType.Rule.Write :: Nil
   }
 
+  case object GetRuleComplianceByDirective extends RuleApi with OneParam with StartsAtVersion24 with SortIndex {
+    val z: Int = implicitly[Line].value
+    val description    = "Get compliance by directive for the given rule"
+    val (action, path) = GET / "rules" / "{id}" / "compliance" / "byDirective"
+    override def dataContainer: Option[String]          = None
+    val authz:                  List[AuthorizationType] = AuthorizationType.Compliance.Read :: Nil
+  }
+
   def endpoints: List[RuleApi] = values.toList.sortBy(_.z)
 
   def values = findValues
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
@@ -41,6 +41,8 @@ import cats.Order
 import cats.data.NonEmptyList
 import cats.syntax.list.*
 import com.normation.cfclerk.domain.ReportingLogic
+import com.normation.errors.Inconsistency
+import com.normation.errors.IOResult
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.policies.DirectiveId
 import com.normation.rudder.domain.policies.RuleId
@@ -48,6 +50,7 @@ import com.normation.rudder.domain.reports.*
 import com.normation.rudder.domain.reports.ReportType.*
 import com.normation.rudder.reports.ComplianceModeName
 import com.normation.rudder.web.services.ComputePolicyMode.ComputedPolicyMode
+import com.normation.utils.Csv
 import enumeratum.*
 import fs2.Chunk
 import io.scalaland.chimney.*
@@ -57,6 +60,7 @@ import org.apache.commons.csv.CSVFormat
 import org.apache.commons.csv.QuoteMode
 import scala.collection.immutable
 import zio.json.JsonEncoder
+import zio.syntax.ToZio
 
 /**
  * Here, we want to present two views of compliance:
@@ -454,9 +458,8 @@ object CsvCompliance {
 
   def recurseComponent(
       component: ByRuleComponentCompliance,
-      block:     List[String]
-  ): Seq[(List[String], String, String, String, String, String)] = {
-    import com.normation.rudder.rest.data.ComplianceApiData.*
+      block:     List[ComponentField]
+  ): Seq[RuleComponentResult] = {
 
     component match {
       case component: ByRuleValueCompliance =>
@@ -464,18 +467,18 @@ object CsvCompliance {
           node.values.flatMap { value =>
             value.messages.map { report =>
               (
-                block,
-                component.name,
-                node.name,
-                value.componentValue,
-                report.reportType.serialize,
-                report.message.getOrElse("")
+                BlockField(block),
+                ComponentField(component),
+                NodeField(node),
+                ValueField(value),
+                StatusField(report),
+                MessageField(report)
               )
             }
           }
         }
       case component: ByRuleBlockCompliance =>
-        component.subComponents.flatMap(c => recurseComponent(c, block ::: (component.name :: Nil)))
+        component.subComponents.flatMap(c => recurseComponent(c, block ::: (ComponentField(component) :: Nil)))
     }
   }
 
@@ -486,14 +489,93 @@ object CsvCompliance {
           c                                            <- r.components
           (block, component, node, value, status, msg) <- recurseComponent(c, Nil)
         } yield {
-          List(r.name, block.mkString(","), component, node, value, status, msg)
+          List(r.name, block, component, node, value, status, msg)
         }
       }
-      // csvFormat take care of the line separator
+      // csvFormat takes care of line separators
       val out      = new lang.StringBuilder()
       csvFormat.printRecord(out, "Rule", "Block", "Component", "Node", "Value", "Status", "Message")
       csvLines.foreach(l => csvFormat.printRecord(out, l*))
       out.toString
+    }
+  }
+
+  /** Trait that contains Csv[A] and Csv.Header.One[A] instances 
+   * that are common to all values that can be converted to CSV */
+  trait CsvField[A <: String] {
+    given Csv[A] = Csv.instance(a => a)
+    given Csv.Header.One[A] with {}
+  }
+
+  opaque type DirectiveField = String
+  object DirectiveField extends CsvField[DirectiveField] {
+    def apply(directive: ByRuleDirectiveCompliance) = directive.name
+  }
+
+  opaque type BlockField = String
+  object BlockField extends CsvField[BlockField] {
+    def apply(block: List[ComponentField]) = block.mkString(",")
+  }
+
+  opaque type ComponentField = String
+  object ComponentField extends CsvField[ComponentField] {
+    def apply(value: ByRuleValueCompliance) = value.name
+    def apply(block: ByRuleBlockCompliance) = block.name
+  }
+
+  opaque type NodeField = String
+  object NodeField extends CsvField[NodeField] {
+    def apply(node: ByRuleNodeCompliance) = node.name
+  }
+
+  opaque type ValueField = String
+  object ValueField extends CsvField[ValueField] {
+    def apply(value: ComponentValueStatusReport) = value.componentValue
+  }
+
+  opaque type StatusField = String
+  object StatusField extends CsvField[StatusField] {
+    import com.normation.rudder.rest.data.ComplianceApiData.serialize
+    def apply(report: MessageStatusReport) = report.reportType.serialize
+  }
+
+  opaque type MessageField = String
+  object MessageField extends CsvField[MessageField] {
+    def apply(report: MessageStatusReport) = report.message.getOrElse("")
+  }
+
+  type RuleComponentResult =
+    (block: BlockField, component: ComponentField, node: NodeField, value: ValueField, status: StatusField, message: MessageField)
+
+  case class RuleComplianceCsv(
+      directive: DirectiveField,
+      block:     BlockField,
+      component: ComponentField,
+      node:      NodeField,
+      value:     ValueField,
+      status:    StatusField,
+      message:   MessageField
+  ) derives Csv
+  object RuleComplianceCsv {
+    import com.normation.rudder.rest.data.CsvCompliance.RuleComponentResult
+
+    given (using directive: ByRuleDirectiveCompliance): Transformer[RuleComponentResult, RuleComplianceCsv] = {
+      Transformer
+        .define[RuleComponentResult, RuleComplianceCsv]
+        .withFieldConst(_.directive, DirectiveField(directive))
+        .buildTransformer
+    }
+
+    given Transformer[ByRuleRuleCompliance, Seq[RuleComplianceCsv]] = (rule: ByRuleRuleCompliance) => {
+      rule.directives.flatMap(d => {
+        for {
+          c   <- d.components
+          res <- recurseComponent(c, Nil)
+        } yield {
+          given ByRuleDirectiveCompliance = d
+          res.transformInto[RuleComplianceCsv]
+        }
+      })
     }
   }
 }
@@ -921,6 +1003,30 @@ object ComplianceApiData {
         .withFieldRenamed(_.compliance, _.complianceDetails)
         .buildTransformer
     }
+
+    final case class RuleComplianceByDirectiveJson(
+        id:                RuleId,
+        name:              String,
+        compliance:        Double,
+        mode:              ComplianceModeName,
+        policyMode:        ComputedPolicyMode,
+        complianceDetails: ComplianceSerializable,
+        directives:        Option[Seq[ByRuleDirectiveComplianceApi]]
+    ) derives JsonEncoder
+
+    given ruleComplianceByDirectiveJson(using
+        precision: CompliancePrecision,
+        level:     Int
+    ): Transformer[ByRuleRuleCompliance, RuleComplianceByDirectiveJson] = {
+      Transformer
+        .define[ByRuleRuleCompliance, RuleComplianceByDirectiveJson]
+        .withFieldRenamed(_.compliance, _.complianceDetails)
+        .withFieldComputed(
+          _.directives,
+          x => if (level < 2) None else Some(x.directives.map(_.transformInto[ByRuleDirectiveComplianceApi]))
+        )
+        .buildTransformer
+    }
   }
 
   // GROUPS
@@ -1172,4 +1278,47 @@ object ComplianceApiData {
 
   }
 
+}
+
+object ComplianceUtils {
+  def extractComplianceLevel(params: Map[String, List[String]]): IOResult[Option[Int]] = {
+    params.get("level") match {
+      case None | Some(Nil) => None.succeed
+      case Some(h :: tail)  => // only take into account the first level param is several are passed
+        try {
+          Some(h.toInt).succeed
+        } catch {
+          case ex: NumberFormatException =>
+            Inconsistency(s"level (displayed level of compliance details) must be an integer, was: '${h}'").fail
+        }
+    }
+  }
+
+  def extractPercentPrecision(params: Map[String, List[String]]): IOResult[Option[CompliancePrecision]] = {
+    params.get("precision") match {
+      case None | Some(Nil) => None.succeed
+      case Some(h :: tail)  => // only take into account the first level param is several are passed
+        for {
+          extracted <- try {
+                         h.toInt.succeed
+                       } catch {
+                         case ex: NumberFormatException =>
+                           Inconsistency(s"percent precision must be an integer, was: '${h}'").fail
+                       }
+          level     <- CompliancePrecision.fromPrecision(extracted).toIO
+        } yield {
+          Some(level)
+        }
+
+    }
+  }
+
+  def extractComplianceFormat(params: Map[String, List[String]]): IOResult[ComplianceFormat] = {
+    params.get("format") match {
+      case None | Some(Nil) | Some("" :: Nil) =>
+        ComplianceFormat.JSON.succeed // by default if no there is no format, should I choose the only one available ?
+      case Some(format :: _)                  =>
+        ComplianceFormat.fromValue(format).left.map(Inconsistency.apply).toIO
+    }
+  }
 }
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
@@ -96,45 +96,6 @@ class ComplianceApi(
     readDirective:     RoDirectiveRepository
 ) extends LiftApiModuleProvider[API] {
 
-  def extractComplianceLevel(params: Map[String, List[String]]): IOResult[Option[Int]] = {
-    params.get("level") match {
-      case None | Some(Nil) => None.succeed
-      case Some(h :: tail)  => // only take into account the first level param is several are passed
-        try { Some(h.toInt).succeed }
-        catch {
-          case ex: NumberFormatException =>
-            Inconsistency(s"level (displayed level of compliance details) must be an integer, was: '${h}'").fail
-        }
-    }
-  }
-
-  def extractPercentPrecision(params: Map[String, List[String]]): IOResult[Option[CompliancePrecision]] = {
-    params.get("precision") match {
-      case None | Some(Nil) => None.succeed
-      case Some(h :: tail)  => // only take into account the first level param is several are passed
-        for {
-          extracted <- try { h.toInt.succeed }
-                       catch {
-                         case ex: NumberFormatException =>
-                           Inconsistency(s"percent precision must be an integer, was: '${h}'").fail
-                       }
-          level     <- CompliancePrecision.fromPrecision(extracted).toIO
-        } yield {
-          Some(level)
-        }
-
-    }
-  }
-
-  def extractComplianceFormat(params: Map[String, List[String]]): IOResult[ComplianceFormat] = {
-    params.get("format") match {
-      case None | Some(Nil) | Some("" :: Nil) =>
-        ComplianceFormat.JSON.succeed // by default if no there is no format, should I choose the only one available ?
-      case Some(format :: _)                  =>
-        ComplianceFormat.fromValue(format).left.map(Inconsistency.apply).toIO
-    }
-  }
-
   def schemas: ApiModuleProvider[API] = API
 
   /*
@@ -175,8 +136,8 @@ class ComplianceApi(
       given qc: QueryContext = authzToken.qc
 
       (for {
-        level     <- extractComplianceLevel(req.params)
-        precision <- extractPercentPrecision(req.params)
+        level     <- ComplianceUtils.extractComplianceLevel(req.params)
+        precision <- ComplianceUtils.extractPercentPrecision(req.params)
         rules     <- complianceService.getRulesCompliance(level)
       } yield {
         // by default, all details are displayed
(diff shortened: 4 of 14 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -489,13 +489,32 @@ object CsvCompliance {
...
+          (block, component, node, value, status, msg) <- recurseComponent(c, Nil)
+        } yield {
+          List(d.name, block.mkString(","), component, node, value, status, msg)
+        }
+      })
+      // csvFormat takes care of line separators
+      val out      = new lang.StringBuilder()
+      csvFormat.printRecord(out, "Directive", "Block", "Component", "Node", "Value", "Status", "Message")
+      csvLines.foreach(l => csvFormat.printRecord(out, l*))
+      out.toString
```
clarktsiory: we have CSV utils added since this PR : https://github.com/Normation/rudder/pull/6808

It could help normalizing CSV, avoiding typos in header names, and normalizing our CSV to use the utils in this PR. Using it would require mapping from `ByRuleRuleCompliance` to a new case class deriving the `Csv` typeclass...

clarktsiory: (by the way, `implicit class` are to be replaced with no-cost `extension` in scala 3)

skaerg:
Commit modified

webapp/sources/rudder/rudder-web/src/main/elm/sources/Rules.elm
```diff
@@ -697,6 +698,15 @@ update msg model =
       in
       handleOutMsg model groupsTable tabMsg outMsgOpt
 
+    Export ruleId res ->
+      case res of
+        Ok content ->
+          (model, File.Download.string (ruleId.value ++ ".csv") "text/csv" content)
```
clarktsiory: this filename is not the conventional one : we need `rudder_` as prefix, and the current date as a suffix (I see that it was probably inspired from the directive export, but in fact it needs to be renamed to the same convention now)

clarktsiory (changes requested):
LGTM except for the name of the file

skaerg:
Commit modified

webapp/sources/rudder/rudder-web/src/main/elm/sources/DirectiveCompliance/ViewRulesCompliance.elm
```diff
@@ -53,6 +53,7 @@ displayRulesComplianceTable model =
           , i [class ("fa " ++ (if complianceFilters.showComplianceFilters then "fa-minus" else "fa-plus"))][]
           ]
         , div [class "ms-auto my-auto"]
+          {- here ! -}
```
clarktsiory:
```suggestion
```

webapp/sources/rudder/rudder-web/src/main/elm/sources/Rules.elm
```diff
@@ -697,6 +699,28 @@ update msg model =
...
 
+    -- Export compliance table of a single rule to CSV
+    ExportComplianceCsvWithCurrentDate ruleId time ->
+        let
+          timeStr =
+            time
+            |> Time.DateTime.fromPosix
+            |> Time.Iso8601.fromDateTime
+            -- remove millis
+            |> String.toList |> List.take 10 |> String.fromList
```
clarktsiory: From [this ADR](https://docs.rudder.io/devel/adr/webapp/28524-use-elm-time.html), we should avoid any usage of `Time.DateTime` and `Time.Iso8601` : they come from deprecated time library, and we have the plan to remove them. Replacements would be directly using [`Iso8601`](https://package.elm-lang.org/packages/rtfeldman/elm-iso8601-date-strings/latest/) module which should be available

clarktsiory (changes requested):
Usage of deprecated time library should no longer be added since we have an ADR against that.

Also, there is something I find confusing in the messages, from what I understand, we are doing : 
```
button click > ExportComplianceCsvWithCurrentDate > CallApi > ExportComplianceCsvWithFilename
```

Problems I see are : 
* ExportComplianceCsvWithFilename is a result from API call : it shouldn't be a "verb", but an event in past tense, _Exported_
* there is additional overhead on `ExportComplianceCsvWithCurrentDate > CallApi`:
  * `CallApi` as a message should not be in the loop (such generic message should not even exist, since the update loop should consist of wiring `ExportComplianceCsvWithCurrentDate` directly with `getCSVExport`command, the CallApi seems to be doing nothing special more...)


Could you remove the additional message, and rename the result to `Exported...` ?

skaerg:
Commit modified

webapp/sources/rudder/rudder-web/src/main/elm/sources/Rules.elm
```diff
@@ -683,20 +682,42 @@ update msg model =
...
 
+    -- Export compliance table of a single rule to CSV
+    ExportRuleComplianceByDirective ruleId time ->
+        let
+          timeStr =
+            time
+            |> Iso8601.fromTime
+            -- remove time
+            |> String.toList |> List.take 10 |> String.fromList
+          filename = "rudder_rule_" ++ (ruleId.value) ++ "_compliance_by_directive_" ++ timeStr ++ ".csv"
```
skaerg, line 739: i added `by_directive` to the filename so that exports of the "by directive" and "by node" views are clearly separated/will be less likely to be confused with each other by users

clarktsiory: Nice !
I found out that [justinmimbs/date](https://package.elm-lang.org/packages/justinmimbs/date/latest/Date#format) has a date formatter, that can be used with (untested !!) `time |> Date.fromPosix zone |> Date.format "yyyy-MM-dd"` : 
* I think it would be the cleanest solution, compared to conversion to Iso8601 + truncating
* I also think it changes the semantic, in a better way : IIRC, Iso8601 posix are always formatted with `Z`, in UTC timezone; however, the user wants to export relative to the current `zone`

clarktsiory: the difference in semantics is quite important to consider : 
* `2026-01-01T01:00:00+02:00` when formatted with Iso8601 would be truncated and exported with `2025-12-31T23:00:00Z` so `2025-12-31`
* it should be `2026-01-01` in the user timezone

clarktsiory: (the example has silly dates, I'm not sure users would be using this features in that time 😅 but you see the point...)

skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -489,13 +490,36 @@ object CsvCompliance {
...
-      // csvFormat take care of the line separator
+      // csvFormat takes care of line separators
       val out      = new lang.StringBuilder()
       csvFormat.printRecord(out, "Rule", "Block", "Component", "Node", "Value", "Status", "Message")
       csvLines.foreach(l => csvFormat.printRecord(out, l*))
       out.toString
     }
   }
+
+  case class RuleComplianceCsv(
```
P4uline, line 550: > [!TIP]
> Avoid using strings everywhere, use value object instead.

Could you use stronger types rather than simple strings for properties ?
It would make your code safer, it would avoid bugs of inverting parameters : 
```
RuleComplianceCsv(d.name, block.mkString(","), component, node, value, msg, status)
```
vs 
```
RuleComplianceCsv(d.name, block.mkString(","), component, node, value, status, msg)
```
This is cumbersome to debug.

clarktsiory: I agree that it would be a nice addition, especially that `def recurseComponents: (String, String, ...)` returns `String` "business objects" : it would be better to have a strongly typed result of recursion, e.g. "RuleComponentResult" both for the domain, and for easier fields-mapping with the CSV part

skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -489,13 +490,54 @@ object CsvCompliance {
...
+            ComponentField(component),
+            NodeField(node),
+            ValueField(value),
+            StatusField(status),
+            MessageField(msg)
+          )
+        }
+      })
+    }
+  }
```
skaerg, line 580: like this @P4uline ? is there a better/less verbose way to do it ?

P4uline: Yes it's way better ! @clarktsiory will know if there is a better way to do it

clarktsiory: That is only at the "CSV level", but as I mentioned in [this comment](https://github.com/Normation/rudder/pull/7110#discussion_r3187241861), the _String API_ issue occurs yet in the previous level, the one of `recurseComponent`.

My suggestion would be to change the resulting type of `def recurseComponent` so that it is already strongly typed.
So the strong typing could even be applied there by : 
* having a dedicated shape, with named fields : `def recurseComponent(...): RuleComponentResult` and `case class RuleComponentResult(...)` (or even a [named tuple](https://docs.scala-lang.org/scala3/reference/other-new-features/named-tuples.html), since it is only needed for naming fields, not for other specificities of case classes such as hash code/equals/serialization)
* each "String value" could be a dedicated type, but I no longer see the benefit of that if we already have such shape. Having a dedicated type instead of `String` is already what you proposed here, and the most idiomatic way and efficient way to do it now with Scala 3 is [opaque types](https://docs.scala-lang.org/scala3/book/types-opaque-types.html) : again it would avoid defining costly "case class wrappers", which is exactly what opaque types are designed to replace)

clarktsiory: if it is applied only at the level of the CSV conversion, then the initial issue is still there : nothing prevents confusing 
```scala
          (block, component, node, value, status, msg) <- recurseComponent(c, Nil)
```
with
```scala
          (block, component, node, value, msg, status) <- recurseComponent(c, Nil)
```
, even with such wrappers

skaerg:
Commit modified

skaerg:
Commit modified

skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
clarktsiory: A full example of what it would look like with all pieces brought together (I'm sorry as I should have provided with a clearer example of what I meant earlier :sweat_smile:) : 
```scala
opaque type DirectiveField = String
object DirectiveField {
  // one single CSV column : directive name
  def apply(directive: Directive) = directive.name
  given Csv[DirectiveField] = Csv.instance(a => a)
  given Csv.Header.One[DirectiveField] with {}
}

/* ... all other "opaque types" ... */

// RecurseComponent
type RuleComponentResult = (block: BlockField, component: ComponentField, node: NodeField, value: ValueField, status: StatusField, message: MessageField)
object RuleComponentResult {
  given (using directive: Directive): Transformer[RuleComponentResult, RuleComplianceCsv] =
    Transformer
      .define[RuleComponentResult, RuleComplianceCsv]
      .withFieldConst(_.directive, DirectiveField(directive))
      .buildTransformer
}

def recurseComponent(...): Seq[RuleComponentResult] = ...

// CSV
case class RuleComplianceCsv(directive: DirectiveField, block: BlockField, ...) derives Csv
object RuleComplianceCsv {
  given Transformer[ByRuleRuleCompliance, Seq[RuleComplianceCsv]] = (rule: ByRuleRuleCompliance) => {
      rule.directives.flatMap(d => {
        for {
          c   <- components
          res <- recurseComponent(c, Nil)
        } yield {
          given Directive = d
          res.transformInto[RuleComplianceCsv]
        }
      })
  }
}
```

skaerg: could you explain what [this suggestion](https://github.com/Normation/rudder/pull/7110#discussion_r3196417632) would add concretely ? with my latest commit, the problem that was mentioned above can no longer happen, i.e. it is no longer possible to write 
`(block, component, node, value, msg, status) <- recurseComponent(c, Nil)` 
without having a compilation error

clarktsiory: my bad, I was reviewing the wrong commit: indeed having 
```scala
def recurseComponent(...): Seq[(BlockField, ComponentField, NodeField, ValueField, StatusField, MessageField)] = ...
```
already resolves the issue.

That suggestion was about yet having a structure which is easier to map to `RuleComplianceCsv`, by using chimney, that is a named tuple: 
```scala
type RuleComponentResult = (block: BlockField, component: ComponentField, node: NodeField, value: ValueField, status: StatusField, message: MessageField)
def recurseComponent(...): Seq[RuleComponentResult] = ...
```

The simplification is purely about syntax, the main point is about avoiding deconstruction of a "tuple6" then "applying" the `RuleComplianceCsv` constructor on each individual tuple elements (even if it is already strongly typed): 
```scala
        for {
          c   <- components
          // (block, component, node, value, status, msg) <- recurseComponent(c, Nil) // no need to deconstruct tuple6
          res <- recurseComponent(c, Nil)
        } yield {
          // RuleComplianceCsv(DirectiveField(d.name), block, component, node, value, status, msg) // no need to pass deconstructed values 
          given Directive = d
          res.transformInto[RuleComplianceCsv]
        }
``` 

So, it's simply some syntax sugar, and it eliminates the boilerplate on "tuple" and on creating the `RuleComplianceCsv`. That suggestion is completely optional and a matter of personal taste 😄 (I like to use chimney features at their full extent, since it simplifies developer experience around creating instances, but people may prefer using "native Scala API" e.g. around TupleX and manually applying constructors, I won't nitpick on that)

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_compliance.yml
```diff
@@ -5062,10 +5062,10 @@ response:
   type: text
   content: |
     "Rule","Block","Component","Node","Value","Status","Message"
-    "R3","","directive2-component-br3-bn1","node1.localhost","directive2-component-value-br3-bn1","error",""
-    "R3","","directive2-component-br3-bn2","node1.localhost","directive2-component-value-br3-bn2","error",""
-    "R3","","directive2-component-r3-n1","node1.localhost","directive2-component-value-r3-n1","error",""
-    "R3","","directive2-component-r3-n2","node1.localhost","directive2-component-value-r3-n2","error",""
+    "R3","BlockField()","ComponentField(directive2-component-br3-bn1)","NodeField(node1.localhost)","ValueField(directive2-component-value-br3-bn1)","StatusField(error)","MessageField()"
```
clarktsiory: it looks like `toString` method or `Csv` instances are not the right ones

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -486,16 +487,52 @@ object CsvCompliance {
...
+    def apply(s: String): Field = s
+  }
+
+  case class DirectiveField(directive: String) extends CsvField
+  case class BlockField(block: String)         extends CsvField
+  case class ComponentField(component: String) extends CsvField
+  case class NodeField(node: String)           extends CsvField
+  case class ValueField(value: String)         extends CsvField
+  case class StatusField(status: String)       extends CsvField
+  case class MessageField(message: String)     extends CsvField
```
clarktsiory: the correct usage of opaque types, along with `Csv` typeclass (see [this PR](https://github.com/Normation/rudder/pull/6889/changes#diff-8d5fee09f8d841f643a0136aff17719b10c498b61e84a2c76c15ca937f7afed0R56)), would be : 

```scala
opaque type ValueField = String
object ValueField {
  def apply(value: String) = value
  given Csv[ValueField] = Csv.instance(a => a)
  given Csv.Header.One[ValueField] with {}
}

**Review by clarktsiory — COMMENTED** [review]
a few things about correctly using opaque types, along with the CSV utils (the tests added in this PR would be useful to understand usage of CSV utils: https://github.com/Normation/rudder/pull/6889)

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala (hunk 9 above)
Comment (clarktsiory) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala]: sadly, since those opaque types are scalar values, I don't see a way to generate the boilerplate of Csv.

I don't see where the `CsvField` super-interface could be used: it could be [refined](https://docs.scala-lang.org/scala3/reference/other-new-features/opaques-details.html#opaque-types-in-transparent-inline-methods) with `opaque type ValueField <: CsvField = String`, but I don't see any benefit from that, so it may need to be removed

**Comment by skaerg** [comment]
PR updated with a new commit

File: webapp/sources/.scalafmt.conf [hunk 10]
```diff
@@ -2,6 +2,7 @@ version = "3.7.17"
 maxColumn = 130
 runner.debug = true
 runner.dialect = scala213Source3
+runner.dialectOverride.withAllowDependentFunctionTypes = true
```
fanf: can you put it just below the comment ? (and check if the comment is still pointing to the correct place, or update it if it's not the case?)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
fanf: I don't get why they are introduced here. The semantic of these things is defined in the their host business objects, not here. 

If it's just a tool for the CSV case, I don't think opaque types for each elements are the best suited: they are rather verbose for just defining the shape of the CSV, and don't bring much compared to a case class / tuple with named argument. 

On the other hand, if they are used in base business object, they add clarity about what that string is everywhere. 

I do understand it's a bigger refactoring - it can/should be done in two steps, but if it was the idea, it must be advertised

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -486,16 +486,102 @@ object CsvCompliance {
...
+    given Csv[ComponentField] = Csv.instance(a => a)
+    given Csv.Header.One[ComponentField] with {}
+  }
+
+  opaque type NodeField = String
+  object NodeField {
+    def apply(node: ByRuleNodeCompliance) = node.name
+    given Csv[NodeField] = Csv.instance(a => a)
+    given Csv.Header.One[NodeField] with {}
+  }
```
fanf, line 529: these "name" opaque types have all exactly the same shape. Can that be factored out somehow ? It's quite boilerplaty and we don't see easely the ones that are not simple `name` disambiguators

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -158,6 +158,14 @@ object ComplianceApi       extends Enum[ComplianceApi] with ApiModuleProvider[Co
     val dataContainer: Some[String]            = Some("rules")
     val authz:         List[AuthorizationType] = AuthorizationType.Compliance.Read :: Nil
   }
+  case object GetRulesComplianceIdByDirective
+      extends ComplianceApi with GeneralApi with OneParam with StartsAtVersion24 with SortIndex {
+    val z: Int = implicitly[Line].value
+    val description    = "Get compliance by directive for the given rule"
+    val (action, path) = GET / "compliance" / "rules" / "byDirective" / "{id}"
```
P4uline: It would be nice to have the url this way : `GET / "rules" / "{id}" / "compliance" / "byDirective"`

(My guide line examples to design REST endpoints url:
- first: think about what are resources and sub resources involved
- examples : 
`<METHOD> resource ? offset=0&limit=10 `
`<METHOD> resource / {action}`
 `<METHOD> resource / {resource-id}`
`<METHOD> resource / {resource-id} / subresource ? offset=0&limit=10 `
`<METHOD> resource / {resource-id} / subresource / {action}`
`<METHOD> resource / {resource-id} / subresource / {subresource-id}`



)


and also
+ deprecation of old endpoints 
+ update feature page 
+ plan old version and new version of the api
+ ... etc...
(everything we discussed)

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -486,16 +486,93 @@ object CsvCompliance {
...
-      // csvFormat take care of the line separator
+      // csvFormat takes care of line separators
       val out      = new lang.StringBuilder()
       csvFormat.printRecord(out, "Rule", "Block", "Component", "Node", "Value", "Status", "Message")
       csvLines.foreach(l => csvFormat.printRecord(out, l*))
       out.toString
     }
   }
+
+  class CsvField[A <: String] {
```
fanf: I think it should be a `trait`

fanf: with a comment for the intent of what it does/why it exists

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
P4uline: So @skaerg, for this bigger refactoring, could you create a ticket so I will add it to our next mob programming session topic list. @clarktsiory proposed to work on opaque types so this would be the perfect case.

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -486,16 +486,95 @@ object CsvCompliance {
...
+
+  /** Traits that contains Csv[A] and Csv.Header.One[A] instances 
+   * that are common to all values that can be converted to CSV */
+  trait CsvField[A <: String] {
+    given Csv[A] = Csv.instance(a => a)
+    given Csv.Header.One[A] with {}
+  }
+
+  opaque type DirectiveField = String
+  object DirectiveField extends CsvField[DirectiveField] {
```
clarktsiory: 🚀 👍

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -289,6 +342,53 @@ class RuleApi(
...
+        params:     DefaultParams,
+        authzToken: AuthzToken
+    ): LiftResponse = {
+      given qc: QueryContext = authzToken.qc
+
+      import ComplianceApi.*
+
+      (for {
+        t1 <- currentTimeMillis
+
```
clarktsiory: the `currentTimeMillis` API could be replaced with `zio.Clock.instant` and `Duration.fromInterval`, you can have a look at nice examples here : https://github.com/Normation/rudder/pull/6873

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/RuleApi.scala
```diff
@@ -50,30 +50,82 @@ import com.normation.rudder.batch.AsyncDeploymentActor
...
+  }
+
+  def extractComplianceFormat(params: Map[String, List[String]]): IOResult[ComplianceFormat] = {
+    params.get("format") match {
+      case None | Some(Nil) | Some("" :: Nil) =>
+        ComplianceFormat.JSON.succeed // by default if no there is no format, should I choose the only one available ?
+      case Some(format :: _)                  =>
+        ComplianceFormat.fromValue(format).left.map(Inconsistency.apply).toIO
+    }
+  }
```
clarktsiory: The `extract` methods look like they are plain duplicates from `ComplianceApi` : it would be more maintainable and cleaner to refactor them (e.g. in a util singleton)

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/ComplianceApi.scala
```diff
@@ -1635,3 +1596,46 @@ class ComplianceAPIService(
     }
   }
 }
+
+object ComplianceUtils {
```
clarktsiory: I'm not convinced about `ComplianceUtils` if the structure is in the `ComplianceApi.scala` file.

In general, I wouldn't put something shared by 2 files in one of the two : it looks like `ComplianceUtils` is biased for use in `ComplianceApi.scala` (which it shouldn't because it will be used as importantly in RuleApi, DirectiveApi, etc.).

I wouldn't go as far as saying that "1 class = 1 file" (I'm looking at you, java...) is better, but in this specific case, I don't like the `ComplianceUtils` being in the `ComplianceApi` file, and I find the discoverability and reasoning over files simpler with : 

```mermaid
graph TD
    A[ComplianceApi.scala] --> C[ComplianceUtils.scala]
    B[RuleApi.scala] --> C[ComplianceUtils.scala]
```

clarktsiory: what do you think @P4uline ?

webapp/sources/rudder/rudder-web/src/main/elm/elm.json
```diff
@@ -69,4 +69,4 @@
         "direct": {},
         "indirect": {}
     }
-}
+}
```
clarktsiory: there's a recurring issue here...

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -1052,6 +1052,14 @@ object RuleApi       extends Enum[RuleApi] with ApiModuleProvider[RuleApi]
     val authz:                  List[AuthorizationType] = AuthorizationType.Rule.Write :: Nil
   }
 
+  case object GetRuleComplianceByDirective extends RuleApi with OneParam with StartsAtVersion24 with SortIndex {
+    val z: Int = implicitly[Line].value
+    val description    = "Get compliance by directive for the given rule"
+    val (action, path) = GET / "rules" / "{id}" / "compliance" / "byDirective"
+    override def dataContainer: Option[String]          = None
+    val authz:                  List[AuthorizationType] = AuthorizationType.Rule.Read :: AuthorizationType.Compliance.Read :: Nil
```
clarktsiory: It seems that all existing `ComplianceApi` endpoint have `compliance_read`.

However here it's `rule_read` **OR** `compliance_read` i.e. having the right to see "Rules" would be enough to see applied directives and their compliance, which is a security issue !

It should simply be 
```suggestion
    val authz:                  List[AuthorizationType] = AuthorizationType.Compliance.Read :: Nil
```

clarktsiory (commented):
Could you add some YAML tests for the CSV (like for `/api/compliance/directives` in `api_compliance.yml` but in `api_rules.yml` since the new endpoint is in `RuleApi`) ?

Also the API authz should be `Compliance.Read` only

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

clarktsiory (approved):
All LGTM now ! 
Despite the long discussions we now have a great template for future additions of compliance in CSV format !

PR #7115, opened 2026-04-28
Fixes #28157: Change logs : fix search

P4uline:
https://issues.rudder.io/issues/28157

The search was already working well.
Some column were ignored and now are taken in account in the query: id, actor (principal) and date.

After some discussion to decide if we keep or remove the id from the query, it was decided to keep it with an exact equals instead of a like, not to pollute more relevant research.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala
@@ -429,7 +429,9 @@ object EventLogJdbcRepository {
       None
     } else {
       val v = "%" + search.value + "%"
-      Some(fr"(temp1.filter ilike ${v} OR temp1.eventtype ilike ${v})")
+      Some(
+        fr"(temp1.filter ilike ${v} OR (temp1.id::text) like ${search.value} OR temp1.principal ilike ${v} OR (temp1.creationdate::text) ilike ${v})"
+      )
     }
   }
 }


P4uline:
PR rebased

P4uline:
PR rebased

P4uline:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala
```diff
@@ -429,7 +429,9 @@ object EventLogJdbcRepository {
       None
     } else {
       val v = "%" + search.value + "%"
-      Some(fr"(temp1.filter ilike ${v} OR temp1.eventtype ilike ${v})")
+      Some(
+        fr"(temp1.filter ilike ${v} OR temp1.id = (${search.value})::integer OR temp1.principal ilike ${v} OR (temp1.creationdate::text) ilike ${v})"
```
clarktsiory: I am surprised that `(${search.value})::integer` wouldn't fail when the search is not an integer...
Should it not be `temp.id::text = ${search.value}` instead ?

clarktsiory (commented):
adding a lot of `OR` makes me worry about performance, but I don't know about how it behave with millions of event logs (we don't have a way to test that unless with a load testing machine) ...

An optimization that could be applied for the `id` part, to take advantage of the index could be something like : 
```sql
select ... from temp1 where id = ${searchValueInt}
UNION ALL
select  ... from temp1 where ... # other filters
```

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala
P4uline: yes, fixed

P4uline:
> adding a lot of `OR` makes me worry about performance, but I don't know about how it behave with millions of event logs (we don't have a way to test that unless with a load testing machine) ...
> 
> An optimization that could be applied for the `id` part, to take advantage of the index could be something like :
> 
> ```sql
> select ... from temp1 where id = ${searchValueInt}
> UNION ALL
> select  ... from temp1 where ... # other filters
> ```

I planned to manage perf in the next milestone, fine for you ?

P4uline:
PR updated with a new commit

clarktsiory (approved):
Okay, performances issues will need to be addressed later

PR #7117, opened 2026-04-28
Fixes #28825: Enhance graalvm sandboxing and disable internationalization

fanf:
https://issues.rudder.io/issues/28825

So, I'm not sure it worth it: 
- we can disable internationnalisation but not remove the dependency to icu4j, so no gain here, 
- if we disable it, we can't run in anything but "trusted" to change the setting, 
- we can't go beyond `CONSTRAINED` because we don't have the proprietary GraalVM


fanf:
So, this is going nowhere, because graalvm embed its own version of package, leading to noClassDefFound at runtime. 
the only viable option would be to define our own restricted langage or use one designed for embeding.

PR #7122, opened 2026-04-30
Fixes #28775: Rule deletion/disable/enable has no notice about change requests when workflow are enabled

clarktsiory:
https://issues.rudder.io/issues/28775

* There was many inconsistencies with all Modal, which is fixed by having a single `viewModal` :
  * it enforces that the button is always a `btn-primary` with `"Open request"` text, and there is an alert message when change requests are enabled
  * it has the logic of checking if the mandatory (or not) input message 

* Fixing the behavior consists in reading the `changeRequestId` from the result of the "DELETE API call" (the same exact thing is done for rule updates)
---
**Further improvements:**
* this PR only changes the Create/Update/Disable/Enable/Delete action modal, but we should take "Delete category" modal into account
* we could make a separate module `Modal.elm`, but it would have made the diff harder, and I do not know if `ModalState` should belong there

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Rules.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Rules.elm
@@ -20,6 +20,7 @@ import Rules.View exposing (view)
 import Rules.ViewUtils exposing (..)
 
 import Ui.Datatable exposing (SortOrder(..), Category, SubCategories(..))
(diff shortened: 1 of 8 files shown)


clarktsiory:
Commit modified

RaphaelGauthier:
LGTM ! :clap:

RaphaelGauthier (approved):

clarktsiory:
OK, merging this PR

PR #7128, opened 2026-05-07
Fixes #28855: there is no way to manage technique resources via API

P4uline:
https://issues.rudder.io/issues/28855


P4uline:
PR updated with a new commit

PR #7129, opened 2026-05-07
Fixes #28720: The groups All nodes managed by 'xxxx' policy server don't show the windows agent

fanf:
https://issues.rudder.io/issues/28720

The change is only adding the word `Linux`, but longest is the simple migration. 

In the migration, the description is not changed, because it looks like they have changed more often, and the meaning is OK for it once the display name is correct. 

UPMERGE NOTE: the migration part was not merged in Rudder 9.2 since it will useless here.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
@@ -637,8 +637,8 @@ object PolicyServerConfigurationObjects {
     )
     NodeGroup(
       idGroupHasPolicyServer(nodeId),
-      s"All nodes managed by '${nodeId.value}' policy server",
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/servers/PolicyServerManagementService.scala
```diff
@@ -637,8 +637,8 @@ object PolicyServerConfigurationObjects {
     )
     NodeGroup(
       idGroupHasPolicyServer(nodeId),
-      s"All nodes managed by '${nodeId.value}' policy server",
-      s"All nodes known by Rudder directly connected to the '${nodeId.value}' server. This group exists only as internal purpose and should not be used to configure Nodes.",
+      s"All Linux nodes managed by '${nodeId.value}' policy server",
+      s"All Linux nodes known by Rudder directly connected to the '${nodeId.value}' server. This group exists only as internal purpose and should not be used to configure Nodes.",
       Nil,
```
fanf, line 642: That's the correction

clarktsiory (approved):
LGTM!

fanf:
OK, merging this PR

PR #7130, opened 2026-05-07
Fixes #28856: There is no documentation on special tenants value

clarktsiory:
https://issues.rudder.io/issues/28856

* Add a special `tenant-ids` schema in API doc, and the expected format (used only in API accounts for now, but it should later be used in user management too) : 
 
Screenshot From 2026-05-21 16-30-33

* Mention the default `tenant="*"` in the demo users.xml file

### DIFF
--- a/webapp/sources/api-doc/components/schemas/api-account-new.yml
+++ b/webapp/sources/api-doc/components/schemas/api-account-new.yml
@@ -39,9 +39,7 @@ properties:
     format: rfc3339
     example: 2025-05-10T23:59:59Z
   tenants:
-    description: a comma separated list of tenants allowed for that account. See tenant documentation for syntax.
-    type: string
-    example: "*"
+    $ref: './tenant-ids.yml'
   authorizationType:
     description: >-
       Authorization type for that account: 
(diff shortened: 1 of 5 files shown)


fanf (changes requested):
I think you need to differentiate between a `tenant-id`, which must be non empty alpha-chars, and `tenant-ids` parameter, which is a either a comma separated list of `tenant-id`s or one of the special chars `*` or `-`. 
It is absolutly forbidden to use `* ` or `-` as a tenant ID.

webapp/sources/api-doc/components/schemas/tenant-id.yml
```diff
@@ -0,0 +1,14 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2026 Normation SAS
+type: string
+description: |
+  A tenant identifier.
+  
+  Special values:
+  - "*" → access to all tenants
+  - "-" → access to no tenants
```
fanf: these special values are for permissions, not for tenant ID, when you can pass a list of tenant ids.

clarktsiory:
PR updated with a new commit

webapp/sources/api-doc/components/schemas/tenant-id.yml
clarktsiory: I renamed the file to `tenant-ids.yml` as it's a CSV list of tenants, with expected usage being having isolated '*' or '-' : 

Screenshot From 2026-05-21 16-30-33

fanf (approved):

PR #7133, opened 2026-05-11
Fixes #26574: Missing edit button on node documentation

skaerg:
https://issues.rudder.io/issues/26574

This PR aims to add a text input on every node's page so that its documentation can be modified

read mode :

image

edit mode : 

image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeDescription/View.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeDescription/View.elm
@@ -1,0 +1,77 @@
+module NodeDescription.View exposing (view)
+
+import Dom exposing (addClass, appendNodeList, element)
+import Html exposing (Html, div, i, label, span, text, textarea)
+import Html.Attributes exposing (class, for, id, maxlength, name, title)
+import Html.Events exposing (onClick, onInput)
+import Markdown
+import NodeDescription.DataTypes exposing (..)
+
+
+view : Model -> Html Msg
+view model =
+    div [] [ descriptionTextArea model ]
+
+
+descriptionTextArea : Model -> Html Msg
+descriptionTextArea model =
+    let
+        inputClass =
+            "form-control col-xs-12"
+
+        textArea =
+            case model.editMode of
+                HasNodeWriteAuth Write ->
+                    textarea
+                        [ name "node-description"
+                        , id "node-description"
+                        , maxlength 255
+                        , onInput EditNodeDescription
+                        , class inputClass
+                        ]
+                        [ text model.newDescription ]
+
+                _ ->
+                    Dom.render
+                        (element "div"
+                            |> addClass "markdown"
+                            |> appendNodeList (Markdown.toHtml Nothing model.newDescription)
+                        )
+
+        editButtons =
+            case model.editMode of
+                NoNodeWriteAuth ->
+                    text ""
+
+                HasNodeWriteAuth Read ->
+                    i
+                        [ class "fa fa-pencil ms-2 text-primary cursorPointer edit-node-documentation"
+                        , onClick ToggleMode
+                        , title "Edit node documentation"
+                        ]
+                        []
+
+                HasNodeWriteAuth Write ->
+                    i
+                        [ class "fa fa-check ms-2 text-primary cursorPointer save-node-documentation"
+                        , onClick ToggleMode
+                        , title "Save node documentation"
+                        ]
+                        []
+    in
+    div
+        [ id "node-description" ]
+        [ div
+            [ class "row" ]
+            [ label
+                [ for "node-description-label", class "col-xs-12" ]
+                [ span [ class "fw-normal" ]
+                    [ text "Documentation"
+                    , editButtons
+                    ]
+                ]
+            , div
+                [ class "col-xs-12" ]
+                [ textArea ]
+            ]
+        ]
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
@@ -275,7 +275,7 @@ object DisplayNode extends Loggable {
           <button class="nav-link" data-bs-toggle="tab" data-bs-target={
         htmlId_#(jsId, "sd_fs_")
       } type="button" role="tab" aria-controls={htmlId_#(jsId, "sd_fs_")}>File systems</button>
-      </li> ::
+        </li> ::
       <li class="nav-item">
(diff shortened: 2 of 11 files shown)


skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/package.json
```diff
@@ -8,6 +8,7 @@
     "datatables.net-plugins": "^1.13.5",
     "datatables.net-rowgroup": "^1.3.1",
     "diff": "^8.0.2",
+    "elm-format": "^0.8.8",
```
clarktsiory: This should not be part of this PR, there is an upcoming "global enforcement" of Elm code formatting (check https://issues.rudder.io/issues/28989)
For now you can just remove it and keep your code formatted (not mandatory, we will apply formatting of all filesvery soon anyway)

webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeDescription/View.elm
```diff
@@ -0,0 +1,75 @@
...
+
+descriptionTextArea : Model -> Html Msg
+descriptionTextArea model =
+    let
+        inputClass =
+            "form-control col-xs-12"
+
+        attributes =
+            [ name "NodeDescription"
+            , id "NodeDescriptionLabel"
```
clarktsiory: the PascalCase naming for HTML attributes is very unconventional, we prefer kebab-case (we could even have an ADR for that)

clarktsiory: is there a reason for that ?
I'm not surprised if the nodes page has other occurrences for that, but since you are adding new HTML that is used only locally, it isn't a problem and it would reduce inconsistency with the rest of the codebase

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -661,15 +661,28 @@ object DisplayNode extends Loggable {
...
+                                |var nodeDocumentation = document.getElementById("nodedocumentation-app")
+                                |var initValues = {
+                                |  nodeId : "${nodeFact.id.value}",
+                                |  contextPath : contextPath,
+                                |  nodeWrite : CanWriteNode
+                                |};
+                                |var app = Elm.NodeDescription.init({node: nodeDocumentation, flags: initValues});
+                                |app.ports.errorNotification.subscribe(function(str) {
+                                |  createErrorNotification(str)
+                                |});
```
clarktsiory: there are no reasons to use `var` instead of `const` locally

clarktsiory: we don't have an ADR yet also, but it's the more modern and "safe" JS 😅

clarktsiory (commented):
The structure looks good, I addressed only some minor issues.

We discussed the `hasWriteRights` flag pattern, which seems to not going into the direction of minimizing Elm/Lift coupling, but we have no solution to "check users rights" for now (it would be a greater architectural change). So adding the flag will do for now

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeDescription/View.elm
```diff
@@ -0,0 +1,75 @@
...
+                ]
+        in
+        div [ id "save-node-description" ]
+            [ input
+                attrList
+                [ text "Update" ]
+            ]
+
+    else
+        div [ id "save-node-description" ] []
```
clarktsiory: is there a reason why this is not `text ""` ?

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/DisplayNode.scala
```diff
@@ -661,15 +661,28 @@ object DisplayNode extends Loggable {
...
+
         </div>
         <div class="rudder-info">
           <h3>Documentation</h3>
-          <div class="markdown" id="nodeDocumentation">
-          </div>
           {
+      <div id="nodedocumentation-app"></div> ++
       WithNonce.scriptWithNonce(
-        Script(OnLoad(JsRaw(s"generateMarkdown(${Str(nodeFact.documentation.getOrElse("")).toJsCmd}, '#nodeDocumentation')")))
```
clarktsiory, line 672: We seems to have lost markdown support for the display

webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeDescription/View.elm
```diff
@@ -0,0 +1,75 @@
...
+                        ]
+                   )
+    in
+    div
+        [ id "node-description" ]
+        [ div
+            [ class "row" ]
+            [ label
+                [ for "node-description-label", class "col-xs-12" ]
+                [ span [ class "fw-normal" ] [ text "Description" ] ]
```
clarktsiory: the label "Description" is confusing, we already have the "Documentation" title, we should avoid the confusion be simply keeping "documentation"

webapp/sources/rudder/rudder-web/src/main/elm/sources/NodeDescription/View.elm
```diff
@@ -0,0 +1,75 @@
...
+                        [ maxlength 255
+                        , onInput EditNodeDescription
+                        , class inputClass
+                        ]
+
+                    else
+                        [ class inputClass
+                        , disabled True
+                        , readonly True
+                        ]
```
clarktsiory, line 59: Actually node description is markdown, so we have lost support for displaying pretty markdown.

There are some examples of markdown display in Elm, e.g. in the technique editor.

The UX solution would be to have a "crayon" button next to the button, to toggle edit mode (@RaphaelGauthier can correct me otherwise)

clarktsiory (approved):
Could you add a comment in the model that the `documentation` is actually same field as the `description` ?

Also the markdown support is lost, so users would always see editable "raw markdown".
Could you have a look at how to fix that, is it possible to do without taking too long ?

clarktsiory (changes requested):
I mistakenly approved, so I put it in draft to prevent merge

skaerg:
PR updated with a new commit

clarktsiory (approved):
LGTM !

skaerg:
OK, squash merging this PR

PR #7134, opened 2026-05-11
Fixes #28873: UserId list parameter should be userIds

clarktsiory:
https://issues.rudder.io/issues/28873

Simply rename `userId` to `userIds`, since it's a `List[String]`


fanf (approved):
LGTM

PR #7132, opened 2026-05-07
Port more joda-time uses to java-time

mbaechler:
You should ignore the flake.nix commit which is only useful to develop locally for me.

As I'm pushing this PR in a rush tonight, I didn't took the time to create a new issue regarding this migration. 
I'll do it first thing in two weeks.

PR #7135, opened 2026-05-11
Fixes #28874: Migrate to java-time offset and zoned date time 

clarktsiory:
https://issues.rudder.io/issues/28874

Several porting to java-time "offset datetime" serialisation, and some ZonedDateTime too.

For the history of commits, see the original PR from @mbaechler: https://github.com/Normation/rudder/pull/7132.

The next steps would be about removing the `DateFormaterService#serialize(dateTime: org.joda.datetime.DateTime)` method (which was "Deprecated" in the original PR, but still used for now, we would still need to check plugins too)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CleanupUsers.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CleanupUsers.scala
@@ -45,8 +45,7 @@ import com.normation.utils.CronParser.*
 import com.normation.utils.DateFormaterService
 import com.normation.zio.*
 import cron4s.CronExpr
-import org.joda.time.DateTime
-import org.joda.time.DateTimeZone
+import java.time.Duration
 import zio.*
 
 /**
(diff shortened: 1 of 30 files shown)


clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/CleanupUsers.scala
```diff
@@ -85,21 +84,20 @@ class CleanupUsers(
...
       getNonAdminUserIds
-        .flatMap(
+        .flatMap(userIds => {
           userRepository
             .disable(
-              _,
-              Some(d.minus(disableInactive.toMillis)),
-              Nil,
-              trace(s"User was inactive since last '${disableInactive.render}'")
+              userIds = userIds,
```
P4uline, line 95: In a previous PR I was asked not to use named argument... Should this case be an exception ? If so why ?

mbaechler: I can't say why somebody told you so, but my opinion is :

- if parameters semantic is obvious, you can omit named parameters
- if local variables have the same names as parameters, you can omit named parameters
- in all other cases, they help the reader so they worth the additional verbosity

clarktsiory:
PR rebased

fanf (approved):
Lgtm, seems to work as before

PR #7137, opened 2026-05-12
Fixes #28878: The export of archive is broken

fanf:
https://issues.rudder.io/issues/28878

So, we are using a filter that only matches the end of the path, and it stops on the first matches. If you happen to have two techniques with the same end, it never work correctly for the first one. 

The problem is all around `GitParseRudderObjects.scala` file, which was create to deal with technique at a certain revision point, and hopefuly that we don't use a lot. 

Moreover, in the specific case of getting technique resources, we rewrote the git filter logic, which we should really avoid to do, because it's not the kind of horror that we want to duplicate. 

So the whole correction is small: 
- replace the ad-hoc filtered tree traversal by `GitFindUtils.listFiles`
- add leading `/` in `listFiles` so that we disambiguate between `/blabla_something/foo/bar` and `/something/foo/bar`

Most of the other changes are related to the unit test and set-uping the repo with two techniques.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
@@ -65,7 +65,6 @@ import com.normation.rudder.domain.policies.Rule
 import com.normation.rudder.domain.policies.RuleTargetInfo
 import com.normation.rudder.domain.policies.RuleUid
 import com.normation.rudder.domain.properties.GlobalParameter
-import com.normation.rudder.git.FileTreeFilter
 import com.normation.rudder.git.GitCommitId
 import com.normation.rudder.git.GitFindUtils
 import com.normation.rudder.git.GitRepositoryProvider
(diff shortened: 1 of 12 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
```diff
@@ -168,7 +166,7 @@ class GitParseRules(
     for {
       treeId <- GitFindUtils.findRevTreeFromRevString(repo.db, rev.value)
       // rules are just under "rules", but use list to check is one exists on that revtree
-      rules  <- GitFindUtils.listFiles(repo.db, treeId, List(rulesDirectory.directoryPath), uid.value + ".xml" :: Nil)
+      rules  <- GitFindUtils.listFiles(repo.db, treeId, List(rulesDirectory.directoryPath), s"/${uid.value}.xml" :: Nil)
```
fanf, line 169: it shouldn't be necessary for UUIDs, but well, that's even more correct

VinceMacBuche (approved):

PR #7141, opened 2026-05-19
Fixes #28831: Export Rule compliance by node to CSV

skaerg:
https://issues.rudder.io/issues/28831

This PR aims to add an "export to CSV" button for the node compliance tab of a given rule.

image

Commit history : 

1. implement `GetRuleComplianceByNode` endpoint in JSON format ; refactor parts that are common to this endpoint and the `GetRuleComplianceByDirective` endpoint ; add the "export to CSV" button in the Elm app
2. implement `GetRuleComplianceByNode` endpoint in CSV format ; add API tests for the `GetRuleComplianceByNode` endpoint for both the JSON and CSV formats ; add a tooltip in both the "nodes" and "directives" pages next to the export button in order to indicate that user filters are not taken into account when exporting the rule's compliance to CSV (i.e. the full table will be exported)
3. API documentation for the 2 new endpoints
4. review
5. review : refactor parts of the API doc schemas
6. review
7. add a tooltip when the user hovers over the export button instead of having a separate :information_source: icon next to the button for the tooltip


skaerg:
PR updated with a new commit

skaerg:
i would like your opinion about these specific points @P4uline @clarktsiory  : 

- in the implementation of the `GetRuleComplianceByNode` endpoint in CSV format, i operated under the premise that the result of the "by directive" and "by node" views are identical, with the only difference being that the order of the columns is different. For this reason, i did not re-implement the `recurseComponent` function for the "by node" case and instead used the existing function that is used in the "by directive" sort, and simply swapped the order of the columns at the end. is this implementation okay ?
- since there have been talks of deprecating the Compliance API and instead migrate all of its endpoints to the APIs that correspond to the resource type that is handled (e.g. the rule compliance endpoints should be moved to the Rules API), I was wondering if the rule compliance endpoints should already get copied from the Compliance API to the Rule API (especially since the two new endpoints that are added to the Rules API could get confused with the existing endpoints of the Compliance API) ? If so :
  - I think the rule compliance endpoints should be copied to the Rules API in another PR in order to make the change more explicit/traceable ; should I create an issue for this change ? 
  - in which version should the Compliance API be deprecated ?

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Agentpolicymode/View.elm
```diff
@@ -109,6 +109,7 @@ view model =
           , text ", or contact your Rudder administrator about this."
           ]
   in
+  -- fixme
```
clarktsiory:
```suggestion
```

webapp/sources/rudder/rudder-web/src/main/elm/elm.json
```diff
@@ -69,4 +69,4 @@
         "direct": {},
         "indirect": {}
     }
-}
+}
```
clarktsiory:
```suggestion
}

```

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -544,10 +544,10 @@ object CsvCompliance {
     def apply(report: MessageStatusReport) = report.message.getOrElse("")
   }
 
-  type RuleComponentResult =
+  type RuleComponentByDirectiveResult =
     (block: BlockField, component: ComponentField, node: NodeField, value: ValueField, status: StatusField, message: MessageField)
```
clarktsiory, line 548: @skaerg to answer you first question:
> simply swapped the order of the columns at the end. is this implementation okay ?

Swapping the columns order looked like the "expected behavior", so the implementation looks okay to me.
Except that I don't see the point of renaming `RuleComponentResult` to `RuleComponentByDirectiveResult`, since, by having the same "result" format, we derive two CSV formats from it.

The result here will always be the same, and there are 2 new transformers from the result to `RuleComplianceByDirectiveCsv` and `RuleComplianceByNodeCsv`, so the result should not be renamed (but the CSV format are already correctly named)

clarktsiory:
for the other concerns : 
> should I create an issue for this change ?

:arrow_right: yes it would be nice ! (we should be careful too about the lifecycle of the duplicate implementation)

> in which version should the Compliance API be deprecated ?

great question, WDYT @fanf @P4uline ?

fanf:
We can deprecate it in 9.2 and remove the public interface in 10.0 for the last API version

P4uline:
>       * in which version should the Compliance API be deprecated ?

It has to be handled on a couple of minor versions at least.

IIRC what we talked about a couple of weeks ago, the process could look like something like this:

1. Publish and announce new endpoint, and announce date of deprecation of the old endpoint 9.x
2. Deprecate endpoint, but the endpoint is actually still usable in 9.(x+1)
3. Endpoint is deleted in 9.(x+2) (or maybe 10.0 ?)

skaerg:
PR updated with a new commit

webapp/sources/api-doc/components/responses/rule-compliance-by-directive.yml
```diff
@@ -0,0 +1,278 @@
...
+        type: object
+        properties:
+          successAlreadyOK:
+            type: number
+            format: float
+            example: 48.68
+          noReport:
+            type: number
+            format: float
+            example: 36.18
```
clarktsiory: Such very long object, which is the same used everywhere in our Compliance API really deserves a dedicated schema.

You could have a separate file e.g. `api-doc/components/schemas/compliance-details.yml` with: 
```yaml
type: object
properties:
  ...  # all known compliance fields and example
```

and reuse it here with 
```suggestion
      complianceDetails:
        $ref: '../schemas/compliance-details.yml'
```

That would reduce the number of lines in this PR diff significantly, and would be very useful in the future (and could also be used in many other existing places)

skaerg:
PR updated with a new commit

webapp/sources/api-doc/components/responses/rule-compliance-by-directive-csv.yml
```diff
@@ -0,0 +1,8 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2013-2026 Normation SAS
+type: string
+format: binary
+example: >-
+  Directive, Block, Component, Node, Value, Status, Message\n
+  directive-copyGitFile, None, directive-copyGitFile-component-br6-bn4, node1.localhost, directive-copyGitFile-component-value-br6-bn4, successAlreadyOK, None\n
+  directive-copyGitFile, None, directive-copyGitFile-component-br6-bn5, node1.localhost, directive-copyGitFile-component-value-br6-bn5, successAlreadyOK, None\n
```
clarktsiory: output seems to not be the right one: 

Image


Additionally, the `application/json` content type is wrong, it looks like we return text/plain instead of application/json, so there is also something wrong with the `content:` of the parent file

webapp/sources/api-doc/paths/rules/compliance-by-directive.yml
```diff
@@ -0,0 +1,32 @@
...
+    - $ref: ../../components/parameters/rule-id.yml
+  responses:
+    "200":
+      description: Success
+      content:
+        application/json:
+          schema:
+            oneOf:
+              - $ref: ../../components/responses/rule-compliance-by-directive.yml
+              - $ref: ../../components/responses/rule-compliance-by-directive-csv.yml
```
clarktsiory: this should likely be `application/json:` with the json yaml, and `text/plain` for the csv yaml

clarktsiory (commented):
The OpenAPI doc still has some errors, is it possible to create a multi-line example text, and not a binary one ? (the older one for directives compliance is also wrong and we will to open another issue to fix it)

skaerg:
PR updated with a new commit

clarktsiory (commented):
Nice !
The doc looks all good, but it needs to be rebased

skaerg:
PR updated with a new commit

clarktsiory (approved):

PR #7145, opened 2026-05-21
Fixes #20614: VMmemory attribute should be an integer, not a string

fanf:
https://issues.rudder.io/issues/20614

So just changing the schema seems to be enought. I didn't switch to "memory" type, because it was already switched to a "long" comparator, and so the change remains minimal.


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM, it works fine

PR #7146, opened 2026-05-21
Fixes #28906: Missing context URL in result

fanf:
https://issues.rudder.io/issues/28906

We can't use `nodeLink` (and similar version) of `LinkUtil` because it relies on `S.contextPath` which is not set in a non-lift API. 
So we need to use our own version of context path passed in the API management context.


clarktsiory (approved):

fanf:
PR updated with a new commit

PR #7152, opened 2026-05-27
Fixes #28885: Refactoring with opaque types 

P4uline:
https://issues.rudder.io/issues/28885

Refactoring with opaque types made in mob programming with @fanf @P4uline @clarktsiory and @skaerg 

Things we learned:
* opaque types visibility, when not being careful about the scope, can lead to subtle errors
  * when the opaque type is NOT at top-level, or not without nested object `object OpaqueTypes { ... }`, anything at the SAME LEVEL as the opaque type will see the underlying type of the opaque type
  * so in that case, it not longer acts as an opaque type but as a type alias (which is not obvious when reading compilation errors)
* union types can be misleading, on opaque types: for example on `type ComponentName = BlockName | ValueName` then we have union of two "string" opaque type
  * then the compiler erasure phase fail when we pattern match `match { case b: BlockName => ...; case v: ValueName }` because that's both on strings

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
@@ -49,6 +49,8 @@ import com.normation.rudder.domain.policies.RuleId
 import com.normation.rudder.domain.reports.*
 import com.normation.rudder.domain.reports.ReportType.*
 import com.normation.rudder.reports.ComplianceModeName
+import com.normation.rudder.rest.data.CsvCompliance.*
+import com.normation.rudder.rest.data.CsvCompliance.CsvComplianceOpaqueTypes.*
 import com.normation.rudder.web.services.ComputePolicyMode.ComputedPolicyMode
 import com.normation.utils.Csv
 import enumeratum.*
@@ -59,6 +61,7 @@ import java.lang
 import org.apache.commons.csv.CSVFormat
 import org.apache.commons.csv.QuoteMode
 import scala.collection.immutable
+import zio.json.JsonCodec
 import zio.json.JsonEncoder
 import zio.syntax.ToZio
 
@@ -125,7 +128,7 @@ final case class ByDirectiveByRuleCompliance(
 
 final case class ByDirectiveNodeCompliance(
     id:         NodeId,
-    name:       String,
+    name:       NodeName,
     policyMode: ComputedPolicyMode,
     compliance: ComplianceLevel,
     rules:      Seq[ByDirectiveByNodeRuleCompliance]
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -230,42 +241,42 @@ final case class ByRuleNodeCompliance(
...
 final case class ByRuleByNodeByDirectiveByBlockCompliance(
-    name:           String,
+    name:           BlockName,
     reportingLogic: ReportingLogic,
     subComponents:  Seq[ByRuleByNodeByDirectiveByComponentCompliance]
 ) extends ByRuleByNodeByDirectiveByComponentCompliance with BlockCompliance[ByRuleByNodeByDirectiveByComponentCompliance] {
   override def subs: List[ByRuleByNodeByDirectiveByComponentCompliance & ComponentCompliance] = subComponents.toList
 
-  override def componentName: String = name
+  override def componentName: String = name.value
```
fanf, line 271: why isn't this typed?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -230,42 +241,42 @@ final case class ByRuleNodeCompliance(
...
 }
 
 final case class ByRuleByNodeByDirectiveByValueCompliance(
-    name:       String,
+    name:       ValueName,
     compliance: ComplianceLevel,
     values:     Seq[ComponentValueStatusReport]
 ) extends ByRuleByNodeByDirectiveByComponentCompliance {
-  override def componentName: String = name
+  override def componentName: String = name.value
```
fanf, line 279: same question here ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
P4uline: It would involve to modify a trait with a very large impact on other classes

clarktsiory (commented):
I wonder why this makes the API tests fail on JSON compliance with : 

```
-                        "value" : "e9a1a909-2490-4fc9-95c3-9d0aa01717c9-component-value-br4-bn3",
+                        "value" : "e9a1a909-2490-4fc9-95c3-9d0aa01717c9-component-value-br4-bn5",
```

, do you have an idea of why @skaerg ?

skaerg:
> I wonder why this makes the API tests fail on JSON compliance with :
> 
> ```
> -                        "value" : "e9a1a909-2490-4fc9-95c3-9d0aa01717c9-component-value-br4-bn3",
> +                        "value" : "e9a1a909-2490-4fc9-95c3-9d0aa01717c9-component-value-br4-bn5",
> ```
> 
> , do you have an idea of why @skaerg ?


this looks weird indeed. i'm not sure why, either :sweat:

fanf:
You can't understand by just looking at a part of the error. There is many more changes than just that one. 
It looks like the ordering changed, which is not good: either it was missing (json array/map need to be sorted to avoid change like that), or the sorting doesn't work anymore

image

So, before merging that PR, you need to: 
- enforce ordering in `getDirectivesCompliance`, `getRulesCompliance` etc - in an other PR if missing
- rebase that change on the PR
- see if there is still change. If there is, then understand why the ordering changed.

fanf (changes requested):
The sorting of results in compliance must be stable upon this change.

P4uline:
> The sorting of results in compliance must be stable upon this change.

I think we should do it with TDD : first add a test checking the order, then provide a fix that will fix the test.

fanf:
whatever, but it needs to be done. If the PR is not mergeable nor change addressed, put it back into draft

PR #7164, opened 2026-06-01
Fixes #28943: Add an opaque type for the composed reports of node reports

fanf:
https://issues.rudder.io/issues/28943

The main change is creating an opaque type for `Map[PolicyTypeName, AggregatedReport]`. In addition that allows to factor allow common filtering operation here. 
We also very often switch from a list of `NodeStatusReport`s to a list of proto-`RuleStatusReport`s - proto because we need to do a bit more filtering then. I factored out that into a conversion object. 

Actually, it looks like other repeated complicated operation on `ComplianceApi` should likely be in similar transormation objects, but I wanted to keep that change not too big.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -184,33 +184,6 @@ trait BlockComplianceByNode[Sub <: ComponentCompliance] extends ComponentComplia
 
 sealed trait StatusReport extends HasCompliance
 
-/**
- * Define two aggregated view of compliance: by node (for
- * a given rule) and by rules (for a given node).
- */
-final class RuleStatusReport private (
-    val forRule: RuleId,
-    val report:  AggregatedStatusReport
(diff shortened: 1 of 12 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -184,33 +184,6 @@ trait BlockComplianceByNode[Sub <: ComponentCompliance] extends ComponentComplia
...
-
-  /*
-   * Build rule status reports from node reports, deciding which directives should be "skipped"
-   */
-  def fromNodeStatusReports(ruleId: RuleId, nodeReports: Map[NodeId, NodeStatusReport]): RuleStatusReport = {
-    val toKeep = nodeReports.values.flatMap(_.reports.flatMap(_._2.reports)).filter(_.ruleId == ruleId).toList
-    RuleStatusReport(ruleId, toKeep)
-  }
-}
-
```
fanf, line 213: moved below next to its counterpart `NodeStatusReport`

clarktsiory (approved):
Nice ! LGTM

PR #7170, opened 2026-06-03
Fixes #28962: The node compliance tab is not displayed since #28943

clarktsiory:
https://issues.rudder.io/issues/28962

The `rudderBase` fallback seems very common, so standard usage of collection constructors should look like `PolicyTypes.from(...).or(fallback)` (there are already `fromTypes(head, tail)` for the non empty chunk)

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
@@ -591,12 +591,7 @@ object DirectiveStatusReport {
   def merge(directives: Iterable[DirectiveStatusReport]): Map[DirectiveId, DirectiveStatusReport] = {
     directives.groupBy(_.directiveId).map {
       case (directiveId, reports) =>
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/reports/StatusReports.scala
```diff
@@ -591,12 +591,7 @@ object DirectiveStatusReport {
   def merge(directives: Iterable[DirectiveStatusReport]): Map[DirectiveId, DirectiveStatusReport] = {
     directives.groupBy(_.directiveId).map {
       case (directiveId, reports) =>
-        val tags          = {
-          reports.flatMap(_.policyTypes.types) match {
-            case Nil          => PolicyTypes.rudderBase
-            case c @ ::(_, _) => PolicyTypes.fromCons(c)
-          }
```
clarktsiory, line 598: this was the culprit of the MatchError

clarktsiory:
Commit modified

fanf (approved):
LGTM

PR #7171, opened 2026-06-03
Fixes #28965: Property creation on root server is not peristed after rebooting

fanf:
https://issues.rudder.io/issues/28965

So, this was entertening. 

There was several problems: 

- `logIgnoredException` is really a `ZIO` effect, and it was never ran, so we didn't have a chance to see that their was ignored exception in LDAP layer; 
- said ignored exception doesn't seem to make sense anymore. Here, it would report a success (what was happengin) while actually, there was a big problem: the save wasn't done. I think they used to be useful before we had ZIO, but now, they are counter effective. I removed them. It might cause problem - I don't know if they would be bigger than not saving something and saying all is good. 
- we were triggering that ignored exception because we were trying to add both object class `ruddernode` and `rudderNode`. This is because at some point in time, when looking for parents of an object, we didn't stepped to their object classes to get their normalized name. So since key in the "all object classes" map are lower case, we had inconsistencies. 

There was also some useless `.toSeq`.


clarktsiory (approved):
LGTM

PR #5705, opened 2024-06-02
Fixes #24964: Remove unused tables about compliance and eventlogs

fanf:
https://issues.rudder.io/issues/24964

PR #7173, opened 2026-06-04
Fixes #28967: Enable exhaustive pattern matching checks in Scala 3

clarktsiory:
https://issues.rudder.io/issues/28967

Add the `-Ycheck-all-patmat` compiler flag (which is a private one, so I have no measure of the compilation performance impact, but it helped raise actual issues).

Most affected code blocks are old ones, and occur on some `String`:
* in a "select"/"radio" fields
* in a parser, but with values already checked in pre-conditions 

All bits of omitted cases were "invalid cases", that were mostly already validated (hence we had very few, if any, `MatchError` in production), **except** for the `"Archive"` choice, which was removed in [this PR](https://github.com/Normation/rudder/pull/5705) but it was no longer a valid choice.

---
N.B.:
* I couldn't see any noticeable performance penalty during compilation (the compilation time is around the same as before on my machine)
* I checked plugins, there were no such occurrence, they all compile fine

### DIFF
--- a/webapp/sources/pom.xml
+++ b/webapp/sources/pom.xml
@@ -174,6 +174,7 @@ limitations under the License.
             <arg>-Wunused:locals</arg>          <!-- Warn if a local definition is unused. -->
             <arg>-Wunused:implicits</arg>       <!-- Warn if an implicit parameter is unused. -->
             <arg>-Wunused:privates</arg>        <!-- Warn if a private member is unused. -->
+            <arg>-Ycheck-all-patmat</arg>       <!-- Check exhaustivity and redundancy of all pattern matching (meant for debugging) -->
             <arg>-Ysemanticdb</arg>
           </args>
           <jvmArgs>
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
@@ -154,6 +154,7 @@ object OrderedComparators extends ComparatorList {
 sealed trait CriterionType extends ComparatorList {
 
   // Base validation, subclass only have to define validateSubCase
+  // Since validated value is often already parsed, it could be changed from String to a dedicated type representation of the value
   def validate(value: String, compName: String): PureResult[String] = comparatorForString(compName) match {
     case Some(c) =>
       c match {
@@ -643,9 +644,10 @@ case object MachineComparator extends LDAPCriterionType {
 
   override def buildFilter(attributeName: String, comparator: CriterionComparator, value: String): Filter = {
     val v = value match {
-      // the machine can't belong to another type
+      // the machine is already validated, so it can't belong to another type, fallback will not be used
       case "Virtual"  => OC_VM
       case "Physical" => OC_PM
+      case _          => OC_MACHINE
     }
     comparator match {
       case Equals => IS(v)
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateCategoryOrGroupPopup.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateCategoryOrGroupPopup.scala
@@ -180,6 +180,8 @@ class CreateCategoryOrGroupPopup(
       </span>
       case "dynamic" =>
         <span title="Nodes will be automatically added and removed so that the list of members always matches this group's search criteria.">Dynamic</span>
+      case _         =>
+        NodeSeq.Empty
     },
     Some(5)
   ) {
@@ -201,6 +203,8 @@ class CreateCategoryOrGroupPopup(
           <span id="textGroupRadio">Group</span>
         case "Category" =>
           <span id="textCategoryRadio">Category</span>
+        case _          =>
+          NodeSeq.Empty
       },
       Some(1)
     ) {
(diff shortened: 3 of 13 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -643,9 +644,10 @@ case object MachineComparator extends LDAPCriterionType {
 
   override def buildFilter(attributeName: String, comparator: CriterionComparator, value: String): Filter = {
     val v = value match {
-      // the machine can't belong to another type
+      // the machine is already validated, so it can't belong to another type, fallback will not be used
       case "Virtual"  => OC_VM
       case "Physical" => OC_PM
+      case _          => ""
```
fanf: you still need to use something legit. We have an `OC_MACHINE` for that.

webapp/sources/utils/src/main/scala/com/normation/utils/Version.scala
```diff
@@ -217,16 +217,17 @@ object ParseVersion {
...
-      case '-' => Separator.Minus
-      case '+' => Separator.Plus
-      case ',' => Separator.Comma
-      case '.' => Separator.Dot
+      case '~' => Some(Separator.Tilde)
+      case '-' => Some(Separator.Minus)
+      case '+' => Some(Separator.Plus)
+      case ',' => Some(Separator.Comma)
+      case '.' => Some(Separator.Dot)
+      case _   => None
```
fanf: I think we want to error on an unknown separator, not just ignore it. It would be a source of strange and creative bugs like "1.2#4" being transformed into version "1.24"

fanf (changes requested):
Two small things to change, but a good addition once it's done!

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
clarktsiory: Okay ! I didn't know about `OC_MACHINE`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/queries/CmdbQuery.scala
```diff
@@ -643,9 +644,10 @@ case object MachineComparator extends LDAPCriterionType {
 
   override def buildFilter(attributeName: String, comparator: CriterionComparator, value: String): Filter = {
     val v = value match {
-      // the machine can't belong to another type
+      // the machine is already validated, so it can't belong to another type, fallback will not be used
```
clarktsiory, line 647: We would really want to pass a validated `value: "Virtual" | "Physical"`, because it is the actual type of the value

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/CreateCategoryOrGroupPopup.scala
```diff
@@ -244,72 +248,71 @@ class CreateCategoryOrGroupPopup(
...
+          logger.error("An error occurred while saving the group: " + m)
+          formTracker.addFormError(error(m))
+          onFailure & onFailureCallback()
       }
     }
+    (piItemType.get, formTracker.hasErrors) match {
+      case (_, true)           => onFailure & onFailureCallback()
+      case ("Group", false)    => onGroup
+      case ("Category", false) => onCategory
+      case (_, false)          => onFailure & onFailureCallback()
```
clarktsiory, line 314: same thing here, we would really want to represent a `WBRadioField` with an inner `type ValueType = "Group" | "Category"` to avoid the last case, which is validated in pre-conditions (and would be better off when handled in actual types)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/utils/src/main/scala/com/normation/utils/Version.scala
```diff
@@ -217,16 +218,16 @@ object ParseVersion {
...
+  val toSeparator:      PartialFunction[Char, Separator] = {
+    case '~' => Separator.Tilde
+    case '-' => Separator.Minus
+    case '+' => Separator.Plus
+    case ',' => Separator.Comma
+    case '.' => Separator.Dot
   }
-  def separators[A: P]: P[IndexedSeq[Separator]] = P(CharsWhile(separatorChar).!).map((s: String) => s.toSeq.map(toSeparator))
+  def separators[A: P]: P[List[Separator]]               = P(CharsWhile(separatorChar).!).flatMap((s: String) =>
+    s.toList.traverse(toSeparator.lift(_)).fold(Fail(s"Invalid char in ${s}"))(Pass(_))
```
fanf, line 229: can you add an invalid char in test to demo that case?

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/utils/src/main/scala/com/normation/utils/Version.scala
clarktsiory: I added a test, but essentially it it will not go into the failure branch, because the parser already matches valid separator chars (`CharsWhile(separatorChar)`)

fanf: ah yes

webapp/sources/pom.xml
```diff
@@ -174,6 +174,7 @@ limitations under the License.
             <arg>-Wunused:locals</arg>          <!-- Warn if a local definition is unused. -->
             <arg>-Wunused:implicits</arg>       <!-- Warn if an implicit parameter is unused. -->
             <arg>-Wunused:privates</arg>        <!-- Warn if a private member is unused. -->
+            <arg>-Ycheck-all-patmat</arg>       <!-- Check exhaustivity and redundancy of all pattern matching (meant for debugging) -->
```
fanf, line 177: how much does it impact compiler time?

clarktsiory: I've seen no significant impact at all, the compilation time is around the same (sometimes it is even a few seconds faster with the flag)

fanf (approved):

clarktsiory:
OK, squash merging this PR

PR #7174, opened 2026-06-04
Fixes #28948: Group page is empty with group_read permission with nodes in tenants

fanf:
https://issues.rudder.io/issues/28948

This look like the minimal change that can be done to restore non-tenant groups:
- remove tenant filtering in group repo, 
- remove associated tests

I tested that with that change, a user with a tenant get back access to all groups.


clarktsiory (approved):
LGTM !

PR #7176, opened 2026-06-04
Fixes #28976: Rule disable and enable still have no notification about change requests

clarktsiory:
https://issues.rudder.io/issues/28976

In https://github.com/Normation/rudder/pull/7122, we had a generic "saveStatus" to save a disabled/enabled rule, so the previous "Disable" message was nonsense (it is "enable"/"disable" under the same msg). 

What we should do upon the msg is [the same as for "Delete"](https://github.com/Normation/rudder/pull/7122/changes#diff-cfee51f6a0dfec5fc4bf5f733384b58567a3505ea4a2028df197a7ad44ca1e5dR473-R514): check if there is a change request ID, in which case emit the change request notification and keep the current rule for display


clarktsiory:
Commit modified

clarktsiory:
Commit modified

P4uline (approved):

PR #7178, opened 2026-06-05
Fixes #28978: The jinja2 python script is never updated

amousset:
https://issues.rudder.io/issues/28978

We only check for file existence so any content would be considered valid. Plus we use a predictable path, might as well use a real non-persisted tempfile that would prevent all concurrence/toctou issues.


amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

m4rtinh4rt (approved):

PR #7183, opened 2026-06-08
Fixes #29007: When a generation is processing, we don't know the last generation status

fanf:
https://issues.rudder.io/issues/29007

So all the business/backend was here to keep the last status, but it wasn't displayed, which was a pitty. 

I added in the header bar the last status, so that we know even without opening the details if we are generating based on a valid or error state. 

Screenshot_20260608_130238

Screenshot_20260608_130154

With short generation, it is not very nice to use, since the details menu is regenerated when status change, but the feature is not very usefull for short generation, so it doesn't seem important - it just makes screenshot harder to do :)


clarktsiory (approved):
LGTM

PR #7184, opened 2026-06-08
Fixes #28945: Add tenant management to configuration objects

fanf:
https://issues.rudder.io/issues/28945

:information_source: Disclaimer: the changes in that repo were done with the assist of claude code. Most of the systematic signature update (adding query/change context everywhere) were kept. Still, I reviewed each line, with a particular attention to the core filtering logic in ldap repos, the call at edge, and the serialization bits. Even with direction, claude was not good at:
- avoiding `systemQC`, it really likes thing that Just Compile - I think I only let legit one, but it will need care, 
- typical usage of serialization with zio-json, enumeratum, chimney. It might need a skill here
- boilerplate. It churn out methods and methods and methods for everything. Sometime it helps readability, sometimes it stinks javaisme. 

# Feats
(yes, feats, given how tedious it was)

- 1/ multi-tenants implemented for core configuration objects (directives and active techniques, rules, global properties, groups again)
- 2/ add a "read-only" kind of tenants so that user can see but not modify things, 
- 3/ API / UI kind of work (there will be likely a zillion things to adapt)

This is a big change, but most of it is just theading `given/using QueryContext/ChangeContext`.

Interesting parts are in package `com.normation.rudder.repository.TenantFiltering.scala` where is the actual filtering logic (usage of `TenantService` and `TenantCheckLogic`). 

The general design choice for LDAP respository is similar to what was done for `NodeFactRepository`, but softer:
- old LDAP repository are kept as they are. They become the persistance layer, 
- new `Ro/WoTenantXXX` repository are created. They have the tenant filtering logic and use the LDAP repo, or in memory mock for test, as a persistance layer. 

For snippet, I added a `snippetQC` in secure snippet that can be use for things out of `dispatch`. 

# Open questions / TODO

These questions are put here to give context, but they can/must be resolved in other PR. That PR is for infrastructure only (make everything works end to end, including API tests). 

## what is the core logic for tenant visibility and modification ? 

There is two parts here:  visibility when there is several tenants, and tenant update. 

## Visibility of object with several tenants

What is visibility we want when there is several tenant on an object ? The two options are: 
     - 1/ an user must have a super-set of all tenants applied (intersection == set of tenant on object). In that case, when an user create an object and set its tenants, removing tenant open up its visibility, 
     - 2/ an user must have at least one tenant (non empty intersection). In that case, adding tenant open up its visibility. 

It seems that the semantic (2) of "more tenants means more visibility" is the one expected by humans, but I fear it is unsound taken in a whole with nodes. So we need to do the math to know what is actually possible to remain consistent. 

## Entity's tenant update

It would be very nice to have tenants assigned at entity creation and then never change. The model would be easy to reason about. A change in tenant would mean a clone. 
But that would make migration and tenant evolution extremely complicated for users: 
- going from a non-tenant infra to a tenant one becomes a nightmare of directive and rule redefinition. Since it would be the kind of migration already painful, it becomes a show stopper, 
- assigning tenant during migration will be extremely error prone. So it needs to be amendable. 

But if tenant can be changed, it bears two more questions: 
- who can do it ? Admin or anybody ?
- if anybody can, then what tenant can be changed ? Typically, if an object has "A, B" tenant, and user has "A", can he remove B ? Make object open ? Etc. 

So we really need to do the math here, two, to keep a whole consistent behavior. 

## Check `open` security tag 

It doesn't seems to work when applied in directives and parent categories. Perhaps it's just a deserialisation problem, perhaps it's worse. 

## Auto set security tag value to ncf directive categorie

It seems that saving a technique when you have tenants correctly set its tenants, but it looks like the parent category (`ncf_techniques`) also get the tenants. We don't want to do that, we should not be allows to change tenant in that case.

### DIFF
--- a/adr/webapp/multi-tenants/28945-object-tenant-tag-lifecycle.md
+++ b/adr/webapp/multi-tenants/28945-object-tenant-tag-lifecycle.md
@@ -1,0 +1,40 @@
+# Object tenant tag lifecycle: creation, change, and monotonic growth
+
+* Status: accepted (monotonic-growth part: implementation pending)
+* Deciders: FAR
+* Date: 2026-07-03
+
+## Context
+
+We must define how a configuration object gets its tenant tag, who may change it, and how historical data
+(the event log) stays sound when a tag changes over time. Migration requires that pre-existing (untagged)
+objects can be assigned tenants, so tags cannot be made fully immutable from the start.
+
+## Decision
+
+**Creation.** When an object is created without an explicit tag, it inherits the creator's *writable*
+tenants (`restrictToWrite.toSecurityTag`), never the read-only ones - a read-only tenant must not leak into a
+created object. When the tenant feature is disabled, created objects are `None`. An explicitly provided tag
+is still validated: the creator must be able to write it.
+
+**Change.** Only an administrator (all-tenants grant) may change an object's tenant list. A non-admin's
+attempt to change the list is ignored and the existing tag kept. This also makes a read-modify-write
+round-trip safe: a tenant only ever reads the tenants it owns, so re-saving the object cannot accidentally
+drop the tenants it could not see.
+
+**Monotonic growth (accepted; implementation pending).** An object may only *gain* visibility, never lose it
+(`None ⊂ ByTenants(S ⊆ S') ⊂ Open`). To narrow an object's tenants, one duplicates it into a fresh object and
+chooses the narrower list at creation time. This law makes event-log filtering sound: record the object's
+*pre-change* tag alongside each event and filter event reads by `grant.canSee(T_event)`. Because tags only
+grow, `T_event ⊆ T_now`, so anyone who can see an event can currently see the object (no leak), while a
+tenant added *after* a change never sees that change's pre-membership values (no historical-value leak).
+Deleting a whole tenant is safe (it is also removed from every grant), and reusing a tenant id is an intended
+recovery path for an accidentally deleted tenant (to be documented).
+
+## Consequences
+
+* Predictable, leak-free creation and admin-only re-scoping.
+* The monotonic-growth law reverses the earlier "an admin can drop a tenant from an object" behavior and its
+  test; those must be rewritten when the law is implemented.
+* Event-log soundness additionally requires persisting the per-event tenant tag (an event-log schema change +
+  migration of existing rows to admin-only). This is deferred to its own change.
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
@@ -324,7 +324,7 @@ object JsonQueryObjects {
       value:       Option[ConfigValue],
       description: Option[String],
       inheritMode: Option[InheritMode],
-      security:    Option[SecurityTag] = None
+      security:    Option[SecurityTag]
   ) {
     def updateParameter(parameter: GlobalParameter): GlobalParameter = {
       // provider from API is force set to default.
@@ -962,7 +962,7 @@ class ZioJsonExtractor(queryParser: CmdbQueryParser & JsonQueryLexer) {
       value       <- params.parse2("value", GenericProperty.parseValue(_))
       inheritMode <- params.parse2("inheritMode", InheritMode.parseString(_))
     } yield {
-      JQGlobalParameter(params.optGet("id"), value, params.optGet("description"), inheritMode)
+      JQGlobalParameter(params.optGet("id"), value, params.optGet("description"), inheritMode, None)
     }
   }
 
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
@@ -483,7 +483,10 @@ object JsonQueryObjects {
               q.map(Some(_)),
               dynamic,
               enabled,
-              security
+              // like JQRule/JQDirective/JQGlobalParameter: if the request does not carry a security
+              // tag, keep the existing one (do not wipe it). The actual tenant-change authorization
+              // is enforced later by TenantCheckLogic.manageUpdate.
+              security.orElse(group.security)
             )
           )
       }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
@@ -76,6 +76,7 @@ import com.normation.rudder.services.queries.StringCriterionLine
 import com.normation.rudder.services.queries.StringQuery
 import com.normation.rudder.services.servers.AllowedNetwork
 import com.normation.rudder.services.servers.DeleteMode
+import com.normation.rudder.tenants.SecurityTag
 import com.typesafe.config.ConfigValue
 import enumeratum.Enum
 import enumeratum.EnumEntry
@@ -234,7 +235,8 @@ object JsonQueryObjects {
       policyMode:       Option[Option[PolicyMode]] = None,
       tags:             Option[Tags] = None, // for clone
 
-      source: Option[DirectiveId] = None
+      source:   Option[DirectiveId] = None,
+      security: Option[SecurityTag] = None
   ) {
     val onlyName: Boolean = displayName.isDefined &&
       shortDescription.isEmpty &&
@@ -248,20 +250,22 @@ object JsonQueryObjects {
       tags.isEmpty // no need to check source or reason
 
     def updateDirective(directive: Directive): Directive = {
-      directive.patchUsing(
-        PatchDirective(
-          id,
-          techniqueVersion,
-          parameters.flatMap(_.get("section").map(s => SectionVal.toMapVariables(s.toSectionVal._2))),
-          displayName,
-          shortDescription,
-          policyMode,
-          longDescription,
-          priority,
-          enabled,
-          tags
+      directive
+        .patchUsing(
+          PatchDirective(
+            id,
+            techniqueVersion,
+            parameters.flatMap(_.get("section").map(s => SectionVal.toMapVariables(s.toSectionVal._2))),
+            displayName,
+            shortDescription,
+            policyMode,
+            longDescription,
+            priority,
+            enabled,
+            tags
+          )
         )
-      )
+        .copy(security = security.orElse(directive.security))
     }
   }
 
@@ -276,7 +280,8 @@ object JsonQueryObjects {
       enabled:                             Option[Boolean] = None,
       tags:                                Option[Tags] = None, // for clone
 
-      source: Option[RuleId] = None
+      source:   Option[RuleId] = None,
+      security: Option[SecurityTag] = None
   ) {
 
     val onlyName: Boolean = displayName.isDefined &&
@@ -308,7 +313,8 @@ object JsonQueryObjects {
         directiveIds = updateDirectives,
         targets = updateTargets,
         isEnabledStatus = updateEnabled,
-        tags = updateTags
+        tags = updateTags,
+        security = security.orElse(rule.security)
       )
     }
   }
@@ -317,11 +323,17 @@ object JsonQueryObjects {
       id:          Option[String],
       value:       Option[ConfigValue],
       description: Option[String],
-      inheritMode: Option[InheritMode]
+      inheritMode: Option[InheritMode],
+      security:    Option[SecurityTag] = None
   ) {
     def updateParameter(parameter: GlobalParameter): GlobalParameter = {
       // provider from API is force set to default.
-      parameter.patch(PatchProperty(id, value, Some(PropertyProvider.defaultPropertyProvider), description, inheritMode))
+      val patched =
+        parameter.patch(PatchProperty(id, value, Some(PropertyProvider.defaultPropertyProvider), description, inheritMode))
+      security match {
+        case None      => patched
+        case Some(tag) => patched.withSecurity(Some(tag))
+      }
     }
   }
 
@@ -368,7 +380,8 @@ object JsonQueryObjects {
       properties:  Option[List[GroupProperty]],
       query:       Option[Option[Query]],
       isDynamic:   Option[Boolean],
-      _isEnabled:  Option[Boolean]
+      _isEnabled:  Option[Boolean],
+      security:    Option[SecurityTag]
   )
 
   final case class JQGroupCategory private[apidata] (
@@ -433,7 +446,8 @@ object JsonQueryObjects {
       dynamic:                             Option[Boolean] = None,
       enabled:                             Option[Boolean] = None,
       @jsonAliases("category") categoryId: Option[NodeGroupCategoryId] = None,
-      source:                              Option[NodeGroupId] = None
+      source:                              Option[NodeGroupId] = None,
+      security:                            Option[SecurityTag] = None
   ) {
 
     val onlyName: Boolean = displayName.isDefined &&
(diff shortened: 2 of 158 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
```diff
@@ -317,11 +323,17 @@ object JsonQueryObjects {
       id:          Option[String],
       value:       Option[ConfigValue],
       description: Option[String],
-      inheritMode: Option[InheritMode]
+      inheritMode: Option[InheritMode],
+      security:    Option[SecurityTag] = None
```
clarktsiory: Why does the `security` has a default `= None` here, where other fields do not ?

webapp/sources/rudder/rudder-web/src/main/scala/net/liftweb/http/SecureDispatchSnippet.scala
```diff
@@ -56,11 +56,21 @@ trait SecureDispatchSnippet extends DispatchSnippet {
     CurrentUser.queryContext.withQCOr(loggedInsecureDispatch)(secureDispatch)
   }
 
+  private def logInvaliAccess: Unit = ApplicationLoggerPure.Auth.logEffect.warn(
```
clarktsiory:
```suggestion
  private def logInvalidAccess: Unit = ApplicationLoggerPure.Auth.logEffect.warn(
```

webapp/sources/rudder/rudder-web/src/main/scala/net/liftweb/http/SecureDispatchSnippet.scala
```diff
@@ -56,11 +56,21 @@ trait SecureDispatchSnippet extends DispatchSnippet {
...
+    s"Snippet '${this.getClass.getName}' can't be accessed in current security context (user is not authenticated)"
+  )
+
   private def loggedInsecureDispatch: DispatchIt = {
-    ApplicationLoggerPure.Auth.logEffect.warn(
-      s"Snippet '${this.getClass.getName}' can't be accessed in current security context (user is not authenticated)"
-    )
+    logInvaliAccess
```
clarktsiory:
```suggestion
    logInvalidAccess
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/SecurityContext.scala
```diff
@@ -229,6 +282,9 @@ object QueryContext {
   // for system queries (when rudder needs to look-up things)
   implicit val systemQC: QueryContext = QueryContext(eventlog.RudderEventActor, TenantAccessGrant.All)
 
+  // for no right queries
+  implicit val noneQC: QueryContext = QueryContext(EventActor("none"), TenantAccessGrant.None)
```
clarktsiory, line 304: this is exactly what led to an "unknown" user being logged (see [this PR](https://github.com/Normation/rudder/pull/6958/changes#diff-bfb5f55c5a131de1d92348d9de2b4cc53c339c05862bc25e3c2d42ab68a7895cL168-L169))

clarktsiory: the way we found in that PR to avoid this anti-pattern is to "always" propagate the implicit query context, down to where it is needed.

At some point, by turning things such as:
```scala
val rules:        List[Rule]   = roRuleRepo.getAll(false)(using snippetQC).toBox.getOrElse(Seq()).toList
```
into
```scala
def rules(using qc: QueryContext):        List[Rule]   = roRuleRepo.getAll(false)(using qc).toBox.getOrElse(Seq()).toList
```
, there is a crossing point where we can see the `QueryContext` already provided by the `SecureDispatchSnippet`:

```scala
override def secureDispatch: QueryContext ?=> PartialFunction[String, NodeSeq => NodeSeq] = {
    case "getRules"              => rules  // it automatically uses the provided QueryContext
```

clarktsiory (commented):
I reviewed the core changes in the model of RW/RO tenants (not the ldap yet and I haven't tested).

I have some comments for now about the anti-patttern when declaring QC in snippets: we shouldn't create any more "implicit"/"given" QC: they are already all provided in `SecureDispatchSnippet`, and we just need to "use" it ([this ADR](https://docs.rudder.io/devel/adr/webapp/28452-avoid-currentuser-for-querycontext-in-lift-snippets.html) states that decision). I provided an example that I hope explains the alternative

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/SecurityContext.scala
fanf: No, there's a lot of cases where the query context is used in stateful class fields, and is needed at class instanciation - which is the pattern I exactly asked for in the previous pr. So, without rewriting half of snippets : how do we handle that ?

fanf: In addition to the fact that solution is a big rewrite on its own, this solution is not ok (and so the ADR not enought) for at least two other aspects: 
- changing a `val` to a `def` means that - of course - you change the call semantic. It means that you introduce a new call each time, which can have 1/ performance impact, 2/ atomicity/scoping impact, 
- in the case of a statefull snippet, for ex. for form management, well lift actually needs `val`, it can't be a def. 

So the ADR needs to address the real quetion of what is the lifecycle of the session and snippet, and how and when the current user is correctly defined given that combined lifecycles.

clarktsiory: Indeed we didn't address the case of stateful snippets specifically.

The problem I'm pointing is especially that, when introducing a widely usable `implicit` with wrong values, we will tend to forget that it's still an exception, and it broadens the possibilities to misuse it.

In that case, and from the context of ZIO VS Lift state management mentioned in the ADR, a better option then would be simply to inline the optional query context (as in https://github.com/Normation/rudder/blob/branches/rudder/9.1/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/NodeGroupForm.scala#L127): 
* the `Option` forces us to use a specific fallback value of the state (e.g. empty list of rules)
* this makes it more traceable to know which places are "stateful", i.e. need an initial query context to fetch state

fanf: Well, I don't see the point. Either having the implicit in `SecureSnippet` works, or it doesn't. If it works, then we can use it. If it doesn't, then we need to provide a framework that is always ok, and that doesn't put responsibility on the user, who won't know what to do in exceptional case. 
Plus, the inline version is plainly horrible, so it won't be used and it will be workedaround. It won't be sustained. The least effort choice needs to be the correct one.

fanf: We really need to understand the why. Why sometimes `CurrentUser` is undefined ? This is a terrible bug that need to be corrected, not worked around. We need to principle mental model of what is happening.

clarktsiory: Okay we need a framework, which I would ideally describe like so:
* it should be similar to having an `Option`, propagated down to the edge where we display values, or an error in the snippet
  * when we know we have access, just use the value
  * when we don't, if possible have an error to display, or use a fallback empty value

This may also be about throwing a custom exception (which may be even less intuitive to use), but I'm not sure we would like to do that

But my initial point is: it should be based on something else than `implicit val user = "none"`, because that one was the initial security issue

fanf: it's not a security issue. It restricts user right *more* than it should.

clarktsiory: > Why sometimes CurrentUser is undefined 

This is explained in the ADR: 

> a variable lookup would sometimes return no user, even if there is actually an authenticated user (when used with the ZIO framework in latest versions, see [the corresponding issue](https://issues.rudder.io/issues/26605)).


We tried once searching for a workaround, IIRC we gained no insight at all at

clarktsiory: It [was](https://issues.rudder.io/issues/28404) indeed a security issue: we couldn't audit which user did some action in the UI because we have a `qc` (implicit and usable everywhere), on which snippets could simply do `qc.actor` and log an action with an `none`/`unknown` actor.

I agree we should find a better solution for that, but the technical aspects of mangling with ZIO state VS Lift state was not a trivial one, we may as well provide "type system"/"error handling" guards

fanf: it's not the same case. Here it's access to resource, not change

fanf: So, given https://issues.rudder.io/issues/29034#note-2, I think the `snippetQC` is not a problem once the filter is changed, if we are sure it's evaluated out of the zio effect - either by being a val (on the trait, it will be run before constructor of children). 

We can even make it fail with a bigger error message than what is currently printed if we want.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/test/resources/api-tenant/api_tenant_groups.yml
```diff
@@ -0,0 +1,426 @@
...
+user: zoneB
+method: DELETE
+url: /api/latest/groups/tenant-group-1
+response:
+  code: 500
+  content: >-
+    {
+      "action": "deleteGroup",
+      "result": "error",
+      "errorDetails": "Error when deleting target GroupTarget(NodeGroupId(NodeGroupUid(tenant-group-1),Revision(default))). All dependent rules where updated (); cause was: Unexpected: Inconsistency: Object 'tenant-group-1' can't be deleted by zoneB"
```
clarktsiory: it looks like an unserialized group target

webapp/sources/rudder/rudder-web/src/main/elm/sources/Groups/DataTypes.elm
```diff
@@ -40,6 +40,10 @@ type alias ComplianceSummaryValue =
   , complianceDetails : ComplianceDetails
   }
 
+type SecurityTag
+  = Open
+  | ByTenants (List String)
+
```
clarktsiory: duplication of the model and its decoders + the policy mode badge across multiple Elm apps could be avoided, by putting it all under a `elm/sources/Tenants/SecurityTag.elm`

webapp/sources/rudder/rudder-rest/src/test/resources/api-tenant/api_tenant_techniques.yml
```diff
@@ -0,0 +1,253 @@
...
+              ]
+            }
+          }
+        ]
+      }
+    }
+---
+description: zoneC user sees no active technique (no object has zoneC tenant)
+user: zoneC
+method: GET
```
clarktsiory, line 227: Updating a technique would also have tenants write restrictions (there seems to be missing tests to DELETE/POST/PUT cases)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
```diff
@@ -244,50 +316,50 @@ class DefaultTenantCheckLogic extends TenantCheckLogic {
             // creation when feature disabled: set securityTag to "none".
             case TenantStatus.Disabled         =>
               action(updated.updateSecurityContext(None))
-            // in the case of creation, we force the user tenant to its tenant
+            // in the case of creation, we force the user tenant to its (writable) tenant
             case TenantStatus.Enabled(tenants) =>
-              if (cc.accessGrant.canSee(updated)) {
+              if (writeGrant.canSee(updated)) {
```
clarktsiory: there is no check against existing `tenants` at creation (there is indeed one during update with a check of tenant `ts.forall(t => tenants.contains(t))`) i.e. admins cannot create/update objects with unknown tenants, but other users can.
That seems to be a security risk like: user creates a new object with `zone-all`which doesn't exist (yet)...

I think this should be validated too, I didn't see test case covering that scenario

clarktsiory (commented):
On backend side there is only a main issue about check of existing tenants (and some tests that could be added).

On frontend side code factorisation into dedicated module would be a best practice to avoid repetition

webapp/sources/rudder/rudder-web/src/main/elm/sources/Groups/DataTypes.elm
clarktsiory: this seems to have been removed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
clarktsiory: I think this is still relevant

adr/webapp/multi-tenants/28945-object-tenant-tag-lifecycle.md
clarktsiory, line 1: @amousset would the extra 'multi-tenants' directory be compatible for the ADR doc build ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
fanf: Right, I think it's safer as you say.

webapp/sources/rudder/rudder-web/src/main/elm/sources/Groups/DataTypes.elm
fanf: :scream:

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/apidata/JsonQueryObjects.scala
fanf: I've just done the same thing everywhere, and I'm not sure why some objects do have default param None and other not. It will be for another day, let's keep consistency by objects.

webapp/sources/rudder/rudder-rest/src/test/resources/api-tenant/api_tenant_groups.yml
fanf: it was an oooold logger

webapp/sources/rudder/rudder-web/src/main/elm/sources/Groups/DataTypes.elm
fanf: I brought them back

webapp/sources/rudder/rudder-rest/src/test/resources/api-tenant/api_tenant_techniques.yml
fanf: We can't for now, because the technique file format didn't change yet, only the "active techniques", ie once the technique is loaded in LDAP. 

That magically works for techniques from the editor, because IIRC the change context is correctly passed in the rest end points and below, down to the creation of the active technique, but we can't for ex have an archive.

And we can't really test that in current test framework, the bridge between technique API and LDAP repo is complicated (and not done - ncfTechniqueWriter, fakeUpdatePTLibService just return success)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
fanf: :heavy_check_mark:  done

clarktsiory (commented):
Looks good to me, it would just need a rebase
As discussed there are further changes to be addressed e.g. for techniques

fanf:
OK, merging this PR

PR #7188, opened 2026-06-10
Fixes #29034: Correct CurrentUser initialisation problem

fanf:
https://issues.rudder.io/issues/29034

The description of the problem is available here: https://issues.rudder.io/issues/29034#note-2

I didn't see any occurence of `f(...)(using CurrentUser.queryContext).[toBox, runNow, etc]` so we should be safe on that part. 

I tested the other case by first changing the `doFilter` on an other thread pool with a `IOResult.attempt`, and yes, the security context is lost; and it remains with the filter on the same thread. So that should makes everything work up to the lift `RequestVar` in all cases. 

I didn't see any other things on that path that are boxed into a ZIO effect after authentication filter (which we know works). I looked at anything defined in `web.xml` plus other filters. 

I've also update the corresponding ADR to make the root cause clear and better explain what is forbidden. 
The decision doesn't change, so I don't think it needs another ADR.

### DIFF
--- a/adr/webapp/28452-avoid-currentuser-for-querycontext-in-lift-snippets.md
+++ b/adr/webapp/28452-avoid-currentuser-for-querycontext-in-lift-snippets.md
@@ -3,7 +3,7 @@
 * Status: accepted
 * Deciders: CAN, PIO, FAR
 * Date: 2025-03-01
-
+* Updated for details about root cause explanation: 2026-06-10
 
 ## Context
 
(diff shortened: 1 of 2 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

adr/webapp/28452-avoid-currentuser-for-querycontext-in-lift-snippets.md
```diff
@@ -38,6 +42,22 @@ trait SecureDispatchSnippet extends DispatchSnippet {
 }
 ```
 
+For snippet `val` instantiated by snipped constructor (for example for stateful snippets) via a `f(x)(using qc:QueryContext)` function, 
+the call to `CurrentUser.queryContext` MUST be done in another expression than the one passed to `runNow`: 
+
+```scala 3
+given qc: QueryContext = CurrentUser.queryContext
```
clarktsiory: That snippet is inherently wrong since in the `CurrentUser` we have:

```scala
  // should generally not be used, if you need a `QueryContext` in snippets, you should inherit `SecureDispatchSnippet`
  def queryContext: Option[QueryContext] = getCurrentUser.map(_.qc)
```


Also, in the previous ADR, the last bullet point in "Consequences" is:
> Prevent using the CurrentUser#queryContext in other potentially unsafe places.


Is this a new ADR ?
And is there a proposition to change the current `CurrentUser.queryContext` implementation ?

fanf: The snippet will be corrected. 
I don't think the consequences changed, `Prevent using the CurrentUser#queryContext in other potentially unsafe places.` is still right and it means "not using CurrentUser in the context of a thread switch specifically, ie with a runNow". 
Perhaps I just don't get what you mean ?

fanf:
PR updated with a new commit

adr/webapp/28452-avoid-currentuser-for-querycontext-in-lift-snippets.md
clarktsiory: okay, we seem to agree on what is corresponds to in the current ADR, and that it would be further corrected

clarktsiory (approved):
LGTM !

PR #7195, opened 2026-06-11
Fixes #28984: Custom roles with underscore are not allowed but error is misleading 

fanf:
https://issues.rudder.io/issues/28984


clarktsiory (approved):
LGTM !

PR #7198, opened 2026-06-11
Fixes #29012: The reason why Trigger agent inventory and Trigger agent buttons are greyed out on windows agents is not explained in the webapp

skaerg:
https://issues.rudder.io/issues/29012

This PR aims to add a tooltip on the "Trigger agent" and "Trigger agent inventory" buttons for windows nodes which explain why the buttons are disabled

image

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ReportDisplayer.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ReportDisplayer.scala
@@ -442,7 +442,10 @@ class ReportDisplayer(
                           <div id={"triggerAgent" + defaultOrInventory} class="mb-3">
             <button id={
                             "triggerBtn" + defaultOrInventory
-                          } class="btn btn-primary btn-trigger" disabled="disabled" title="This action is not supported for Windows node">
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/services/ReportDisplayer.scala
```diff
@@ -442,7 +442,7 @@ class ReportDisplayer(
                           <div id={"triggerAgent" + defaultOrInventory} class="mb-3">
             <button id={
                             "triggerBtn" + defaultOrInventory
-                          } class="btn btn-primary btn-trigger" disabled="disabled" title="This action is not supported for Windows node">
+                          } class="btn btn-primary btn-trigger" data-bs-toggle="tooltip" title="This action is not supported for Windows nodes" disabled="disabled" style="pointer-events: auto !important;">
```
clarktsiory: The `style` is banned from our frontend PR checklist in the case that [the equivalent Bootstrap class exists](https://getbootstrap.com/docs/5.3/utilities/interactions/#pointer-events): could you use that one instead ?

skaerg:
Commit modified

clarktsiory (approved):

PR #7199, opened 2026-06-11
Fixes #29051: Only compile the Rust regexes once

amousset:
We were quite inconsistent. Let's harmonize.

https://issues.rudder.io/issues/29051


amousset:
PR updated with a new commit

PR #7213, opened 2026-06-16
Fixes #29066: Add a script to help secure relays rsync command

peckpeck:
https://issues.rudder.io/issues/29066

### DIFF
--- a/relay/sources/Makefile
+++ b/relay/sources/Makefile
@@ -113,6 +113,8 @@ install: build
 	install -m 755 ../../target/release/rudder-package $(DESTDIR)/opt/rudder/bin/rudder-package
 	install -m 644 ../../target/rudder.bash $(DESTDIR)/etc/bash_completion.d/rudder-package.sh
 	install -m 644 rudder-package/tools/rudder_plugins_key.asc $(DESTDIR)/opt/rudder/etc/rudder-pkg/rudder_plugins_key.asc
+	# rsync command filter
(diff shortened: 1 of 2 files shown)


relay/sources/Makefile
```diff
@@ -113,6 +113,8 @@ install: build
 	install -m 755 ../../target/release/rudder-package $(DESTDIR)/opt/rudder/bin/rudder-package
 	install -m 644 ../../target/rudder.bash $(DESTDIR)/etc/bash_completion.d/rudder-package.sh
 	install -m 644 rudder-package/tools/rudder_plugins_key.asc $(DESTDIR)/opt/rudder/etc/rudder-pkg/rudder_plugins_key.asc
+	# rsync command filter
+	install -m 755 rsync-command-filter $(DESTDIR)/opt/rudder/bin/rsync-command-filter
```
amousset, line 117: I'm not a fan of filling `/opt/rudder/bin` with random scripts

amousset: If this is supposed to be used by the system technique why not put it there?

peckpeck: Where should i put it then ?

amousset: in the technique that uses it?

peckpeck: There is no technique that use it, when a user wants to use rsync to synchronize relay data, he can use the given script to restrict rsync usage in ssh authorized_keys

PR #7216, opened 2026-06-18
Fixes #28958: Titles in main dashboard Score breakdown should be centered for each donut graph

skaerg:
https://issues.rudder.io/issues/28958

image


skaerg:
do you think the titles should be centered above the donut graph (instead of both the donut graph and the legend on the right side as is the case right now) @RaphaelGauthier ?

i.e. should it be like this 
image

or like that ? 
image

or is there another better option ?

RaphaelGauthier:
I think the titles should be centered just above the donut graph, not the legend. It might look weird if the legend text is long! This can be done easily by setting the width of the h4 tags to the same as that of the doughnuts:
```
.node-chart h4 {
width : $doughnut-size;
...
}
```

skaerg:
PR updated with a new commit

RaphaelGauthier (approved):

skaerg:
OK, squash merging this PR

PR #7217, opened 2026-06-18
Fixes #29075: Error when changing expiration date of an expired API token

clarktsiory:
https://issues.rudder.io/issues/29075

When an old API account, without the attribute for "last used", is modified on other fields (e.g. expiration), the schema version can change: it may receive a new version v2 (added in this PR: https://github.com/Normation/rudder/pull/6852).

So it's mainly about intercepting the event, and produce the correct diff on the "last used" date.
Also, event logs for that are not displayed correctly, so just add the missing bits for display of event logs

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDiffMapper.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDiffMapper.scala
@@ -697,8 +697,16 @@ class LDAPDiffMapper(
                                 diff.map(
                                   _.copy(modAPITenants = Some(SimpleDiff(oldAccount.tenants.serialize, mod.getOptValueDefault("-"))))
                                 )
-
-                              case x => Left(Err.UnexpectedObject("Unknown diff attribute: " + x))
+                              case A_API_SCHEMA_VERSION          =>
+                                val newVersion = ApiAccountSchemaVersion(mod.getOptValue())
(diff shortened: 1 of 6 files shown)


clarktsiory:
Commit modified

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDiffMapper.scala
```diff
@@ -697,8 +697,16 @@ class LDAPDiffMapper(
...
+                              case A_API_SCHEMA_VERSION          =>
+                                val newVersion = ApiAccountSchemaVersion(mod.getOptValue())
+                                val oldDate    = oldAccount.lastAuthentication.date
+                                val newDate    = AccountLastAuthentication.from(newVersion, oldDate)
+                                // keep old date, if version changed having impact on date, then we have a diff
+                                val lastDiff   = if (oldDate == newDate) None else Some(SimpleDiff(oldDate, newDate))
+                                diff.map(
+                                  _.copy(modLastAuthentication = lastDiff)
+                                )
+                              case x                             => Left(Err.UnexpectedObject("Unknown diff attribute: " + x))
```
clarktsiory, line 709: This case was the one causing the error

fanf (approved):
LGTM, looks like all the bit for a diff displayed are here

clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

PR #7219, opened 2026-06-18
Fixes #29083: Cannot import archive

fanf:
https://issues.rudder.io/issues/29083

So, it was "just" that we forgot that we can import old fashion techniques and so it happens that we save these one. 
The API tests didn't catch it because they don't go so low level (the save is a mock not actually calling rudderc)


clarktsiory (approved):
LGTM

PR #7221, opened 2026-06-18
Fixes #28895: Update a directive while tree filter is not empty breaks the filtering

skaerg:
https://issues.rudder.io/issues/28895

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Filters.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Filters.elm
@@ -91,4 +91,12 @@ update msg model =
         encodedFilters = encodeFilters newModel
       in
       (newModel, searchTree encodedFilters)
-      
+
+    ResetFilters ->
+      let
+        (prevModel, updateTags) =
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-web/src/main/elm/sources/Filters.elm
```diff
@@ -91,4 +91,12 @@ update msg model =
...
\ No newline at end of file
+
+    ResetFilters ->
+      let
+        (model1, updateTags) =
+          update (UpdateTags Tags.DataTypes.Remove []) model
+        (model2, updateFilter) =
+          update (UpdateFilter "") model1
+      in
+        (model2, Cmd.batch [updateTags, updateFilter])
```
clarktsiory: 1 and 2 are really the naming convention that's confusing me the most 😅
How about `prevModel` and `newModel` ?

clarktsiory: It could also be avoided by using `Tuple.mapBoth` I think

skaerg:
PR updated with a new commit

clarktsiory (approved):

PR #7223, opened 2026-06-19
Fixes #29025: Implement TOTP for Rudder users

clarktsiory:
https://issues.rudder.io/issues/29025

OTP implementation:
* add the config `rudder.auth.otp.enforce=false` by default
* adding new `userstotp` table with its migration
* create a new `/secure/otp.html` Elm page (inspired from login page, which needed some restructuration) and:
  * a dedicated API to get the OTP status (if enrollement is needed)
  * a dedicated Spring filter `OtpAuthenticationFilter` for processing the verification request
  * all that is handled with a new special Spring role: `ROLE_PRE_AUTH`, intercepting the authentication until the code is verified (in which case `ROLE_USER` is put back in authenticated principal)  
* add the OTP status into user management page, and allow to reset a user's OTP from there

Some screens:
* Enrollement + login:
image
image

* OTP login:
image

* User management:
image
image
image

---
Next steps:
* add unit tests
* some other security enforcement, like "rejecting 3 bad OTP code", add a timeout for filling the OTP page, especially during enrollement where secret is displayed

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
@@ -75,6 +75,14 @@ CREATE TABLE Users (
 , otherInfo      jsonb -- general additional user info
 );
 
+
+CREATE TABLE userstotp (
+  user_id TEXT PRIMARY KEY REFERENCES users(id) ON DELETE CASCADE
+, secret TEXT NOT NULL
+, created_at TIMESTAMPTZ NOT NULL DEFAULT now()
+);
+
+
 CREATE TABLE UserSessions (
   userId       text NOT NULL CHECK (userId <> '')
 , sessionId    text NOT NULL CHECK (sessionId <> '')
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
@@ -1,0 +1,69 @@
+/*
+ * *************************************************************************************
+ * Copyright 2026 Normation SAS
+ * *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * *************************************************************************************
+ */
+package com.normation.rudder.users
+
+import java.time.Instant
+
+opaque type UserId = String
+object UserId {
+  def apply(s: String): UserId = s
+
+  extension (self: UserId) {
+    def value: String = self
+  }
+}
+
+// Need to be a class because opaque type doesn't allow toString override
+final class TotpSecret private (secret: String) extends AnyVal {
+  // Avoid printing the value
+  override def toString: String = "[REDACTED TotpSecret]"
+
+  def exposeSecret(): String = {
+    secret
+  }
+}
+object TotpSecret {
+  def apply(secret: String) = new TotpSecret(secret)
+}
+
+/**
+ * Represents a TOTP (Time-based One-Time Password) configuration for a user.
+ * Stores the secret and creation timestamp.
+ */
+case class Totp(
+    secret:  TotpSecret,
+    created: Instant
+)
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/TotpRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/TotpRepository.scala
@@ -1,0 +1,98 @@
+/*
+ * *************************************************************************************
+ * Copyright 2026 Normation SAS
+ * *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * *************************************************************************************
+ */
+package com.normation.rudder.users
+
+import com.normation.errors.IOResult
+import com.normation.rudder.db.Doobie
+import doobie.*
+import doobie.implicits.*
+import doobie.postgres.implicits.*
+import doobie.util.*
+import zio.interop.catz.*
+
+trait TotpRepository {
+  def getByUserId(userId: UserId): IOResult[Option[Totp]]
+  def create(userId:      UserId, totp: Totp): IOResult[Unit]
+  def delete(userId:      UserId): IOResult[Unit]
+  def getEnabledUsers(): IOResult[Set[UserId]]
+}
+
+object JdbcTotpRepository {
+
+  given Meta[UserId]     = Meta[String].imap(UserId(_))(_.value)
+  given Meta[TotpSecret] = Meta[String].imap(TotpSecret(_))(_.exposeSecret())
+  given Read[Totp]       = Read.derived
+  given Write[Totp]      = Write.derived
+
+  def getByUserIdSQL(userId: UserId): Query0[Totp] =
+    sql"SELECT secret, created_at FROM userstotp WHERE user_id = ${userId.value}".query[Totp]
+
+  def upsertQuerySQL(userId: UserId, totp: Totp): Update0 =
+    sql"INSERT INTO userstotp (user_id, secret, created_at) VALUES (${userId.value}, ${totp}) ON CONFLICT (user_id) DO UPDATE SET secret=EXCLUDED.secret, created_at=EXCLUDED.created_at".update
+
+  def deleteQuerySQL(userId: UserId): Update0 =
+    sql"DELETE FROM userstotp WHERE user_id=${userId.value}".update
+
+  def getAllByUserSQL(): Query0[String] =
+    sql"SELECT user_id FROM userstotp".query[String]
+}
+
+class JdbcTotpRepository(doobie: Doobie) extends TotpRepository {
+  import JdbcTotpRepository.*
+  import doobie.*
+
+  def getByUserId(userId: UserId): IOResult[Option[Totp]] = {
+    transactIOResult(s"Error when getting TOTP for user ${userId.value}")(xa => getByUserIdSQL(userId).option.transact(xa))
+  }
+
+  def create(userId: UserId, totp: Totp): IOResult[Unit] = {
+    transactIOResult(s"Error when creating TOTP for user ${userId.value}")(
+      upsertQuerySQL(userId, totp).run.transact(_).unit
+    )
+  }
+
+  def delete(userId: UserId): IOResult[Unit] = {
+    transactIOResult(s"Error deleting TOTP for user ${userId.value}")(
+      deleteQuerySQL(userId).run.transact(_).unit
+    )
+  }
+
+  def getEnabledUsers(): IOResult[Set[UserId]] = {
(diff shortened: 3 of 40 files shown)


clarktsiory:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/TotpService.scala
```diff
@@ -0,0 +1,225 @@
...
+
+trait TotpVerificationService {
+  def verify(userId: String, code: String): IOResult[Unit]
+}
+
+/**
+ * Status of OTP enrollment for a user
+ */
+case class TotpStatus(
+    needEnrollment: Boolean
```
fanf, line 93: At first read it's unclear if "need enrolment" means that totp is already enabled (user need to enrol at each connection) or if it's yet to be enrolled in the service.

Prefer to use an enum for the states. Moreover it's more extendible, for ex if one day wd have needs like "the user need to regenerate an OTP" or whatever we would need for migration

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/OtpApi.scala
```diff
@@ -0,0 +1,174 @@
...
+            .orElse(
+              otpGeneratorService.verify(userId, code) *>
+              ApplicationLoggerPure.Auth.debug(s"User '${userId}' submitted a valid TOTP code")
+            )
+            .chainError(
+              "Could not verify user OTP, pleasy retry with a valid code, generate a new OTP, or ask admin to reset your OTP if you lost it"
+            )
+
+        // FIXME: Weird place to change auth, but it's actually valid since it's called from logged in user
+        _     = {
```
fanf: Yes, it's not done like that.

Well, you're right it's a weird place, and it breaks the design rule that once you're in the API part, authentification must not change. 

It will breaks the security model, because it means that you have a not fully authenticated user reaching that point, which certainly authorize async requests or other cases.
Plus, it forces to couple spring security with hierarchy of the code that are independent from it. 
Authentication change happens in spring security filters, else it does not know what to do. 

So you need a custom filter after the form login that will : 
- check if OTP second factor is required,
- check if the user already did the challenge and gas correct authority (so the kind of logic below with authorities)
- if the ito code token isin the request, do the check you did here, 
- else if it has authority, let the request proceed,
- if not, redirect OTP page

The filter must be after form, and before intercept filter I believe, so something like :  

```
<security:http>
    <security:custom-filter ref="otpCheckFilter" before="FILTER_SECURITY_INTERCEPTOR" />
</security:http>
<bean id="otpGateCheckFilter" class="com.normation.....OtpCheckFilter"/>
```

And I think the ito page just set a code parameter filled by the user, like the form page just set a user and password, 

```
  <security:custom-filter ref="otpAuthenticationFilter" after="FORM_LOGIN_FILTER" />


public class OtpAuthenticationFilter extends AbstractAuthenticationProcessingFilter {


    public OtpAuthenticationFilter() {
        super(new AntPathRequestMatcher("/login/otp/verify", "POST"));
    }

    @Override
    public Authentication attemptAuthentication(HttpServletRequest req, HttpServletResponse res) {
        Authentication firstFactor = SecurityContextHolder.getContext().getAuthentication();
        String code = req.getParameter("code");
        return getAuthenticationManager().authenticate(new OtpAuthenticationToken(firstFactor, code));
    }
}```

**Review by fanf — DISMISSED** [review]
You're right, you can't do it like that :)

**Comment by clarktsiory** [comment]
PR rebased

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/OtpApi.scala (hunk 2 above)
Comment (clarktsiory) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/OtpApi.scala]: I added the corresponding filter, finding the `AbstractAuthenticationProcessingFilter` was not so obvious, but it indeed appears to be the right tool for the job !

**Comment by clarktsiory** [comment]
PR updated with a new commit

**Comment by clarktsiory** [comment]
PR updated with a new commit

**Comment by clarktsiory** [comment]
PR updated with a new commit

**Comment by clarktsiory** [comment]
PR updated with a new commit

**Comment by clarktsiory** [comment]
PR updated with a new commit

**Comment by clarktsiory** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql [hunk 3]
```diff
@@ -75,6 +75,14 @@ CREATE TABLE Users (
 , otherInfo      jsonb -- general additional user info
 );
 
+
+CREATE TABLE userstotp (
```
amousset, line 79: it looks like we use camelcase for the db and field names on other tables

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/TotpService.scala
```diff
@@ -0,0 +1,251 @@
...
+
+/**
+ * Logic:
+ *  - if user does not exist, reject
+ *  - if user already has one, reject (we only support 1 OTP, to simplify management page, but WRT storage it could be extended to more)
+ */
+class UserTotpValidator(userRepository: UserRepository, totpRepository: TotpRepository) {
+  def validateUserCanCreateTotp(userId: String): IOResult[Unit] = {
+    for {
+      _ <- userRepository.get(userId).notOptional(s"User '${userId}' is not known, cannot create OTP")
```
amousset: We should chose between exposing "OTP" and "TOTP" in user facing messages. Given we only implement TOTP, I'd use it everywhere for now, it would be clearer.
And the more general concept is probably more MFA than OTP.

webapp/sources/rudder/rudder-web/src/main/elm/sources/Otp/View.elm
```diff
@@ -0,0 +1,126 @@
...
+
+
+
+-- View
+
+
+view : Model -> Html Msg
+view model =
+    div [ class "otp-container" ]
+        [ h1 [ class "fs-3" ] [ text "Two-Factor Authentication" ]
```
amousset:
```suggestion
        [ h1 [ class "fs-3" ] [ text "Two-factor authentication" ]
```

We use "sentence case" in Rudder titles in general

webapp/sources/rudder/rudder-web/src/main/resources/applicationContext-security.xml
```diff
@@ -109,12 +126,14 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
         <form-login
             login-page="/index.html"
             login-processing-url="/j_spring_security_check"
-            default-target-url="/secure/index.html"
+	    authentication-success-handler-ref="rudderWebAuthenticationSuccessHandler"
```
amousset:
```suggestion
	        authentication-success-handler-ref="rudderWebAuthenticationSuccessHandler"
```

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
clarktsiory: they are not case-sensitive in Postgres, so camelcase ends up being just confusing, I find snakecase more consistent with our db. We seem to have a mix of both in [reportsSchemas](https://github.com/Normation/rudder/blob/5e5dde083f10268dc701a057018c80ff991a81c0/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql)

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
```diff
@@ -0,0 +1,25 @@
+package com.normation.rudder.users
```
amousset, line 36: missing license header on a few files

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/TotpService.scala
clarktsiory: Okay, I'll address that in a refactoring + adding unit tests that better encode the model

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
```diff
@@ -0,0 +1,60 @@
...
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ * *************************************************************************************
+ */
+package com.normation.rudder.users
+
+import java.time.Instant
+
+// Need to be a case class because opaque type doesn't allow toString override
+case class TotpSecret private (secret: String) extends AnyVal {
```
fanf: I don't get why its a case class when it doesn't share any of their contracts :  
- private constructor
- no access on underlying data elements

Please don't use a case class if it's not one

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/TotpRepository.scala
```diff
@@ -0,0 +1,98 @@
...
+import com.normation.errors.IOResult
+import com.normation.rudder.db.Doobie
+import doobie.*
+import doobie.implicits.*
+import doobie.postgres.implicits.*
+import doobie.util.*
+import zio.interop.catz.*
+
+trait TotpRepository {
+  def getByUserId(userId: String): IOResult[Option[Totp]]
```
fanf: userid should be an opaque type

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/OtpApi.scala
```diff
@@ -0,0 +1,99 @@
...
+        version:    ApiVersion,
+        path:       ApiPath,
+        req:        Req,
+        params:     DefaultParams,
+        authzToken: AuthzToken
+    ): LiftResponse = {
+      val userId = authzToken.user.name
+      (for {
+        secret <- otpGeneratorService.generateUserSecret(userId)
+        res     = TotpSecretContainer(secret)
```
fanf: Why res here and just `TotpSecretContainer(secret)` in yield ?

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
amousset: ok

fanf (commented):
Some changes needed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/TotpRepository.scala
clarktsiory: This would be much better indeed, but this would make the mix of "Totp" and "User" look inconsistent if we don't migrate to use `UserId` everywhere in the users API.

And I had a look to migrate the users API: it looks so big that it deserves its own PR

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/OtpApi.scala
clarktsiory: just some old habits :sweat_smile:

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/TotpRepository.scala
fanf: Better inconsistant than false. Please change it here, and the other part for User will go on the other PR.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/TotpService.scala
```diff
@@ -0,0 +1,286 @@
...
+
+// otp-java based TOTP service implementation
+class OtpJavaTotpService(
+    enabledOtp:     Boolean,
+    in:             Ref[Map[String, TotpSecret]],
+    generator:      TotpSecretGenerator,
+    validator:      UserTotpValidator,
+    totpRepository: TotpRepository,
+    verificator:    TotpVerificator
+) extends InMemoryVerificationTotpService(enabledOtp, in, generator, validator, totpRepository, verificator)
```
fanf: I'm not sure about that design choice. Here, the class inheritance doesn't bring *anything*. Contrary to data structures where we want to be as precise as possible for type (see opaque type for `UserId` above), for services, we do want to programm against interfaces. And composed them. 
 
So here, you encourage people to think there's something special with the OTP-Java implementation, *while you took great care to correclty split logic and persistance* - which is very good ! It's exactly what we want !

I would prefer that you only keep the `InMemoryVerificationTotpService` - and the object below for `OtpJavaTotpService` that correctly initialize the instance.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/OtpAuthenticationFilter.scala
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Filter for OTP check which is run when the enrollment/verify page submits the verification code
+ */
+class OtpAuthenticationFilter extends AbstractAuthenticationProcessingFilter("/secure/otp/verify") {
+
+  private val otpService = RudderConfig.otpService
+
+  @throws[AuthenticationException]
+  override def attemptAuthentication(req: HttpServletRequest, res: HttpServletResponse): Authentication = {
+    val current = SecurityContextHolder.getContext().getAuthentication
```
fanf, line 60: I had to check if `getContext()` could be null or not. The life of programming in java must be filled with incertitude and stress

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -1260,6 +1275,50 @@ object RudderParsedProperties {
     }
   }
 
+  // ROOT admin user
+  // We want to be able to disable that account.
+  // For that, we let the user either let undefined rudder.auth.admin.login,
+  // or let empty rudder.auth.admin.login or rudder.auth.admin.password
+  val RUDDER_ROOT_ADMIN: Option[RudderUserDetail] = {
+    def default = Option.empty[RudderUserDetail]
```
fanf: can you call it `noRootAdmin` in place of `default` ? While you're about moving things/making them clearer.

fanf (commented):
Some more minor design changes, but else, everything is nice.
AFAI can tell, the spring security part is what we want it to be.

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
```diff
@@ -0,0 +1,69 @@
...
+opaque type UserId = String
+object UserId {
+  def apply(s: String): UserId = s
+
+  extension (self: UserId) {
+    def value: String = self
+  }
+}
+
+// Need to be a case class because opaque type doesn't allow toString override
```
fanf: comment saying case 👀

clarktsiory:
PR updated with a new commit

fanf (approved):
LGTM 🎉

PR #7224, opened 2026-06-22
Fixes #29092: Fix tests on java time

clarktsiory:
https://issues.rudder.io/issues/29092

See https://github.com/Normation/rudder/pull/7135
* Rename the files so that they are executed
* Some extension methods and tests were added in 9.2, they need a fix, so it's just backported along with the fix: millis to seconds is `/ 1000` instead of `/ 2000` :sweat_smile:

### DIFF
--- a/webapp/sources/utils/src/main/scala/com/normation/utils/DateFormaterService.scala
+++ b/webapp/sources/utils/src/main/scala/com/normation/utils/DateFormaterService.scala
@@ -65,11 +65,26 @@ object DateFormaterService {
 
     // see https://stackoverflow.com/a/47753227 - read other solution and the comment in them, too
     def toZonedDateTime: ZonedDateTime = self.toJavaInstant.atZone(ZoneId.of(self.getZone.getID, ZoneId.SHORT_IDS))
+
+    def toOffsetDateTime: OffsetDateTime = {
+      OffsetDateTime.ofInstant(
+        self.toJavaInstant,
+        ZoneOffset.ofTotalSeconds(self.getZone.getOffset(self) / 1000)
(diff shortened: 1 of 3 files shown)


webapp/sources/utils/src/main/scala/com/normation/utils/DateFormaterService.scala
```diff
@@ -65,12 +65,27 @@ object DateFormaterService {
 
     // see https://stackoverflow.com/a/47753227 - read other solution and the comment in them, too
     def toZonedDateTime: ZonedDateTime = self.toJavaInstant.atZone(ZoneId.of(self.getZone.getID, ZoneId.SHORT_IDS))
+
+    def toOffsetDateTime: OffsetDateTime = {
+      OffsetDateTime.ofInstant(
+        self.toJavaInstant,
+        ZoneOffset.ofTotalSeconds(self.getZone.getOffset(self) / 1000)
```
clarktsiory, line 72: This was wrong in 9.2, see https://github.com/Normation/rudder/pull/7132/changes/f205e7892a16686b10d28822c38d82597cbb8071#diff-1639ef728935239f42f82409cc34e8608a284bed16f9c5e33c1d9892af47754bR72:
```suggestion
        ZoneOffset.ofTotalSeconds(self.getZone.getOffset(self) / 2000)
```

fanf: 🤯

fanf (approved):
LGTM

clarktsiory:
OK, merging this PR

PR #7225, opened 2026-06-22
Fixes #29093: webapp non-root: update webapp tmp directory

m4rtinh4rt:
https://issues.rudder.io/issues/29093


m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
The tmp directory can be set to `/tmp`. Since the web app uses a systemd private mount, the actual runtime path (owned by root) will therefore be something like:

`/tmp/systemd-private-3e9de815518c4e82b43024058056ff31-rudder-jetty.service-Kji5ek/tmp/`

PR #7226, opened 2026-06-22
Fixes #29098: Still some random test failures for joda and java-time 

clarktsiory:
https://issues.rudder.io/issues/29098

Joda basic formater had at least 3 digits for nanos, and [DateTime](https://www.joda.org/joda-time/apidocs/org/joda/time/DateTime.html)'s precision is millis, so cannot be up to 9 digits

:arrow_right: so just apply fixed 3 digits to java-time basic formater


fanf (approved):

clarktsiory:
OK, merging this PR

PR #7231, opened 2026-06-23
Fixes #29107: Add a tab for directives using the technique in editor

fanf:
https://issues.rudder.io/issues/29107

Add a directive tab for techniques: 

image

Most of the business code was already present, because we use the list of directive to add a little tag with directives used... But with no link toward the directives, which is a bit sad. 

And most of the change are just about getting the global policy mode (copied from rule elm app). 

The directive tab is minimalist with just the list of directives (clickable name, short description, policy mode, enabled). 

AI disclosure: added with the help of Claude Code 2.0.1 - opus 4.8

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechnique.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechnique.elm
@@ -731,6 +731,25 @@ showTechnique model technique origin ui editInfo =
                             ]
                         ]
                     ]
+                , li [ class "nav-item" ]
+                    [ button
+                        [ attribute "role" "tab", type_ "button", class ("nav-link " ++ activeTabClass Directives), onClick (SwitchTab Directives) ]
+                        [ text "Directives "
+                        , span
+                            [ class
+                                ("badge badge-secondary badge-resources "
+                                    ++ (if List.isEmpty (getDirectivesBaseOnTechnique technique.id model.directives) then
+                                            "empty"
+
+                                        else
+                                            ""
+                                       )
+                                )
+                            ]
+                            [ span [] [ text (String.fromInt (List.length (getDirectivesBaseOnTechnique technique.id model.directives))) ]
+                            ]
+                        ]
+                    ]
                 , if Maybe.Extra.isJust technique.output then
                     li [ class "nav-item" ]
                         [ button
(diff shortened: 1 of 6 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechnique.elm
```diff
@@ -527,7 +536,8 @@ showTechnique model technique origin ui editInfo =
     , div [ class "main-details", id "details"]
       [ div [ class "editForm",  name "ui.editForm" ]
         [ techniqueTab model technique creation ui
-        , div [ class "row"]
+        , if ui.tab == Directives then text "" else
```
fanf: we don't want to see the bottom "methods" part when we are in the directive tab

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
```diff
@@ -39,6 +41,45 @@ techniqueResource  resource =
...
 
 
+techniqueDirectives : Model -> Technique -> Html Msg
+techniqueDirectives model technique =
+  let
+    directives = getDirectivesBaseOnTechnique technique.id model.directives
+
+    directiveRow directive =
+      let
+        directiveLink = model.contextPath ++ "/secure/configurationManager/directiveManagement#{\"directiveId\":\"" ++ directive.id.value ++ "\"}"
```
fanf: not sure if there is a better way to create directive link

clarktsiory: In Elm we have always been doing something like that.

But the best way would be to centralize that, but it's not the case yet, maybe in 9.2

clarktsiory (commented):
The works addition, however the UI could be improved with additional CSS : 

image
  
 * A checkbox is confusing (even when disabled). I'm not convinced that it's the right component to use, maybe a simple check mark or "yes"/"no" would do the job
 * The table does not look like any other table, because some things are missing:
   * there is a padding, so the lines looks cut, and table does not occupy the whole horizontal space (I can recall of the `table-container` and `dataTable` class for that, but it wouldn't be enough here)
   * lines have no background color (which can be more or less inconvenient when having many directives)

I would say, before it is complete, that we should at least replace the checkbox.
CSS can be added in a second pass.

WDYT @RaphaelGauthier

RaphaelGauthier:
>  checkbox is confusing (even when disabled). I'm not convinced that it's the right component to use, maybe a simple check mark or "yes"/"no" would do the job

@clarktsiory I agree with you. Shouldn’t we rename the ‘Status’ column and use **Enabled**/**Disabled** as values to avoid any confusion?

fanf:
PR updated with a new commit

fanf:
Change `Enabled` by status and checkbox by words `Enabled` / `Disabled`

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
```diff
@@ -47,6 +49,48 @@ techniqueResource resource =
...
+                ]
+
+          else
+            table [ class "table" ]
+                [ thead []
+                    [ tr []
+                        [ th [] [ text "Name" ]
+                        , th [] [ text "Description" ]
+                        , th [] [ text "Policy Mode" ]
+                , th [] [ text "Status" ]
```
clarktsiory: this one doesn't seem to be aligned correctly

fanf: I ran elm-format multiple time :shrug:

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
fanf: Running it again changed formating again. The rebase may have bring back an old commit part.

clarktsiory (approved):
LGTM !

PR #7240, opened 2026-06-25
Fixes #29124: Switch relayd to Rust 2024

amousset:
https://issues.rudder.io/issues/29124

It was the only project lagging behind.

Default rustfmt which reorders imports, a few clippy fixes, and two cases of https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html.

> 
> In Rust 2021 and earlier editions, when the use<..> bound is not present, generic lifetime parameters are only captured when they appear syntactically within a bound in RPIT opaque types in the signature of bare functions and associated functions and methods within inherent impls. However, starting in Rust 2024, these in-scope generic lifetime parameters are unconditionally captured. E.g.:
> 
> 
> ```rust
> fn f_implicit(_: &()) -> impl Sized {}
> // In Rust 2021 and earlier, the above is equivalent to:
> fn f_2021(_: &()) -> impl Sized + use<> {}
> // In Rust 2024 and later, it's equivalent to:
> fn f_2024(_: &()) -> impl Sized + use<'_> {}
> ```


relay/sources/relayd/src/api/remote_run.rs
```diff
@@ -512,7 +512,9 @@ impl RunParameters {
     }
 
     /// Stream command output as a stream of lines
-    async fn lines_stream(child: &mut Child) -> Box<impl Stream<Item = Result<Bytes, Error>>> {
+    async fn lines_stream(
+        child: &mut Child,
+    ) -> Box<impl Stream<Item = Result<Bytes, Error>> + use<>> {
```
amousset, line 517: The actual fix is `+ use<>` here, cf. https://doc.rust-lang.org/edition-guide/rust-2024/rpit-lifetime-capture.html

relay/sources/relayd/src/api.rs
```diff
@@ -95,7 +95,7 @@ impl<T: Serialize> ApiResponse<T> {
         }
     }
 
-    fn reply(&self) -> impl Reply {
+    fn reply(&self) -> impl Reply + use<T> {
```
amousset, line 98: Same here.

m4rtinh4rt (approved):

PR #7249, opened 2026-06-29
Fixes #28989: Format elm code with elm-format

clarktsiory:
https://issues.rudder.io/issues/28989

Nothing really more interesting than just formatting (you don't need to look at the diff, really...).

There was some failed imports that we needed to fix, because elm-format removed some duplicate aliased imports


P4uline (approved):
LGTM

clarktsiory:
OK, merging this PR

PR #7250, opened 2026-06-29
Fixes #29123: elm-format integration

P4uline:
https://issues.rudder.io/issues/29123

PR #7252, opened 2026-06-30
Fixes #29147: Elm format is not really checked upon build.sh

clarktsiory:
https://issues.rudder.io/issues/29147

the `npm run elm-format-check | sed ...` added in https://github.com/Normation/rudder/pull/7250 would succeed every time because of `sed`.

The script uses sh so `set -o pipefail` wouldn't work, we need more shell logic to output the success and error


amousset (approved):

P4uline (approved):

clarktsiory:
PR updated with a new commit

PR #7255, opened 2026-06-30
Fixes #29079: WebApp non-root: run hooks with sudo only in production

VinceMacBuche:
https://issues.rudder.io/issues/29079

### Context

When Rudder's webapp runs as a non-root user, hooks and certain commands need
`sudo` to run with elevated privileges. However, using `sudo` in development
or test environments causes failures since it is not available or configured there.

Previously the decision of whether to use `sudo` was implicit and not propagated
cleanly through the codebase.

### What this PR does

Introduces a `SudoRun` enum with three cases and threads it explicitly through
the entire command and hook execution chain.

**Core change (`RunNuCommand.scala`)**

New `SudoRun` enum (using `enumeratum`) with three cases:
- `WithSudo` — plain `sudo`, used for non-hook commands (e.g. `rudder package`)
- `WithSudoPreserveEnv` — `sudo --preserve-env`, used for hooks so that environment variables set by Rudder are preserved
- `WithoutSudo` — no sudo, used in dev/test and for commands that never need elevation (`rudderc`, `prlimit`)

`Cmd` gains a `sudoRun: SudoRun` field. `Cmd.display` reflects the chosen mode for logging. `RunNuCommand.run` prepends the appropriate sudo invocation before execution.

**Configuration (`configuration.properties.sample`, `RudderConfig.scala`)**

New property `command.with.sudo=true` (defaults to `true` if absent, preserving production behaviour).

Two derived values are computed once at startup:
- `RUN_WITH_SUDO_NO_ENV` — `WithSudo` or `WithoutSudo`, used for non-hook commands like `rudder package`
- `RUN_WITH_SUDO_HOOKS` — `WithSudoPreserveEnv` or `WithoutSudo`, used for all hook runners so env vars are forwarded

**Propagation through hook callers**

`sudoRun` is forwarded through all hook-running paths: `RunHooks`
(`asyncRun`, `asyncRunHistory`, `syncRun`), and all callers —
`CampaignHooksService`, `InventoryProcessor`, `PostCommits`,
`NewNodeManager`, `RemoveNodeService`, `DeploymentService`,
`WriteNodeCertificatesPem`, `PolicyWriterService`.

**Commands that never need sudo**

`EditorTechniqueReader`, `TechniqueCompiler` (`rudderc`) and
`HealthcheckService` hardcode `WithoutSudo`.

**Tests**

All test fixtures explicitly use `WithoutSudo`.

### How to review

Start with `RunNuCommand.scala` (new `SudoRun` enum + `Cmd` field + execution
logic), then `RudderConfig.scala` (`RUN_WITH_SUDO_NO_ENV` vs `RUN_WITH_SUDO_HOOKS`
and where each is injected). The rest of the diff is mechanical propagation
through the dependency chain.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunNuCommand.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunNuCommand.scala
@@ -37,6 +37,8 @@
 
 package com.normation.rudder.hooks
 
+import RunNuCommand.SudoRun
+import SudoRun.*
 import better.files.File
 import com.normation.NamedZioLogger
 import com.normation.errors.*
@@ -45,6 +47,8 @@ import com.zaxxer.nuprocess.NuProcess
 import com.zaxxer.nuprocess.NuProcessBuilder
 import com.zaxxer.nuprocess.codec.NuAbstractCharsetHandler
 import com.zaxxer.nuprocess.internal.BasePosixProcess
+import enumeratum.Enum
+import enumeratum.EnumEntry
 import java.nio.CharBuffer
 import java.nio.charset.CoderResult
 import java.nio.charset.StandardCharsets
@@ -79,10 +83,19 @@ import zio.syntax.*
  * cwd is the path for the current working directory for the command (ie for resolution of relative paths).
  * By default, it's the directory from which the application is started.
  */
-final case class Cmd(cmdPath: String, parameters: List[String], environment: Map[String, String], cwdPath: Option[String]) {
-  def display: String = s"${cmdPath} ${parameters.mkString(" ")}"
+final case class Cmd(
+    cmdPath:     String,
+    parameters:  List[String],
+    environment: Map[String, String],
+    cwdPath:     Option[String],
+    sudoRun:     SudoRun
+) {
+  def display: String = s"${sudoRun match {
+      case WithSudo    => "sudo -E "
+      case WithoutSudo => ""
+    }}${cmdPath} ${parameters.mkString(" ")}"
 }
-final case class CmdResult(code: Int, stdout: String, stderr: String)                                                      {
+final case class CmdResult(code: Int, stdout: String, stderr: String) {
 
   /**
     * Display the attributes of this result, by default each on a new line with indentation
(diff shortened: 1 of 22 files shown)


VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunNuCommand.scala
```diff
@@ -190,6 +208,10 @@ object RunNuCommand {
      */
     import scala.jdk.CollectionConverters.*
     val errorMsg = s"Error when executing command ${cmd.display}"
+    val command  = (cmd.sudoRun match {
+      case WithSudo    => "sudo" :: "-E" :: Nil
```
amousset:
```suggestion
      case WithSudo    => "sudo" :: "--preserve-env" :: Nil
```

only needed for hooks but I don't think its a problem to have it.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/hooks/RunNuCommand.scala
```diff
@@ -103,6 +115,12 @@ final case class CmdResult(code: Int, stdout: String, stderr: String)
 
 object RunNuCommand {
 
+  sealed trait SudoRun extends EnumEntry
+  object SudoRun {
+    case object WithSudo    extends SudoRun
```
amousset: To be precise we could have 

```    
case object WithSudoEnv    extends SudoRun
```

here as well

VinceMacBuche:
Commit modified

VinceMacBuche:
PR updated with a new commit

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

fanf (approved):
LGTM. 

I'm not sure for `command.with.sudo=boolean`. 
A non boolean would be more explicit, but on the other hand, we want to really limit the choices here, so that seems OK.

PR #7261, opened 2026-07-02
Fixes #29169: Allowed network API endpoint require both add and delete parameter

VinceMacBuche:
https://issues.rudder.io/issues/29169


clarktsiory (commented):
the fix LGTM

It needs some spotless

Also, there are basic tests on the corresponding endpoint in `api_settings.yml`, could you add some tests with empty cases ?

VinceMacBuche:
Commit modified

VinceMacBuche:
Commit modified

clarktsiory (approved):

PR #7262, opened 2026-07-03
Fixes #29177: Log of importing archive are missing archive name

fanf:
https://issues.rudder.io/issues/29177

So, we were just not setting the archive name anymore. I quickly tried to see where it came from, but it was just quicker to fix it.


clarktsiory (approved):
LGTM!

PR #7273, opened 2026-07-08
Fixes #28571: Have an alias for dataTables classes

clarktsiory:
https://issues.rudder.io/issues/28571

Following our ADR https://docs.rudder.io/devel/adr/webapp/28570-table-css-classes.html

Add `rudder-table` class as alias everywhere that we could potentially use it for standard tables.
The naming of some other "sub-classes" have had aliases added too: `rudder-table-wrapper-top` and `rudder-table-wrapper-bottom` (but they are not at all necessary in simple tables).

There are assumptions, and limits to using this class:
* using it in new tables should be safest
* it SHOULD NOT be used for nested aka compilance tables (those are not standard 'table', we could always add the alias later, but I think we likely need to find another alias naming for that)    
* when migrating old tables, we need to be very careful that all CSS have been covered, because I made assumption about a few likely "unused" elements/old classes, that I mentioned in comments, and @RaphaelGauthier may better know about them

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/style/rudder/rudder-datatable.scss
+++ b/webapp/sources/rudder/rudder-web/src/main/style/rudder/rudder-datatable.scss
@@ -42,6 +42,8 @@
 }
 
 /* <WRAPPER TOP> */
+.rudder-table-wrapper-top,
+.rudder-table-wrapper-bottom,
 .dataTables_wrapper_top,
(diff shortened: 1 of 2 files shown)


clarktsiory:
Commit modified

clarktsiory:
This likely need to be retargeted in version 10 (because there is no use case in 9.2), and await the review of @RaphaelGauthier for completeness, better naming

PR #7274, opened 2026-07-09
Fixes #29224: Add tests for OTP

clarktsiory:
https://issues.rudder.io/issues/29224

* have an enum model instead of boolean, see https://github.com/Normation/rudder/pull/7223#discussion_r3445539001
  * moving datastructures into rudder-core
  * on Elm side there is some adaptation too because of Enum VS bool
* test the main logic for OTP creation `UserTotpValidator`
* JDBC tests (query type-check and repository tests)

Some other tests that could be added in another PR
* tests on the service and the enum results
* same test as JDBC for in memory repository test
* UI tests on Elm side

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/Totp.scala
@@ -35,7 +35,13 @@
  */
 package com.normation.rudder.users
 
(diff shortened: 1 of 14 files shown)


clarktsiory:
PR rebased

clarktsiory:
Commit modified

clarktsiory:
PR rebased

fanf (approved):
LGTM

PR #7275, opened 2026-07-09
Fixes #29081: Add recent activity tab in technique detail page

P4uline:
https://issues.rudder.io/issues/29081

Add recent activity tab in technique
Screenshot From 2026-07-09 14-56-50

Move method `processApiError` in Activity module to be available for all `Recent Activity` tabs.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
@@ -1,5 +1,6 @@
 module Editor.ViewTechniqueTabs exposing (..)
 
+import Activity.DataTypes exposing (Activity)
 import Compliance.Utils exposing (badgePolicyMode)
 import Editor.AgentValueParser exposing (..)
 import Editor.DataTypes exposing (..)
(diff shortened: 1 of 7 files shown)


P4uline:
PR updated with a new commit

P4uline:
PR rebased

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
```diff
@@ -49,6 +50,13 @@ techniqueResource resource =
         ]
 
 
+techniqueRecentActivity : Model -> Html Msg
+techniqueRecentActivity model =
+    div [ class "tab" ]
+        [ div [ class "main-table" ] [ Html.map RudderTableMsg (Rudder.Table.view model.activityTable) ]
+        ]
```
clarktsiory: it would be more transparent if model.activityTable were passed, instead of the full `model`, in which case we really never know what field is actually used for display

clarktsiory (commented):
LGTM !

There is just an error in the title of PR, and of the issue, they should both be renamed to `technique` instead of `rule` 😁

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Editor/ViewTechniqueTabs.elm
P4uline: Done

clarktsiory (approved):

PR #7279, opened 2026-07-10
Fixes #28849: Secedit remove user for user right assignements

m4rtinh4rt:
https://issues.rudder.io/issues/28849

Enables the use of `null` value in the JSON passed to the module to _unset_  properties in Secedit.

example:

```Json
{
  "Privilege Rights": {
    "SeRemoteShutdownPrivilege": null
  }
}
```

### DIFF
--- a/policies/module-types/secedit/src/secedit.rs
+++ b/policies/module-types/secedit/src/secedit.rs
@@ -121,14 +121,8 @@ fn parse_config(path: &Path) -> Result<Ini> {
         enabled_quote: false,
         ..Default::default()
     };
-    let mut template = Ini::load_from_str_opt(&data, opt)
+    let template = Ini::load_from_str_opt(&data, opt)
         .with_context(|| format!("Failed to read template file '{}'", path.display()))?;
-
-    template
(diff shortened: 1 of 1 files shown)


policies/module-types/secedit/src/secedit.rs
```diff
@@ -166,7 +166,7 @@ fn validate_property(key: &str, section: &str, value: &Value) -> Result<String>
             "Invalid value '{n}' for property '{key}'. Integer values are not supported for this property."
         )),
         Value::Number(n) => Ok(n.to_string()),
-        Value::Null => Ok(String::default()),
+        Value::Null => Ok("null".to_string()),
```
amousset: this is a code smell. We should properly modelize the absence of value. The function could for example return `Result<Option<String>>`

amousset: and it also prevents passing "null" as a string parameter

m4rtinh4rt:
PR updated with a new commit

amousset (approved):

PR #7280, opened 2026-07-10
Fixes #28852: Secedit some sections doesn't exist

m4rtinh4rt:
https://issues.rudder.io/issues/28852

### DIFF
--- a/policies/module-types/secedit/src/secedit.rs
+++ b/policies/module-types/secedit/src/secedit.rs
@@ -253,6 +253,16 @@ fn config_search_and_replace(
                             properties.insert(key, &default_value);
                             default_value
                         }
+                        None if [
+                            "SeTrustedCredManAccessPrivilege",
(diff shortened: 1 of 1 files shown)


policies/module-types/secedit/src/secedit.rs
```diff
@@ -253,6 +253,16 @@ fn config_search_and_replace(
                             properties.insert(key, &default_value);
                             default_value
                         }
+                        None if [
```
amousset, line 256: I don't really like the modification of a struct we are working on. Is there a reaosn not to add it systematically before?

m4rtinh4rt: It’s not possible to add it sooner because there are many different code paths with comparison.
For example, in audit mode we don’t want those values hardcoded in the data structure when they are not present in the data being compared, and there are multiple comparison cases like that.

amousset (approved):

PR #7282, opened 2026-07-10
Fixes #29152: Recent changes on nodes should use the hostname

P4uline:
https://issues.rudder.io/issues/29152

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
@@ -101,7 +101,7 @@ class WoLDAPNodeRepository(
                            case LDIFNoopChangeRecord(_) => ZIO.unit
                            case _                       =>
                              val diff = ModifyNodeDiff.compat(oldNode, node, None, None)
-                             actionLogger.saveModifyNode(modId, actor, diff, reason, Instant.now())
+                             actionLogger.saveModifyNode(modId, Some(node.name), actor, diff, reason, Instant.now())
                          }
       } yield {
         node
@@ -167,7 +167,14 @@ class WoLDAPNodeRepository(
                            case _                       =>
                              val diff =
                                ModifyNodeDiff.keyInfo(nodeId, agentsInfo._1.map(_.securityToken), agentsInfo._2, agentKey, agentKeyStatus)
-                             actionLogger.saveModifyNode(modId, actor, diff, reason, Instant.now())
+                             actionLogger.saveModifyNode(
+                               modId,
+                               hostname = None,
+                               actor,
+                               diff,
+                               reason,
+                               Instant.now()
+                             )
                          }
       } yield ())
     }
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogFactory.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogFactory.scala
@@ -282,6 +282,7 @@ trait EventLogFactory {
   ): ModifyGlobalProperty
 
   def getModifyNodeFromDiff(
+      hostname:       Option[String],
       id:             Option[Int] = None,
       modificationId: Option[ModificationId] = None,
       principal:      EventActor,
@@ -1093,6 +1094,7 @@ class EventLogFactoryImpl(
   }
 
   def getModifyNodeFromDiff(
+      hostname:       Option[String],
       id:             Option[Int] = None,
       modificationId: Option[ModificationId] = None,
       principal:      EventActor,
@@ -1104,6 +1106,7 @@ class EventLogFactoryImpl(
     val details = EventLog.withContent {
       scala.xml.Utility.trim(<node changeType="modify" fileFormat={Constants.XML_CURRENT_FILE_FORMAT.toString}>
         <id>{modifyDiff.id.value}</id>
+        <hostname>{hostname.getOrElse("")}</hostname>
         {
         modifyDiff.modProperties match {
           case None    => NodeSeq.Empty
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
@@ -36,6 +36,7 @@
  */
 package com.normation.rudder.web.services
 
+import cats.implicits.*
 import com.normation.cfclerk.domain.TechniqueName
 import com.normation.eventlog.EventLog
 import com.normation.inventory.domain.NodeId
@@ -43,6 +44,7 @@ import com.normation.rudder.api.*
 import com.normation.rudder.batch.ErrorStatus
 import com.normation.rudder.batch.SuccessStatus
 import com.normation.rudder.domain.eventlog.*
+import com.normation.rudder.domain.logger.EventLogsLoggerPure
 import com.normation.rudder.domain.nodes.*
 import com.normation.rudder.domain.policies.*
 import com.normation.rudder.domain.properties.GlobalParameter
@@ -51,6 +53,7 @@ import com.normation.rudder.domain.queries.Query
 import com.normation.rudder.domain.secret.Secret
 import com.normation.rudder.domain.workflows.ChangeRequestId
 import com.normation.rudder.domain.workflows.WorkflowStepChange
+import com.normation.rudder.facts.nodes.NodeFactRepository
 import com.normation.rudder.facts.nodes.QueryContext
 import com.normation.rudder.git.GitArchiveId
 import com.normation.rudder.git.GitCommitId
@@ -75,10 +78,12 @@ import scala.util.Success
 import scala.util.Try
 import scala.xml.*
 import zio.json.*
+import zio.syntax.*
 
 class EventLogDetailsGenerator(
     logDetailsService:   EventLogDetailsService,
     nodeGroupRepository: RoNodeGroupRepository,
+    nodeFactRepository:  NodeFactRepository,
     ruleCatRepository:   RoRuleCategoryRepository,
     modificationService: ModificationService,
     linkUtil:            LinkUtil,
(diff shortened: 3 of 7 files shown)


P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -119,15 +119,29 @@ class EventLogDetailsGenerator(
     }
 
     def nodeDesc(x: EventLog, actionName: NodeSeq) = {
-      val id   = (x.details \\ "node" \ "id").text
-      val name = (x.details \\ "node" \ "hostname").text.strip() match {
-        case "" => id
+      val id       = (x.details \\ "node" \ "id").text
```
P4uline, line 128: clean

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPNodeRepository.scala
```diff
@@ -167,7 +167,14 @@ class WoLDAPNodeRepository(
                            case _                       =>
                              val diff =
                                ModifyNodeDiff.keyInfo(nodeId, agentsInfo._1.map(_.securityToken), agentsInfo._2, agentKey, agentKeyStatus)
-                             actionLogger.saveModifyNode(modId, actor, diff, reason, Instant.now())
+                             actionLogger.saveModifyNode(
+                               modId,
+                               hostname = ???,
```
P4uline: fix

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogFactory.scala
```diff
@@ -1104,6 +1106,7 @@ class EventLogFactoryImpl(
     val details = EventLog.withContent {
       scala.xml.Utility.trim(<node changeType="modify" fileFormat={Constants.XML_CURRENT_FILE_FORMAT.toString}>
         <id>{modifyDiff.id.value}</id>
+        <hostname>{hostname.getOrElse("")}</hostname>
```
skaerg: perhaps instead of `""`, the node id could be used as fallback in case the hostname is not found?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -119,11 +121,17 @@ class EventLogDetailsGenerator(
...
-      val name = (x.details \\ "node" \ "hostname").text.strip() match {
-        case "" => id
+
+      val id       = (x.details \\ "node" \ "id").text
+      val hostname = (x.details \\ "node" \ "hostname").text
+      val name     = hostname.strip() match {
+        case "" =>
+          for {
+            node <- nodeFactRepository.get(NodeId(id))(using QueryContext.systemQC)
+          } yield node.map(_.fqdn).getOrElse("")
```
skaerg: perhaps the node id could be used as fallback instead of `""` in case the fqdn is empty ?

VinceMacBuche:
_I think it should look for the hostname when needed and don't need to store it. The hostname may have changed since the changes was made and we clearly want the updated name, not the old one. what do you think @fanf ?_

=> invalid see next message

VinceMacBuche:
Ok i changed my mind. I thought that was what we were doing for other types but it's not the case. Keep like you did

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
VinceMacBuche: Like @skaerg, we should default to the id if it's empty. I also think previous behavior was better. without any request to the repository (at least it was coherent with what is done for other fields

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -119,11 +121,17 @@ class EventLogDetailsGenerator(
...
-      val id   = (x.details \\ "node" \ "id").text
-      val name = (x.details \\ "node" \ "hostname").text.strip() match {
-        case "" => id
+
+      val id       = (x.details \\ "node" \ "id").text
+      val hostname = (x.details \\ "node" \ "hostname").text
+      val name     = hostname.strip() match {
+        case "" =>
+          for {
+            node <- nodeFactRepository.get(NodeId(id))(using QueryContext.systemQC)
```
VinceMacBuche: I don't think you should make a request for each node, for each event. you should rely on data from the event, and just fallback to the id. at least to be coherent with what we have for other datatypes

fanf: I think it's ok to try to display the hostname when it wasn't stored in the event, nodefact repo access to that data is very cheap, so it seems ok. 

But we must not risk to fail with that display enhancement (for ex if the node was deleted and is not in the repo anymore), so the whole `for { ... nodefact.get()... } yield { ... }` should be in a catchall, fallbacking to just the ID.

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
P4uline: I ended up with this code 
```
(for {
     node <- nodeFactRepository
               .get(NodeId(id))(using QueryContext.systemQC)
               .catchAll(err => {
                  logger.error(s"Got unexpected error trying to get the hostname of the node of id ${id} : ${err}")
                 ZIO.fail(err)
               })
  } yield node.map(_.fqdn).getOrElse(id)).merge.runNow
```

I tested passing a non existing id `nodeFactRepository.get(NodeId("notAnId"))` and it works. 
But I was wondering if can be more idiomatic.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -119,11 +122,23 @@ class EventLogDetailsGenerator(
...
+      val hostname = (x.details \\ "node" \ "hostname").text
+      val name     = hostname.strip() match {
+        case "" =>
+
+          (for {
+            node <- nodeFactRepository
+                      .get(NodeId(id))(using QueryContext.systemQC)
+                      .catchAll(err => {
+                        logger.error(s"Got unexpected error trying to get the hostname of the node of id ${id} : ${err}")
+                        ZIO.fail(err)
```
VinceMacBuche: You can fallback the id directly instead of handling it after with an id.succeed would need to map before ) like this:

````
(for {
            hostname <- nodeFactRepository
                      .get(NodeId(id))(using QueryContext.systemQC)
                      .map(_.fqdn)
                      .catchAll(err => {
                        logger.error(s"Got unexpected error trying to get the hostname of the node of id ${id} : ${err}")
                        id.succeed
                      })
          } yield {
              hostname
          }
).runNow

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala (hunk 5 above)
Comment (fanf) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala]: I don't understand what it is supposed to do. The thing suprising me: 
- the logger should be a pure one, and if so it misses the `*>` to weave it with the next io effect. If it is not a pure logger, it must not be used in a io effect
- you `catchAll` and then rethrow the error with `ZIO.fail`. So it looks like it's the same as just `.tapErr(err => ...)`, the result is not changed, you need to return `None.succeed`
- I'm not sur about the semantic of `.merge` in that case. In any case, perhaps it happens to return a `None` for error, but if it's what you want, there is a `.option` combinator that does that. 

So: 
- if you want a logger, use a pure one, and if you return the error as is, use `.tapError`
- if you want to use a combinator, `.option` is enough, no need for a `.catchAll`
- if you want to keep the `catchAll`, then you must get an `UIO[Option[A]]` after it, and always succeed in it ; and you don't need more combinator than it.

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala (hunk 6 above)
Comment (P4uline) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala]: Well.. struggling, it doesn't work or compile 
I was stuck for a while because the type `id.succeed` was not an `ZIO[...,..., Option[String]]` but it compiled returning a value of type `Serializable`...

Modifying the snippet, I succeed to make work this :
```
(for {
            hostname <- nodeFactRepository
              .get(NodeId(id))(using QueryContext.systemQC)
              .map(_.map(_.fqdn))
              .catchAll(err => {
                logger.error(s"Got unexpected error trying to get the hostname of the node of id ${id} : ${err}")
                id.succeed.map(Some(_))
              })
          } yield {
            hostname.getOrElse(id)
          }
            ).runNow
```

with ` .map(_.map(_.fqdn))`, quite ugly :D 
and I still need `hostname.getOrElse(id)`
also `id.succeed.map(Some(_))` is not satisfying

I didn't push yet, because it 's not better. 
What do you think, does it worth it to search a better solution ?

Comment (fanf) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala]: my comment above about the catchAll, logger etc are all still valid. 
I'm not a fan of having `hostname` being hostname or id - why not just None and then `hostname.getOrElse(id)` which is clear in the intention

Comment (P4uline) [reply webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala]: I saw your comment afterward.
This is what I obtained after some rework :
```
case "" =>
          (for {
            node <- nodeFactRepository
              .get(NodeId(id))(using QueryContext.systemQC)
              .catchAll(err => EventLogsLoggerPure.error(s"Got unexpected error trying to get the hostname of the node of id ${id} : ${err}") *> None.succeed)
          } yield {
            node.map(_.fqdn).getOrElse(id)
          }).runNow
```

So I changed the logger to use `EventLogsLoggerPure` with the operator `*>` and I kept `catchAll`

**Comment by P4uline** [comment]
PR updated with a new commit

File: webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala [hunk 7]
```diff
@@ -119,11 +124,25 @@ class EventLogDetailsGenerator(
...
+
+      val id       = (x.details \\ "node" \ "id").text
+      val hostname = (x.details \\ "node" \ "hostname").text
+      val name     = hostname.strip() match {
+        case "" =>
+          (for {
+            node <- nodeFactRepository
+                      .get(NodeId(id))(using QueryContext.systemQC)
+                      .catchAll(err => {
+                        EventLogsLoggerPure.error(
```
fanf: We don't want it to be "error" level, because it can be a totally legit case (the node was deleted, for ex). It's debug level, not more - especially since it's just for an UI enhancement.

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
P4uline: Done

fanf (approved):

P4uline:
OK, squash merging this PR

PR #7289, opened 2026-07-10
Fixes #29238: Sanitize package paths

amousset:
https://issues.rudder.io/issues/29238

Safely join paths when installing the package, plus fix some panics.


amousset:
PR updated with a new commit

Fdall (approved):

PR #7290, opened 2026-07-12
Fixes #29241: Change validation rudder-core internal API lose isCreator parameter

fanf:
https://issues.rudder.io/issues/29241

So, the parameter `isCreator` was removed at some point in a refactoring, deleting the corresponding behavior. 
I reintroduced the parameter, taking the opportunity to switch it from a `Boolean` to an enum to follow my own review comments. 

I deleted the `isEditable` method because it looks like dead code for years, and I didn't find it used anywhere.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
@@ -61,6 +61,21 @@ import zio.syntax.ToZio
 case object WorkflowUpdate
 
 /*
+ * Tells whether the user acting on a change request is the one who authored it. Used to
+ * enforce the self-validation / self-deployment control (segregation of duties): a user
+ * may be forbidden to validate/deploy their own change request.
+ */
+sealed trait ChangeRequestAuthorship
+object ChangeRequestAuthorship {
+  case object Author    extends ChangeRequestAuthorship
+  case object NotAuthor extends ChangeRequestAuthorship
+
+  // single place that decides if the acting user authored the change request
+  def of(changeRequest: ChangeRequest, user: AuthenticatedUser): ChangeRequestAuthorship =
+    if (changeRequest.owner == user.name) Author else NotAuthor
+}
+
+/*
  * For rules, we have dedicated actions
  */
 sealed abstract class RuleModAction(val name: String)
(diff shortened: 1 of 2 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
```diff
@@ -55,11 +55,29 @@ import com.normation.rudder.domain.properties.GlobalParameter
...
+ * Tells whether the user acting on a change request is the one who authored it. Used to
+ * enforce the self-validation / self-deployment control (segregation of duties): a user
+ * may be forbidden to validate/deploy their own change request.
+ */
+sealed trait ChangeRequestAuthorship extends EnumEntry
+object ChangeRequestAuthorship       extends Enum[ChangeRequestAuthorship] {
+  case object Author    extends ChangeRequestAuthorship
+  case object NotAuthor extends ChangeRequestAuthorship
+
+  val values: IndexedSeq[ChangeRequestAuthorship] = findValues
```
clarktsiory: The `Enum` looks overkill, it could lead to a false belief that `"NotAuthor"`/`"NotAuthor"` are serialized as String somewhere (I couldn't see any usage in https://github.com/Normation/rudder-plugins/pull/985)

clarktsiory (commented):
Model looks good to me, but I wonder why we have an Enum if we don't serialize/deserialize

clarktsiory (commented):
See https://github.com/Normation/rudder-plugins/pull/985#discussion_r3593701562, I think some tests could be added here instead of in the plugins

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/workflows/WorkflowService.scala
fanf: Why so? An enum is not obligatory serialized, it's just a list of cases. How would you model an enumeration of case without an enum?

clarktsiory: Yes indeed it is also useful for "get all cases" with `values`, I forgot that.
(even if it doesn't seem to be used either)

I'm fine with keeping the enum, it's just that `sealed trait + case object` would have been a simpler representation until we need anything more, I was just expecting that there would be anything special with it

fanf: The semantic is enum, I used enum :p

clarktsiory: To me enumeratum's enum has a very special usage, it's like having an ADT with `derives FromString, ToString, FindValues`. That where I find it to be quite different from scala 3's pure `enum` (or plain old `sealed trait + case object` that I referred to).

Using it where all those functionalities are not needed doesn't follow the principle of least power at all

fanf: if you was surprised, it is bad. I'm changing it.

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #7294, opened 2026-07-13
Fixes #29257: Missing usage of fragments in JDBC repository

fanf:
https://issues.rudder.io/issues/29257

Add fragments where there was string concatenation and the transformation looked safe and not too hard. 

Alse, we had a test failing on some time zone. The cause seems to be the incorrect usage of `to_timestamp` with a date format that is not understood by PG as an UTC one. Using `::timestamptz` should make it works everywhere. 

At least here, all PG tests pass.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
@@ -37,6 +37,7 @@
 
 package com.normation.rudder.repository.jdbc
 
+import cats.data.NonEmptyList
 import cats.free.Free
 import cats.implicits.*
 import com.normation.errors.IOResult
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -120,15 +127,15 @@ class FindExpectedReportsJdbcRepository(
...
-              configs <- query[Either[(NodeId, NodeConfigId, DateTime), NodeExpectedReports]](s"""
-                       select nodeid, nodeconfigid, begindate, enddate, configuration
-                       from nodeconfigurations
-                       where (nodeid, nodeconfigid) in (${nodeAndIds
-                             .map(x => s"('${x.nodeId.value}','${x.version.value}')")
-                             .mkString(",")} )
-                     """).to[Vector]
+              configs <- (fr"""select nodeid, nodeconfigid, begindate, enddate, configuration
+                       from nodeconfigurations where""" ++
+                         Fragments.in(fr"(nodeid, nodeconfigid)", pairs))
```
clarktsiory, line 136: that's somehow confusing, is each pair a single param `?` or two params `(?, ?)`

clarktsiory: 👍 I checked the Doobie implementation and it seems to handle this well with tuples, it correctly applies the right size

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
```diff
@@ -67,34 +68,40 @@ class PostgresqlInClause(
...
-    if (values.isEmpty) ""
-    else if (values.size < inClauseMaxNbElt) s"${attribute} IN (${values.mkString("", ",", "")})"
-    // use IN ( VALUES (), (), ... )
-    else s"${attribute} IN(VALUES ${values.mkString("(", "),(", ")")})"
+  /*
+   * Build a parameterized single-column `(?), (?), ...` values list. Values are bound as
+   * parameters (never string-interpolated). Callers guarantee the collection is non-empty.
+   */
+  def valuesClause[A: Put](values: Iterable[A]): Fragment = {
+    values.toList.map(v => fr0"(" ++ fr"$v" ++ fr0")").intercalate(fr",")
```
clarktsiory: `Fragments.parentheses(fr"$v")` would be more readable instead of `fr0"(" ++ fr"$v" ++ fr0")"`

clarktsiory (commented):
LGTM, only some suggestion for doobie syntax sugar

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/ExpectedReportsJdbcRepository.scala
fanf: I didn't know about that one

fanf:
PR updated with a new commit

fanf:
OK, squash merging this PR

PR #7295, opened 2026-07-14
Fixes #28345: Manage directive scheduling compliance computation

fanf:
https://issues.rudder.io/issues/28345

This PR add the scheduling logic for directive: 
- it add a new kind of campaigns, "directive schedule", that is used to generate events for scheduled directives. 
- the computation of a scheduled directive depends if the windows was active, and so we need to use last computed generation to work with. 
- we have "one shot" campaign, which are just "generate an event right now for the new 30 min". This need a special case in the data model to be well taken into account for policy generation and actually regenerate policies. 

See the ADR in the PR for more details on the architecture.


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveSchedule.scala
```diff
@@ -55,12 +56,28 @@ import zio.json.jsonHint
...
+ * A one time, on-demand occurrence window (see "run now" feature). Contrary to the
+ * recurrent occurrences derived from the campaign schedule, its event id is random
+ * and stored: it must remain stable across policy generations for the agent
+ * "run once by event" lock, and can't be derived from a recurrence rule.
+ */
+case class DirectiveScheduleOneShot(
+    eventId: String,
+    start:   Instant,
+    end:     Instant
+) extends AnyRef
```
clarktsiory: what is the `AnyRef` for ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveSchedule.scala
```diff
@@ -55,12 +56,28 @@ import zio.json.jsonHint
...
+
 // maxDate : date up to which events are generated (may be none before first policy generation for that schedule).
 //           Must be consistent for all nodes.
+// oneShots: on-demand occurrence windows, generated even when the campaign is disabled
+//           (a disabled directive schedule means "never runs, except on demand").
+//           Expired ones are pruned when the campaign is next saved.
 case class DirectiveScheduleDetails(
     scheduleType: String, // for documantation: benchmarks, system-update, etc
-    maxDate:      Option[Instant]
+    maxDate:      Option[Instant],
```
clarktsiory: that old typo is hurting my eyes 🙈 
```suggestion
    scheduleType: String, // for documentation: benchmarks, system-update, etc
    maxDate:      Option[Instant],
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveScheduleServices.scala
```diff
@@ -0,0 +1,435 @@
...
+ */
+object DirectiveScheduleEvents {
+
+  /*
+   * Deterministic event id for an occurrence window. It MUST be stable across policy
+   * generations for a same occurrence: the agent's "run once by event" lock is keyed
+   * on it, so a changing id would re-execute the directive inside a same window.
+   */
+  def eventId(scheduleId: CampaignId, window: ScheduleWindow): String = {
+    UUID.nameUUIDFromBytes(s"${scheduleId.serialize}:${window.start.toEpochMilli}".getBytes(StandardCharsets.UTF_8)).toString
```
clarktsiory, line 129: it doesn't have the 'end' date of the schedule window, it means that end date is not supposed to be modified for the same campaign ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveScheduleServices.scala
```diff
@@ -0,0 +1,435 @@
...
+    def loop(cursor: Instant, acc: List[ScheduleWindow], remaining: Int): PureResult[List[ScheduleWindow]] = {
+      if (remaining <= 0) Right(acc.reverse)
+      else {
+        // nextCampaignDate is still a Joda-Time API: bridge at that boundary only
+        CampaignDateScheduler.nextCampaignDate(schedule, cursor.toJodaDateTime) match {
+          case Left(err)                 => Left(err)
+          case Right(None)               => Right(acc.reverse)
+          case Right(Some((start, end))) =>
+            val w = ScheduleWindow(start.toJavaInstant, end.toJavaInstant)
+            if (!w.end.isAfter(cursor)) {
```
clarktsiory: isn't that exactly `w.isClosedAt(cursor)` ?

adr/webapp/scheduling/28181-benchmark-schedule-storage-and-ui.md
clarktsiory, line 1: I see that there are existing ADRs related to scheduling (ex. https://docs.rudder.io/devel/adr/webapp/28535-passing-discrete-scheduled-events-from-webapp-to-agent.html) which are not in the `scheduling` directory. I'm fine with subdirectories and categories, but then it needs to be consistent and discoverable. 

More generally we should find a way to make it possible to organize or browse ADRs, current ones in `multi-tenants` are not visible, @amousset would there be an option to have subcategories with our static ADR generation tool ?

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/schedule/DirectiveScheduleEventsTest.scala
```diff
@@ -0,0 +1,312 @@
...
+    }
+  }
+
+  "event ids" should {
+    val id = CampaignId("sched1")
+    val w1 = ScheduleWindow(date(10, 4), date(10, 6))
+    val w2 = ScheduleWindow(date(11, 4), date(11, 6))
+
+    "be deterministic for a same occurrence" in {
+      DirectiveScheduleEvents.eventId(id, w1) === DirectiveScheduleEvents.eventId(id, w1.copy())
```
clarktsiory:
```suggestion
      DirectiveScheduleEvents.eventId(id, w1) === DirectiveScheduleEvents.eventId(id, w1)
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignRepository.scala
```diff
@@ -83,22 +91,41 @@ class CampaignRepositoryImpl(
...
-                                  c <-
-                                    campaignSerializer
-                                      .parse(json.contentAsString)
-                                      .chainError(s"Error when parsing campaign file at '${json.pathAsString}'")
-                                } yield {
-                                  c
-                                }).either
+                                withReadLock(CampaignId(json.nameWithoutExtension)) {
+                                  campaignSerializer
+                                    .parse(json.contentAsString)
```
clarktsiory, line 126: I have no heuristics about ZIO STM locks overhead, compared here with 'getAll' usage: I assume this is fine with the only call sites being: 
* an initial `getAll` at startup on all campaigns for async events processing
* another one for specific API calls to the campaigns API

So it looks fine to me, but it may be worth evaluating the system behavior with a high number of campaigns

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
```diff
@@ -1177,7 +1239,10 @@ object ExecutionBatch extends Loggable {
       modes:                  NodeModeConfig,
       ruleExpectedReports:    RuleExpectedReports,
       timer:                  ComputeComplianceTimer,
-      overrides:              List[OverriddenPolicy]
+      overrides:              List[OverriddenPolicy],
+      schedules:              Map[CampaignId, JsonDirectiveSchedule] = Map.empty,
+      previous:               PreviousCompliance = PreviousCompliance.empty,
+      configBeginDate:        Option[Instant] = None
```
clarktsiory: I don't understand the need for default values, it's only used once with all values provided now, could you remove them ?

clarktsiory (commented):
Globally it looks nice and well-covered with tests

I had only a few questions about the implementation, and a few minor suggestions

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveSchedule.scala
fanf: bad remaining extends from an other copy/pasted code :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveScheduleServices.scala
fanf: Yes, in that case, the "end" date is not really a end for a campaign, it's just the spread to avoid every node chech benchmarks at the same time. So it is not meant to be modifier afterward.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/schedule/DirectiveScheduleServices.scala
fanf: maybe :sweat: - there was a lot of iterative change :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/CampaignRepository.scala
fanf: Yes, we didn't have anything, which is Bad. We have something similar for LDAP, so I think it should be really fine compared to even just a disk access.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/ExecutionBatch.scala
fanf: ok t - I don't think they still have any usage.

fanf:
PR updated with a new commit

PR #7296, opened 2026-07-15
Fixes #29222: Export Group compliance by Node to CSV

skaerg:
https://issues.rudder.io/issues/29222

This PR aims to add an "Export to CSV" button to for both the "global" and "targeted" scopes in a Group's node compliance tab.

global : 
image

targeted : 
image

commit history : 
1. implement new endpoints, i.e. `GET /groups/{groupId}/compliance/global/byNode` and `GET /groups/{groupId}/compliance/targeted/byNode` with a "format" query parameter for both, and add API tests for both the json and csv formats
2. add export button in elm in the "node compliance" tab of a group, for both the "global" and "targeted" scopes
3. update API documentation

### DIFF
--- a/webapp/sources/api-doc/paths/groups/global-compliance-by-node-id.yml
+++ b/webapp/sources/api-doc/paths/groups/global-compliance-by-node-id.yml
@@ -1,0 +1,96 @@
+# SPDX-License-Identifier: CC-BY-SA-2.0
+# SPDX-FileCopyrightText: 2026 Normation SAS
+get:
+  summary: Global compliance details of a group by node
+  description: Get compliance results for all rules that apply to a node within the group, sorted by node
+  operationId: getGlobalNodeGroupComplianceByNodeId
+  parameters:
+    - in: query
+      name: format
+      schema:
+        type: string
+        enum:
+          - csv
+          - json
+        default: json
+      description: format of export
+      style: form
+      explode: false
+    - $ref: ../../components/parameters/compliance-level.yml
+    - $ref: ../../components/parameters/compliance-percent-precision.yml
+    - $ref: ../../components/parameters/target-or-node-group-id.yml
+  responses:
+    "200":
+      description: Success
+      content:
+        application/json:
+          schema:
+            type: object
+            required:
+              - result
+              - action
+              - data
+            properties:
+              result:
+                type: string
+                description: Result of the request
+                enum:
+                  - success
+                  - error
+                example: success
+              action:
+                type: string
+                description: The id of the action
+                enum:
+                  - getGlobalNodeGroupComplianceByNodeId
+              data:
+                type: object
+                required:
+                  - nodes
+                properties:
+                  nodes:
+                    type: array
+                    items:
+                      $ref: ../../components/schemas/group-node-compliance.yml
+        text/plain:
+          schema:
+            type: string
+          example: |
+                "Node","Rule","Directive","Block","Component","Value","Status","Message"
+                "node1.localhost","R1","25. Testing blocks","","directive-techniqueWithBlocks-component-r1-n1","directive-techniqueWithBlocks-component-value-r1-n1","successAlreadyOK",""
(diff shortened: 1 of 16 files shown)


skaerg:
Commit modified

skaerg:
Commit modified

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/Compliance.scala
```diff
@@ -512,6 +512,31 @@ object CsvCompliance {
     }
   }
 
+  private def recurseComponent(
+      component: ComponentStatusReport,
+      block:     List[ComponentField]
```
P4uline: It would be more meaningful to rename `block` to something like `blockAccumulator` as the purpose of this `recurseComponent` seem to perform a traversal tree algorithm collecting all blocks together.

webapp/sources/api-doc/paths/groups/global-compliance-by-node-id.yml
```diff
@@ -0,0 +1,71 @@
...
+          - csv
+          - json
+        default: json
+      description: format of export
+      style: form
+      explode: false
+    - $ref: ../../components/parameters/compliance-level.yml
+    - $ref: ../../components/parameters/compliance-percent-precision.yml
+    - $ref: ../../components/parameters/target-or-node-group-id.yml
+  responses:
```
P4uline, line 22: Error cases are missing. 
For instance it would be useful to the user to know what error status we get when a parameter group id doesn't exist. Is it 404 ? Is it 500 ?

webapp/sources/api-doc/paths/groups/targeted-compliance-by-node-id.yml
```diff
@@ -0,0 +1,70 @@
...
+          - json
+        default: json
+      description: format of export
+      style: form
+      explode: false
+    - $ref: ../../components/parameters/compliance-level.yml
+    - $ref: ../../components/parameters/compliance-percent-precision.yml
+    - $ref: ../../components/parameters/target-or-node-group-id.yml
+  responses:
+    "200":
```
P4uline, line 23: same here, there no error cases documented

webapp/sources/rudder/rudder-rest/src/test/resources/api/api_groups.yml
```diff
@@ -4004,6 +4004,244 @@ response:
   content: |
     "Rule","Directive","Block","Component","Node","Value","Status","Message"
     "R3","directive2","","directive2-component-r3-n1","node1.localhost","directive2-component-value-r3-n1","error",""
+---
+description: Get a node group's global compliance by node in JSON format
+method: GET
+url: /api/latest/groups/g3/compliance/global/byNode
+response:
+  code: 200
```
P4uline, line 4012: As for documentation, a couple of tests about error cases would be nice.

P4uline:
Also it would be nice to have some unit test with very simple examples with simple trees, it would cover the case of nested blocks, and also cover limit cases.

clarktsiory (commented):
I had a quick look and it seems to be a nice implementation, similar to previous ones (I didn't pay attention to details at all)

For existing comments of @P4uline about YAML documentation, it could be painful to maintain all error examples, AFAIK there is no way to factorize it (yet - maybe that's a bigger subject...)

skaerg:
PR updated with a new commit

P4uline:
> I had a quick look and it seems to be a nice implementation, similar to previous ones (I didn't pay attention to details at all)
> 
> For existing comments of @P4uline about YAML documentation, it could be painful to maintain all error examples, AFAIK there is no way to factorize it (yet - maybe that's a bigger subject...)

There was a live discussion about this in parallel. To sum up, this is not about documenting all potential existing error cases, it's about identifying the relevant cases to document, for instance if we pass a non existing group id, what kind of response will we get ? 404, 500 with some message in the body ?

P4uline (approved):

skaerg:
OK, squash merging this PR

PR #7304, opened 2026-07-16
Fixes #29267: Use activity module in dashboard

P4uline:
https://issues.rudder.io/issues/29267

Refactor `Dashboard.elm` to use `Activity` module and get rid of `Dashboard` module which is mostly duplicated code.

Move `Search` and `FilterTypes` under the same type `BodyParameters`.
Move ports, data types, `init` and `view` from `Dashboard` module to `Dashboard.elm`


P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR rebased

P4uline:
PR rebased

webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/JsonEncoder.elm
```diff
@@ -1,24 +1,52 @@
...
+                    ]
+
+            else
+                object
+                    [ ( "draw", int 1 )
+                    , ( "start", int 0 )
+                    , ( "length", int 20 )
+                    , ( "order", list encodeEventLogFilterOrder [ eventLogFilterOrder ] )
+                    , ( "typeFilter", object [ ( "include", list string bodyParameters.filterTypes ) ] )
+                    ]
```
fanf: These 4 blocks are almost identical but for one parameter. That leads to two problems : 
- it's hard to see the logic and what actually change,
- the common parts will likely diverge and in any case it makes maintenance a burden.

Try to create a common object and only change on it what 8s relevant in each branches

fanf (changes requested):
Nice cleaning but sole code is still duplicated 4 times and it should not

webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/JsonEncoder.elm
P4uline: Agree, I already quickly had a look, it's not that simple because of `Json.Encode` api. I will look deeper, there is for sure a way to make it cleaner.

fanf: I was just thinking as something like: 
```
object
    (List.filterMap identity
        [ Just ( "draw", int 1 )
        , Just ( "start", int 0 )
        , Just ( "length", int 20 )
        , Just ( "order", list encodeEventLogFilterOrder [ eventLogFilterOrder ] )
        , if search2String bodyParameters.search /= "" then
            Just ( "search", object [ ( "value", string searchStr ) ] )
          else
            Nothing
        , if List.isEmpty bodyParameters.filterTypes then
            Nothing
          else
            Just ( "typeFilter", object [ ( "include", list string bodyParameters.filterTypes ) ] )
        ]
    )
```

P4uline:
PR updated with a new commit

P4uline:
PR rebased

fanf (approved):
LGTM!

P4uline:
OK, squash merging this PR

PR #7311, opened 2026-07-17
Fixes #29305: Target computation must take into account tenant

fanf:
https://issues.rudder.io/issues/29305

Correct root cause of security-benchmark https://issues.rudder.io/issues/29246 (https://github.com/Normation/rudder-plugins-private/pull/1437). 
The main problem was that `target` (for rules, benchmarks, system updates...) was not tenant-aware. 
Looking at correcting `getNodeIdsChunk` show some other problems: 
- we have duplicate code (`getNodeIdsChunk` and `getNodeIds` that do almost the same thing, but diverged, 
- we compute again and again and again the set of policy servers for a bunch of bad reason: not out of loops, because we use a `MapView` that is realised thousands of time, etc. 

So the correction: 
- delete `getNodeIds`, Set-based, so with very poor performance, 
- Create an opaque type for `(allNodeIds, policyserverId)` : `NodeAndServerIds` so that we know what is what, 
- add in `NodeFactRepository` a method to directly get policy server ID and the set of NodeId with the correct query context. It returns the previous opaque type. That's addressing the root cause of the problem  
- the set of policy server is stored in a cache, because they almost never change and are few . It is updated along with the node count (added in delete method that was missing), 
- use everywhere that `NodeAndServerIds` in place of a map of `nodeId -> isPolicyServer`. 
- where possible, compute the list of target out of a loop to avoid computing it again and again, 
- for all the method like `getNodeIdsChunk`, always use the parameter order `target, groups, nodeAndServerIds` because we had a bit of everything, and it was hard to follow. 

Everything else are just ripples of these changes.

### DIFF
--- a/.claude/skills/rudder-scala/600-security-in-depth.md
+++ b/.claude/skills/rudder-scala/600-security-in-depth.md
@@ -33,6 +33,34 @@ type class to read/update it (see `HasSecurityTag[ActiveTechnique]` in
 - filter by scope in the repository/persistence layer (the right place to enforce —
   e.g. tenant-filtering proxy repositories), not only in the UI.
 
+### Node-id sets: always resolve against QueryContext-filtered facts
+
+Group `serverList`s (and anything derived from them: rule targets, campaign targets…)
+are **not** tenant-filtered and can also contain deleted nodes. Any node-id set
+returned to a user must therefore be **intersected with the facts visible in the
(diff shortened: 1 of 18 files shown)


VinceMacBuche (approved):

VinceMacBuche:
OK, merging this PR

PR #7169, opened 2026-06-03
Fixes #28964: Update test system technique to current templates

fanf:
https://issues.rudder.io/issues/28964

This is uterly non interesting. 
The process followed was: 
- take the last `techniques/system` in branche 9.2 in technique repository, 
- replace existing `rudder-core/src/test/resources/configuration-repository/techniques/system`
- run the `WriteTechniquesTest.scala` test, 
- copy outputs from `/tmp/test-rudder-config-repo-2026-06-03T16:23:57.665+02:00/*` at the correct place in `rudder-core/src/test/resources/configuration-repository/expected-share`
- correct other tests not working (REST for technique version, etc)
- run tests in all plugins to see if there is any problem anywhere (none found)

PR #7312, opened 2026-07-18
Fixes #29316: Replace string template with a fastparse native implementation

fanf:
https://issues.rudder.io/issues/29316

Replace StringTemplate by our own fastparse parser. 

The parser is the hardest, biggest and most uninteresting part (from a rudder architecture PoV) of the change. 
It was claude-fable generated based on our good coverture of policy generation tests. I don't think I would have been able to implement it without weeks of tedious work. Happy it was done in a couple hours. The result is actually pretty nice and it looks like the king of things the bot really excel to do (one file, no big architecture choices, can be compared to other output...). 

It gives byte-for-byte to the string template one on our test corpus which is rather extensive and at least tests all system technique for 9.2 since they were updated for https://github.com/Normation/rudder/pull/7169. 

I kepted the two implementations for now, and they can be configured with a `rudder-web.properties` option:
```
rudder.policy.template.engine=string-template | rudder-fastparse
```
And for now, I let string template being the default. If we are happy with result in alpha, we can change it for beta. 

# Code evolution

The main changes are rather trivial appart the parser by itself (the arch
- create a trait `PolicyTemplateService` for the parsing logic API to isolate string template 
- update `PolicyWriterService` to use the trait 
- create an implementation for string template and for fastparse of the traint, 
- create the feature switch that given what user chose init in `RudderConfig` the string template or fastparse implementation, 

For tests: 
- add a "fastparse" version of our policy writing tests, 
- create a differential test to compare between string template and our parser output of our template cases,

Everything is green. 

# Security relief

We can remove string template (at least in 9.2 not use it), an outdated, not maintained, huge dependency that is a security liability for years now, Just that is a resounding success. 

# Safety - no more mutable template

We had several really sad bugs in the past due to string template being mutable, reusing templates, etc. For example, at some point in time, we had variable from one directive used in an other when load was heavy.  
With an immutable, FT-minded parser, we can't have that anymore by construction. 

# Performance improvements

In addition to the security and safety improvment, we seems to have a nice side effect of: 
- around 30x speed-up in pur template reading/filling, 
- around 1.3x (30%) improvment on policy writing. 

The second one is amazing if it holds in more complex set-up than my tiny env (120 nodes and not that many directives), and on servers that don't have an nvme disk. 

The policy writing should be massively dominated by actually writing the policies on disk. Here, we saw that there was so much contention due to string template not being thread-safe (so we needed mutex) that we win a lot.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
@@ -70,9 +70,8 @@ import com.normation.rudder.services.policies.NodeConfiguration
 import com.normation.rudder.services.policies.ParameterEntry
 import com.normation.rudder.services.policies.Policy
 import com.normation.rudder.services.policies.PolicyId
-import com.normation.templates.FillTemplatesService
-import com.normation.templates.FillTemplateThreadUnsafe
 import com.normation.templates.FillTemplateTimer
+import com.normation.templates.PolicyTemplateService
 import com.normation.templates.STVariable
 import com.normation.zio.*
 import java.nio.charset.Charset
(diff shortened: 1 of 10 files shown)


fanf:
OK, merging this PR

PR #7321, opened 2026-07-22
Fixes #29340: Add link on recent activity id items leading to item detail on change log page

P4uline:
https://issues.rudder.io/issues/29340

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/ActivityTable.elm
+++ b/webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/ActivityTable.elm
@@ -1,23 +1,66 @@
 module Activity.ActivityTable exposing (..)
 
-import Activity.DataTypes exposing (Activity, ActivityMsg)
+import Activity.DataTypes exposing (Activity, ActivityMsg, ContextPath(..))
 import Activity.HtmlParserAdapter exposing (toHtml, toString)
-import Html exposing (text)
-import Html.Attributes exposing (class)
+import Html exposing (Html, a, text)
+import Html.Attributes exposing (class, href)
 import List.Nonempty as NonEmptyList
 import Ordering
 import Rudder.Table exposing (ColumnName(..), buildConfig, buildCustomizations, buildOptions)
 import Time exposing (Zone)
-import Utils.DateUtils exposing (posixToString)
+import Utils.DateUtils exposing (posixToString, posixToStringWithHoursMinutesAndSecondsTo0, posixToStringWithoutTimeZoneOffset)
 
 
-initTable : Zone -> Rudder.Table.Model Activity msg
-initTable timezone =
+initTable : ContextPath -> Zone -> Rudder.Table.Model Activity msg
+initTable (ContextPath contextPath) timezone =
     let
+        {-
+           Add a link on the id to navigate to the detail of this activity log on change log page.
+           Build the json parameters to query on this activity log with the directive id as search value and activity
+           date as start date:
+           {
+             "start":0,
+             "length":5,
+             "search":{"value":"123e4567-e89b-12d3-a456-426614174000","regex":false,"fixed":[]},
+             "startDate":"2026-07-29 10:49:48",
+             "draw":1
+           }
+        -}
+        idWithLink : Activity -> Html msg
+        idWithLink activity =
+            let
+                search id =
+                    "{\"value\":\"" ++ String.fromInt id ++ "\",\"regex\":false,\"fixed\":[]}"
+            in
+            a
+                [ href
+                    (contextPath
+                        ++ "/secure/configurationManager/changeLogs#"
+                        ++ "{\"search\":"
+                        ++ search activity.id
+                        ++ ",\"startDate\":"
+                        ++ "\""
+                        ++ posixToStringWithHoursMinutesAndSecondsTo0 timezone activity.date
+                        ++ "\""
+                        ++ ",\"endDate\":"
+                        ++ "\""
+                        ++ posixToStringWithoutTimeZoneOffset timezone activity.date
+                        ++ "\""
+                        ++ ",\"draw\":1"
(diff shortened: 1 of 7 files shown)


P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/ActivityTable.elm
```diff
@@ -1,23 +1,66 @@
...
+        idWithLink activity =
+            let
+                search id =
+                    "{\"value\":\"" ++ String.fromInt id ++ "\",\"regex\":false,\"fixed\":[]}"
+            in
+            a
+                [ href
+                    (contextPath
+                        ++ "/secure/configurationManager/changeLogs#"
+                        ++ "{\"search\":"
```
amousset, line 39: Isn't there a cleaner way to build a JSON string? this looks brittle.

skaerg: i think that `Json.Encode.encode` could be used to build the JSON filter; here is an excerpt from the documentation of `encode` : 
```
import Json.Encode as Encode

tom : Encode.Value
tom =
    Encode.object
        [ ( "name", Encode.string "Tom" )
        , ( "age", Encode.int 42 )
        ]

compact = Encode.encode 0 tom
-- {"name":"Tom","age":42}
```

webapp/sources/rudder/rudder-web/src/main/elm/sources/Utils/DateUtils.elm
```diff
@@ -60,6 +62,39 @@ posixToString zone time =
...
+        ++ "-"
+        ++ monthToNmbString (Time.toMonth zone time)
+        ++ "-"
+        ++ padded (Time.toDay zone time)
+        ++ " "
+        ++ padded (Time.toHour zone time)
+        ++ ":"
+        ++ padded (Time.toMinute zone time)
+        ++ ":"
+        ++ padded (Time.toSecond zone time)
```
skaerg, line 95: i think it is possible to do the same thing with existing functions in the `DateTime` and `Iso8601` libraries : [example](https://github.com/Normation/rudder/pull/6825/changes#diff-82e5b4f27d8f6f8a88da47afa3539e99181f2f20184be60b87aaa881f526a582R177) (`List.take 10` should just be replaced with `List.take 19` i think)

skaerg (commented):
also, there was a discussion about whether it would be possible to have the collapsible details pane immediately be open when one clicks on the link to a change log; i.e. we would like for the page to look like this :

Image

instead of like this i.e. with the pane closed (as is currently the case) : 

Image

however, i am not sure whether this can easily be achieved; what do you think @RaphaelGauthier @clarktsiory ?
i also feel like the current display is plenty good enough, at least for now, so i do not think this feature is necessary to merge this PR (though it could always be added later, since it makes the webapp nicer to browse)

webapp/sources/rudder/rudder-web/src/main/elm/sources/Utils/DateUtils.elm
clarktsiory: This shouldn't even exist, such format is evil: it is very misleading to not have timezone information, we have enforced in [our ADR](https://docs.rudder.io/devel/adr/webapp/28515-use-java-time.html) that it should ALWAYS be present. 

It's only passed to the REST API (not even for display) and assumed implicitly to be UTC. So, the backend could be easily changed to accept a "full date" with timezone, and still accept "no timezone" and fallback to UTC for compatibility:
https://github.com/Normation/rudder/blob/f346ba4945342dd950a502f89813b32e4c18396f/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala#L146
https://github.com/Normation/rudder/blob/f346ba4945342dd950a502f89813b32e4c18396f/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala#L164-L167

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/changeLogs.js
```diff
@@ -51,6 +51,30 @@ import * as htmlFormatter from '../libs/jsondiffpatch/lib/formatters/html.js'
...
+     * http://localhost:8081/rudder-web/secure/nodeManager/nodes#{"query":{"select":"node","composition":"and","where":[]}}
+     * http://localhost:8081/rudder-web/secure/configurationManager/directiveManagement#{"directiveId":"1c99b738-742b-4444-a964-0759c9fc5b74"}
+     */
+  const param = filterXSS(decodeURIComponent(window.location.hash.substring(1)));
+  function parse(param) {
+      if(Object.keys(param).length === 0)
+          return {};
+      else return JSON.parse(param);
+  }
+  const filterCriteria = parse(param);
```
clarktsiory, line 76: formatting of this JS seems very wrong.
It's very verbose too, it could simply be 
```suggestion
  const filterCriteria = Object.keys(param).length === 0 ? {} : JSON.parse(param);
```

webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/ActivityTable.elm
clarktsiory: I completely agree.
There is only the caveat that we should be careful to use the compact version.
It also seems to work without URL-encoding...

clarktsiory (commented):
It seems to work well, it looks okay overall despite some improvements that could be made. Could you open issues for these @skaerg ?

skaerg (approved):
The suggested improvements will be done as part of this issue : https://issues.rudder.io/issues/29448

skaerg:
OK, squash merging this PR

PR #7324, opened 2026-07-22
Fixes #29351: Error log at boot: Error while deleting ldap entries, error details: Inconsistency: Active technique 'test_import_export_archive' was not found

fanf:
https://issues.rudder.io/issues/29351

So, the problem is that we are not very consistent in the semantic of `delete` a thing, and sometime if the the thing doesn't exist, it's a noop, and other time it is an error. 
The tenant filtering must have the same semantic than the underlying repo, else when the thing is absent, we believe it's because of a right limitation. 

I also changed a `get(id)` to return an option in place of erroring when the thing is absent for rule category, because it makes things simpler to test with tenant and the impact was limited.


P4uline (approved):
LGTM

PR #7331, opened 2026-07-24
Fixes #29365: LDAP error when deleting a node

fanf:
https://issues.rudder.io/issues/29365

Several normalisation things - everything was "working", but we were not doing what is correct on several point: 

- first, we delete the same /var/rudder/share directory  for the node with several process, which is wastfull for nothing. Since the webapp both walk the whole directory for orphan, it can also do the delete - especially since it also call the post hooks, so if there is a problem with the webapp app, hooks are not a safegard. 
- `DeleteMode` and `LdapRemoveNodeBackend` are not used anymore => clean dead code
- I reworked `traceOpenFdOnChange` because the goal of that thing wasn't clear at all. It took me some time to understand that it was some kind of guard to see if the deletion process was increasing a lot the number of open file descriptors, because the default java methods are stupid and do that (so the ZStream approach explained in the comment for `cleanPoliciesRec`). But here, it was a *trace* level log, no chance to be on when needed. And repeated for each directory, so extremelly noisy for nothing. So now, we log a trace level log one time, and a warn if more than 100 new fd are consumed. 
- and finally, the bug correction in `LDAPConnection`: for an LDAP delete operation, a "no such object" is semantically a success, not an error that we don't want to pop-up.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/node-post-deletion/10-clean-policies
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/node-post-deletion/10-clean-policies
@@ -22,7 +22,7 @@
 # and /var/rudder/backup/node-id
 
 # we delete the parent of current (i.e. parent of(/.../node-id/rules)
-NODE_DIR="$(dirname ${RUDDER_POLICIES_DIRECTORY_CURRENT})"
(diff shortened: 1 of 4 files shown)


P4uline (approved):
Walkthrough with @fanf, LGTM

fanf:
OK, merging this PR

PR #7332, opened 2026-07-24
Fixes #29374: Rename fastparse class to actually fastparse

fanf:
https://issues.rudder.io/issues/29374

It is just a renaming

### DIFF
--- a/webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FastparseTemplateParser.scala
+++ b/webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FastparseTemplateParser.scala
@@ -107,7 +107,7 @@ object TemplateAst {
 
 final case class ParsedTemplate(parts: Seq[TemplateAst])
 
-object AmpersandTemplate {
+object FastparseTemplateParser {
   import TemplateAst.*
 
   def parse(content: String): PureResult[ParsedTemplate] = Parser.parse(content)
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-templates/src/main/scala/com/normation/templates/FastparseTemplateParser.scala
```diff
@@ -107,7 +107,7 @@ object TemplateAst {
 
 final case class ParsedTemplate(parts: Seq[TemplateAst])
 
-object AmpersandTemplate {
+object FastparseTemplateParser {
```
amousset, line 110: I think I prefer keeping it named after the template type/syntax rather than the library implementing it (maybe both?)

fanf: I think it's better because we already had an ampersand grammar, and what I think is important from an arch pov is the change to a new engine, because I think when I hear people talk about, them use "st (or string template) technique templates". The fact that the template use "&" seems secondary - it's what they always were. 
 
I want that when we say between us "there is a problem in the xxxx parser", we can look without much more question. 

And finally, usually we don't name things based on what is used to implement it because that say little of the purpose, and implementation changes. But here it's a special case: we are in the "rudder-template" project, whose only goal is to parse ".st" files with ampersand tags and file them. So relatively to that scope, the parser engine feel like an interesting knowledge point - especially if it changes only very 15 years.

skaerg: > I want that when we say between us "there is a problem in the xxxx parser", we can look without much more question.

If so, could this class not be called `TemplateParser.scala` ?

fanf: yes, but it's usefull to be clear into which version of the template. We naturally talk about "the string template parser"* for the pre-9.2 one, so it was about porting that de facto nomenclature to the new instance. 

(*) and not the "string template template parser", because it's strange, but it's what we really mean.

fanf:
So, we discussed it internaly, and it looks like it's ok like that.

PR #7334, opened 2026-07-25
Fixes #29376: Add tenant management to change requests and logs

fanf:
https://issues.rudder.io/issues/29376

Add tenant management to event logs and change requests. 

# Model 

## HasSecurityTag 

Same `HasSecurityTag` than before, added to change events. 
We needed one more aspect for monotony of tenants addition, see below. 

## ReaderScope

We also needed a transient represention of what a user can see to adapt SQL queries with that information, so we introduced the `ReaderScope` object. It is a cleaner separation of the `queryContext.canSee` function. 

The usage is now in `QueryContext.canSee` and for SQL queries, in `TenantSql`. 

Since that thing is used zillions on time in every hot path of Rudder, some care (and imperative code) was used to make it a zero-cost, no allocation abstraction. And so, it's an `opaque type` in place of an `enum` (which would have been simpler and clearer).  

## Change request

A change request can theorically, one day, have several changes on several objects. 
The logic chosen is that an user can see a change request IIF he can see each changes (AND logic). 
Appart from that, nothing special. 

# Serialisation in base of securoty tag for events

The serialisation is in a new `securityTag` column of `EventLog`. 
I chose that to: 
- store in jsonb (same serialisation than other place, events use XML), 
- be able to check for NULL easlely (for pre-tenant cases)
- and have an index and do easy query when getting events for a user.

It also has a gin index for the filtering on tenant IDs on query requests. 

The column is added in `MigrateEventLogAddSecurityTag`. 

Nothing very interesting appart that, the serialisation is in `securityTagMeta`. 

# Monotic growth of tenants on configuration objects

That PR also implements the monotonic growth of tenant on configuration objects, so that a configuration object can only gain tenants, never remove them. That is necessary to keep old change events non viewable, but still allow to change tenants, espacially from "no tenants" (for ex pre-9.2) to "some tenants". 
 
That rule does not apply to a node. Admin can freely change tenant on a node. For ex,that allows to redefine the tenant topology (for configuration objects, you can clone them on such a case). And we don't have the same problem with node, since node history is not read throught the same path as configuration histories. 

This is implemented in `manageUpdate`, which is starting to become quite complicated. 
We have new tests for that, and some older tests that used to change tenants were updated. 

# UI

Well, nothing - the change event snippet is just extending `SecureSnippet` now.


fanf:
PR updated with a new commit

clarktsiory (approved):

PR #7338, opened 2026-07-27
Fixes #29390: Multi-tenants: close filtering gaps (core)

fanf:
https://issues.rudder.io/issues/29390

So, this is a big PR that closes (I think) all remaining multi-tenant filtering gap. 
It's always a big the same thing: add `QueryContext` - then use it. 
It is based on https://github.com/Normation/rudder/pull/7334

# NodeGroup serialised list of nodes

They need to be filtered in API in particular to avoid node id leaks. 

# Node reports repository and logger and change service

- add `QueryContext` in `ReportsJdbcRepository` and `NodeChangesService`,
- also change signature from `Box` to `IOResult`, 
- and remove dead code. 

# "isApplied" for rules

It's more complicated than before tenant, because it depends of the node actually in user scope. 

# API

- keep the group tenant in "group properties" cache (in `TenantScopedGroupProps`) so that it can be used in apis, 
- several API were consolidated around tenant (checking, filtering content, weaving qc...)

# Tenant "delegableTo"

Encode the idea that when you have a set of tenant, you should not give more - even when you are admin.

# Deployment status

Filter nodes and there error based on user tenant.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
@@ -42,6 +42,7 @@ import com.normation.inventory.domain.InventoryError.Inconsistency
 import com.normation.inventory.domain.NodeId
 import com.normation.rudder.facts.nodes.NodeFactRepository
 import com.normation.rudder.tenants.QueryContext
+import com.normation.rudder.tenants.TenantAccessGrant
 import com.normation.zio.*
 import zio.*
 import zio.syntax.ToZio
@@ -155,8 +156,17 @@ class ScoreServiceImpl(
     }
   }
 
+  // write-side tenant guard: a tenant-restricted actor must not mutate the score of a node it can
+  // not see. `All` (the system paths - node-deletion callbacks) skips the check, so cleanup still runs even
+  // when the node fact has already been removed.
+  private def checkNodeWritable(nodeId: NodeId)(implicit qc: QueryContext): IOResult[Unit] = {
+    if (qc.accessGrant == TenantAccessGrant.All) ZIO.unit
+    else nodeFactRepo.get(nodeId).notOptional(s"Node '${nodeId.value}' is not accessible").unit
+  }
+
   def deleteNodeScore(nodeId: NodeId)(implicit qc: QueryContext): IOResult[Unit] = {
     for {
+      _         <- checkNodeWritable(nodeId)
       _         <- cache.update(_.removed(nodeId))
       newScores <- scoreCache.updateAndGet(_.removed(nodeId))
       _         <- scoreRepository.deleteScore(Seq(nodeId), None)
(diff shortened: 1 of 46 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/score/ScoreService.scala
```diff
@@ -155,8 +156,17 @@ class ScoreServiceImpl(
...
   }
 
+  // write-side tenant guard (H7 shape): a tenant-restricted actor must not mutate the score of a node it can
+  // not see. `All` (the system paths - node-deletion callbacks) skips the check, so cleanup still runs even
+  // when the node fact has already been removed.
+  private def checkNodeWritable(nodeId: NodeId)(implicit qc: QueryContext): IOResult[Unit] = {
+    if (qc.accessGrant == TenantAccessGrant.All) ZIO.unit
+    else nodeFactRepo.get(nodeId).notOptional(s"Node '${nodeId.value}' is not accessible").unit
+  }
+
```
clarktsiory, line 166: it's weird that "write" check only checks read access, shouldn't there be a `canModify` instead of just `nodeFactRepo.get` ?

clarktsiory: so I guess it's because it's nodes, and not configuration objects, so nodes write and read tenants attribution are equivalent

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/tenants/ChangeRequestTenantTest.scala
```diff
@@ -142,4 +142,22 @@ class ChangeRequestTenantTest extends Specification {
       checkTenant.checkChangeRequestModify(crZoneA, zoneAro.newCC()).either.runNow must beLeft
     }
   }
+
+  // B4: on create, the tenant tag an admin chooses must reference only EXISTING tenants (like the update path),
+  // otherwise a dangling/phantom tenant tag is created.
+  "[create] admin-chosen tenant tag must reference existing tenants (B4)" should {
+    val existing = TenantStatus.Enabled(Set(TenantId("zoneA"), TenantId("zoneB")))
+    def create(r: Rule): Either[?, ?] =
```
clarktsiory, line 150: it's too bad to have a `Either[?, ?]`, I guess that leaving it to type inference was not possible ?

clarktsiory (commented):
Only a question and nitpicking, LGTM

I'll rebase since #7334 was merged

clarktsiory:
PR rebased

clarktsiory (approved):

clarktsiory:
OK, squash merging this PR

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/tenants/ChangeRequestTenantTest.scala
fanf: I honestly don't remember

PR #7341, opened 2026-07-27
Fixes #29398: Sanitize inventory signature path

fanf:
https://issues.rudder.io/issues/29398

- resolve symlink on the deepest existing ancestor 
- extract the signature name part to make it testable
- sanitize both inventory and signature path

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/InventoryApi.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/InventoryApi.scala
@@ -62,28 +62,23 @@ object InventoryApi {
 
   val sigExtension = ".sign"
 
-  /*
-   * Compute the on-disk name to use for the uploaded signature file.
-   *
-   * SECURITY:`rawSigFileName` is the attacker-controlled multipart file name. We only ever work
-   * on its basename (`File(_).name`), so that a path such as `pwn/../../../etc/cron.d/pwn` can never
-   * escape the incoming inventory directory. For a legitimate upload (a bare file name, no directory)
-   * this is byte-identical to the previous behaviour.
-   *
-   * We keep the `.gz` compression marker and ensure a `.sign` extension when the uploaded signature
-   * does not already carry one.
-   */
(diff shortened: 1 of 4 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/InventoryApi.scala
```diff
@@ -50,13 +50,44 @@ import com.normation.rudder.rest.RestUtils.effectiveResponse
...
+   *
+   * We keep the `.gz` compression marker and ensure a `.sign` extension when the uploaded signature
+   * does not already carry one.
+   */
+  def signatureFileName(rawSigFileName: String): String = {
+    val sigFilename = File(rawSigFileName).name
+    // remove gz extension for sig name comparison
+    val simpleOrig  =
+      if (sigFilename.endsWith(".gz")) sigFilename.substring(0, sigFilename.length - 3) else sigFilename
+    if (sigFilename.startsWith(simpleOrig)) { // assume extension is ok
```
amousset: this is always true, isn't it?

fanf: Yes, it wasn't cleaned in the move and doesn't make sense anymore

fanf: Reading more carefully, the old code make no sense whatesoever. It seems that at some point a "originalFilename" was swapped with a "signatureFile" - the idea was to compare inventory name and its signature file name, not signature with itself

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/lift/InventoryApi.scala
fanf: There, now it makes some sense. And I tested with all the combination of .. and gz I thought of.

amousset (approved):

PR #7343, opened 2026-07-28
Fixes #29405: Remove DeleteMode and clean-up removed node LDAP branch

fanf:
https://issues.rudder.io/issues/29405

This is a loooong overdue cleaning that should have done in ... 8.0. 
We used to move deleted inventories in a "ou=Removed Inventories" branch. That was a bad idea, so circa Rudder 5, we stared deleting them (configurable with a parameter). Then circa 8.0.0, we removed the parameter (https://github.com/Normation/rudder/pull/5047) and hardcoded the erase choice. 
But we didn't removed all the dead code and dead logic of removed inventories. 

Since we removed even more things in https://github.com/Normation/rudder/pull/7331, there's really no reason to keep all that. 

The changes: 
- delete `ou=Removed Inventories,ou=Inventories,cn=rudder-configuration` with a migration
- remove logic around purge of inventories in that branch, 
- delete any reference of that `ou` in bootstrap ldif, tests, etc
- delete `removedDIT` parameter everywhere, 
- delete `RemovedDit` status and change `getStatus` to `Option[NodeId]` when node isn't present, 
- remove `deleteMode` parameter in API and doc (it was already unused actually)
- for API returned values, keep a `JRRemovedInventory` that say "deleted" for the node info after a delete action


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM, `PurgeDeletedNode.scala` is a remnant too, I'll make another PR to delete it

PR #7345, opened 2026-07-28
Fixes #29409: Allow to isolate global properties on a target (a set of nodes)

fanf:
https://issues.rudder.io/issues/29409

Add a `scope` parameter to global properties so that they can be restricted to a given target. A scoped global property is thus only available on nodes that belong to the target, and not on all node. 

The ADR describes in details the chosen design, but in summary: 
- add a `Target` king of property in the merge algo, 
- for groups, it does change anything so that user 
- for a node, the merge is done and then we filter out any scoped properties is the node is not in the target. 

Note: a user which would override a scoped global property overridden on a group, if the node is not in the target, would not have the property after merge since it is filtered out. That's the desired behavior but could be surprising in the general case; 
For us and for now, it can't happen because only benchmark can use scoped global properties and the benchmark UI take care of that and benchmark properties are hidden.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/properties/NodePropertiesService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/properties/NodePropertiesService.scala
@@ -38,14 +38,19 @@
 package com.normation.rudder.properties
 
 import com.normation.errors.*
+import com.normation.inventory.domain.NodeId
 import com.normation.rudder.domain.logger.NodePropertiesLoggerPure
 import com.normation.rudder.domain.properties.FailedNodePropertyHierarchy
+import com.normation.rudder.domain.properties.GlobalParameter
+import com.normation.rudder.domain.properties.ParentProperty
 import com.normation.rudder.domain.properties.SuccessNodePropertyHierarchy
 import com.normation.rudder.facts.nodes.NodeFactRepository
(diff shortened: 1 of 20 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/properties/NodePropertiesService.scala
```diff
@@ -120,8 +132,28 @@ class NodePropertiesServiceImpl(
...
+  )(implicit qc: QueryContext): IOResult[(Map[String, GlobalParameter], List[(Set[NodeId], ParentProperty.Target)])] = {
+    // the qc does the tenant filtering, and policy servers are cached in the repository
+    nodeFactRepository.getNodeAndServerIds().map { ids =>
+      val split = params.toList.map { p =>
+        p.scope match {
+          case None    => Left((p.name, p))
+          case Some(t) => Right((groups.getNodeIds(Set(t), ids), ParentProperty.Target(t, p, None)))
+        }
+      }
+      (split.collect { case Left(x) => x }.toMap, split.collect { case Right(x) => x })
```
clarktsiory, line 156: looks a lot like partitionMap, which is the name of the method, but it's okay to keep it like that since there is an additional `toMap` of left-side here

clarktsiory (approved):
LGTM, looks like the right design from the ADR

PR #7350, opened 2026-07-29
Fixes #29413: Remove interpolation of $rudder.parameters in webapp and rudderc

fanf:
https://issues.rudder.io/issues/29413

So, this is the first part of the 9.2 and 10.0 migration out of rudder parameters to only keep inherited node properties, both in server side and on agent side. 

It handles the removing of the different kind of ${rudder.parameters} syntax interpolation in webapp and rudderc. 
In that first step, we keep `rudder-parameters.json` so that on the agent side, nothing change. 

# rudderc

On rudderc, it's rather simple and mechanic: it means transforming the parsing of `${rudder.parameters` and `${rudder_parameters` (a new one) into technique compilation error - they should go up to rudder policy generation notification zone. 

# webapp

On webapp, it's much more impactful, but in a good way: we can clean a lot of code. 

## Erroring on parsing

The real correction is replacing the parsing case for `${rudder.param.X}` and `${rudder.parameters[X]}` as error failing the policy generation, like for rudderc. 

## Interpolation value compiler simplification

But most of it is in [‎webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/InterpolatedValueCompiler.scala‎](https://github.com/Normation/rudder/pull/7350/changes#diff-19db967ed8a6fae2a6eb1c7d6ebb0ec2285e32a0c0268854209243a888a496e8) where lives the logic to do something with `${rudder.param.X}` and `${rudder.parameters[X]}`. 

This was interpolated in the resolution of global parameters (because a global parameters used to be able to use the expanded value of another parameter) and in directive parameters. 

All that code is now useless and removed. 
`translateToAgent` is a special case: it was already dead code and is now cleaned-up. 

# visibility == hidden case

In addition since it's the same code logic and as a side effect, we remove hidden global properties to end up in `rudder-parameters.json`. This only impact benchmarks properties, which are huge and in any case the one used live in `properties.d/properties.json` because we need the overridden ones. So we just remove a useless, almost duplication.

### DIFF
--- a/policies/rudderc/src/compiler.rs
+++ b/policies/rudderc/src/compiler.rs
@@ -388,6 +388,11 @@ fn lint_expression(s: &str) -> Result<()> {
             warn!("Error parsing '{}': {:?}", s, e);
         }
         Ok(e) => {
+            // unlike the other lints, using a removed global parameter fails the compilation
+            if let Err(err) = e.check_no_global_parameter() {
+                error!("Error checking '{}': {:?}", s, err);
+                user_error()
+            }
             if let Err(e) = e.lint() {
                 error!("Error checking '{}': {:?}", s, e);
             }
--- a/policies/rudderc/src/ir/value.rs
+++ b/policies/rudderc/src/ir/value.rs
@@ -15,7 +15,7 @@
 
 use std::{cmp::Ordering, fmt::Debug, str::FromStr, sync::OnceLock};
 
-use anyhow::{Error, Result, bail};
+use anyhow::{Error, Result, anyhow, bail};
 use nom::{
     Finish, IResult, Parser,
     branch::alt,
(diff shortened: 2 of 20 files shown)


clarktsiory:
LGTM for the webapp change, I tested the error message and generation errors seems to work as expected.
I didn't check technique compilation and Rust part, could you have a look @amousset ?

amousset:
The Rust side is ok for me.

amousset (approved):

clarktsiory:
PR rebased

clarktsiory (approved):

clarktsiory:
PR updated with a new commit

clarktsiory:
OK, squash merging this PR

PR #7351, opened 2026-07-29
Fixes #29417: Rename the session cookie

amousset:
https://issues.rudder.io/issues/29417

* Add the `__Secure-` prefix to harden a bit the cookie on the brwoser side, see https://developer.mozilla.org/en-US/docs/Web/Security/Practical_implementation_guides/Cookies#name
* Rename it to avoid being flagged at each audit.

Only in 9.2 as the security improvement is small, and the impact could not be totally invisible (tough we don't seem to hardcode `JSESSIONID` anywhere, which is good)


fanf (approved):
The end of a period

PR #7352, opened 2026-07-29
Fixes #29233: Add recent activity table in global parameters details

P4uline:
https://issues.rudder.io/issues/29233

Screenshot From 2026-07-30 11-13-56

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/ParameterManagement.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/ParameterManagement.scala
@@ -204,10 +204,16 @@ class ParameterManagement extends SecureDispatchSnippet with Loggable {
             "pageLength": 25
           });""") &
       JsRaw(s"""
-        /* Formating function for row details */
+
+          /* Formating function for row details */
           function fnFormatDetails(id) {
             const sOut = '<span id="'+id+'" class="parametersDescriptionDetails"/>';
             return sOut;
+          };
+
+          function fnFormatActivity(id) {
+            // FIXME : add some CSS for activity table from elm app
+            return '<span><ul class="ms-2"><li><b>Recent activity:</b></li></ul></span><div class="parametersDescriptionDetails"><div id="globalPropertiesRecentActivityApp"></div></div>'
           };
 
           ${jsVarNameForId(gridName)}.rows().nodes().to$$().each( function () {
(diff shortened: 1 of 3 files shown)


P4uline:
PR updated with a new commit

P4uline:
PR rebased

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/ParameterManagement.scala
```diff
@@ -204,12 +204,18 @@ class ParameterManagement extends SecureDispatchSnippet with Loggable {
...
-        /* Formating function for row details */
+
+          /* Formating function for row details */
           function fnFormatDetails(id) {
             const sOut = '<span id="'+id+'" class="parametersDescriptionDetails"/>';
             return sOut;
           };
 
+          function fnFormatActivity(id) {
+            // FIXME : add some CSS for activity table from elm app
```
skaerg, line 215: indeed, i think the CSS should be modified : i think the table should have a border and empty space all around it in order to separate it from the rest

Image

did you have other ideas @P4uline ?

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/ParameterManagement.scala
```diff
@@ -231,12 +237,45 @@ class ParameterManagement extends SecureDispatchSnippet with Loggable {
                     color = 'color2';
                   const row = ${jsVarNameForId(
           gridName
-        )}.row(this).child(fnFormatDetails(jsid), color + ' parametersDescription details').show();
+        )}.row(this);
+
+                  const globalPropertyName = jTr.find('td.name').find('b').html();
```
skaerg, line 242: it would be nice to fetch the name of the property in a more robust way eventually.
fetching the name of the property in this way would not be necessary had the entire global properties table been contained in a single Elm app, but migrating it to Elm is not really feasible with the current state of the `Rudder.Table` module as it does not support many of the key features of the properties table. 

in the meantime, do you see a cleaner way of fetching the property name @clarktsiory @RaphaelGauthier ?

skaerg (commented):
eventually (not in this PR of course), the entire global properties table could be migrated to Elm, so that usages of datatable can be removed from this page as well. 

we could start by also migrating the rest of the content of the dropdown (i.e. the description field) so that it can be part of the same Elm app as the recent activity table, and, eventually, the entire table of global properties could be migrated to Elm. 

we cannot migrate the entire global properties table right now with the current `Table` module of the Elm library, because some key features of the global properties table would be missing : 
- each row, when clicked, opens a new pane that contains the description and recent activity table of the given global property. the row can be clicked again to collapse the pane.
- when the details pane is open, there is a small downwards triangle on the bottom of the row : 
Image

- the current global properties table allows the user to only display a certain number of entries (i.e. paging)
Image



do you see any other features that must be added to the Table module before the global properties table can be migrated to Elm @clarktsiory ?

clarktsiory:
Global properties are indeed specific, and as you reported need some additional table features:
* collapsible per-row content
* pagination

There are also other tables that would need these in order to be migrated in Elm (for example change logs).


But the UI may also need to be revised: it has quite diverged from the node and group properties UI (which are the same Elm app). I'd favor consistency with other "properties UI" to plain migration of an "old UI".

So, I don't think we should prioritize having a "generic table", to having "consistent properties UI across webapp". This should be discussed with @RaphaelGauthier

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/ParameterManagement.scala
clarktsiory: it's acceptable that you modify it in another PR, that's a kind of iterative fix/improvement you can do in alpha

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/configuration/ParameterManagement.scala
clarktsiory: I completely agree that it's not ideal to rely on potentially unstable CSS selectors (such as `b`) to find the name of the global property... The current `td.name` looks fine to me, but I'd prefer directly `find('td.name').text()` with further sanitization.

As for a cleaner way to do that, I don't know any other solution than CSS selectors at this level...

skaerg (approved):
The suggested improvements will be done as part of this issue : https://issues.rudder.io/issues/29447

skaerg:
OK, squash merging this PR

PR #7354, opened 2026-07-31
Fixes #29432: Small hardening fixes in relayd

amousset:
https://issues.rudder.io/issues/29432

Pass `--` before service names in systemctl invocations, so a name can never be parsed as an option.
    
Shared files:
    
- Derive the metadata path from the walked entry in the cleanup job instead of rebuilding it with with_extension("metadata"), which replaced the last extension of a dotted file id: for `application.properties` it removed the content file but left the metadata behind, so head kept answering 200 for a file that no longer existed, and the sender only re-uploads on 404.
- Set the expires field on the parsed metadata instead of appending a line to its serialization. Uploaded metadata already containing `expires` produced a duplicate header, which the parser rejects: the file never expired and its HEAD answered 500 forever. The date computed from the ttl parameter always overrides the one sent by the client.
- Reject a file_id using the reserved `.metadata` extension. It collided with the metadata of another file of the same source, letting an upload overwrite it and making the cleanup job read file content as metadata.
- Reject `.` and `..` as source_id, target_id or file_id. They match the allowed characters but do not name a file, and moved the path they are built into.
- Answer 400 instead of 500 for these invalid ids, and log them as warnings: they are client errors, not internal ones.
    
Watched directories:
    
- Handle errors per entry in the catch-up listing instead of returning from the task. A file removed between the listing and the metadata read, which happens as processed files are deleted concurrently, killed the only mechanism recovering the files inotify missed, silently, as the JoinHandle is dropped. Those reports then sat until the retention sweep deleted them.
- Give up the cleanup pass on a listing error rather than trying the next entry, which spun at full speed and grew the logs for as long as the error lasted. The next tick lists the directory again.


relay/sources/relayd/src/http_client.rs
```diff
@@ -85,7 +85,7 @@ impl HttpClientBuilder {
 
     pub fn no_verify(self) -> Result<HttpClient, Error> {
         debug!("Creating HTTP client with no certificate verification");
-        Ok(HttpClient::System(
+        Ok(HttpClient::NoVerify(
```
amousset, line 88: copy-paste error

m4rtinh4rt (approved):

amousset:
OK, merging this PR

PR #7356, opened 2026-07-31
Fixes #29435: Escape js in rule category

fanf:
https://issues.rudder.io/issues/29435


clarktsiory (approved):
LGTM, works as expected
I tested other fields and they seem to work as documented in comments

clarktsiory:
OK, merging this PR

PR #7360, opened 2026-08-03
Fixes #29445: webapp non-root: disable chown of group policy owner

m4rtinh4rt:
https://issues.rudder.io/issues/29445

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/20-mv-policies
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/20-mv-policies
@@ -1,0 +1,3 @@
+#!/usr/bin/env bash
+
+/usr/bin/rsync -a --delete --chown=root:root /var/rudder/share/policies/ /var/rudder/cfengine-community/inputs/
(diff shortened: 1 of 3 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/PolicyWriterService.scala
amousset: we can do a bit of cleanup by removing the useless option

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

m4rtinh4rt:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-finished/20-mv-policies
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+/usr/bin/rsync -a --delete --chown=root:root /var/rudder/share/policies/ /var/rudder/cfengine-community/inputs/
```
amousset, line 3: `/var/rudder/share/policies/` this path is not great it does not reflects its content at all.

amousset:
OK, squash merging this PR

PR #7369, opened 2026-08-06
Fixes #29154: nodelastcompliance table can get *really* big

clarktsiory:
https://issues.rudder.io/issues/29154

Schedule a "vacuum full" every day, using existing `databasecleaner` configuration values (by default at midnight)

The architecture of the change is about:
* adding a generic `JdbcVacuum` generic trait, which may be useful for other future vacuum on other tables
* adding a `JdbcVacuumFull` generic implementation with `VACUUM FULL` command
* adding an implementation with 'vacuum full' for nodelastcompliance

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/JdbcVacuum.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/JdbcVacuum.scala
@@ -1,0 +1,82 @@
+/*
+ *************************************************************************************
+ * Copyright 2026 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
+ * Versions) and 6 (6. Conveying Non-Source Forms.) of the GNU General
+ * Public License version 3, when you create a Related Module, this
+ * Related Module is not considered as a part of the work and may be
+ * distributed under the license agreement of your choice.
+ * A "Related Module" means a set of sources files including their
+ * documentation that, without modification of the Source Code, enables
+ * supplementary functions or services in addition to those offered by
+ * the Software.
+ *
+ * Rudder is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
+ *
+ *************************************************************************************
+ */
+
+package com.normation.rudder.repository.jdbc
+
+import cats.syntax.apply.*
+import com.normation.errors.IOResult
+import com.normation.rudder.db.Doobie
+import doobie.*
+import doobie.implicits.*
+import zio.*
+import zio.interop.catz.*
+
+/**
+ * Interface for vacuuming/maintaining a table:
+ * this is a Postgres-specific maintenance operation, which can be combined,
+ * e.g. as batch database maintenance operations defined in the webapp
+ */
+sealed trait JdbcVacuum {
+  def vacuum(): IOResult[Unit]
+}
+
+object JdbcVacuum {
+  // all the vacuum operations to run in the database maintenance schedule
+  def all(doobie: Doobie): List[JdbcVacuum] = List(
+    new JdbcVacuumFull("NodeLastCompliance")(doobie)
+  )
+}
+
+/**
+ * Postgres VACUUM FULL implementation: rewrites the entire table, acquiring a lock,
+ * but clears more disk space.
+ *
+ * Regular VACUUM only flags reusable (dead) tuples which makes the table keep its size.
+ * VACUUM FULL actually rewrites the table and truncates the file, which is what
+ * reclaims the actual disk space on disk.
+ */
+class JdbcVacuumFull(table: String)(doobie: Doobie) extends JdbcVacuum {
+
+  import doobie.*
+
+  override def vacuum(): IOResult[Unit] = {
+    val query = s"VACUUM FULL ${table}"
+
+    transactIOResult(s"error when vacuuming full table ${table}")(xa =>
+      (FC.setAutoCommit(true) *> Update0(query, None).run <* FC.setAutoCommit(false)).transact(xa)
+    ).unit
+  }
+}
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
@@ -3431,6 +3431,14 @@ object RudderConfigInit {
       )
     }
 
+    lazy val jdbcMaintenanceSchedule: JdbcMaintenanceSchedule = {
+      JdbcMaintenanceSchedule.make(
+        JdbcVacuum.all(doobie),
+        hour = RUDDER_BATCH_DATABASECLEANER_RUNTIME_HOUR,
+        minute = RUDDER_BATCH_DATABASECLEANER_RUNTIME_MINUTE
+      )
+    }
+
     lazy val techniqueLibraryUpdater = new CheckTechniqueLibrary(
       techniqueRepositoryImpl,
       techniqueStatusService,
@@ -3980,6 +3988,7 @@ object RudderConfigInit {
     cleanOldInventoryBatch.start()
     gitFactRepoGC.start()
     gitConfigRepoGC.start()
+    jdbcMaintenanceSchedule.start()
     rudderUserListProvider.registerCallback(UserRepositoryUpdateOnFileReload.createCallback(userRepository))
     userCleanupBatch.start()
 
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/JdbcMaintenanceSchedule.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/batch/JdbcMaintenanceSchedule.scala
@@ -1,0 +1,95 @@
+/*
+ *************************************************************************************
+ * Copyright 2026 Normation SAS
+ *************************************************************************************
+ *
+ * This file is part of Rudder.
+ *
+ * Rudder is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * In accordance with the terms of section 7 (7. Additional Terms.) of
+ * the GNU General Public License version 3, the copyright holders add
+ * the following Additional permissions:
+ * Notwithstanding to the terms of section 5 (5. Conveying Modified Source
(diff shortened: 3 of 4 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
```diff
@@ -284,4 +292,13 @@ class JdbcNodeStatusReportStorage(doobie: Doobie, jdbcBatchSize: Int) extends No
     transactIOResult(s"error when saving compliance for nodes")(xa => query.transact(xa)).unit
   }
 
+  override def vacuum(): IOResult[Unit] = {
```
ncharles: it should probably be called vacuum_full rather than vacuum, otherwise we might assume tit is only a vacuum and use it by mistake

ncharles (changes requested):
the logic sounds ok, but i'd rather have vacuum called vacuum full
someone else should review the scala

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -205,6 +205,9 @@ CREATE TABLE NodeLastCompliance (
 , details             jsonb NOT NULL
 );
 
+ALTER TABLE NodeLastCompliance SET (autovacuum_vacuum_threshold = 0);
+ALTER TABLE NodeLastCompliance SET (autovacuum_vacuum_scale_factor = 0.05);
```
fanf: Can you add a comment explaining what is the goal of that? Something like: 
"Vaccum analysis is `dead_tuples > vacuum_threshold + vacuum_scale_factor * reltuples`, with default value `dead_tuples > 50 + 0.2 * reltuples`. Here, we want purely proportionnal, 5% bloat leading to eligibility".

fanf: But I'm not sure these value are important for the patch. The problem was that no vaccum was done (or able to complete), why the defaul values aren't OK?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
```diff
@@ -102,6 +103,9 @@ trait NodeStatusReportStorage {
 
   // this method erase these node status reports
   def delete(nodes: Iterable[NodeId]): IOResult[Unit]
+
+  // this method vacuums the node status reports as storage maintenance operation (e.g. postgres jdbc)
+  def vacuum(): IOResult[Unit]
```
fanf: I don't think that should be on the storage trait. This is specific not only of jdbc, but even on the specific case of postgresql - it's really an implementation details and not the generic storage interface.

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -3437,6 +3437,15 @@ object RudderConfigInit {
       )
     }
 
+    lazy val nodeStatusReportVacuum: NodeStatusReportVacuum = {
+      val storage = new JdbcNodeStatusReportStorage(doobie, RUDDER_JDBC_BATCH_MAX_SIZE)
```
fanf: We already have a storage instance in `nodeStatusReportRepository` and we don't declare several instance of these things. They are meant to be singleton. 
But this is just a symptom of the other design problem.

fanf (changes requested):
I think the design is coupling things that should remain independant. 

We have a clue in the actual current implement:  
- vacuum, it's really not a storage that is needed. You don't use *any* of the storage API in the vaccuum. 
- this is reciprocal: the storage interface doesn't need vaccum, it's a completely orthogonal subject to its goal. 

So from a design / interface standpoints, we really have two interfaces responsible of different things, that should just not be artificially coupled: 
- the `NodeStatusReportStorage` interface handle action for storing reports *whatever the actual storage implementation*. 
- we need a new API for a service *meant to handle database vacuum* - at least for that table, but certainly for other too. It is something that knows that we are dealing with Postgresql for Rudder specifically. The API is the design contract (the service is started here, etc) and impementation handle the properties. For exemple, we could imagine have several implementation, more or less agressive, that could be configured via rudder param, or whatever. 

And I'm not sure we need to change the PostgreSQL defaults. The bug was likely that no vacuum full was done. 
Which means we should try to understand why no autovacuum was set by default, and what table / general strategy should be used. 

The current vacuum parameters looks rather ad-hoc in the SQL schema file, and the same goes for scheduler services for that. 

So perhaps we need to handle that in two time: 
- this bug correction: just the `NodeStatusReportVacuum` interface + implem
- another one, likely targeting 9.2: normalize under a common "postgresql cleaning service" all vacuum and other operations needed to handle each tables, with a sane default applied to all table and specific tables having specific needs all in one place.

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
clarktsiory: I added the description, which explains our choice: it's quite a small table (number of rows is the total number of nodes, except from the server), it would be an optimization for the autovacuum

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/reports/NodeStatusReportRepository.scala
clarktsiory: It ended up being in the class name: `NodeLastComplianceVacuumFull`

clarktsiory:
@fanf I addressed the comments, the suggestion was great to have generic "jdbc vacuum", I added a generic interface for that, with the "vacuum full" implementation for NodeLastCompliance

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
fanf: OK, I still don't get why we want that to be proportionnal to the number of nodes. 
It means that we accept to have more bloat (pur, useless bloat since things are just wholly rewritten in that table) for bigger base - but why ? It's linked to the number of time we trigger the auto vacuum, but given the table, is it really better to wait longer ? (I really don't now, but that doesn't seem obvious. I would have say something like pure threshold base, every 100 dirty column or something like that - does it needs a full table scan ?)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/JdbcVacuum.scala
```diff
@@ -0,0 +1,76 @@
...
+
+private class JdbcVacuumFull(table: String)(doobie: Doobie) extends JdbcVacuum {
+
+  import doobie.*
+
+  override def vacuum(): IOResult[Unit] = {
+    val query = s"VACUUM FULL ${table}"
+
+    transactIOResult(s"error when vacuuming full table ${table}")(xa =>
+      (FC.setAutoCommit(true) *> Update0(query, None).run <* FC.setAutoCommit(false)).transact(xa)
```
fanf, line 79: why do you need autocommit here ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/JdbcVacuum.scala
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * Interface for vacuuming/maintaining on a table:
+ * this is a Postgres-specific maintenance operation, which can be combined,
+ * e.g. as batch database maintenance operations defined in the webapp
+ */
+sealed trait JdbcVacuum {
+  def vacuum(): IOResult[Unit]
+}
+
+private class JdbcVacuumFull(table: String)(doobie: Doobie) extends JdbcVacuum {
```
fanf: I think that that service should take a list of table. I would much prefer to have all the table needing a specific postgresql maintainance set-up be groupes together, so that we can analyse that on one place only. And so, no `NodeLastComplianceVacuumFull`.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/JdbcVacuum.scala
clarktsiory: VACUUM is a DDL statement, it returned an error that it cannot be run inside a transaction (the doobie default): here is the doobie documentation for that exact topic: https://typelevel.org/doobie/docs/17-FAQ.html#how-do-i-run-something-outside-of-a-transaction-

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
clarktsiory: it's already heavily reduced from the quoted default (`50 + 0.2 * rows` default VS `0 + 0.05 * rows` now), it is like so by default in the postgres autovacuuming model surely because of the complexity of vacuum: it requires full scan, so full scan of large tables shouldn't be dependent on a fixed threshold potential too small and causing full scan to be costly...
In any case, for a small base (mostly the case for ours here), proportional is aggressive and so would relieve the "vacuum full" during the declared maintenance window, which was the initial purpose for tuning the auto-vacuum in this PR

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/JdbcVacuum.scala
clarktsiory: I agree but that kind of model would be the one we want for 9.2 because it is very generic and out of scope for this bugfix (implementing that would require abstracting also over the schedule, in addition to the list of tables)

> likely targeting 9.2: normalize under a common "postgresql cleaning service" all vacuum and other operations needed to handle each tables, with a sane default applied to all table and specific tables having specific needs all in one place.

fanf: why abstracting over the schedule? All table can have the same maintenance here (it's likely just that one in any case), and if we need more configurabitility, we can do that in a minor, yes.

clarktsiory: Okay then, if we have the same maintenance window for all tables it can be changed to something like "JdbcMaintenanceSchedule"

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
fanf: I don't want to now about the default etc, I want to know why `0 + 0.05 * rows` is the correct configuration for that case, the rationnal leading to these numbers, hopefully with reference to metrics or postgresql doc or reasoning, that THAT justification is written down. 

The goal is to be able to review it in the future so that it's actionnable: do the hypothesis changed ? Some important aspect of the data ? Is there thresholds ? etc. 

And it the number is pure guessing, that the comment say so ("the number are guessed and should be changed if any metrics lead to thing they are incorrect").

clarktsiory:
PR updated with a new commit

clarktsiory:
as discussed IRL, we will for now correct the main issue with only the "daily vacuum full".
Then we will do some analysis for the autovacuuming part, which seems is another concern, not necessarily linked to the reported issue

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
clarktsiory: As discussed, it will be addressed in another issue/PR because it has not been analyzed, changing the defaults in this PR has been based on no concrete evidence and autovacuum ought to be tested on its own

fanf (approved):
LGTM :)

clarktsiory:
OK, squash merging this PR

PR #7370, opened 2026-08-07
Fixes #29476: We allow to use empty passwords

amousset:
https://issues.rudder.io/issues/29476

* Prevent any blank password to be accepted on the encoder side
* Correctly consider blank passwords as not modified in the user management backend

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
@@ -56,6 +56,7 @@ import com.normation.zio.*
 import enumeratum.*
 import java.io.InputStream
 import java.security.SecureRandom
+import org.apache.commons.lang3.StringUtils
 import org.springframework.security.crypto.password.PasswordEncoder
 import org.xml.sax.SAXParseException
 import scala.collection.immutable.SortedMap
@@ -104,7 +105,11 @@ object PasswordEncoderType extends Enum[PasswordEncoderType] {
 }
 
 class PasswordEncoderDispatcher(bcryptCost: Int, argon2Params: Argon2EncoderParams) {
-  def dispatch(encoderType: PasswordEncoderType): PasswordEncoder = {
+  /*
+   * The returned encoders are the raw hash algorithms: they don't enforce any Rudder password policy
+   * They must only be used through `RudderPasswordEncoder`, which is why this stays package-private.
+   */
+  private[users] def dispatch(encoderType: PasswordEncoderType): PasswordEncoder = {
     encoderType match {
       case PasswordEncoderType.BCRYPT   => RudderPasswordEncoder.bcryptEncoder(bcryptCost)
       case PasswordEncoderType.ARGON2ID =>
@@ -128,8 +133,7 @@ object RudderPasswordEncoder {
 
   private val secureRandom = new SecureRandom()
 
-  // Proper password hash functions
-  class bcryptEncoder(cost: Int)                          extends PasswordEncoder {
+  private[users] class bcryptEncoder(cost: Int)                          extends PasswordEncoder {
     override def encode(rawPassword: CharSequence):                           String  = {
       val salt: Array[Byte] = new Array(16)
       secureRandom.nextBytes(salt)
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -199,9 +202,15 @@ case class RudderPasswordEncoder(
   }
 
   override def matches(rawPassword: CharSequence, encodedPassword: String): Boolean = {
+    // Defense in depth: an empty password must never authenticate, whatever is stored for the user.
+    !isBlankPassword(rawPassword) &&
     subPasswordEncoder(encodedPassword).matches(rawPassword, encodedPassword)
   }
 
+  private def isBlankPassword(rawPassword: CharSequence): Boolean = {
+    rawPassword == null || rawPassword.toString.strip().isEmpty
```
amousset: wider than an empty string but I think it won't hurt and can prevent misconfigurations.

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserManagement.scala
```diff
@@ -460,13 +460,19 @@ final case class JsonUserFormData(
...
+  private def isPasswordUnchanged(json: JsonUserFormData): Boolean = json.password.strip().isEmpty
+
   given transformer(using passwordEncoder: UserPasswordEncoder[SecretUserPassword]): Transformer[JsonUserFormData, User]           = {
     Transformer
       .define[JsonUserFormData, User]
       .withFieldComputed(
         _.password,
         json => {
-          if (json.isPreHashed) UserPassword.unsafeHashed(json.password)
+          if (isPasswordUnchanged(json)) UserPassword.unknown
```
amousset, line 475: seems to work but I don't know this part of the code well

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -199,9 +202,15 @@ case class RudderPasswordEncoder(
...
 
   override def matches(rawPassword: CharSequence, encodedPassword: String): Boolean = {
+    // Defense in depth: an empty password must never authenticate, whatever is stored for the user.
+    !isBlankPassword(rawPassword) &&
     subPasswordEncoder(encodedPassword).matches(rawPassword, encodedPassword)
   }
 
+  private def isBlankPassword(rawPassword: CharSequence): Boolean = {
+    rawPassword == null || rawPassword.isEmpty
+  }
```
clarktsiory: We'd better use the battle-tested apache commons [StringUtils.isBlank](https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/StringUtils.html#isBlank(java.lang.CharSequence)) implementation

clarktsiory (commented):
The correction looks good to me, there is only a suggestion on using a well-known util for testing against a blank string

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/UserManagement.scala
clarktsiory: Yes the check looks legitimate, the `isPasswordUnchanged` applies the same sanitization (strip().isEmpty) that we use in the `unsafeHashed` implementation for checking against an empty password.

We should likely reorganize the constructors to mutualize the checks in the future

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
amousset: why not, but it is a larger restriction (though acceptable probably)

amousset:
PR updated with a new commit

clarktsiory (approved):
great !

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
clarktsiory: Indeed, there is confusion if we check for empty where should check blank, here it seems legitimate to disallow blankness because it's an empty password too

amousset:
OK, squash merging this PR

PR #7372, opened 2026-08-07
Fixes #29477: Make the argon2 implementation more reliable

amousset:
https://issues.rudder.io/issues/29477

- Catch exceptions from bouncy castle (like bcrypt does)
- Read the tag and salt sizes from the stored hash instead of assuming the current defaults, so a hash created with other sizes can still be verified
- Enforce UTF-8 when encoding the password to avoid variability depending on the JVM config as it would break password matching
- Remove the actual hash value from error messages
- Enforce minimal values for all parameters

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
@@ -84,17 +84,23 @@ sealed abstract class PasswordEncoderType(override val entryName: String) extend
 object PasswordEncoderType extends Enum[PasswordEncoderType] {
   case object BCRYPT   extends PasswordEncoderType("BCRYPT")   {
     val defaultCost = 12
+    // From OWASP recommendations
+    val minimumCost = 10
 
     // 16 bytes = 128 bits
     val saltSize = 16
   }
   case object ARGON2ID extends PasswordEncoderType("ARGON2ID") {
-    // Defaults for user-controlled settings
+    // Defaults and minimums for user-controlled settings. The minimums are the ones of the
+    // lowest configuration OWASP considers acceptable (m=19MiB, t=2, p=1), and they are also
+    // above the values BouncyCastle refuses outright (t=0 and p=0 make it throw).
     // 19 MiB, from OWASP recommendations
     val minimumMemory      = Argon2Memory(19 * 1024)
     // 128 MiB
     val defaultMemory      = Argon2Memory(128 * 1024)
+    val minimumIterations  = Argon2Iterations(2)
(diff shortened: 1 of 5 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/users/RudderUserDetailsFile.scala
```diff
@@ -195,7 +200,7 @@ case class RudderPasswordEncoder(
 
     // We could (and likely should) use RudderPasswordProvider for encoding but we need to store the algorithm to use
     // (as DelegatingPasswordEncoder does) based on the user configuration.
-    subPasswordEncoder(rawPassword.toString).encode(rawPassword)
```
amousset, line 202: currently not used but made no sense, detecting the password type from its raw value

clarktsiory:
does not seem to be correctly linked with the corresponding bugtracker issue (I added it in the description): https://issues.rudder.io/issues/29477

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -790,74 +790,62 @@ object RudderParsedProperties {
...
+          s"Property '$property' is configured to $value$u, which is lower than the minimum secure value of $minimumValue$u. Using the default value instead: $defaultValue$u."
         )
         defaultValue
-      } else cost
+      } else value
     } catch {
       case ex: ConfigException =>
         ApplicationLogger.debug(
-          s"Property 'rudder.bcrypt.cost' is absent or empty in rudder.configFile. Default cost to $defaultValue."
+          s"Property '$property' is absent or empty in rudder.configFile. Using the default value: $defaultValue$u."
```
clarktsiory: string interpolation should be with `${property}`, `${default}`, `${u}` (that's our convention)

clarktsiory: Since `getInt` could throw a `ConfigException`, the message could be misleading:
* it's only visible as `debug`, I think it's an `info`-level log (there are similar examples that have the `info` level)  
* either the message should include the case for "invalid integer", or the sub-type `ConfigException.WrongType` could be caught beforehand

clarktsiory (commented):
Overall LGTM, my only concern is about details of implementation in the generic parsing: the error could be clearer and better handled

amousset:
PR updated with a new commit

amousset:
PR updated with a new commit

clarktsiory (commented):
Great !
Now there is just some merge conflict before merging

amousset:
Fixed!

PR #7381, opened 2026-08-12
Fixes #29496: Prepare the inventory module for the content to come

amousset:
The base of the module, before any section is added:

* The binary becomes a shim over a library, so that everything it does can be exercised from tests, with the command line in its own module.
* Logging through `rudder_cli`, with `--quiet` and `-d`/`-dd` (fusion compat)
* Fix the base sections to match Fusion


policies/module-types/inventory/src/hardware.rs
```diff
@@ -0,0 +1,307 @@
...
+/// parses it back from.
+const LAST_DATE: &str = "%a %b %e %H:%M";
+
+/// The last user to have logged in, and when.
+fn last_logged_user() -> (Option<String>, Option<String>) {
+    if find_in_path(LAST).is_none() {
+        debug!("No '{LAST}', reporting no last logged user");
+        return (None, None);
+    }
+    match cmd(LAST, &[]) {
```
Fdall, line 96: I feel like we should try to use the `-w` option if available

amousset: for now I do things like fusion in general, we can improve it later

policies/module-types/inventory/README.md
```diff
@@ -1,31 +1,131 @@
...
+Non-goals: replacing a _real_ inventory tool, inventorying hardware components (memory slots,
+controllers, ports, sound or video cards, batteries, peripherals), or producing anything the server
+does not use.
 
-## Non-goals
+## What it reads
 
-* Replace Fusion Inventory or any other _real_ inventory tool.
-* In particular, there is absolutely no support for hardware information.
+**No command is required**: one that is missing costs only what it feeds. A run fails only on the
```
Fdall, line 15: I do not understand this

Fdall (approved):

policies/module-types/inventory/README.md
amousset: it means the inventory degrades on missing commands but none is required

amousset: but this is a bug LLMy I agree, I'll fix in the next PR.

PR #7384, opened 2026-08-13
Fixes #29514: Use dedicated event log endpoint in rules recent activity tab

skaerg:
https://issues.rudder.io/issues/29514

This PR aims to use a new `POST eventlog/rules` endpoint in the rules' recent activity page.
This endpoint is identical to `POST eventlog`, except `POST eventlog/rules` only returns rule-related event logs, which removes the need for the event logs to be filtered by type in the elm app.
in addition, the rules' recent activity page can now be viewed so long as the user has the `rule_read` authorization, while it could previously only be viewed by `administrator`s

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
@@ -40,6 +40,9 @@ package com.normation.rudder.rest.internal
 import com.normation.errors.*
 import com.normation.eventlog.*
 import com.normation.rudder.api.ApiVersion
+import com.normation.rudder.domain.eventlog.AddRuleEventType
+import com.normation.rudder.domain.eventlog.DeleteRuleEventType
+import com.normation.rudder.domain.eventlog.ModifyRuleEventType
 import com.normation.rudder.domain.logger.EventLogsLoggerPure
 import com.normation.rudder.domain.properties.NodeProperty
 import com.normation.rudder.repository.EventLogRepository
@@ -57,6 +60,7 @@ import com.normation.zio.UnsafeRun
 import io.scalaland.chimney.syntax.*
 import net.liftweb.http.LiftResponse
 import net.liftweb.http.Req
+import zio.NonEmptyChunk
 import zio.ZIO
 import zio.syntax.*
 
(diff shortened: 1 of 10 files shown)


skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/Activity/ApiCalls.elm
```diff
@@ -8,14 +8,22 @@ import Http.Detailed as Detailed
...
+getActivities bodyParameters (ContextPath contextPath) resourceTypeOpt =
     let
+        url =
+            case resourceTypeOpt of
+                Just resourceType ->
+                    contextPath :: "secure" :: "api" :: "eventlog" :: [ resourceType ]
+
+                Nothing ->
+                    contextPath :: "secure" :: "api" :: [ "eventlog" ]
+
```
clarktsiory, line 21: this is a complicated syntax for a list like: `[ contextPath, "secure", "api", "eventlog", resourceType ]`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -182,6 +187,55 @@ class EventLogAPI(
...
+                           )
+                         )
+                       })
+        totalRecord <- coreService.getUserEventLogCount(filter = None)
+        totalFilter <- coreService.getUserEventLogCount(filter = Some(filter))
+        events      <- coreService.getUserEventLogs(filter = Some(filter))
+        res          = EventLogSlice(events, totalRecord, totalFilter)
+      } yield {
+        (restFilter.draw, res)
+      }).chainError("Error when fetching event logs")
```
clarktsiory, line 222: This looks very much like the original code without `rule` filter: it would be more scalable and maintainable to factor this code with a parameterized type filter (the only difference between rules, directives, etc. and no filter is an additional `typeFilter`).

This will do in this first PR, but it should be factored out in next PRs

clarktsiory (approved):
OK, but there are some things to consider for next PRs:
* code factorization will be a must, we don't want to duplicate the endpoint implementation
* other things we shouldn't forget after the batch of PRs:
  * to add unit tests for all /eventlogs/{resource} endpoint (with the test set-up that comes along)
  * to handle a 403 error in every recent activity API call: there is still a nasty error when we don't have enough rights

skaerg:
OK, squash merging this PR

PR #7385, opened 2026-08-13
Fixes #29501: The rules' recent changes tab may contain an empty "recent changes" dropdown

RaphaelGauthier:
https://issues.rudder.io/issues/29501

The form is hidden if there are no recent changes, and its appearance has been improved:

Capture d’écran du 2026-08-13 11-31-06


clarktsiory (approved):
LGTM !

PR #7387, opened 2026-08-13
Fixes #29507: changelog includes the hash of the token of API account

skaerg:
https://issues.rudder.io/issues/29507

### api account added : 

image

### api account deleted : 

image

### api account modified when the token is not regenerated : 

image

### api accound modified when the token is regenerated : 

image

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
@@ -42,6 +42,7 @@ import com.normation.errors.Inconsistency
 import com.normation.errors.PureResult
 import com.normation.rudder.facts.nodes.NodeSecurityContext
 import enumeratum.*
+import io.scalaland.chimney.Transformer
 import java.nio.charset.StandardCharsets
 import java.security.MessageDigest
 import org.bouncycastle.util.encoders.Hex
@@ -486,6 +487,26 @@ final case class ApiAccount(
   }
 }
 
+object ApiAccount {
+  given transformer: Transformer[ApiAccount, ApiAccountNoToken] = {
+    Transformer
+      .define[ApiAccount, ApiAccountNoToken]
+      .buildTransformer
+  }
+}
+
+case class ApiAccountNoToken(
+    id:                  ApiAccountId,
+    kind:                ApiAccountKind, // Authentication token. It is a mandatory value, and can't be ""
+    // If a token should be revoked, use isEnabled = false.
+    name:                ApiAccountName, // used in event log to know who did actions.
+    description:         String,
+    isEnabled:           Boolean,
+    creationDate:        DateTime,
+    tokenGenerationDate: DateTime,
+    tenants:             NodeSecurityContext
+) {}
+
 /**
  * An API principal, containing the secret, to be used just after creation, and never stored.
  */
@@ -500,18 +521,4 @@ final case class NewApiAccount(
     creationDate:        DateTime,
     tokenGenerationDate: DateTime,
     tenants:             NodeSecurityContext
-) {
-  def toApiAccount(): ApiAccount = {
-    ApiAccount(
-      id,
-      kind,
-      name,
-      token.map(_.toHash()),
-      description,
-      isEnabled,
-      creationDate,
-      tokenGenerationDate,
-      tenants
-    )
-  }
-}
+) {}
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDiffMapper.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDiffMapper.scala
@@ -71,6 +71,7 @@ import com.unboundid.ldif.LDIFAddChangeRecord
 import com.unboundid.ldif.LDIFChangeRecord
 import com.unboundid.ldif.LDIFModifyChangeRecord
 import com.unboundid.ldif.LDIFModifyDNChangeRecord
+import io.scalaland.chimney.syntax.*
 import net.liftweb.common.*
 import scala.util.control.NonFatal
 
@@ -568,7 +569,7 @@ class LDAPDiffMapper(
           val e = LDAPEntry(add.toAddRequest().toEntry)
           for {
             param <- mapper.entry2ApiAccount(e)
-          } yield AddApiAccountDiff(param)
+          } yield AddApiAccountDiff(param.transformInto[ApiAccountNoToken])
         case _ => Left(Err.UnexpectedObject(s"Bad change record type for requested action 'Add Api Account': ${change}"))
       }
     } else {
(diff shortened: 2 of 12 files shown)


skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -1500,7 +1511,6 @@ class EventLogDetailsGenerator(
       <ul class="evlogviewpad">
         <li><b>Rudder ID: </b><value id="id"/></li>
         <li><b>Name:&nbsp;</b><value id="name"/></li>
-        <li><b>Token:&nbsp;</b><value id="token"/></li>
```
clarktsiory, line 1503: if this is removed the `#token` replacement won't work, you should leave it if you want to still display `Token changed` text

skaerg: this is the template for the token added/deleted cases. the token value only changes during regeneration, which produces a token modified change log, which uses the `apiAccountModDetailsXML` template just below this one

clarktsiory: okay then (I only saw the diff), nice catch !

amousset:
Ideally the hash should never leave the LDAP, and only be compared in queries, this would prevent misuse as we never need to do anything with it (but ok for this simple fix now).

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/marshalling/XmlUnserialisationImpl.scala
```diff
@@ -838,8 +838,6 @@ class ApiAccountUnserialisationImpl extends ApiAccountUnserialisation {
       _              <- TestFileFormat(apiAccount)
       id             <- (apiAccount \ "id").headOption.map(_.text) ?~! (s"Missing attribute 'id' in entry type API Account : ${entry}")
       name           <- (apiAccount \ "name").headOption.map(_.text) ?~! (s"Missing attribute 'name' in entry type API Account : ${entry}")
-      token          <-
```
amousset, line 841: don't we lose the information a token's value has been changed?

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

P4uline (commented):
Nice

I think the migration script of the XML data of eventlog table should be part of this PR

skaerg:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/api/DataStructures.scala
```diff
@@ -486,6 +487,26 @@ final case class ApiAccount(
   }
 }
 
+object ApiAccount {
+  given transformer: Transformer[ApiAccount, ApiAccountNoToken] = {
+    Transformer
+      .define[ApiAccount, ApiAccountNoToken]
+      .buildTransformer
+  }
```
clarktsiory, line 495: define + buildTransformer is same as  `derive`
```suggestion
  given transformer: Transformer[ApiAccount, ApiAccountNoToken] = Transfomer.derive
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/ldap/LDAPDiffMapper.scala
```diff
@@ -597,7 +598,14 @@ class LDAPDiffMapper(
                                 }
                               case A_API_TOKEN                   =>
                                 nonNull(diff, mod.getOptValueDefault("")) { (d, value) =>
-                                  d.copy(modToken = Some(SimpleDiff(oldAccount.token.flatMap(_.exposeHash()).getOrElse(""), value)))
+                                  d.copy(modToken = {
+                                    Some(
+                                      SimpleDiff(
+                                        (oldAccount.token.flatMap(_.exposeHash()).getOrElse("").substring(0, 6) ++ "******"),
```
clarktsiory, line 604: the `substring` combined with `getOrElse("")` here produces an IndexOutOfBoundsException

In general we should avoid the evil `substring` method on non-tested code/without a comment that explains why it cannot throw.

Here from the local code there is an assumption that the has 6 or more characters, but that's not for the diff class to guess, it should be encoded in the hash type.
For what you are doing, I suggest you take the example of `ApiTokenSecret.exposeSecretBeginning` e.g. with a new `exposeHashBeginning` which does the proper matching on the hash

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -944,7 +944,14 @@ class EventLogDetailsGenerator(
                     </ul>{
                     (
                       "#name" #> mapSimpleDiff(apiAccountDiff.modName) &
-                      "#token" #> mapSimpleDiff(apiAccountDiff.modToken) &
+                      "#token" #> {
+                        apiAccountDiff.modToken match {
+                          case Some(_) =>
+                            <ul class="evlogviewpad"><li><b>Token regenerated</b></li></ul>
```
clarktsiory, line 950: not sure it's always "regenerated" in that case: it could have been a creation

clarktsiory (changes requested):
There is some string processing that needs to be fixed.
Can you also add a test case e.g. something like
```
"never serialise a token in the event log" in {
  serialise(...) must not(contain("token"))
}
```
?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
skaerg: > it could have been a creation

i don't think so; when a user creates a token, the type of the change log that gets created is `CreateAPIAccountEventLog`, not `ModifyAPIAccountEventLog` (this code is only used in the `ModifyAPIAccountEventLog` case).
i listed all the important change log cases with accompanying screenshots in the PR description ; is there a case i missed ?

clarktsiory: mmmmh... not always, there is a dedicated scenario where the account token is not created, IIRC that's why it's an `Option` from the start, the account ID when defined makes the token empty : 
image


But we can always override that by creating a token despite the "account ID" during a regeneration event, which therefore leads to token creation despite being in the `Modify` branch

PR #7389, opened 2026-08-14
Fixes #29505: Fix trivy branch checks

amousset:
https://issues.rudder.io/issues/29505

Do like rudder-plugins which is known to work.


amousset:
Commit modified

PR #7390, opened 2026-08-14
Fixes #29529: Users table query has wrong column userid

clarktsiory:
https://issues.rudder.io/issues/29529

An `id` has been replaced by `userId`. Just revert it back, and add a covering postgres test with sample cases

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
@@ -39,6 +39,8 @@ package com.normation.rudder.users
 
 import cats.ApplicativeError
 import cats.data.NonEmptyList
+import cats.syntax.applicative.*
+import cats.syntax.functor.*
 import com.normation.errors.Inconsistency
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -39,6 +39,8 @@ package com.normation.rudder.users
 
 import cats.ApplicativeError
 import cats.data.NonEmptyList
+import cats.syntax.applicative.*
```
clarktsiory, line 42: this imports `.pure` syntax

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/users/UserRepository.scala
```diff
@@ -39,6 +39,8 @@ package com.normation.rudder.users
...
+import cats.syntax.functor.*
```
clarktsiory, line 43: this imports `void` which is `.map(_ => ())` (equivalent of ZIO's `.unit`)

amousset (approved):

PR #7399, opened 2026-08-14
Fixes #29515: Use dedicated event log endpoint in directives recent activity tab

skaerg:
https://issues.rudder.io/issues/29515

This PR aims to use a new `POST eventlog/directives` endpoint in the directives' recent activity page.
This endpoint is identical to `POST eventlog`, except `POST eventlog/directives` only returns directive-related event logs, which removes the need for the event logs to be filtered by type in the elm app.
in addition, the directives' recent activity page can now be viewed so long as the user has the `directive_read` authorization, while it could previously only be viewed by `administrator`s

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
@@ -40,9 +40,7 @@ package com.normation.rudder.rest.internal
 import com.normation.errors.*
 import com.normation.eventlog.*
 import com.normation.rudder.api.ApiVersion
-import com.normation.rudder.domain.eventlog.AddRuleEventType
-import com.normation.rudder.domain.eventlog.DeleteRuleEventType
-import com.normation.rudder.domain.eventlog.ModifyRuleEventType
+import com.normation.rudder.domain.eventlog.*
 import com.normation.rudder.domain.logger.EventLogsLoggerPure
 import com.normation.rudder.domain.properties.NodeProperty
 import com.normation.rudder.repository.EventLogRepository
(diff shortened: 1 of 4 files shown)


skaerg:
Commit modified

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -187,6 +191,62 @@ class EventLogAPI(
...
+      filter       = queryFilter.copy(typeFilter = {
+                       Some(EventLogRequest.TypeFilter(include = Some(include), exclude = None))
+                     })
+      totalRecord <- coreService.getUserEventLogCount(filter = None)
+      totalFilter <- coreService.getUserEventLogCount(filter = Some(filter))
+      events      <- coreService.getUserEventLogs(filter = Some(filter))
+      res          = EventLogSlice(events, totalRecord, totalFilter)
+    } yield {
+      (restFilter.draw, res)
+    }).chainError("Error when fetching event logs")
```
clarktsiory: could we have the "types" that we are fetching in the error message ?
or else it would be the exact same error as for the main endpoint...

Since we are using list of multiple types, having `include.mkString(",")` would be okay

clarktsiory (commented):
LGTM, but could the error message be adapted ?

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -40,8 +40,11 @@ package com.normation.rudder.rest.internal
 import com.normation.errors.*
 import com.normation.eventlog.*
 import com.normation.rudder.api.ApiVersion
+import com.normation.rudder.domain.eventlog.AddDirectiveEventType
 import com.normation.rudder.domain.eventlog.AddRuleEventType
+import com.normation.rudder.domain.eventlog.DeleteDirectiveEventType
 import com.normation.rudder.domain.eventlog.DeleteRuleEventType
+import com.normation.rudder.domain.eventlog.ModifyDirectiveEventType
```
clarktsiory: also, you could import all at once with `...eventlog.*`, it would be easier (maybe we want that as a scalafmt rule starting from 3-4 imports, but the rule doesn't seem to apply yet)

skaerg:
PR updated with a new commit

skaerg:
PR updated with a new commit

clarktsiory (approved):

skaerg:
OK, squash merging this PR

PR #7401, opened 2026-08-14
Fixes #29516: Use dedicated event log endpoint in techniques recent activity tab

skaerg:
https://issues.rudder.io/issues/29516

This PR aims to use a new `POST eventlog/editorTechniques` endpoint in the techniques' recent activity page.
This endpoint is identical to `POST eventlog`, except `POST eventlog/editorTechniques` only returns editor technique-related event logs, which removes the need for the event logs to be filtered by type in the elm app.
in addition, the techniques' recent activity page can now be viewed so long as the user has the `technique_read` authorization, while it could previously only be viewed by `administrator`s


clarktsiory (approved):

skaerg:
OK, merging this PR

PR #7403, opened 2026-08-14
Fixes #18567: Event log should have a rollback button which only reverts an item

clarktsiory:
https://issues.rudder.io/issues/18567

Adding a new main rollback section, putting the "rollback of all changes" in its own:
image

Changes of the feature consists, downwards from the frontend, of:
* adding an `action=item` in the frontend API call, the HTML and JS needed some reorganisation (see screenshot above for the result)
* handling the `Item` action, to rollback one given item type, this is the most interesting part:
  * event log type needs to be dispatched (pattern matched) to corresponding repository for the item, and appropriate method :
    * rolling back an "_add_" event results in corresponding "_delete_" in repository
    * "_modify_"/"_delete_" roll back as "_save_" method of the repository
  * there are existing "archive parsers" that lookup the git repo for a given item ID, we just need to extract the item ID from the event log details
* unit tests have been added for some bugs corrected in existing parsers, and for the rollback of each supported event type

Some things related to tenants:
* the rollback tenants was a `systemQC` with access to all tenants. But it should be restricted to the tenants of the user making the rollback, which is a bug, missed by https://github.com/Normation/rudder/pull/7334
* remove `toBox`/`toIO` by rewriting Box into IOResult

Things to note : techniques are a special case, they have another problem for rollback even in earlier branches (https://issues.rudder.io/issues/29500), so they can't be really tested in reproductible way, but the unit tests are there to ensure the item rollback should work

--- 

This is my first contribution with Claude, I used it quite iteratively to check for defects in my own initial implementation (I started by hand to do the rollbackItem architecture and the HTML + JS change), and eventually to mostly to generate corresponding tests in my own

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
@@ -49,10 +49,7 @@ import com.normation.cfclerk.domain.TechniqueName
 import com.normation.cfclerk.domain.TechniqueVersion
 import com.normation.cfclerk.xmlparsers.TechniqueParser
 import com.normation.errors.*
-import com.normation.rudder.configuration.DirectiveRevisionRepository
 import com.normation.rudder.configuration.GroupAndCat
-import com.normation.rudder.configuration.GroupRevisionRepository
-import com.normation.rudder.configuration.RuleRevisionRepository
 import com.normation.rudder.domain.logger.ConfigurationLoggerPure
 import com.normation.rudder.domain.nodes.NodeGroupCategoryId
 import com.normation.rudder.domain.nodes.NodeGroupUid
@@ -133,7 +130,7 @@ class GitParseRules(
     ruleUnserialisation: RuleUnserialisation,
     val repo:            GitRepositoryProvider,
     rulesRootDirectory:  String // relative name to git root file
-) extends ParseRules with GitParseCommon[List[Rule]] with RuleRevisionRepository {
+) extends ParseRules with GitParseCommon[List[Rule]] {
 
   val rulesDirectory: GitRootCategory = getGitDirectoryPath(rulesRootDirectory)
 
@@ -281,7 +278,11 @@ class GitParseGroupLibrary(
     libRootDirectory:     String, // relative name to git root file
 
     categoryFileName: String = "category.xml"
-) extends ParseGroupLibrary with GitParseCommon[NodeGroupCategoryContent] with GroupRevisionRepository {
+) extends ParseGroupLibrary with GitParseCommon[NodeGroupCategoryContent] {
+
+  // the root category is stored in the group library directory itself, so contrary to the other
+  // categories, its ID can not be read back from the path of the files it holds
+  private val rootCategoryId = NodeGroupCategoryId("GroupRoot")
 
   val groupsDirectory: GitRootCategory = getGitDirectoryPath(libRootDirectory)
 
@@ -376,20 +377,17 @@ class GitParseGroupLibrary(
       // nodegroups are any where in the subtree with parent directory name == uuid of the group category.
       // So we need to find the group by name, and split path to get its category. Be careful, the name of root
       // category is not "groups" but "GroupRoot"
-      groups <- GitFindUtils.listFiles(repo.db, treeId, List(groupsDirectory.directoryPath), s"/{uid.value}.xml" :: Nil)
+      groups <- GitFindUtils.listFiles(repo.db, treeId, List(groupsDirectory.directoryPath), s"/${uid.value}.xml" :: Nil)
       res    <- groups.toList match {
                   case Nil      => None.succeed
                   case h :: Nil =>
                     for {
                       xml   <- GitFindUtils.getFileContent(repo.db, treeId, h)(is => ParseXml(is, Some(h)))
                       group <- groupUnserialiser.unserialise(xml).toIO
-                      // h is path relative to config-repo, so may contains only one "/" (rules/ruleId.xml)
-                      catId  = h.split('/').toList.reverse match {
-                                 case Nil | _ :: Nil       => // assume root category even if Nil
-                                   NodeGroupCategoryId("GroupRoot")
-                                 case group_ :: catId :: _ =>
-                                   NodeGroupCategoryId(catId)
-                               }
+                      // h is path relative to config-repo, so may contains only one "/" (groups/groupId.xml)
+                      parent = h.substring(0, h.lastIndexOf('/') + 1)
+                      catId  = if (parent == groupsDirectory.directoryPath) rootCategoryId
+                               else NodeGroupCategoryId(parent.split('/').last)
                       // we need to correct ID revision to the one we just looked-up.
                       // (it's normal to not have it serialized, since it's given by git, it's not intrinsic)
                     } yield Some(GroupAndCat(group.modify(_.id.rev).setTo(rev), catId))
@@ -407,7 +405,6 @@ class GitParseGroupLibrary(
 trait TechniqueRevisionRepository {
   /*
    * Get the technique with given name, version and revision from history
-   * (todo: what about head?)
    */
   def getTechnique(
       name:    TechniqueName,
@@ -663,7 +660,7 @@ class GitParseActiveTechniqueLibrary(
 
     uptcFileName: String = "category.xml",
     uptFileName:  String = "activeTechniqueSettings.xml"
-) extends ParseActiveTechniqueLibrary with GitParseCommon[ActiveTechniqueCategoryContent] with DirectiveRevisionRepository {
+) extends ParseActiveTechniqueLibrary with GitParseCommon[ActiveTechniqueCategoryContent] {
 
   /**
    * Get a directive for the specific given revision;
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemArchiveManagerImpl.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemArchiveManagerImpl.scala
@@ -38,28 +38,18 @@
 package com.normation.rudder.repository.xml
 
 import com.normation.errors.*
-import com.normation.eventlog.EventActor
-import com.normation.eventlog.EventLog
-import com.normation.eventlog.ModificationId
+import com.normation.eventlog.*
 import com.normation.rudder.batch.AsyncDeploymentActor
 import com.normation.rudder.batch.AutomaticStartDeployment
-import com.normation.rudder.domain.Constants.FULL_ARCHIVE_TAG
+import com.normation.rudder.domain.Constants.*
 import com.normation.rudder.domain.eventlog.*
 import com.normation.rudder.domain.logger.GitArchiveLoggerPure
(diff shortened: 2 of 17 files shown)


clarktsiory:
Commit modified

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
```diff
@@ -376,20 +377,17 @@ class GitParseGroupLibrary(
...
-                      catId  = h.split('/').toList.reverse match {
-                                 case Nil | _ :: Nil       => // assume root category even if Nil
-                                   NodeGroupCategoryId("GroupRoot")
-                                 case group_ :: catId :: _ =>
-                                   NodeGroupCategoryId(catId)
-                               }
+                      // h is path relative to config-repo, so may contains only one "/" (groups/groupId.xml)
+                      parent = h.substring(0, h.lastIndexOf('/') + 1)
+                      catId  = if (parent == groupsDirectory.directoryPath) rootCategoryId
+                               else NodeGroupCategoryId(parent.split('/').last)
```
clarktsiory, line 390: this was a bug, which is now covered in `TestGitParseRudderObjects`

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

fanf:
> the rollback tenants was a systemQC with access to all tenants. But it should be restricted to the tenants of the user making the rollback, which is a bug

Not at all. A full rollback doesn't any sensible meaning in a tenant context. It checkout the config of everything, without regard for tenants, and load it back. This is why it was - and it must remain - a pure administrator-with-all-tenants action. 

And it's also why the by-item rollback must follow another architecture.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
```diff
@@ -376,20 +377,17 @@ class GitParseGroupLibrary(
       // nodegroups are any where in the subtree with parent directory name == uuid of the group category.
       // So we need to find the group by name, and split path to get its category. Be careful, the name of root
       // category is not "groups" but "GroupRoot"
-      groups <- GitFindUtils.listFiles(repo.db, treeId, List(groupsDirectory.directoryPath), s"/{uid.value}.xml" :: Nil)
+      groups <- GitFindUtils.listFiles(repo.db, treeId, List(groupsDirectory.directoryPath), s"/${uid.value}.xml" :: Nil)
```
fanf, line 380: that looks like a bug that need to be corrected in previous version

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/xml/TestGitParseRudderObjects.scala
```diff
@@ -95,4 +177,37 @@ class TestGitParseRudderObjects extends Specification with Loggable with AfterAl
...
+
+      res.map(_.group.id.uid) must beSome(rootCatGroupId)
+      res.map(_.categoryId) must beSome(rootCatId)
+      res.map(_.categoryId) must not(beSome(NodeGroupCategoryId("groups")))
+    }
+
+    // revisions are given by git and not serialized in the archive, so the looked-up one is stamped
+    // back into the group ID. Users of that method (like an item rollback) need to know it.
+    "stamp the revision it was looked-up at into the group ID" in {
+      val res = parseGroupLib.getGroupRevision(rootCatGroupId, Revision("master")).runNow
```
fanf, line 202: I'm not sure you are testing anything here, because master is default on most git, so also our default revision.

webapp/sources/rudder/rudder-web/src/main/webapp/secure/configurationManager/changeLogs.html
```diff
@@ -47,28 +47,45 @@ <h1>
...
+      <div class='d-flex flex-column gap-2' id="restoreItem">
+        <fieldset class='rollbackFieldSet pt-3 card card-body'>
+          <legend>Rollback this item</legend>
+          <div id='{{rollbackItemId}}' class="d-flex w-auto align-items-center justify-content-center column-gap-2">
+            <span class='alignRollback'>Restore this item its state previous to the change</span>
+            <button class="btn btn-default ms-2" id='{{rollbackBtnId}}'>Restore</button>
+          </div>
+          <div id='{{confirmItemId}}'></div>
+        </fieldset>
+        <fieldset class='rollbackFieldSet pt-3'>
```
fanf, line 61: that part needs to be hidden if the user isn't admin. This is an admin-only action and must remain so. 

The part above (item rollback) is user compatible

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemArchiveManagerImpl.scala
```diff
@@ -569,6 +587,241 @@ class ItemArchiveManagerImpl(
...
+  /*
+   * Rolling back always needs item ID in repos, we can find it in the event log details, at specific XML selector
+   *
+   * Logic of reverting change, depends on event log type:
+   * - reverting an addition is just a deletion
+   * - reverting a deletion or modification is a restore of the item as it is
+   *   - archive needs to be observed e.g. for category
+   *   - and
+   */
+  private[xml] def rollbackOneItem(archiveId: GitCommitId, event: EventLog)(implicit cc: ChangeContext): IOResult[Unit] = {
```
fanf: that design won't work well once we get change logs for campaigns, benchmarks, etc. It will need the same extensible design than other part of change logs.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/modification/ModificationService.scala
```diff
@@ -37,77 +37,75 @@
...
+  // the state just *before* a change is the parent of the commit that change led to
+  private def parentOf(commit: GitCommitId): GitCommitId = GitCommitId(commit.value + "^")
+
   def restoreToEventLog(
       eventLog:         EventLog,
       commiter:         PersonIdent,
       rollbackedEvents: Seq[EventLog],
       target:           EventLog
-  ): Box[GitCommitId] = {
+  )(using cc: ChangeContext): IOResult[GitCommitId] = {
```
fanf: that change context should not be here: 
- it isn't used for the commit info (commiter),
- this is a system-only action, that doesn't support specific tenants (if at some point, we manage rights this way, it will be needed, but for now, it doesn't give the correct signal to code reader).

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/modification/ModificationService.scala
```diff
@@ -37,77 +37,75 @@
...
-
   }
 
   def restoreBeforeEventLog(
       eventLog:         EventLog,
       commiter:         PersonIdent,
       rollbackedEvents: Seq[EventLog],
       target:           EventLog
-  ): Box[GitCommitId] = {
+  )(using cc: ChangeContext): IOResult[GitCommitId] = {
```
fanf: same here

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/modification/ModificationService.scala
```diff
@@ -37,77 +37,75 @@
...
+    for {
+      commit   <- commitOf(eventLog)
+      rollback <- itemArchiveManager.rollback(parentOf(commit), commiter, rollbackedEvents, target, "before")
+    } yield {
+      rollback
+    }
+  }
+
+  def restoreItemEventLog(
```
fanf: given that the mechanism and reach of rollback for one item on one side, before/after even on the other side are totally different, I would prefer that they are in two separated interfaces and implementation.

clarktsiory:
PR updated with a new commit

clarktsiory:
okay, it makes sense then, in the spirit of making rollbacks possible in recent activities too, we will need a different architecture for rights too (and of course tenants too, but they are already present in core methods here)

@fanf I put the `systemQC` back for old rollback

fanf (changes requested):
I think we must enforce in the code that "by item rollback" and "rollback before/after commit" are not the same kind of thing, not even in the same league: 
- 1/ rollback before/after is a global, admin-only, far reaching, tenant agnostic operation. Given what it does, it outreaches any authorization or tenant. It's more a system/backup operation than a functionnal one
- 2/ rollback item is a typical user-operated action in their daily workflow. They are tenant- and authorization-bound. 

The code and architecture must reflect that. In the long term, they won't be used at the same place, and we will perhaps even remove the 1/.

webapp/sources/rudder/rudder-web/src/main/webapp/secure/configurationManager/changeLogs.html
clarktsiory: it's a bug in earlier version, it should be fixed there too then: read-only admin that have access to change logs can see the buttons (the API call returns an error because it requires full admin write rights)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitParseRudderObjects.scala
clarktsiory: yes I'll create an issue to backport that, but it was unused code that I needed to reuse and fix in this PR

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/xml/TestGitParseRudderObjects.scala
clarktsiory: The purpose of the added tests in that file is to "parse" the current revision and assert that groups have expected category, due to the reported 1 char bug (missing `$`) on groups, it couldn't parse correctly

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/modification/ModificationService.scala
clarktsiory: Indeed, it would make distinction and evolution easier, I will change that too

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/xml/TestGitParseRudderObjects.scala
fanf: I understand that - what I mean is that the master branch is the default branch too, so I'm not sure that using `master` as a testing variable gives you more info than none. Here, perhaps that without version parameter, you get the same result.

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

fanf (approved):
Good ! 

And it works as expected, and it's just amazing :)

PR #7404, opened 2026-08-14
Fixes #29518: Use dedicated event log endpoint in global parameter recent activity tab

skaerg:
https://issues.rudder.io/issues/29518

This PR aims to use a new `POST eventlog/parameters` endpoint in the parameters' recent activity page.
This endpoint is identical to `POST eventlog`, except `POST eventlog/parameters` only returns parameter-related event logs, which removes the need for the event logs to be filtered by type in the elm app.
in addition, the parameters' recent activity page can now be viewed so long as the user has the `parameter_read` authorization, while it could previously only be viewed by `administrator`s

:warning: This PR also removes the `BodyParameters` datatype in the `Activity` Elm app since its `filterTypes` field is no longer used; we only need its `search` field


clarktsiory (approved):
Great, I even think this design is better: avoid passing (Add|Remove|Modify)Resource type filters from the frontend, the API endpoint for the dedicated resource already handles the filtering

PR #7409, opened 2026-08-18
Fixes #29552: We embeded useless data in war that bloat it sizes

fanf:
https://issues.rudder.io/issues/29552

Remove dev/test data from war: 

- semanticdb files - only interesting for scalac/scalafix, 
- JS/CSS map - only interesting for debuging - it means we won't have them in prod env
- a test jar that wasn't scoped test :shrug: 

With some maven sorcery I'd have prefered not known. 

We that, we gain 8MB, from 180 to 172 MB


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM

PR #7410, opened 2026-08-18
Fixes #29539: Handle Scala dependency security checks since Trivy

clarktsiory:
https://issues.rudder.io/issues/29539

* fs2 and doobie are tightly linked, just update doobie and zio-cats-interop together so that they both pull latest `fs2:3.12.2` (it's already the case in 9.1). doobie could be `1.0.0-RC11` as in 9.1 branch, to satisfy that constraint, but I kept `1.0.0-RC12` which is n-1 of latest (RC13 is a huge change due to namespace change) 
* apache `commons-lang3` has a deprecation in the patched version (the fix is the same as in 9.1: String.CS.replace)
* a vulnerable version of apache `commons-fileupload2` is pulled by lift-webkit: 2.0.0-M2, the transitive dependency was supposed to be ignored, but it had the wrong name: just exclude the transitive one, the actual one being used is the version in the commons-fileupload variable (and the dependency itself is declared in rudder-rest/pom.xml) i.e. 2.0.0-M4
* others: just a bump of the `logback` version

### DIFF
--- a/.trivyignore
+++ b/.trivyignore
@@ -1,13 +1,14 @@
  # Scala
-CVE-2025-11226 # ch.qos.logback/logback-core: Conditional arbitrary code execution in logback-core
-CVE-2025-58369 # FS2 half-shutdown of socket during TLS handshake may result in spin loop
-CVE-2025-48976 # apache-commons-fileupload: Apache Commons FileUpload DoS via part headers
-CVE-2025-48924 # commons-lang/commons-lang: Uncontrolled Recursion vulnerability in Apache Commons Lang
-CVE-2026-41720 # Spring LDAP has Authentication Bypass with Empty Password
+
 
 # npm
-CVE-2024-1899 # Showdown vulnerable to Regular Expression Denial of Service (ReDoS) in link/anchor parsing, no release patch
-CVE-2025-9910 # jsondiffpatch is vulnerable to Cross-site Scripting (XSS) via HtmlFormatter::nodeBegin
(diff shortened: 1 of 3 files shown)


.trivyignore
```diff
@@ -1,9 +1,4 @@
  # Scala
-CVE-2025-11226 # ch.qos.logback/logback-core: Conditional arbitrary code execution in logback-core
-CVE-2025-58369 # FS2 half-shutdown of socket during TLS handshake may result in spin loop
-CVE-2025-48976 # apache-commons-fileupload: Apache Commons FileUpload DoS via part headers
-CVE-2025-48924 # commons-lang/commons-lang: Uncontrolled Recursion vulnerability in Apache Commons Lang
-CVE-2026-41720 # Spring LDAP has Authentication Bypass with Empty Password
 
 # npm
```
amousset, line 4: we still need to explain why these are acceptable

amousset (approved):

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/pom.xml
```diff
@@ -434,6 +434,7 @@ limitations under the License.
     <cats-version>2.13.0</cats-version>
     <ip4s-version>3.7.0</ip4s-version>
     <doobie-version>1.0.0-RC12</doobie-version>
+    <fs2-version>3.12.2</fs2-version>
```
clarktsiory, line 437: The variable is used in plugins, so I needed to write it back to the updated version

clarktsiory:
OK, squash merging this PR

PR #7411, opened 2026-08-18
Fixes #29563: Impact of #29554 in rudder

skaerg:
https://issues.rudder.io/issues/29563

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
@@ -134,6 +134,16 @@ trait DefaultPluginDef extends RudderPluginDef {
   def pluginMenuEntry:  List[(Menu, Option[String])] = Nil
   def pluginMenuParent: List[Menu]                   = Nil
 
+  def handleDuplicateMenus(m: Menu, newMenu: Menu): Menu = {
+    // We need to avoid collision on name/loc
+    if (m.kids.exists(_.loc.name == newMenu.loc.name)) {
+      PluginLogger.error(s"There is already a menu with id (${newMenu.loc.name}")
+      m
+    } else {
+      Menu(m.loc, (m.kids :+ newMenu).sortBy(_.loc.name)*)
+    }
+  }
+
   override def updateSiteMap(menus: List[Menu]): List[Menu] = {
 
     val updatedMenu = pluginMenuParent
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
```diff
@@ -134,6 +134,16 @@ trait DefaultPluginDef extends RudderPluginDef {
   def pluginMenuEntry:  List[(Menu, Option[String])] = Nil
   def pluginMenuParent: List[Menu]                   = Nil
 
+  def handleDuplicateMenus(m: Menu, newMenu: Menu): Menu = {
+    // We need to avoid collision on name/loc
+    if (m.kids.exists(_.loc.name == newMenu.loc.name)) {
+      PluginLogger.error(s"There is already a menu with id (${newMenu.loc.name}, please contact Plugin team")
```
amousset: > please contact Plugin team

who? It is not helpful for users, nor for devs I think.

skaerg:
Commit modified

P4uline (approved):

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/plugins/PublicPlugin.scala
```diff
@@ -134,6 +134,16 @@ trait DefaultPluginDef extends RudderPluginDef {
...
 
+  def handleDuplicateMenus(m: Menu, newMenu: Menu): Menu = {
+    // We need to avoid collision on name/loc
+    if (m.kids.exists(_.loc.name == newMenu.loc.name)) {
+      PluginLogger.error(s"There is already a menu with id (${newMenu.loc.name}")
+      m
+    } else {
+      Menu(m.loc, (m.kids :+ newMenu).sortBy(_.loc.name)*)
+    }
+  }
```
clarktsiory, line 145: this seems to be a static method, that would belong to an `object RudderPluginDef`.
This would make this testable, and reusable in plugins, as `RudderPluginDef.handleDuplicateMenus`

skaerg: we want to use `DefaultPluginDef.handleDuplicateMenus` in `DefaultPluginDef.updateSiteMap`, which is an override of `RudderPluginDef.updateSiteMap`  which is not static. because of this, is it not impossible to make `handleDuplicateMenus` static without also making `RudderPluginDef.updateSiteMap`static ?

clarktsiory: There already is an `object RudderPluginDef { ... }`, every `def x` method you will put within the `object` will be accessible from anywhere, because it's static.
Companion objects are just the default conventional target for such methods.

An alternative is to put the method at top-level (without having any wrapping `trait`/`class`/`object`): [it's a Scala 3 feature](https://docs.scala-lang.org/scala3/book/taste-toplevel-definitions.html), but we are still used to having namespace (old Scala 2 habit))...
The corresponding import would then be `com.normation.plugins.updateSiteMap` instead of `com.normation.plugins.RudderPluginDef.updateSiteMap`, IDK which version is prefered

P4uline: Here we want an instance method to be able to override it only certain plugins : cve and system-update (because the way we build the menu is a bit specific for those plugins) and keep a default behavior for others.

We can still use a static method or a top-level method to extract the few duplicated code lines inside `handleDuplicateMenus`.

clarktsiory: in fact, I think in the plugins, the duplicated implementation can just be called as `super.handleDuplicateMenus` instead

PR #7412, opened 2026-08-18
Fixes #29562: Remove the cf-promises hook

amousset:
https://issues.rudder.io/issues/29562

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-ready/10-cf-promise-check.disabled
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-ready/10-cf-promise-check.disabled
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+# Additional syntax check on the generated Linux policies. Disabled by default, not needed in general.
+
 # Hooks parameter are passed by environment variable:
 #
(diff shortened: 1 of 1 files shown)


amousset:
Commit modified

webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-ready/10-cf-promise-check.disabled
```diff
@@ -1,5 +1,7 @@
 #!/bin/sh
 
+# Additional syntax check on the generated Linux policies. Disabled by default, not needed in general.
```
fanf, line 3: We have an UI configuration that manages specifically the state of that hook and it's awaiting just a non exec file (in place of .disable extension), see: https://github.com/Normation/rudder/blob/0dd739567d69423540631bdbe5315bf89175be74/webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/PropertiesManagement.scala#L669

fanf (changes requested):

webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-ready/10-cf-promise-check.disabled
amousset: it is probably reset on each upgrade then

PR #7427, opened 2026-08-20
Fixes #29583: Remove scalaj-http dependance in rudder core

fanf:
https://issues.rudder.io/issues/29583

The only big difference is the need to create our own SSL ignored  class, it was provided by  scalaj-http before that.


clarktsiory (approved):
LGTM

PR #7430, opened 2026-08-20
Fixes #29567: Syntax error on scheduled event generation

fanf:
https://issues.rudder.io/issues/29567

Need: https://github.com/Normation/rudder-techniques/pull/1938

So, there was A LOT of things not correct: 

- the idea to pass the json directly in a var in a cfengine file need escaping, which we don't want to handle => swith to a `scheduled_events.json` at the root directory of policies. That hance removes all the system variable `MODULE_PARAM_SCHEDULE` related code, 
- I created a new adr superseeding the previous one, since we were telling "we chose not to use a json file"
- we weren't actually setting the "ifvarclass" for scheduled events. It was set to "always None" in 9.1 for safety, be never changed 
- I added generation tests for the ifvarclass, which create a bunch of new files, which is good for my commit metrics
- actually, use `if =>` and not `ifvarclass =>` - and quote the right hand side
- some names in the generated json keys were not correct: `event`/`events`, `scheduledId`, etc
- the rust part was never setting `not_before` and `not_after` in base.


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/write/BuildBundleSequence.scala
```diff
@@ -470,7 +470,7 @@ class BuildBundleSequence(
         policy.technique.policyTypes,
         policyMode,
         policy.technique.useMethodReporting,
-        None
+        policy.ifVarClass
```
fanf, line 473: "oups"

fanf:
PR updated with a new commit

adr/webapp/scheduling/29567-passing-scheduled-events-through-a-generated-json-file.md
```diff
@@ -0,0 +1,98 @@
...
+
+- Status: accepted, supersedes [28535 - Generation of discrete scheduled event for agent schedule module](28535-passing-discrete-scheduled-events-from-webapp-to-agent.md)
+- Deciders: FAR
+- Date: 2026-08-20
+
+## Context
+
+ADR #28535 chose to pass the list of discrete schedule events from webapp to agent through a system variable, `MODULE_PARAM_SCHEDULE`, and explicitly rejected the "dedicated JSON file" alternative on the ground that parsing a JSON file in CFEngine to then pass parameters to the module would be "extremely painful".
+
+That decision did not survive contact with reality (#29567). The system variable was expanded by StringTemplate into the generated `common.cf`:
```
amousset, line 11: this file reads like pure ai slop. Not sure what to do about it but it makes me a bit sad.

amousset:
The system techniques should really live in this repo. Especially as we are introducing deeper coupling, movingoff from the standard technique interface.

webapp/sources/rudder/rudder-core/src/test/resources/configuration-repository/expected-share/root-with-scheduled-directive/rules/cfengine-community/rpmPackageInstallation/7.0/rpmPackageInstallation.cf
```diff
@@ -0,0 +1,326 @@
+#####################################################################################
```
amousset, line 1: this is funny that this techniques is kept around for webapp tests

amousset:
LGTM except for the ADR.

amousset (approved):

adr/webapp/scheduling/29567-passing-scheduled-events-through-a-generated-json-file.md
fanf: Ah, funny - it was rewritten by agent, I corrected other bits, but that ones felt ok. I will try to rewrite it.

fanf: ok, I didn't commited the one I edited, this one is a previous version not dealing with the last changes we did in path etc.

PR #7436, opened 2026-08-26
Fixes #29631: OTP page input should have focus upon page load

clarktsiory:
https://issues.rudder.io/issues/29631

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/autofocus: to avoid additional click/navigation


RaphaelGauthier (approved):
LGTM!

PR #7440, opened 2026-08-26
Fixes #29549: If one parameter's id contains another parameter's id, the recent activity table of the second parameter will contain event logs that concern the first parameter

P4uline:
https://issues.rudder.io/issues/29549

Screenshot From 2026-08-26 17-13-46

Add the ability to filter by id
- new field id in `RequestEventLog` : `objectId`
- ~~second pass of filtering at the service level
The id is not stored in postgresql as a dedicated column, it's contained in the details field which is an xml content.
So after the full text search postgresql query another pass of filtering in the xml content is needed, using the library scala.xml, to match the exact value of the id watching tags `<name />` of `<id /> `depending on the object.
The second pass of filtering is made in the service layer `EventLogServiceImpl`.~~
It wasn't the right approach :point_up: 
Instead: enrich the SQL query with xpath to be able to filter on XML id/name equal to the object id in XML
- the elm app calling the applications are updated to pass an id and a search


P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/eventlog/EventLogServiceImpl.scala
```diff
@@ -54,12 +55,50 @@ import doobie.*
...
+              val id = (node \\ "id").text
+              id.equals(i.value)
+            case Constants.XML_TAG_DIRECTIVE        =>
+              val id = (node \\ "id").text
+              id.equals(i.value)
+            case Constants.XML_TAG_EDITOR_TECHNIQUE =>
+              val id = (node \\ "id").text
+              id.equals(i.value)
+            case other                              =>
+              EventLogsLoggerPure.warn(s"Filtering by id on event logs we should not have this object type here ${other}")
```
fanf, line 94: I think it's debug level, not warn. Ask yourself: is this a message than an ops need to know and act on, and know what to do (warn) or is it for dev (debug, trace)

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepositoryTest.scala
```diff
@@ -175,6 +175,6 @@ class EventLogJdbcRepositoryTest extends Specification with IOChecker with DBCom
     )
   )
 
-  def defaultFilter = EventLogRequest(0, 10, None, None, None, None, None, None)
+  def defaultFilter = EventLogRequest(0, 10, None, None, None, None, None, None, None)
```
fanf, line 178: It would be nice to have an actual test that used to fail and now works with the correction

fanf (changes requested):
Can you add a test demoing what the new patch does ?

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepositoryTest.scala
P4uline: Yes I should add some tests.
If I keep the implementation as is, then I should rather add a Service test because this change has no effect on the SQL query.

But I've changed my mind. 
I will rework the PR to remove the filtering at the service level and make the filter in the sql query like the other filter critiera. It will make the query a bit more complex: using xpath to query name or id in XML data, but the second pass of filtering in the service level will possibly break the pagination (cubersome to handle).

P4uline:
PR updated with a new commit

P4uline:
- I removed the filtering in the service
- I added an objectId filter in EventLogRequest
- I modified the query to have a where clause with xpath on name and id equal to objectId 
- I modified elm activity module to encode an objectId instead of a search, we're not doing a full text search anymore

I keep it in draft as there is more clean and tests to do

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

P4uline:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/jdbc/EventLogJdbcRepository.scala
```diff
@@ -342,12 +342,37 @@ object EventLogJdbcRepository {
...
+      - /entry/globalParameter/name/text() -> <entry><globalParameter><name>my-global-param</name></globalParameter></entry>
+      - /entry/directive/id/text() -> <entry><directive><id>73a0c118-f0d3-4136-a905-a1c726055ea8</id></directive></entry>
+      - /entry/rule/id/text() -> <entry><rule><id>73a0c118-f0d3-4136-a905-a1c726055ea8</id></rule></entry>
+      - /entry/technique/id/text() -> <entry><technique><id>9b49ec98-b1fe-45ec-90b2-78371ef53155</id></technique></entry>
+      - /entry/nodeGroup/id/text() -> <entry><nodeGroup><id>9b49ec98-b1fe-45ec-90b2-78371ef53155</id></nodeGroup></entry>
+     */
+    val objectId = filter
+      .flatMap(f => f.objectId)
+      .map(objectId =>
+        fr"${objectId.value} = ANY(xpath('/entry//id/text()', data)::text[]) OR ${objectId.value} = ANY(xpath('/entry//name/text()', data)::text[])"
```
clarktsiory, line 356: Because `id` is put first, this makes postgres read the xml and search for an `id` that does not exist for every global parameter event: this has hidden cost that needs to be addressed (by adding some index) or by making that explicit in a comment if it's fine because of scope of the "recent activites request".

I assume that global parameters are not the main types, but it may still have non-negligible load...

clarktsiory (commented):
Have you tried it in a load test machine ? If it's problematic, usage of a gin index seems possible for XML to optimize the query, but it would require a migration...

If it's acceptable and performance tuning in event log is the subject of future optimizations, it should be taken note of, and the PR can be merged

P4uline:
> Have you tried it in a load test machine ? If it's problematic, usage of a gin index seems possible for XML to optimize the query, but it would require a migration...
> 
> If it's acceptable and performance tuning in event log is the subject of future optimizations, it should be taken note of, and the PR can be merged

The work about the performance is in progress as you know (we planned a discussion for next friday), and it's about the full text search.

Here I replaced a full text search using `ilike` searching a text in all XML content to an xpath query selecting an element strictly equal to (not like) the object id. 
So at the best, performance should be better, at the worse performance should be the same. There's no reason this change should degrade performance.

clarktsiory (approved):
OK, so I'm approving this, provided that we make performance improvements later

PR #7442, opened 2026-08-27
Fixes #29647: JSON payload in create technique endpoint is completely wrong

ElaadF:
https://issues.rudder.io/issues/29647

### DIFF
--- a/webapp/sources/api-doc/components/schemas/editor-technique.yml
+++ b/webapp/sources/api-doc/components/schemas/editor-technique.yml
@@ -5,7 +5,7 @@ properties:
   id:
     type: string
     description: Technique id
-    example: security-policy
+    example: security_policy
   name:
     type: string
     description: Technique name
@@ -17,11 +17,17 @@ properties:
   category:
     type: string
     description: category of this technique
-    example: "user_techniques"
+    example: "ncf_techniques"
   description:
     type: string
     description: description of this technique
     example: "This techniques apply generic security policies"
+  documentation:
+    type: string
+    description: documentation in markdown of the technique, exposed on directive side
+    example: "This technique install package based on the *name* through a *technique parameter*"
+  tags:
+    type: object
   source:
     type: string
     description: Source of the technique, always editor here
--- a/webapp/sources/api-doc/components/schemas/technique-block.yml
+++ b/webapp/sources/api-doc/components/schemas/technique-block.yml
@@ -2,6 +2,10 @@
 # SPDX-FileCopyrightText: 2023 Normation SAS
 type: object
 properties:
+  type:
+    type: string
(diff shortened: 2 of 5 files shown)


webapp/sources/api-doc/components/schemas/editor-technique.yml
clarktsiory, line 4: it seems too that many attributes are required: `description`, `documentation`, etc. see https://github.com/Normation/rudder/blob/d0ff7571743f4343783bdc959922eaba5a06afe5/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala#L105-L117

something like:
```suggestion
type: object
required:
  - id
  - version
  - name
  - category
  - description
  - documentation
  -  calls
  - parameters
  - resources 
  - tags 
properties:
```

webapp/sources/api-doc/components/schemas/technique-method-call.yml
```diff
@@ -2,6 +2,10 @@
 # SPDX-FileCopyrightText: 2023 Normation SAS
 type: object
 properties:
```
clarktsiory, line 4: same comment here: there are a lot of required parameters that are not specified, see https://github.com/Normation/rudder/blob/d0ff7571743f4343783bdc959922eaba5a06afe5/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/EditorTechnique.scala#L169-L197

The `type` is required, but also:

```suggestion
type: object
required:
  - type
  - id
  - component
  - reportingLogic
  - condition
  - calls
properties:
```

webapp/sources/api-doc/components/schemas/technique-block.yml
```diff
@@ -2,6 +2,10 @@
 # SPDX-FileCopyrightText: 2023 Normation SAS
 type: object
 properties:
```
clarktsiory, line 4:
```suggestion
type: object
required:
  - type
  - method
  - id
  - parameters
  - condition
  - component
  - disabledReporting
properties:
```

clarktsiory (commented):
Nice ! Could you also add the mandatory parameters as suggested ?

webapp/sources/api-doc/components/schemas/technique-method-call.yml
amousset: this makes no sense to require all these, I think it is a bug in the API instead

clarktsiory: yes it's sadly the default of the current API.
we will address the actual optional VS mandatory values in another issue to correct the documentation: https://issues.rudder.io/issues/29652

clarktsiory (approved):

PR #7445, opened 2026-08-27
Fixes #29661: Remove unused js function in rudder-datatable

VinceMacBuche:
https://issues.rudder.io/issues/29661

  ## What

  Removes unreachable code from `rudder-datatable.js`: **351 lines deleted, 1 file changed**,
  no behaviour change and no API change.

  Removed:

  - **jQuery DataTables factories with no entry point**
    - the whole "Rule compliance details" section: `createComponentTable`,
      `createNodeComponentValueTable`, `createRuleComponentValueTable` (~205 lines).
      `createComponentTable` only called *itself* recursively, which is why it stayed
      invisible to a naive search.
    - `createChangesTable`
    - `createInnerTablerow`, used only by the three above
  - **dead globals**: `anOpen` (only used by `createInnerTablerow`), `inventories`
  - **orphaned helpers**: `computeCompliancePercentFromString`,
    `isApplyingFromComplianceString`, `reportsSum`, `selectInterval`, `changeCursor`

  ## When each stopped being used

  For each symbol, the commit that removed its last call site, and the first release
  containing that commit:

| function(s) | Unused since | 
|---|---|
| `selectInterval`, `changeCursor` | **4.1.0** |
| `inventories` | **6.2.0** | 
| `reportsSum` | **6.2.0** |
| `createChangesTable` | **8.0.0** | 
| `isApplyingFromComplianceString` | **8.1.3** | `
| `computeCompliancePercentFromString` | **9.0.0** | 
| `createComponentTable`, `createNodeComponentValueTable`, `createRuleComponentValueTable`, `createInnerTablerow`, `anOpen` | **9.1.0** | 

  Two things worth noting for reviewers:

  - The largest chunk is recent. The DataTables v2 migration (#27863, 9.1.0) removed the
    entry points — `createDirectiveTable`, `createInnerTable`, `localComponentTable`,
    `localDirectiveTable` — but left the callees behind. This is leftover cleanup from that
    migration, not old sediment.
  - `selectInterval` still manipulated a `c3-bar-highlighted` class, from the c3.js era;
    c3.js has not been a dependency for years.

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
+++ b/webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
@@ -35,11 +35,9 @@
 *************************************************************************************
 */
 
-var anOpen = [];
 var ruleCompliances = {};
 var recentChanges = {};
 var recentChangesCount = {};
-var inventories = {};
 var recentGraphs = {};
 
 $.extend(true, DataTable.defaults, {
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -1532,6 +1325,7 @@ function createTechnicalLogsTable(gridId, nodeId, data, contextPath, refresh, re
 
 }
 
+<<<<<<< Updated upstream
```
fanf: That needs to be removed

fanf (changes requested):
There is a merge marker still present. 

Also for next time: can you reword a bit the description outputed by the agent (remove the verbiage linked to its work thread, only keep what is relevant to that PR - shorter for the same information content)

VinceMacBuche:
Commit modified

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
VinceMacBuche: i did not catch it because it happend at last rebase after i reviewed it locally when creating the branch

VinceMacBuche:
Do you mean the follow up parts ? i thought it was interesting to keep it to give some insights of what could follow 

or the reviewer insights ?

fanf:
Not sure about the follow up - I think I would prefer to have them in a a linked issue than here, where they don't matter and are not very actionnable. 
For the big array with the date, I know: it's too much and it doesn't bring anything interesting. It's a typical agent bias: very precise and detailled information that we don't care about for the task at hand.

fanf (approved):
LGTM

VinceMacBuche:
Ok, in fact i asked explicitly to build the table with the last tag the function was used (maybe the commit column is useless, sure). it was something that make me choose to remove things and accept the change so i was thinking it would be useful

PR #7447, opened 2026-08-27
Fixes #29663: Autovacuuming of nodelastcompliance could be improved

clarktsiory:
https://issues.rudder.io/issues/29663

Found the origin of the issue:
* TOAST table is the culprit (it's the internal PG table that stores the big JSONB compliance with pointers to chunks of binary)
* solution is to adjust toast-specific vacuum options (that will apply only to the toast table, the main table is fine WRT autovacuum)

* ..still tuning the values, I just tried lower ones to prove that it affects the vacuuming with even 2GB, so it should work with higher value...

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
+++ b/webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
@@ -205,6 +205,26 @@ CREATE TABLE NodeLastCompliance (
 , details             jsonb NOT NULL
 );
 
+-- Details of compliance can be very big and occupy TOAST space (depending on directives numbers and size).
+--
+-- Trigger for autovacuum is: autovacuum_vacuum_threshold + autovacuum_vacuum_scale_factor × reltuples
+-- Default is 50 + 0.2 * rows, also applies to TOAST.
+--
+-- We don't want fixed threshold, but scaled factor that would adapt to any number of nodes, behave
+-- the same across different installations.
+--
+-- Finally, we need a factor lower than the default, since by default it leaves quite high number of dead tuples
+-- in the TOAST table without triggering, find it with this query:
+-- SELECT relid::regclass, n_live_tup, n_dead_tup, n_ins_since_vacuum,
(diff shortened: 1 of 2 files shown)


fanf:
OK, nice. IIRC, with PG 14, we have an option to do that automatically by table :)

webapp/sources/rudder/rudder-core/src/main/resources/reportsSchema.sql
```diff
@@ -105,34 +105,34 @@ CREATE TABLE RudderSysEvents (
...
-CREATE INDEX composite_node_execution_idx ON RudderSysEvents (nodeId, executionTimeStamp);
-CREATE INDEX ruleId_idx                   ON RudderSysEvents (ruleId);
+create index executionTimeStamp_idx       on RudderSysEvents (executionTimeStamp);
+create index composite_node_execution_idx on RudderSysEvents (nodeId, executionTimeStamp);
+create index ruleId_idx                   on RudderSysEvents (ruleId);
 
-CREATE INDEX endRun_control_idx ON RudderSysEvents (id) WHERE eventType = 'control' and component = 'end';
-CREATE INDEX changes_executionTimeStamp_idx ON RudderSysEvents (executionTimeStamp) WHERE eventType = 'result_repaired';
+create index endRun_control_idx on RudderSysEvents (id) WHERE eventType = 'control' and component = 'end';
+create index changes_executionTimeStamp_idx on RudderSysEvents (executionTimeStamp) WHERE eventType = 'result_repaired';
```
fanf: I don't understand the partial case change - and if some would be done, I think I would prefer "every SQL key words upper case"

clarktsiory: that's my IntelliJ... :sweat_smile:

clarktsiory:
Closing this because it does not solve the main issue of bloat that prevented the autovacuum.
Tuning with these parameters does not seem to overcome the limitations documented in https://github.com/Normation/rudder-doc/pull/1162, refer to that PR instead

PR #7448, opened 2026-08-27
Fixes #29519: exotic passwords fails vulnerability management

fanf:
https://issues.rudder.io/issues/29519

So, that was far less trivial than initially though, the main reason being that there isn't really a good, small INI parser in JVM that is maintained and matches `serde_ini` parsing/escaping even a bit. 

So it was easier (and more tedious) to just do our own. 

The main points: 
- we have a new INI reader/writer in scala in `util` packages following `serde_ini` semantic in the tiny subset that we use
    - and there is a lot of details about trim (serde doesn't do that), white spaces (they are not the same in java and rust), split (doesn't do what we want), etc. 
    - the general rule is that our file format is very strict (because of `serde_ini`), which is certainly good because it lets very few space for client file having diverged: they don't have the lattitude for that, not even a space in EOL. 
- we have a common test file and common test rules between rust and scala (the test file lives in the rust part), 
- we have a common `PluginSettingService` used everywhere in Rudder, including in CVE (see https://github.com/Normation/rudder-plugins-private/pull/1479)
- `rudder-pkg.conf` file is now a global property exposed in `RudderConfig` and used elsewhere (in plugins),
- we have an explicit list of known keys, in a specific order for `PluginSettings`, other are ignored. 

Most of the code is simple, just... I really would have prefered to not have to write here.

### DIFF
--- a/relay/sources/rudder-package/src/config.rs
+++ b/relay/sources/rudder-package/src/config.rs
@@ -129,6 +129,7 @@ mod tests {
     use std::path::Path;
 
     use pretty_assertions::assert_eq;
+    use secrecy::ExposeSecret;
 
     use crate::config::{Configuration, Credentials, ProxyConfiguration};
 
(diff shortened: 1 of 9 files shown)


fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #7451, opened 2026-08-28
Fixes #29655: Technique change logs miss a link to go the technique view

P4uline:
https://issues.rudder.io/issues/29655

### DIFF
--- a/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
+++ b/webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
@@ -99,6 +99,7 @@ class EventLogDetailsGenerator(
 
   def displayDescription(event: EventLog): NodeSeq = {
     import linkUtil.*
+
     def crDesc(x: EventLog, actionName: NodeSeq) = {
       val id   = RuleId.parse((x.details \ "rule" \ "id").text).getOrElse(RuleId(RuleUid("")))
       val name = (x.details \ "rule" \ "displayName").text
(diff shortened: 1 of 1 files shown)


webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
```diff
@@ -204,8 +205,12 @@ class EventLogDetailsGenerator(
     }
 
     def editorTechniqueDesc(x: EventLog, actionName: NodeSeq) = {
-      val name = (x.details \ "technique" \ "name").text
-      Text(s"Technique ${name} ${actionName}")
+      val id   = (x.details \ "technique" \ "id").text
+      val name = (x.details \ "technique" \ "displayName").text
+      Text("Technique ") ++ {
+        if (id.size < 1) Text(name)
+        else <a href={name}>{name}</a> ++ actionName
```
clarktsiory: The `href` should be a link, we should use `techniqueLink(id)`

Also, I don't understand the conditions and the reason to use the `id`:
* isn't `id` always defined and size always non empty ?
* there seems to be 3 variants for the actual value of `name`: 
  * in case of deletion/addition of technique it's indeed `displayName`: https://github.com/Normation/rudder/blob/92823cc2041cef1271822efe9ff89d9a2fa8dc83/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/eventlogs/EditorTechniqueXmlSerialisation.scala#L48
  * in case of modification, there is no `displayName` but a `previousName`, which is always there: https://github.com/Normation/rudder/blob/92823cc2041cef1271822efe9ff89d9a2fa8dc83/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/ncf/eventlogs/EditorTechniqueXmlSerialisation.scala#L188-L190, we can't use `name` which is the "name diff", see the screenshot:  
technique diff details

P4uline:
PR updated with a new commit

P4uline:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/web/services/EventLogDetailsGenerator.scala
P4uline: > isn't id always defined and size always non empty ?

It should, as there is a check on the id for rule, directive, group I kept the same logic. 
It was probably to cover legacy design. I suppose as the technique change logs are pretty recent, we can get rid of this check.

> there seems to be 3 variants for the actual value of name

Indeed, the link didn't work for modified techniques, I added an additional case.
Here the logic is slightly different for the directive, do you know why ? 
If we migrate the format of all event log XML entries in the future, it could be the opportunity to normalize all elements.

P4uline:
PR updated with a new commit

clarktsiory (approved):

P4uline:
OK, squash merging this PR

PR #7452, opened 2026-08-28
Fixes #29500: Technique modification event log rollback causes sql error

clarktsiory:
https://issues.rudder.io/issues/29500

This one is a crazy combination of bugs, each one needed to be fixed to make it work (see the list below).

It revolves around the `gitcommit` table (which has no constraint, so it allows duplicate pairs): 
```sql
+----------------+------+-----------+
| Column         | Type | Modifiers |
|----------------+------+-----------|
| gitcommit      | text |  not null |
| modificationid | text |           |
+----------------+------+-----------+
Indexes:
    "gitcommit_pkey" PRIMARY KEY, btree (gitcommit)

+------------------------------------------+--------------------------------------+
| gitcommit                                | modificationid                       |
|------------------------------------------+--------------------------------------|
| 4607c5a8d94b7a7ec3351a2285aec89aded6d3fa | 7a714ee5-cd69-4122-971d-1ed5a517501e |
| 9dbc4a0dd759e3068eaad982a90417270337d5e2 | 7a714ee5-cd69-4122-971d-1ed5a517501e |
+------------------------------------------+--------------------------------------+
```

Upon reading the `gitcommit` table with doobie, we assume that we have one single match:
  * I left it untouched, it's a bigger architectural issues on eventlogs that need to be addressed in upper branches   
  * the proposed solution is to ensure we write something that we can read back: write a unique (gitcommit, modificationid) pair 

---
### The bugs that were fixed

1. appending the commit of the technique modifications in `configuration-repository/techniques` to the table was simply forgotten (upon save and deletion, in `TechniqueArchiverImpl`)
    * but simply adding the commit will conflict with number 2:
2. rollback of the technique commit happened on the wrong modification, there was a `modificationid` bound to the wrong commit which changed `configuration-repository/directives`, the culprit is the _callback_ `TechniqueAcceptationUpdater`:
    * solution is to avoid linking the `modificationid` of callbacks to a commit: use a dedicated implementation that do not commit (see https://github.com/Normation/rudder/pull/7452#issuecomment-5512592436)
3. the same solution needs to be applied to `DeleteEditorTechniqueImpl`, which otherwise would generate multiple commits in the table for the same modification
4. with all that, restoration of `techniques/` was not possible until it's correctly added to `PartialArchive`  
5. (a small one:) the semaphore of git repo was not used: it needs to be 

---
### Other changes
* addition of tests when saving a technique, to ensure that the git commit list in memory is correctly updated
* change of utility `newCC`: add the optional `modId`, it looks fine with 9.2
* cleanup of unused methods in the archiver class

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/GitItemRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/GitItemRepository.scala
@@ -267,6 +267,41 @@ trait GitConfigItemRepository extends GitItemRepository {
 }
 
 /**
+ * A `GitConfigItemRepository` that commits without linking the commit to the modification id.
+ *
+ * Sometimes, modId are linked to several operation that each have a linked commit entry (to keep the whole change traceable).
+ * But we don't always want to link the commits to that modId, typically for callback-like changes.
+ *
+ * So, use this mixin to enforce that a modification is not linked to a commit (e.g. for a secondary commit of a user change).
+ *
+ * See https://issues.rudder.io/issues/29500, for the case of technique callbacks
+ */
+trait GitConfigItemRepositoryWithoutModId extends GitConfigItemRepository {
+
+  override def commitAddFileWithModId(
+      modId:         ModificationId,
+      commiter:      PersonIdent,
+      gitPath:       String,
+      commitMessage: String
+  ): IOResult[GitCommitId] = commitAddFile(commiter, gitPath, commitMessage)
+
+  override def commitRmFileWithModId(
+      modId:         ModificationId,
+      commiter:      PersonIdent,
+      gitPath:       String,
+      commitMessage: String
+  ): IOResult[GitCommitId] = commitRmFile(commiter, gitPath, commitMessage)
+
+  override def commitMvDirectoryWithModId(
+      modId:         ModificationId,
+      commiter:      PersonIdent,
+      oldGitPath:    String,
+      newGitPath:    String,
+      commitMessage: String
+  ): IOResult[GitCommitId] = commitMvDirectory(commiter, oldGitPath, newGitPath, commitMessage)
+}
+
+/**
  * Utility trait that factor global commit and tags.
  */
 trait GitArchiverFullCommitUtils extends NamedZioLogger {
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
@@ -38,11 +38,8 @@
 package com.normation.rudder.repository.xml
 
 import com.normation.NamedZioLogger
-import com.normation.cfclerk.domain.RootTechniqueCategoryId
 import com.normation.cfclerk.domain.SectionSpec
-import com.normation.cfclerk.domain.SubTechniqueCategoryId
 import com.normation.cfclerk.domain.Technique
-import com.normation.cfclerk.domain.TechniqueCategoryId
 import com.normation.cfclerk.domain.TechniqueCategoryMetadata
 import com.normation.cfclerk.domain.TechniqueId
 import com.normation.cfclerk.domain.TechniqueName
@@ -65,7 +62,9 @@ import com.normation.rudder.domain.policies.*
 import com.normation.rudder.domain.properties.GlobalParameter
 import com.normation.rudder.git.GitArchiveId
 import com.normation.rudder.git.GitArchiverFullCommitUtils
+import com.normation.rudder.git.GitCommitId
 import com.normation.rudder.git.GitConfigItemRepository
+import com.normation.rudder.git.GitConfigItemRepositoryWithoutModId
 import com.normation.rudder.git.GitPath
 import com.normation.rudder.git.GitRepositoryProvider
 import com.normation.rudder.ncf.ResourceFile
@@ -245,17 +244,6 @@ trait TechniqueArchiver {
       msg:         String
   ): IOResult[Unit]
 
-  /*
-   * Delete a category with all files and techniques it contains.
-   * WARNING: you should probably not do that.
-   */
-  def deleteCategoryRecursively(
-      categoryId: TechniqueCategoryId,
-      modId:      ModificationId,
-      committer:  EventActor,
-      msg:        String
-  ): IOResult[Unit]
-
   def saveTechnique(
       techniqueId:     TechniqueId,
       categories:      Seq[String],
@@ -271,19 +259,6 @@ trait TechniqueArchiver {
       modId:      ModificationId,
       committer:  EventActor,
       msg:        String
-  ): IOResult[Unit]
-
-  /*
-   * Add or update a technique category with all its content, recursively.
-   * In intent, it's a:
-   * `git -a categoryPath; git -u categoryPath; git commit`
-   * If the category path doesn't exists, it's a no-op.
-   */
-  def updateTechniqueCategoryRecursively(
-      category:  TechniqueCategoryId,
-      modId:     ModificationId,
-      committer: EventActor,
-      msg:       String
   ): IOResult[Unit]
 
   /*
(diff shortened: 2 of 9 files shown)


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/TechniqueRepositoryImpl.scala
```diff
@@ -186,9 +187,11 @@ class TechniqueRepositoryImpl(
           changed -- otherToRemove ++ moveToAdd
         }
 
-        val res = Control.bestEffort(callbacks) { callback =>
+        // propagated updates should not be linked to same modId (gitcommit traceability issue)
```
fanf: I'm not sure I agree with that. A modification should be able to trace several changes, functionally linked together. See that more like a trace ID.

It seems ok from a design standpoint to have several commitid (or 0, or 1) linked to the same modid

And this is also partly why there is two concepts, not one (the other being that modid is business level, commitid is persistence level)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/TechniqueRepositoryImpl.scala
```diff
@@ -186,9 +187,11 @@ class TechniqueRepositoryImpl(
...
+        val callbackCC = cc.withModId(ModificationId(uuidGen.newUuid))
```
fanf: Linked to your #2: why keeping the callback commit doesn't break consistency on rollback ? If we for ex remove an added technique, its active technique version must be removed too, no ? (In a revert way, not restore one)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -370,32 +344,20 @@ class TechniqueArchiverImpl(
...
+          rev    <- IOResult.attempt(gitRepo.git.commit.setCommitter(ident).setMessage(msg).call())
+          commit <- IOResult.attempt(GitCommitId(rev.getName))
+        } yield commit)
+      _      <- gitModificationRepository.addCommit(commit, modId)
     } yield {
       s"${relativePath}/${categoryPath}/${techniqueId.serialize}"
     }).chainError(s"error when deleting and committing Technique '${techniqueId.serialize}").unit
   }
 
-  def deleteCategoryRecursively(
```
fanf, line 381: Why deleting that ? Is this dead code ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -245,17 +243,6 @@ trait TechniqueArchiver {
       msg:         String
   ): IOResult[Unit]
 
-  /*
-   * Delete a category with all files and techniques it contains.
-   * WARNING: you should probably not do that.
-   */
-  def deleteCategoryRecursively(
```
fanf, line 252: Why deleting that ? Is this dead code ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -273,19 +260,6 @@ trait TechniqueArchiver {
       msg:        String
   ): IOResult[Unit]
 
-  /*
-   * Add or update a technique category with all its content, recursively.
-   * In intent, it's a:
-   * `git -a categoryPath; git -u categoryPath; git commit`
-   * If the category path doesn't exists, it's a no-op.
-   */
-  def updateTechniqueCategoryRecursively(
```
fanf, line 282: Same

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -503,9 +470,9 @@ class TechniqueArchiverImpl(
   }
 
   def saveTechniqueCategory(
-      categories: Seq[String], // path (inclusive) to the category
+      categories: Seq[String],    // path (inclusive) to the category
       metadata:   TechniqueCategoryMetadata,
-      modId:      ModificationId,
+      modId:      ModificationId, // unused: no gitcommit for technique category save, which has no rollback commit
```
fanf, line 476: Wd may want to trace modid for logs or whatever independently to git commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -530,33 +497,15 @@ class TechniqueArchiverImpl(
...
+                                     gitRepo.git.add.addFilepattern(catGitPath).call()
+                                     gitRepo.git.commit.setCommitter(ident).setMessage(msg).call()
+                                   })
                         } yield ()
                       }
         } yield ()).chainError(s"error when committing technique category '${catGitPath}'").unit
     }
   }
-
-  override def updateTechniqueCategoryRecursively(
```
fanf, line 541: Is this dead code ?

fanf (changes requested):
I'm not convinced that this is the correct design approach.

As a design, I really want to be able to trace one functional change with an identifier, that can span several technical changes, some of which might be linked to git commit. If you force one modid == one commit id, you lode that possibility (and you have two entities for the same opaque id) k

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/TechniqueRepositoryImpl.scala
clarktsiory: AFAIK the current `gitcommit` table is only used for the purpose of rollbacks, and there are strong assumptions in the design that "1 git commit == 1 eventlog", and upon reading the table that "1 modification == 1 eventlog".

So, I'm not sure I understand the need for a trace by now, as it's not visible to users. But I would have assumed that we need another identifier for that, a "batchtraceid" to trace the impact of one change along with all its callbacks...

The alternative if we keep the `modificationid` as the trace ID, is that we have to "revert all the git commits" that are in the table. I'll have a look at that, it should be feasible but I don't know how complex that will be

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/TechniqueRepositoryImpl.scala
clarktsiory: the active technique version is a callback, called for any occurrence of saving a technique: so, as a callback, it's automatically triggered even upon rollback and will converge.

In the current implementation, there is always single commit that we directly rollback: we rollback 1 eventlog with the assumed "1 git commit == 1 event log", with the assumed "1 modification == 1 eventlog" (this may need to change, following the discussion in the previous comment)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/TechniqueRepositoryImpl.scala
fanf: I think we were just in the simple cases where 1 mod == 1 commit hold until technique. Perhaps the design goal was lost because of that

clarktsiory: maybe, I would need some analysis, I didn't check if other items also have separate modification ID (I don't recall that, but it will be worth checking, to help make the decision)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/cfclerk/services/impl/TechniqueRepositoryImpl.scala
fanf: in the case where callback should not be rollbacked even if they have a commit linked to the same modid, I think we miss a distinction between rollbackable and non rollbackable technical change. 
So we have: 
- a modId 
- is an identifier for a change, 
- which is a list of operations, 
- some operations are rollbackable, some not,
- some operations produce git commits, other not

clarktsiory:
PR updated with a new commit

clarktsiory:
Instead of having a new modification ID being generated, I changed it to something even simpler: the instance of archiver that is used in callbacks is an instance that should be producing no extra commit. So we just need to replace it with an archiver that does not touch the `gitcommit` table.

It means we still maintain the `1 modificationId == 1 commit` for techniques (by hand) now. In the future we should make the constraints represented stronger (at type level, with db constraints)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/git/GitItemRepository.scala
```diff
@@ -266,6 +266,41 @@ trait GitConfigItemRepository extends GitItemRepository {
...
 
+/*
+ * A `GitConfigItemRepository` that commits without linking the commit to the modification id.
+ *
+ * A modification is linked to one and only one commit: the one a rollback of the corresponding event
+ * logs is built on. But a change asked by the user can lead to several commits, because everything
+ * archived as a *consequence* of it is archived under the same modification id (to keep the whole
+ * change traceable). Those consequences are archived through an item repository mixing that trait in,
+ * so that only the change the user asked for links its commit.
+ * See https://issues.rudder.io/issues/29500
```
fanf: can you try to retrain claude verbosity ? Or rephrase it in two lines ?
Something like: 
 * A `GitConfigItemRepository` that commits without linking the commit to the modification id.
 * Sometimes, modId are linked to several operation that commits, but we don't want to linked the commits to that modid - typically for callback-like changes. 
 * See: ...

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
```diff
@@ -820,6 +773,13 @@ class UpdatePiOnActiveTechniqueEvent(
 
 /**
  * A specific trait to create archive of an active technique.
+ *
+ * It archives without linking its commits to the modification (`GitConfigItemRepositoryWithoutModId`):
+ * the event logs of an active technique change (`ModifyTechnique`, `DeleteTechnique`) are not
+ * rollback-able, so no rollback can ever be built on its commit, while archiving an active technique
+ * does happen as a consequence of a change on the reference technique library (accepting a new
+ * technique, updating its acceptation datetime, deleting it), under that change modification id.
+ * See https://issues.rudder.io/issues/29500
```
fanf, line 782: same

fanf (commented):
Can you reformulate the comments ? Else, it looks good and much more simple than before 👍

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/GitArchivers.scala
clarktsiory: Indeed, I separated it into sentences as I  and :white_check_mark: 
(I should perhaps add something in my CLAUDE.md for that)

fanf (approved):
LGTM

clarktsiory:
OK, squash merging this PR

PR #7453, opened 2026-08-28
Fixes #29667: Define global vars for group/rule root categories to avoid misnaming

fanf:
https://issues.rudder.io/issues/29667

Introduce `Constants.ROOT_GROUP_CATEGORY` and `Constants.ROOT_RULE_CATEGORY` and use them everywhere in place of the hard-coded string.


clarktsiory (approved):
LGTM

We'll likely need something like that on Elm side too

PR #7455, opened 2026-08-28
Fixes #29625: error 500 when deleting a campaign

fanf:
https://issues.rudder.io/issues/29625

So, we were doing partial function without handling the cases when param is out of domain, which leads to a `MatchError` by construction. 
We didn't see the `MatchError`, because our API error handling function was refactored in 9.2 and doesn't display it correctly in logs. 

So, that PR addresses both - in 9.2, because before 9.2, we didn't had cases out of PF domain (only kind of campaign was update campaigns). 

# Correct the API error log case

When we reach the `catchAllDefect` in `toLiftResponseGeneric`, that means that's an error out of `IOResult`, so most likely a code error. Only dev can do something, but they need to get the info. 
So now, we: 
- display a message in JS console telling people to ask for the admin assistance, because they can't do anything, 
- add a warning log so that the bug (which is a bug at that point, or something strange enought that it needs a ticket) with the stack trace

image

image

# MatchError in partial function

We missed the `orElse(defaultCase)` in the handler chain, so I added on in every place where `delete` was called. 
The default case is `IgnoreAndStop`, ie don't do anything more.

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/campaigns/MainCampaignService.scala
@@ -42,6 +42,7 @@ import com.normation.rudder.campaigns.CampaignEventStateType.*
 import com.normation.utils.StringUuidGenerator
 import org.joda.time.{Duration as JTDuration, *}
 import zio.*
+import zio.syntax.*
 
 /*
  * Method used by API of the campaign service, implemented by the top level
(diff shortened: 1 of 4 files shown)


webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-finished/91-compress-linux-policies
clarktsiory, line 1: should this hook not be executable ? was this change unattended ?

clarktsiory (commented):
LGTM, except for that extra change on hooks

webapp/sources/rudder/rudder-core/src/main/resources/hooks.d/policy-generation-node-finished/91-compress-linux-policies
fanf: completly :facepalm:

fanf:
PR updated with a new commit

clarktsiory (approved):

fanf:
OK, squash merging this PR

PR #7456, opened 2026-08-28
Fixes #29675: Make TenantCheckLogic api easier to use and update

fanf:
https://issues.rudder.io/issues/29675

Rework of the tenant API, with some cleaning and corrections

# API of tenant repo filtering based on `manageXXX` operations

The API for `TenantCheckLogic` for the write part was complicated and brittle from an user point of view. 
So I created a new API that handle all the business logic under function looking like: 
```
manageXXX(...){ object => underlyingOp(object) }
```

This allow to cleanly check for object existence with the correct rights for save, etc. 

So now, `TenantFiltering` (the proxy layer on top of repositories that manages tenant check) is very simple, and the `TenantCheckLogic` is a bit more complex (there was a lot of operation cases), but also more consistent. 

I introduced two types: `Lookup` and `Container` that are here to let the user give us the method to query the underlying object (or container object for move/creation) and let us managed the query context. 

Delete operation are bit special and sometime return noop when object is missing, sometime return an error ; so we need a parameter for that, two. 

Perhaps start with the `TenantFiltering.scala` file, because it shows how the new API is used and so the reason for that whole PR. Then `TenantService.scala`, for how it's done (and the complicated parts). 

Note : I'm not sure about one of the laws.
When an object is created, it gains user tenants, but only the ones read/write, not the read-only ones. It will be easier to change on that way in any case if needed.

## add a `manageRestore` case

For object rollback, we may be not respecting our monotonic law and restore an object with fewer tenant that is had before the restore. This mean that part of the change log may be expose to a tenant that is no longer on the object. 

I don't think it's too grave, but still, it breaks the general case, so it is identified in a dedicated operation. 

# RudderConfig plumbing

That was a whole mess. There was still services instanciated with the old not-tenant filtered repo. 
To make that impossible, I've hidden the raw repo in a block where tenant ones are instantiated. Since perhaps we may need them, they are still accessible in a `xxxRepo.storage` structure, but at least it will be clear that it was a choice. 
I didn't find a case for it, so now all the other services use tenant filtered repo. 

# ActivteTechnique repository

We were missing bits that combined with the previous point made tenant be lost on active techniques and techniques (and the filtering wasn't working correctly either, sometime using the repo directly). 
It's all cleaned and works well now 

# Remove implicit for special QueryContext values

We have several special query context (`todoQC`, `systemQC`, etc). They were noted `implicit` which means they were in the scope and produced implicit resolution while we always explictly call them now. 
So I just removed the implicits.


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -3235,20 +3244,19 @@ object RudderConfigInit {
     lazy val woLDAPParameterRepository: WoParameterRepository = {
       new WoTenantParameterRepo(
         tenantCheckLogic,
```
clarktsiory: it seems parameters are not using the TenantScoped pattern as for other repos

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
```diff
@@ -83,6 +84,55 @@ trait TenantService {
...
+}
+
+/*
+ * Delete operations are inconsistent: sometime, when the object is already absent, it's a noop, sometime an
+ * error. We want to normalize toward noop, but in the meantime, we have to specify it with that enum.
+ */
+enum IfAbsent[+R] {
+  case Noop(result: R)
+  case Fail(msg: String)
+}
```
clarktsiory, line 142: Usage of explicit `IfAbsent` type is great, but looks like a hidden `Either` ?
At least I would have thought about using an Either first, so I would have used an `opaque type IfAbsent[+R] = Either[String, R]`. Since it fits as a `toIO` in `whenAbsent` place.

But since it's not used for standard `map` etc. operations, it's fine as an enum too for now

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/TenantFiltering.scala
```diff
@@ -537,149 +506,132 @@ class RoTenantNodeGroupRepo(
...
+
+  // a restore may put back a narrower tag: that exception is defined in `manageRestore`
+  override def restore(group: NodeGroup)(implicit cc: ChangeContext): IOResult[Option[ModifyNodeGroupDiff]] = {
+    checkTenant.manageRestore(
+      group,
+      roRepo.getNodeGroupOpt(group.id).map(_.map(_._1)),
+      // a group restore does not state a category: reverting the deletion of a group is done by `create`,
+      // which does state one
+      Container.none
+    )(g => underlying.update(g))
```
clarktsiory: isn't that 
```suggestion
    )(g => underlying.restore(g))
```
even if the implementation of underlying is an update ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/TenantFiltering.scala
```diff
@@ -716,67 +667,62 @@ class WoTenantRuleRepo(
...
+    )(r => underlying.update(r))
+  }
+
+  // a restore may put back a narrower tag: that exception is defined in `manageRestore`
+  override def restore(rule: Rule)(using cc: ChangeContext): IOResult[Option[ModifyRuleDiff]] = {
+    checkTenant.manageRestore(
+      rule,
+      roRepo.getOpt(rule.id),
+      roRuleCategory.get(rule.categoryId).notOptional(s"Category with ID '${rule.categoryId.value}' was not found")
+    )(r => underlying.update(r))
```
clarktsiory:
```suggestion
    )(r => underlying.restore(r))
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/TenantFiltering.scala
```diff
@@ -801,44 +747,49 @@ class RoTenantParameterRepo(
...
+
+  // a restore may put back a narrower tag: that exception is defined in `manageRestore`
+  override def restoreParameter(
+      parameter: GlobalParameter
+  )(using cc: ChangeContext): IOResult[Option[ModifyGlobalParameterDiff]] = {
+    checkTenant.manageRestore(
+      parameter,
+      roRepo.getGlobalParameter(parameter.name),
+      Container.none
+    )(p => underlying.updateParameter(p))
```
clarktsiory:
```suggestion
    )(p => underlying.restoreParameter(p))
```

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/TechniqueEditForm.scala
```diff
@@ -476,7 +476,8 @@ class TechniqueEditForm(
                   category.id,
                   technique.id.name,
                   techniqueRepository.getTechniqueVersions(technique.id.name).toSeq,
-                  technique.policyTypes
+                  technique.policyTypes,
+                  None
```
clarktsiory: the technique should be added with the new `USER_LIB_TECHNIQUE_SECURITY_TAG`

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/snippet/administration/TechniqueLibraryManagement.scala
```diff
@@ -507,7 +507,8 @@ class TechniqueLibraryManagement extends SecureDispatchSnippet with Loggable {
                             ActiveTechniqueCategoryId(destCatId),
                             ptName,
                             techniqueRepository.getTechniqueVersions(ptName).toSeq,
-                            policyTypes = PolicyTypes.rudderBase
+                            policyTypes = PolicyTypes.rudderBase,
+                            None
```
clarktsiory: I assume it's `USER_LIB_TECHNIQUE_SECURITY_TAG` too here

webapp/sources/rudder/rudder-web/.vscode/settings.json
clarktsiory, line 1: did you test VSCode ?
is it able to import the project ? 😁

webapp/sources/rudder/rudder-web/src/main/scala/com/normation/rudder/web/components/popup/GiveReasonPopup.scala
```diff
@@ -157,7 +157,8 @@ class GiveReasonPopup(
                         ActiveTechniqueCategoryId(destCatId.value),
                         ptName,
                         techniqueRepository.getTechniqueVersions(ptName).toSeq,
-                        policyTypes = PolicyTypes.rudderBase
+                        policyTypes = PolicyTypes.rudderBase,
+                        None
```
clarktsiory:
```suggestion
                        USER_LIB_TECHNIQUE_SECURITY_TAG
```

clarktsiory (commented):
A minor interrogation about the design of ifabsent, but overall it looks like a great enforcement for bug-catching. Tests looked good to me

Also a few things that need to be grep-ped for replacement (parameter repository, underlying methods, and new constant that are unused)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
fanf: I was going for the simpler enum based on your previous remark that we should strike for it when nothing need more power >_<

webapp/sources/rudder/rudder-web/.vscode/settings.json
fanf: I though I removed it - two times... Not for real IDE work, just as a better leafpad when I want an UI

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
clarktsiory: I guess it's this one: https://github.com/Normation/rudder/pull/7290#discussion_r3593505290

It was specifically because _enumeratum_ does not fit every 'enum', and is overkill when no serialization or exhaustive List of members is needed...

But here the trigger was that the `enum` looks like the enum model of an `Either`, e.g.:
```scala
enum Either[+L, +R] {
  case Left(value: L)
  case Righ(value: R)
}
```

fanf: what opaque type brings here ? It's a zero cost abstraction on either, so ok it's same bytecode and on ram early... But on that case, i don't think we will gain much bytecode or perf. But for the cost of a more complicated inheritance/visibility/implicit story, and a less clear business model. Am I missing something?

clarktsiory: it would better convey the intent to expose only the useful parts of abstraction, within the same scope as the opaque type definition (extension methods & constructors in the companion object):
* a `toIO: IOResult` extension, which is zero-cost and method on Either already exists, so less duplication 
* a constructor of 'noop'/right, also one for 'fail'/left, which could in fact have by-name parameters (the case class don't), it's more consistent with lazy construction of e.g. big error messages

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
fanf: This is never ending. I was missing rule categories too

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/tenants/TenantService.scala
fanf: Why not an opaque type on `PureResult[R]` then ?

clarktsiory: it would even be better indeed, if the underlying error is always a RudderError

clarktsiory (approved):
LGTM 👏 all issues have been addressed (I didn't see the "rule category" case missing the tenant repo pattern too)

clarktsiory (approved):

fanf:
OK, squash merging this PR

PR #7457, opened 2026-08-29
Fixes #29674: org.eclipse.jgit.errors.MissingObjectException maybe link to Jgit multiPackIndex support

fanf:
https://issues.rudder.io/issues/29674

So, this is very strange and I'm not sure I really found the correct culprit after chasing a lot of possibilities. 
It *looks* like setting `core.multiPackIndex` to false made the error stop. But since it was somehow random and non reproducible... Well. We got `org.eclipse.jgit.errors.MissingObjectException` in 9.2 that we didn't before. We switched version of JGit to 7.7, and the big new in 7.7 is the addition of multipack index. A fresh, complex new feature feels like a good culprit. 

So basically, that PR configure our `configuration-repository` with multipack index set to false. 

I made claude find the correct reference in code and check jgit behavior, and it well explains what I was seeing. I used the ref to create an ADR on the choice to disable MPI for us for now. 

That PR also clause others things that could have been the cause (but they are here since ever, so it was unlikely). I let them on 9.2 only, because touching jgit is always a bit scary: 
- close of the open tw that where likely leaking resources at full speed, 
- factor out the logic to find exactly one file and get its content, also closing correctly opened resourced.

### DIFF
--- a/adr/webapp/29674-disable-jgit-multi-pack-index.md
+++ b/adr/webapp/29674-disable-jgit-multi-pack-index.md
@@ -1,0 +1,23 @@
+# Disable the git multi-pack-index in JGit
+
+* Status: accepted
+* Deciders: FAR
+* Date: 2026-08-29
+
+## Context
+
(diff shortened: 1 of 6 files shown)


fanf:
PR updated with a new commit

clarktsiory (approved):
LGTM !

fanf:
OK, squash merging this PR

PR #7458, opened 2026-08-29
Fixes #29678: When we trigger a full policy generation, we have an error about missing hash file (twice)

fanf:
https://issues.rudder.io/issues/29678

Correct issue from https://github.com/Normation/rudder/pull/6983 : 
- an empty/missing file is not an error, it just mean "full regeneration needed"
- use an atomic move so that failure on file writing doesn't lead to full regeneration
- accept that one line can be a decoder error but still get the other ones (so that, we don't regenerate for all nodes, just the ones missing)
- come back to one-node-per-line in the Json hash file as it was before 9.2, because the current format is verbose. Not definitly sure, having one lines per rule is nice too. But it seems better that way, since if you want to explore something, you can grep+jq more easily here

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
@@ -55,10 +55,13 @@ import com.normation.zio.*
 import io.scalaland.chimney.*
 import io.scalaland.chimney.syntax.*
 import java.nio.charset.StandardCharsets
+import java.nio.file.Files
+import java.nio.file.StandardCopyOption
 import java.time.Instant
 import org.joda.time.DateTime
 import zio.*
 import zio.json.*
+import zio.json.ast.Json
 import zio.syntax.*
 
 final case class PolicyHash(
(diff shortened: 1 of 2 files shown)


fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
```diff
@@ -449,28 +482,35 @@ class FileBasedNodeConfigurationHashRepository(path: String) extends NodeConfigu
...
       ) *>
       NodeConfigurationHashes(Nil).succeed
     })
   }
 
+  /*
+   * Atomic write of the JSON file by writing a `.tmp` file then moving it atomically to destination.
+   * The temporary file is in the same directory to guarantee that `rename` stays within one filesystem.
+   */
   private def nonAtomicWrite(hashes: NodeConfigurationHashes): IOResult[Unit] = {
```
clarktsiory: the method name should then be `atomicWrite` now

clarktsiory (commented):
LGTM for the change and I agree that the old format is more practical.
The atomic write method should just be renamed from "nonAtomic" now

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/policies/nodeconfig/NodeConfigurationCacheRepository.scala
fanf: well, it's because the write is not atomic, but I understand why it's not clear at all. 
`def atomicUpdate` ?

fanf:
PR updated with a new commit

clarktsiory (approved):

fanf:
OK, squash merging this PR

PR #7460, opened 2026-09-01
Fixes #29648: Wrong change logs result when comming from recent activity id link

P4uline:
https://issues.rudder.io/issues/29648

Previously the query to get the event log from the Recent Activity was using the full text search feature with a small date range, which is flaky and sometimes caused some bugs, we could have more that one logs or event no logs at all.

Modifications to fix the issue:
- add an id field in the event request object
- modify the sql query to add a eventid=id if the event request id is filled
- change the link on Recent Activity id 
  - remove old full text criteria and date
  - pass the id 
- parse the id in the javascript code to call the endpoint with the id as criteria


fanf (approved):
LGTM

PR #7466, opened 2026-09-03
Fixes #29682: It is possible to override the rudder node property with a String property, which cannot be modified or deleted

fanf:
https://issues.rudder.io/issues/29682

So, two main problems: 
- 1/ we should never be blocked to edit/delete a self-value we set ourselves in a node. That blocked the possibility to reverse the error once the Bad Thing Happened, 
- 2/ `provider` is not inherited like other things, it doesn't make sens on resolved value, only on the real node/group/global properties. 

1/ was only an UI level (that's why we could workaround via REST API), an so it's corrected their by checking the "own value". 

2/ could have been corrected by "ignoring" the inherithed provider, but I prefered to materialize the fact that a resolve property is something else in a dedicated type without that attribute. It seems to be self-contained to inherited property (and not even in plugins) so it looks ok.


VinceMacBuche (approved):

fanf:
OK, merging this PR

PR #6284, opened 2025-03-25
Fixes #26416: JVM platform limit requires splitting services in RudderConfig

fanf:
https://issues.rudder.io/issues/26416

PR #7468, opened 2026-09-04
Fixes #29699: Adding new instance reaches platform limit in RudderConfig

clarktsiory:
https://issues.rudder.io/issues/29699

Putting instances of the `lazy val` in a new `object` is the workaround that prevents the platform limit issue with the most viability (instead of inlining instances, which should only be a temporary solution)

Implementation is similar to the one of https://github.com/Normation/rudder/pull/6284 for APIs, but within an object which defines a "group", instead of a scoped block. 

(showing [the diff without whitespaces](https://github.com/Normation/rudder/pull/7468/changes?w=1) help seeing that nothing else is changed in this PR)


clarktsiory:
Commit modified

fanf (approved):
LGTM

PR #7469, opened 2026-09-04
Fixes #29696: Reading nodes at startup must be done in a bulk request

fanf:
https://issues.rudder.io/issues/29696

The whole idea is to switch from a node-by-node retrieval to a "get all nodes" one. 
It avoid a *lot* of queries (one to get all ids, then 3 per nodes) to only keep 3 (one on nodes branch, and one for pending (resp. accepted) inventories. 

That PR also add a streaming query for LDAP, which should allow to consume less resources during the process. 

We can switch back to previous behavior with a new parameter in rudder-web.properties: `rudder.facts.bulkLoadAtBoot`, true by default. 

Measures on my little instance give ~25% speed up of boot sequence (so certainly a big enhancement for the LDAP loading).

### DIFF
--- a/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
+++ b/webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
@@ -53,6 +53,7 @@ import com.normation.ldap.sdk.LDAPConnectionProvider
 import com.normation.ldap.sdk.LDAPEntry
 import com.normation.ldap.sdk.One
 import com.normation.ldap.sdk.RwLDAPConnection
+import com.normation.ldap.sdk.StreamedSearchResult
 import com.normation.rudder.domain.NodeDit
 import com.normation.rudder.domain.logger.NodeLogger
 import com.normation.rudder.domain.logger.NodeLoggerPure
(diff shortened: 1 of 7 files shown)


fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/facts/nodes/NodeFactStorage.scala
```diff
@@ -969,4 +1002,137 @@ class LdapNodeFactStorage(
...
+   * Bulk load with streaming ldap search.
+   * It can be disabled with `bulkLoadAtBootProperty` set to false.
+   * We only do 3 search (nodes and pending/accepted inventories) and create map of NodeId -> Entry, then
+   * merge them as usual.
+   */
+  override def loadAllCoreNodeFacts(): IOResult[AllCoreNodeFacts] = {
+    if (bulkLoadAtBoot) {
+      // take the read lock since inventories can come, and we want to keep consistency and only proceed them after
+      // initial cache init step
+      NodeLoggerPure.debug(s"Bulk load of node facts ('${LdapNodeFactStorage.bulkLoadAtBootProperty}=true' )")
```
VinceMacBuche: it misses the *> so the log will never be made !!

fanf: arf

fanf:
PR updated with a new commit

VinceMacBuche (approved):

fanf:
OK, squash merging this PR

PR #7470, opened 2026-09-07
Fixes #26191: Impossible to delete a category in the Techniques Tree

fanf:
https://issues.rudder.io/issues/26191

Add technique categeroy in the technique editor UI with corresponding APIs. 
The move part will be done in a second PR, this one was becoming quite large already, even it it's mainly CRUD code with very little business logic. 

# Backend

- there's a new `TechniqueCategoryWriter` that does thing similar to `TechniqueWriter` 
- we take care of path traversal when writting files, 
- the ID of the category (ie what will be used in the FS) is derived from category name, not an UUID. When the name change, we don't change ID to avoid sync problem between env, and so they can be desynchronized. 
    - using a name was prefered for our lib, 
    - user techniques do the same (so consistency of behavior for the "user managed" part). 
    - :yellow_circle:  still, it's extremelly unsetting to have diverging name and ID, so perhaps we need to do something about that at some point. 
- I was quite conservative for the category ID authorized chars (only `[a-z_-]`). Other are transformed to `_`. ID uniqueness is verified in the whole technique lib. 

# API

- :yellow_circle: I didn't use a `PUT` verb for creation. We have long standing will to move away of PUT/POST since we are not using it "correctly", and I so I wanted to see how to do it. The chosen design has an "action" field describing what the API is supposed to do. 
    - I'm not sure I like it. But it's rather straightforward.  
 
```
// create
{
  "action":"create",
  "parent":"ncf_techniques/New_category/U_name_it",
  "name":"Test",
  "description":""
}
```

```
// update
{
  "action":"update",
  "path":"ncf_techniques/New_category/U_name_it/test",
  "name":"test",
  "description":"new description"
}
```

(the fact that we have the path of parent or category is because our endpoint syntaxt is not well tailored for that case, we don't know how to represent "variable resource path", one segment of it being variable)

And in the backend, we have: 

```
  @jsonDiscriminator("action")
  sealed trait JsonTechniqueCategoryAction

  object JsonTechniqueCategoryAction {
    @jsonHint("create")
    final case class Create(parent: String, name: String, description: String = "") extends JsonTechniqueCategoryAction
    @jsonHint("update")
    // absent means "leave as it is", so an empty description removes it
    final case class Update(path: String, name: Option[String] = None, description: Option[String] = None)
        extends JsonTechniqueCategoryAction

    implicit val decoder: JsonDecoder[JsonTechniqueCategoryAction] = DeriveJsonDecoder.gen
  }
```

- the `getAllTechniqueCategories` now mean: get all *the non system categories*. We want to reduce exposition of system objects, which are something different, so lets start here. 
- nothing very interesting appart that

# UX 

The technique tree UX is normalized: 
    - clicking on technique or categories open the right panel, 
    - for technique, it's a new panel with action bouton/form to update & delete it, and a button to create sub categories
    - a colour code inform people of categories that can be managed 

# UI (elm)

- the action buttons part is made common with techniques, 
- new API decoder/endpoints for the new category JSON

# Views 

General view: 
image

Rudder lib categories: 
image

User Techniques
image

User created category:
image

image


P4uline (commented):
I made some tests and I found a bug :
When I created a category in the maintenance/technique tree view, then I don't see it in technique tree of the Technique tree panel. I can see it in the directive tree though.

When I created a sub category from the directive tree, the category is visible everywhere: 
- in the technique tree of the Technique view
- in the technique tree of the maintenance/technique tree view
- in the directive tree

fanf:
Is this the only feedback on that big PR ?

fanf:
@P4uline so that's not a bug, it's a different way of managing categories, and why we want to remove the "Technique Tree" page: 
- technique tree is not creating the corresponding FS information. It's a pure LDAP-side change, which makes it impossible to understand for people, and the reason why we want to move away from it. A category from it only lives in LDAP. The directive page maps LDAP view, so that's what it shows, too 
- technique editor (and people, and archives, and everyone else) expects a category to have a physical existence on FS, with a directory  name and a `category.xml` file. 

So that explains why category created from technique editor from that PR are seen everywhere, but techniques from the "Technique Tree" page are only seen in directives tree (which is useless). 

This is linked to that other bug https://issues.rudder.io/issues/26125 - which is the opposite (a technique only remove from LDAP from Technique Tree page). 

I don't think it's the place to that PR to correct that behavior, because it will need special for existing set-up contrary to what is provided here.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/EndpointsDefinition.scala
```diff
@@ -842,13 +842,31 @@ object TechniqueApi extends Enum[TechniqueApi] with ApiModuleProvider[TechniqueA
...
     val z: Int = implicitly[Line].value
-    val description    = "Get all technique categories"
+    val description    = "Get all non-system technique categories"
     val (action, path) = GET / "techniques" / "categories"
     val authz: List[AuthorizationType] = AuthorizationType.Technique.Read :: Nil
 
     override def name:          String         = "techniqueCategories"
     override def dataContainer: Option[String] = None
   }
+  case object SaveTechniqueCategory     extends TechniqueApiPub with ZeroParam with StartsAtVersion23 with SortIndex {
```
P4uline, line 852: I would be curious to know the motivations of the url design. 
This is the the design I would propose naively for every endpoints: `resources/<resourceId>/subresources/<subresourceId>` expressing a hierachical relation between resource and subresource.

- When I see `techniques/categories` I would suppose techniques own some categories, for instance technique `a` has a list of categories `c1`, `c2`and `c3` and technique `b` has `c4` and `c5`. But the relationship works the other way as the category is a directory where we put techniques. So I would put the category 'before' in the url to express the relationship :
  - GET /categories?path=<path>
  - PUT /categories?path=<path>
  - POST /categories

- I would like to put the id in the path : `resources/<resourceId>`, but for categories the id is a path
  - design with id as a path param `categories/a/b/c` 
    Will I get 404 looking for resource `categories` of id `a` and subresource `b` of id `c` ?
    Or will I get a category of is `a/b/c` ? I'm afraid this design is not resilient enough.
  - design with id as a query param `categories?path=<id>` 
    here there are no ambiguity

So that being said, this is the design I would propose:
```
POST /categories
{
  "path": "ncf_techniques/linux_hardening",
  "name": "Linux hardening",
  "description": "..."
}
```
```
PUT /categories?path=ncf_techniques/linux_hardening
{
  "name": "Linux hardening",
  "description": "..."
}
```
```
GET /categories?path=ncf_techniques/linux_hardening
# response
{
  "path": "ncf_techniques/linux_hardening",
  "name": "Linux hardening",
  "description": "..."
}
```

There is also some work already done in this direction to refactor the compliance of API, and a ADR in progress.
What do you think about this ?

P4uline: Sum-up of live discussion: the urls design is good as is, I will introduce the `namespace` concept in the ADR in progress.

P4uline (approved):
LGTM

fanf:
OK, squash merging this PR

PR #7473, opened 2026-09-09
Fixes #29726: OTP page can be a redirect page of itself

clarktsiory:
https://issues.rudder.io/issues/29726

Just filter out the `/secure/otp.html?redirect=.../secure/otp.html` loop by having empty query params (that will lead to dashboard)

### DIFF
--- a/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
+++ b/webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
@@ -429,9 +429,9 @@ class AppConfigAuth extends ApplicationContextAware {
  */
 class RudderUrlAuthenticationSuccessHandler(enforceOtp: Boolean)
     extends SimpleUrlAuthenticationSuccessHandler("/secure/index.html") {
-  private val otpRedirect = enforceOtp
+  import RudderUrlAuthenticationSuccessHandler.*
 
-  private val otpRedirectUri = "/secure/otp.html"
+  private val otpRedirect = enforceOtp
 
   private val requestCache = new HttpSessionRequestCache()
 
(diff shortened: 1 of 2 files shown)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
```diff
@@ -495,10 +495,13 @@ class RudderUrlAuthenticationSuccessHandler(enforceOtp: Boolean)
...
-        .map(url => {
-          requestCache.removeRequest(request, response) // clean is recommended since we intercept it only in this request
-          s"?redirect=${URLEncoder.encode(url, StandardCharsets.UTF_8)}"
+        .flatMap(url => {
+          // clean is recommended since we intercept it only in this request
+          requestCache.removeRequest(request, response)
+          // prevent redirect to OTP page itself, see https://issues.rudder.io/issues/29726
+          Option.unless(url.takeWhile(_ != '?').endsWith(otpRedirectUri))(url)
         })
+        .map(url => s"?redirect=${URLEncoder.encode(url, StandardCharsets.UTF_8)}")
```
P4uline: If I understand correctly what the code is doing :

If getRedirectUrl is `http:\\redirectionurl/somepath?anything` then nextOtpRedirect returns `?redirect=http:\\redirectionurl/somepath`

Is it possible to add a unit test ? Maybe extracting some code in a method to bypass the requestCache thing...

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/AppConfigAuth.scala
clarktsiory: yes that is right, and the issue was that the redirection page could exactly be the OTP page.
I added tests in https://github.com/Normation/rudder/pull/7473/commits/eb7cbe49cba12502cddc8f1cc569ee93b97523ed

P4uline (approved):

clarktsiory:
OK, squash merging this PR

PR #7477, opened 2026-09-10
Fixes #29730: The text filter for Group compliance by node does not work correctly

skaerg:
https://issues.rudder.io/issues/29730


P4uline (approved):

skaerg:
Commit modified

PR #7480, opened 2026-09-11
Fixes #29226: Upload limit problem in 27413 is back 

clarktsiory:
https://issues.rudder.io/issues/29226

**EDIT: Best way to avoid the initial swallowing of the error is with _frontend file size validation_...**

Handling files with less than 8MB (hardcoded for now) is done on Elm side: we simply need to ignore files that are too big.

But the Elm app had some bugs in handling upload state:
* the previous "multi-attributes in model" was leading to errors/forgotten cases which ended up in broken state after an errored upload (see comments)
* introducing a proper "upload" model solves every case with pattern-matching and proper logic for all cases:
  * several model functions were split into "model invariants", e.g.  `nextUpload` and `currentUpload` state
  * those are tested in a new `ModelTest.elm` file
         

---

This was hard to find and I'm not even sure about the change in Jetty 12 that is the root cause.
The fix that makes the upload error consistently appear is to await the body of the file upload to be consumed even when its too big (apache commons has an util for "consume and forget"). That way, the "broken pipe" is also avoided on apache side, and some options no longer needed (see https://github.com/Normation/rudder-packages/pull/3191).


clarktsiory:
Commit modified

fanf (commented):
Does that mean that we hold the connection even once we have gone beyond the limit ? 
If so, it could be a problem, keeping connection used for uploading 42 TB while it should be stopped much earlier.

fanf:
After some thinkings, that doesn't seem to be the correct approach. We are going to use a client (elm) side validation. 
Plus we would like to make the MAX_FILE_SIZE configurable (we also need to configure both liftRules.maxMimeSize and liftRules.maxMimeFileSize, with some room for the first in addition to the second).

clarktsiory:
Commit modified

webapp/sources/rudder/rudder-web/src/main/elm/sources/FileManager/Update.elm
```diff
@@ -83,41 +85,65 @@ update msg model =
...
             ( model, reload )
 
         Uploaded result ->
             case result of
                 Ok ( _, _ ) ->
-                    case model.uploadQueue of
-                        file :: files ->
-                            ( { model
-                                | filesAmount = model.filesAmount - 1
-                                , uploadQueue = files
```
clarktsiory, line 108: We were forced to compute `filesAmount` along with `uploadQueue` every time, which was error prone: they both represent an "upload" model

webapp/sources/rudder/rudder-web/src/main/elm/sources/FileManager/Model.elm
```diff
@@ -53,16 +132,15 @@ type alias Model =
     , showContextMenu : Bool
     , selectedBin : List FileMeta
     , showDrop : Bool
-    , filesAmount : Int
-    , progress : Http.Progress
```
clarktsiory, line 57: `progress` makes no sense too when there is no "pending upload"

webapp/sources/rudder/rudder-web/src/main/elm/sources/FileManager/View.elm
```diff
@@ -274,6 +280,16 @@ back list =
...
 
 
+renderUploads : UploadStatus File -> List (Html Msg)
+renderUploads status =
+    case status of
+        NoUpload ->
+            []
+
+        PendingUpload { progress, uploadQueue } ->
+            reverse (map (renderUploading progress) (range 0 (NonEmptyList.length uploadQueue - 1)))
```
mbaechler: you can probably simplify this part, too, by providing the right method in model?

clarktsiory: yes indeed, I forgot to change this part, the "composition" syntax is also hard to read, pipes would be much readable

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/FileManager/View.elm
clarktsiory: done in https://github.com/Normation/rudder/pull/7480/commits/2e81aacdf309eba76ee4427359de48ddba3a3d58, the abstraction has been changed to `currentUploadsInProgress`, the view only reverse the order

mbaechler: The only call to this method is for rendering, I would make `currentUploadsInProgress` return the list in the order required by the view.
The model is supposed to be designed for the view.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/elm/sources/FileManager/View.elm
clarktsiory: I thought about that principle too, but I thought that it was more natural to have the current progress at position 0 because of list construction order (it's a stack, feasible in both orders but List naturally add element in first position).

But I quite agree that the stack can also be seen as a reversed list in that case, I changed it: https://github.com/Normation/rudder/pull/7480/commits/e0e6ea13846beeed7841177804600d24b60b82b5

PR #7486, opened 2026-09-16
Fixes #29384: A newline in a server-supplied value writes arbitrary directives

m4rtinh4rt:
https://issues.rudder.io/issues/29384


m4rtinh4rt:
PR updated with a new commit

policies/module-types/secedit/src/secedit.rs
```diff
@@ -211,7 +211,14 @@ fn config_search_and_replace(
         let mut section_diff: Vec<String> = vec![];
         for (key, new_value) in entries {
             let new_value = validate_property(key, section, new_value)?;
-            let new_value = new_value.map(|s| s.replace("\"", ""));
+            let new_value = new_value.map(|s| {
```
amousset: i don't think we want to silently modify the provided user content. We should fail loudly instead in general.
i think the only acceptable case is to trim the string before checking for forbidden chars, if it does not break anything.

m4rtinh4rt: ok

m4rtinh4rt:
PR updated with a new commit

PR #7487, opened 2026-09-17
Fixes #29761: Use scala 3.9.0 in Rudder 9.2

fanf:
https://issues.rudder.io/issues/29761

It seems to Just Work (modulo a real error and 3 now useless type ascription)


webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/RudderConfig.scala
```diff
@@ -3198,8 +3198,6 @@ object RudderConfigInit {
       TenantScopedWrite(new WoTenantDirectiveRepo(tenantCheckLogic, storage, directiveRead.storage), storage)
     }
 
-    lazy val directiveWrite.repository: WoDirectiveRepository = directiveWrite.repository
```
fanf, line 3201: "oups"

VinceMacBuche (approved):

PR #7489, opened 2026-09-17
Fixes #29767: PluginLogger is duplicated

P4uline:
https://issues.rudder.io/issues/29767


P4uline:
PR updated with a new commit

P4uline:
PR rebased

fanf (changes requested):
There isn't any description, nor information wrt to potentiel impact on plugins. Please add that, without it I can't know if it's really mergeable, I don't know if you did the necessary homeworks for that change

PR #7494, opened 2026-09-18
Fixes #29776: Make it possible to rollback item after a commit 

clarktsiory:
https://issues.rudder.io/issues/29776

Chang is not that different from the existing logic:
* the main change is a pattern match between "after" and "before", since restoring add/modify/delete is a different logic depending on that
* the API needs a better splitting: `action=before/after` needs the `type=item/all-configuration` to distinguish both (by default it's "item" now)
* frontend uses the same radio select buttons, and pass the "item" type now


clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemRollbackRepositoryImpl.scala
```diff
@@ -138,40 +142,42 @@ class ItemRollbackRepositoryImpl(
...
       case e: DirectiveEventLog =>
         for {
           sid <- rolledBackItemId(e, XML_TAG_DIRECTIVE)
           id  <- DirectiveId.parse(sid).toIO
-          _   <- e match {
-                   case _: AddDirective => woDirectiveRepository.delete(id.uid).unit
-                   case _: DeleteDirective | _: ModifyDirective => restoreDirective(archiveId, id.uid)
+          _   <- (rollbackType, e) match {
+                   case ("before", _: AddDirective) | ("after", _: DeleteDirective) => woDirectiveRepository.delete(id.uid).unit
+                   case _                                                           => restoreDirective(archiveId, id.uid)
```
clarktsiory, line 161: the logic mainly changes to this one for all items

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
```diff
@@ -0,0 +1,26 @@
...
+ *
+ *************************************************************************************
+ */
+
+package com.normation.eventlog
+
+/**
+ * The state an event log is rolled back to
+ */
+type RollbackType = "before" | "after"
```
fanf: make it an `enum`, there is no reason to use higher-stringly-type for modeling domain

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/services/modification/ItemRollbackService.scala
```diff
@@ -82,11 +84,17 @@ class ItemRollbackServiceImpl(
...
+  ): IOResult[GitCommitId] = {
     for {
       commit   <- commitOf(eventLog)
+      archiveId = rollbackType match {
+                    case "before" => parentOf(commit)
+                    case "after"  => commit
+                  }
       // an item restore is about that one event log, so it is both the only event rolled back and the target
-      rollback <- itemRollbackRepository.rollbackItem(parentOf(commit), commiter, Seq(eventLog), eventLog)
+      rollback <- itemRollbackRepository.rollbackItem(archiveId, commiter, Seq(eventLog -> rollbackType), eventLog, rollbackType)
```
fanf: we (almost) never use `->` to create a tuple, I would prefer to keep it so

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/xml/ItemRollbackRepositoryImplTest.scala
```diff
@@ -436,7 +536,7 @@ private object ItemRollbackRepositoryImplTest {
   }
 
   private def emptyNodeGroupCategory =
-    NodeGroupCategory(Constants.ROOT_GROUP_CATEGORY, "GroupRoot", "", Nil, Nil, isSystem = false, security = None)
+    NodeGroupCategory(NodeGroupCategoryId("GroupRoot"), "GroupRoot", "", Nil, Nil, isSystem = false, security = None)
```
fanf: no, keep the constant, it's here for avoiding that.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemRollbackRepositoryImpl.scala
```diff
@@ -121,14 +121,18 @@ class ItemRollbackRepositoryImpl(
   override def rollbackItem(
       archiveId:        GitCommitId,
       commiter:         PersonIdent,
-      rollbackedEvents: Seq[EventLog],
-      target:           EventLog
+      rollbackedEvents: Seq[(EventLog, RollbackType)],
```
fanf: I think `(EventLog, RollbackType)` should get it's own type, it looks central to that business. `UnitRollbackQuery` I think.

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemRollbackRepositoryImpl.scala
```diff
@@ -121,14 +121,18 @@ class ItemRollbackRepositoryImpl(
...
+      target:           EventLog,
+      rollbackType:     RollbackType
```
fanf: I don't get what are the role of these two. I know it's not a change from that PR, but still. 
If it is: there is a "main" event and potentially depending event, the parameters should be: 
```
def rollbackItem(
      archiveId:        GitCommitId,
      commiter:         PersonIdent,
      rollbackedEvent: UnitRollbackQuery,
      dependingEvents: Seq[UnitRollbackQuery]
)
```

And the implementation should deal with the main event. 

Plus, you are discarding the `rollbackType` line 129 below, so does that really make sense to have pairs? 

It is not very clear either if the parameter GitCommitId is the commit to revert to (the target commit id), or the commit to revert.

It look like what you have, is a query for a rollback with git id, commiter, a rollback type (unique for the whole query), and then you have (I guess?) a main event, and a seq of potentiel dependent event to rollback ?  

The semantic of that method must be very clear, because it's the whole interface for `ItemRollbackRepository`. 
And I don't think you can have both a commit ID and events at that point.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala
```diff
@@ -192,21 +194,35 @@ object RestEventLogDetails {
...
 
     implicit val encoder: JsonEncoder[Action] = JsonEncoder[String].contramap(_.entryName)
+
+    implicit val toRollbackType: Transformer[Action, RollbackType] = {
+      case After  => "after"
+      case Before => "before"
+    }
+  }
+
+  sealed trait Type extends Hyphencase
```
fanf: please avoid `Type` as an object name, it's far too generic and in the context of scala, it is easy to misunterpret it.

fanf: Why hyphen case ? It's because of elm convention ? Can you write it down because it's a difference wrt to domain model serialization.

webapp/sources/rudder/rudder-web/src/main/javascript/rudder/rudder-datatable.js
```diff
@@ -1363,10 +1366,10 @@ function getRadioChecked(radios, validate = s => s) {
  return null;
 }
 
-function confirmRollback(id, action) {
+function confirmRollback(id, action, type) {
   $.ajax({
     type: "POST",
-    url: contextPath + '/secure/api/eventlog/' + id + "/details/rollback?action=" + action,
+    url: contextPath + "/secure/api/eventlog/" + id + "/details/rollback?action=" + action + "&type=" + type,
```
fanf: why not passing the type in the body of the POST ? Here, the rollback query is mixed between post and query parameter, that breaks locality.

fanf (changes requested):
You will need to beter surveil what the bot is doing. And the design must be yours, bot are absolutly abysmall on that part.

webapp/sources/rudder/rudder-core/src/test/scala/com/normation/rudder/repository/xml/ItemRollbackRepositoryImplTest.scala
clarktsiory: this was a rebase issue, it seems to not have been resolved the right way

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
clarktsiory: okay I intended to use literal types to see if it provided the same guarantees and it was used as String, but it can be an enumeratum's entry name with some explanation of why serialization is in lowercase

clarktsiory:
Fixed rebase leftovers

clarktsiory:
PR updated with a new commit: make rollbackType an enum

clarktsiory:
PR updated with a new commit: use enum entryName for serialization

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/repository/xml/ItemRollbackRepositoryImpl.scala
clarktsiory: Indeed the method inherently describes two scenarios:
* it has still `rollbackedEvents: Seq[...]` from an historical signature which could describe "many potential events to rollback"
* the additional `target` (and now rollback type) describe the single "main event to rollback"

For the main event, I also had the idea of creating a case class with a dedicated self-contained model for the commit rollback: `BeforeAt(archiveCommitInfo)|AfterAt(archiveCommitInfo)`, which seems to be the most faithful representation for the rollback of the main event, which would make the interface unambiguous.
But I found out we would lose ability to log the archiveId from current method, so I did not went down to the path of refactoring the data that is passed to the methods.
As for the sequence, it's always a sequence with a single element, it deserves to be removed, as you say

If it's the right scope of this PR, I can pursue with that refactoring.

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala
clarktsiory: it's because URLs are not case sensitive and I tend to prefer kebab-case for parameters, but it shouldn't in fact be passed as params, we avoid them so I will change it to pass a JSON body along with the POST, instead of query params

clarktsiory:
PR updated with a new commit

clarktsiory (commented):
I added some business objects, simplified the interface, and added tests to the service. All comments should have been adressed now

clarktsiory:
PR rebased (not completely yet)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
```diff
@@ -0,0 +1,76 @@
...
+import enumeratum.*
+import enumeratum.EnumEntry.Lowercase
+import io.scalaland.chimney.Transformer
+import io.scalaland.chimney.syntax.*
+
+/**
+ * The state an event log is rolled back to.
+ * Serialization is in lowercase in the eventlog details
+ */
+sealed trait RollbackType extends EnumEntry with Lowercase
```
fanf: if there is a serialization value, make it explicit by overriding entry name. We want to be able: 
- to search for what is in the serialized data and find it in the code source, 
- not link the serialized value to a scala name that can be refactored without notice. 

Write the corresponding ADR if it makes thing clearer and more formal for you :)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
```diff
@@ -0,0 +1,76 @@
...
+  def apply(eventLogId: Long): RollbackEventId = eventLogId
+  extension (self:      RollbackEventId) {
+    def eventLogId: Long = self
+  }
+}
+
+/**
+ * The rollback of item needs a commit, which is the target we want to restore them to.
+ */
+enum RollbackTarget(val archiveCommit: GitCommitId) {
```
fanf, line 62: hey ! Enumeratum, opaque type and enum in 15 lines, I'm sure you won some prize

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
```diff
@@ -0,0 +1,76 @@
...
+object RollbackTarget {
+  given Transformer[RollbackTarget, RollbackType] = Transformer.derive[RollbackTarget, RollbackType]
+
+  extension (self: RollbackTarget) {
+    def rollbackType: RollbackType = self.transformInto[RollbackType]
+  }
+
+  def from(rollbackType: RollbackType, current: GitCommitId): RollbackTarget = rollbackType match {
+    // the state just *before* a change is the parent of the commit that change led to
+    case RollbackType.Before => Before(GitCommitId(current.value + "^"))
```
fanf: is this what we do in the historical logic? I believe it can lead to problem for merge commit, but we shouldn't ever be in that case, so it should be ok.

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/eventlog/ArchiveEventLog.scala
```diff
@@ -396,7 +396,7 @@ object Rollback extends EventLogFilter {
           }
           val main   = {
             <main>
-                <rollbackType>{rollbackType}</rollbackType>
+                <rollbackType>{rollbackType.entryName}</rollbackType>
```
fanf: I think we want a `.serialize` which is entry name in that case, so that we can complexify the `.serialize` even if entry name evolves. It happened with rule/directive/etc ID for ex. And we use that convention elsewere, see line 401 below. You can add that to the ADR.

fanf (changes requested):
Appart for the serialization of the enumeratum, that looks ok and much more clean

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
clarktsiory: Ah sorry it's my fault because I messed up my rebase :(

:arrow_right: I already fixed it in the new version, and I'm planning the ADR, the current version is indeed:

```scala
/**
 * The state an event log is rolled back to.
 * Serialization is in lowercase in the eventlog details
 */
sealed trait RollbackType(val value: String) extends EnumEntry          {
  override def entryName: String = value
  def serialize:          String = value
}
object RollbackType                          extends Enum[RollbackType] {
  case object Before extends RollbackType("before")
  case object After  extends RollbackType("after")

  override def values: IndexedSeq[RollbackType] = findValues
}
```

What you spot is worth an ADR for NOT doing pure enumeratum but explicit `serialize`, here it's lowercase (so I removed the `Lowercase` mixin).
So now we are using:
* parameterized constructor with the explicit serialization value as a string
* `withNameInsensitive` from enumeratum

clarktsiory: In fact I'm not completely satisfied, because it means we are introducing enumeratum, by intendedly ignoring its `enumEntry`:
* we only use their `findValues` macro to search for all enum values 
* and the `withNameInsensitive` helpers for parsing

A solution to avoid leaking the public `entryName` would be to make it a `protected`, but we can't change the lib for that, it already relies heavily (and by experience too I tend to use the already provided field).

I think using enumeratum would only be confusing because of that.
Having an in-house "enum helper" that provides only the "find values" and "parse helpers" we would best suit our usage.
WDYT ?

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
clarktsiory: I'm thinking more and more that we want to not use enumeratum but scala 3 derivations, see https://github.com/Normation/rudder/pull/7494#discussion_r4069428334

Then it'll be worth a prize for "using the greatest scala 3 features" :wink:

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
clarktsiory: yes it's historically like that, it's covered in a test too for now, so we should be able to evolve that as needed

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/eventlog/ArchiveEventLog.scala
clarktsiory: see https://github.com/Normation/rudder/pull/7494#discussion_r4069372968, it's exactly what I find misleading about using enumeratum, we always have the inherited `entryName` even if we defined `value` and `serialize`

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
fanf: no, we don't reinvent what framework with an history of finding and dealing with the corner case we don't see already do. This is a bad practive refusing to see the value of accumulated knowledge of field tested code. 

And I don't get the problem with entryName ? Why not just overriding it as documented? 

```
sealed trait RollbackType(override val entryName: String) extends EnumEntry          {
  // this by convention and uncoupling from implementation detail through enumeratum
  def serialize:          String = value
}
object RollbackType                          extends Enum[RollbackType] {
  case object Before extends RollbackType("before")
  case object After  extends RollbackType("after")

  override def values: IndexedSeq[RollbackType] = findValues
}
```

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/eventlog/ArchiveEventLog.scala
fanf: we don't have to define value, entryName is totally fine.

fanf: value as no intent/action, while serialize brings the idea that it is the conventionnal way to get the string representation of the object.

fanf: (like `parse` for the opposite action)

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
clarktsiory: OK I'm keeping enumeratum and the override of the entryName for now, it should be very explicit in the ADR then

But IIRC before migrating to zio-json, we had many "enum" types as plain `sealed trait + case object` without enumeratum and with an associated `parse` method (and I also think we shouldn't go back to that)

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/rudder/domain/eventlog/ArchiveEventLog.scala
clarktsiory: :white_check_mark: done in https://github.com/Normation/rudder/pull/7494/commits/6dae8553014d20d68f831122a2b5238c71e9aab7

webapp/sources/rudder/rudder-core/src/main/scala/com/normation/eventlog/Rollback.scala
fanf: yes, they predate enumeratum. We introduced it for migration to scala 3 and the fact that `sealerate`, which we used to use for getting all case objects, wasn't ported to it.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala
```diff
@@ -191,23 +192,47 @@ object RestEventLogDetails {
...
-    action: RestEventLogRollback.Action,
-    id:     String
-)
+sealed trait RestRollbackAction(val value: String) extends EnumEntry                {
+  override def entryName: String = value
+  def serialize:          String = value
+}
+object RestRollbackAction                          extends Enum[RestRollbackAction] {
+  case object Before extends RestRollbackAction("before")
+  case object After  extends RestRollbackAction("after")
```
fanf: if it's exactly the same encoding, and there isn't any uncoupling necessary, you can completely remove that DTO and directly use `RollbackType`. 
Before there is a divergence, DTO are mainly boiler plate. We just need to be atuned to look for divergence, even small ones like a name difference in one the enum cases. 
I don't see any here. 

Which also let me see that `RestRollbackAction` is before/after, like `RollbackType`, but `RestRollbackType` is `Item`/`AllConfigurations`, so we have an inconsistency in naming. 

Perhaps: `RestRollbackType` remains as it is; `RollbackType` becomes ``RollbackPosition` ; `RollbackTarget` remains what it is; `RestRollbackAction` is deleted.

fanf (changes requested):
Almost there! The PR is starting to look really nice, the code is simple with few boilerplate. Some name inconsistency remained.

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala
clarktsiory: I did not see that the action was already named `type` in the domain, and that it conflicted... :see_no_evil:

Yes there is no divergence, it's also probably the simplest model of enum, that should in fact serve as example of the ADR https://github.com/Normation/rudder/pull/7498

clarktsiory: I'm renaming `action(before|after)` to `position(before|after)` for the DTO, and the main object for commit handling remains a `target(before(atcommit)|after(atcommit))`

clarktsiory:
PR updated with a new commit

clarktsiory:
PR updated with a new commit

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/data/EventLog.scala
clarktsiory: :white_check_mark: renamed to `position` everywhere for consistency, there should not be a confusing `rollbackType` anymore, except in the API for `RestRollbackType(item|allConfiguration)`

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
```diff
@@ -167,21 +167,12 @@ class EventLogAPI(
...
-                     case Some(actions)       =>
-                       Inconsistency(
-                         s"Only one action is excepted, ${actions.size} found in request : ${actions.mkString(",")}"
-                       ).fail
-                     case None                => Inconsistency("Empty action").fail
-                   }
-        details <- service.rollback(evId, action)
+        evId        <- id.toLongOption.notOptional("event log ID is not a long integer : " + id)
+        rollbackReq <- req.fromJson[RestRollbackRequest].toIO
+        rollbackType = rollbackReq.rollbackType.getOrElse(Item)
```
fanf, line 172: I think we missed an important bit here. If we want to be able to have regular user rollback the item they can see, that needs to have a different endpoint than the old rollback, with accurate rights - for now, perhaps any of the config ones, since IIRC event logs are correctly scoped.

fanf (changes requested):
We need to split the two rollback kind in two different endpoint to be able to give them different rights. Else the by-item rollback remains admin only, which defeat a fair bit of its interest. 

I'm not sure if we want another PR for that, or just not commit the changes here related to rollback type.

clarktsiory:
PR rebased

webapp/sources/rudder/rudder-rest/src/main/scala/com/normation/rudder/rest/internal/EventLogAPI.scala
clarktsiory: This PR was supposed to apply the new "before/after" option to the only existing place where we have per-item rollback: in the "change logs" page (which needs full admin rights as per the internal endpoint definition).

If we want per-item rollback in other places, we can just add the corresponding endpoints in another PR, but that's a whole new feature, it would require:
* as you say, config endpoints with accurate rights (that's at least ~5 endpoint per current API authz definition)
* the corresponding UI (if we want any, e.g. in "history" tab of config items, which needs to show the diff and the button with significant additions in the current Elm architecture)


I updated the current PR description with that to make it clearer

fanf (approved):
OK, LGTM

PR #7495, opened 2026-09-18
Fixes #29701: We need to have a startup timing log 

fanf:
https://issues.rudder.io/issues/29701

The main strangeness is that we have to build something completly aside our standard Lift/RudderConfig/ZIO habits because it needs to observe what could do wrong there, so it has its own object and `Thread`. 

Appart that, we just have `BootPhase` that are "main big chunk" of identified things in Rudder startup, and `BootStep` that are a smaller sub-step in a phase. 

Then, we have combinator to record progress, and the tread daemon that checks that progress happens regularly (and notify when something looks like it's stalling every XX seconds).


webapp/sources/rudder/rudder-web/src/main/resources/logback.xml
```diff
@@ -306,11 +306,12 @@ along with Rudder.  If not, see <http://www.gnu.org/licenses/>.
   <!--
     Startup
     =======
-    Information like timing each operation of the boot to understand
-    what is sooooo long - and should not.
-    Set to debug or trace to have information.
   -->
-  <logger name="bootstrap" level="info" additivity="false">
```
fanf, line 313: this one is a leftover from the time we used java package name for logger. There is nothing interesting at debug or trace level, and we don't use it anymore in code.

fanf:
PR updated with a new commit

webapp/sources/rudder/rudder-web/src/main/scala/bootstrap/liftweb/BootProgress.scala
```diff
@@ -0,0 +1,363 @@
...
+      val content = List(
+        s"state=${state}",
+        s"steps=${steps.get()}",
+        s"phase=${step.phase.name}",
+        s"detail=${step.detail}",
+        s"elapsed_sec=${elapsed.toSeconds}",
+        s"no_progress_sec=${stalledFor.toSeconds}"
+      ).mkString("", "\n", "\n")
+      val tmp     = File(progressFile.path.resolveSibling(progressFile.pathAsString + ".tmp"))
+      progressFile.createDirectories()
```
VinceMacBuche: This does not work, because it create the progress-file as a directory and it will fail the move after (a file cannot replace a directory with a move)

You should rather create a file with createFileIfNotExists with parent flag to true

VinceMacBuche: or you should create the parent with createDirecrtories

progressfile.parent.createDirectories

fanf: ah, I changed to use better file in place of java file late. Ah, I mixed up the method names between the two.

fanf: actually I'm not sure it makes sense that we create parent here, `/var/rudder/run` is likely managed by packaging

fanf:
PR updated with a new commit

fanf:
Changed. Also passed the sleep at the end of the loop so that the file is written at least once at the very begining of Rudder start, so that we can observe that the file is created then deleted.

webapp/sources/rudder/rudder-web/src/main/resources/configuration.properties.sample
```diff
@@ -794,6 +794,26 @@ rudder.relayd.reload=/opt/rudder/bin/rudder relay reload -p
...
+# It's useful as a live-monitoring check, when a boot is slow (for example
+# because of a migration).
+# In addition a log line in `/var/log/rudder/webapp/webapp.log`, the probe
+# write metric logs in var/rudder/run/rudder-boot-progress
+#
+rudder.boot.progress.logInterval=10 seconds
+
+#
+# Time after which a boot without progress (ie same step) is considered stalled
+# and a thread dumb is produced.
```
VinceMacBuche: thread dumb 😆

fanf: that's what happens when you write your comments by hand like it's 2024

fanf:
PR updated with a new commit

